PR #22, opened 2014-10-24
Make the DOMException constructor accept a name

domenic:
It will automatically set the code from the name (if it is one of the legacy names that have codes).

It now just calls `super(message)` to initialize the message and all of the internal error-related state.

The "creating exceptions" section was refactored to work in terms of the new, fully-functional constructor.

Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=27062.


bzbarsky:
I am planning to land support for this in Gecko pretty soon (not exactly Domenic's spec, but the existence of the constructor and its taking two arguments, etc).  @heycam, are you still planning to merge this one?

heycam:
Domenic's patch looks good.  @bzbarsky when you say "not exactly", what differences?  Should we make sure we align the spec and the code?

bzbarsky:
In Gecko DOMException is an interface so far.  So there are all sorts of differences in terms of properties being accessors on the prototype, not own value props.

The IDL I'm using at the moment for the constructor is:

    Constructor(optional DOMString message = "", optional DOMString name)

and mapping the name, if passed, to the right code, etc.  So apart from where the props live, I _think_ I'm matching the proposed spec.

bzbarsky:
I expect that in Gecko we'll add some WebIDL extensions to allow DOMException to continue being an interface but have behavior identical to the spec's... (e.g. IDL support for own value properties).

domenic:
ping

domenic:
@heycam @bzbarsky ping. Rebased on top of the current stuff and made some minor updates for the NewTarget stuff in ES.

heycam:
Thanks Domenic, and sorry again for the delay.  This looks good.

I'm not sure if there is great consistency in the ES spec about whether exception-ignoring [[DefineOwnProperty]] calls or DefinePropertyOrThrow should be used in general, but the latter (which you've changed to, for setting `message` and `code`) does look more common.  I wonder if there are other places in Web IDL where we should switch to DefinePropertyOrThrow?

Now that you have introduced the «1, 2, 3» syntax for abstract list values, there are probably a bunch of places in the spec that we should update to use it.

bzbarsky:
In theory, DefinePropertyOrThrow is identical to [[DefineOwnProperty]] if the object's [[DefineOwnProperty]] is the one defined by http://people.mozilla.org/~jorendorff/es6-draft.html#sec-ordinary-object-internal-methods-and-internal-slots-defineownproperty-p-desc and the object was just created (so doesn't have a property with that name).

But yes, I would think that what engines use internally is closer to DefinePropertyOrThrow, in that if it fails (e.g. due to OOM, which the spec doesn't really consider) an exception will be thrown.

heycam:
Yeah, that's fair. Though the author could replace window.Object with a constructor that defines some non-configurable properties on the instance after it's created?

bzbarsky:
That doesn't affect objects created via object literal `{}` (which is what e.g. dictionary IDL-to-ES conversion is supposed to use) or via other invocations of ObjectCreate with an empty internalSlotsList.

heycam:
Ah indeed, great.

PR #37, opened 2015-01-13
Fix https://www.w3.org/Bugs/Public/show_bug.cgi?id=23133

bzbarsky:
By hooking up interface object prototype chains to follow the interface inheritance chain, which is how ES6 classes that extend each other work.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
   <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/W3C-ED" type="text/css" /></head>
 
   <body>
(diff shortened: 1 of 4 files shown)


heycam:
Looks good, thanks.  (I was going to ask whether you thought the interface objects for `[ArrayClass]` interfaces should have their `[[Prototype]]` pointing to `Array`, but thought better of it, remembering that `[ArrayClass]` will go away soon.)

bzbarsky:
Yeah, I considered the ArrayClass thing, but those are really not subclasses of Array in any meaningful senses, and it would be pretty weird to put Array on their interface object proto chain.

PR #38, opened 2015-01-13
Followup for https://www.w3.org/Bugs/Public/show_bug.cgi?id=23133 to make sure we don't have to deal with defining prototype chains for interface objects when [NoInterfaceObject] is involved.

bzbarsky:


domenic:
I don't understand why you would skip the intermediate ones. The only thing [NoInterfaceObject] does is remove the global property; the interface still exists.

Another way of looking at it, is that this breaks the correpsondence between the constructor-side `__proto__`-chain and the .prototype-side `__proto__`-chain

bzbarsky:
> The only thing [NoInterfaceObject] does is remove the global property; the interface still exists.

Sorry, but no.  The spec says, at <http://heycam.github.io/webidl/#NoInterfaceObject>:

    If the [NoInterfaceObject] extended attribute appears on an interface, it indicates that an interface
    object will not exist for the interface in the ECMAScript binding.

And then in http://heycam.github.io/webidl/#interface-prototype-object it says:

    If the [NoInterfaceObject] extended attribute was not specified on the interface, then the interface
    prototype object MUST also have a property named “constructor” 

You can see this in Gecko:

    (new XMLHttpRequest).__proto__.__proto__.hasOwnProperty("constructor")

tests false.  Blink has a different behavior, but that's a bug in Blink per current spec.

> this breaks the correpsondence 

Not any more than if you had an object with no "constructor" property on a prototype chain in general...

We could change things around so that constructor objects are in fact required to exist but simply not be exposed on the global with [NoInterfaceObject].  That's not how things work right now, however.

Of course this whole business of having [NoInterfaceObject] stuff on proto chains is pretty broken to start with, imo.  The only case I where this happens (at least in Gecko's IDL) is XMLHttpRequestEventTarget, which isn't even implemented anywhere except Blink and Gecko as far as I can tell.  So maybe we can just drop the [NoInterfaceObject] from there and then disallow having [NoInterfaceObject] ancestor interfaces for something that has an interface object?  @annevk?

domenic:
Ah, I see, I was mistaken. I guess [NoInterfaceObject] is more about abandoning the class model entirely and just doing prototypal inheritance.

bzbarsky:
Really, I think the uses of [NoInterfaceObject] fall into the following buckets right now:

1. XMLHttpRequestEventTarget for some reason.
2. Interfaces that appear on the RHS of "implements",  aka mixins.  These would benefit from a specific mixin syntax in IDL.
3. Compat concerns about exposing a name on a global (don't know specific examples here).
4. Cases where people thought it was best practice for reasons that escape me (to avoid global pollution?  Unclear).

In my ideal world we'd have a mixin syntax, drop the annotation on bucket 4, and see what remains in bucket 3, if anything.

domenic:
Of all the "my ideal world"s I've seen recently, that one seems pretty achievable. First step, fix 1?

bzbarsky:
Yep.  Just talked to Anne, filed <https://www.w3.org/Bugs/Public/show_bug.cgi?id=27823>.  He will try to do that tomorrow.  Once that's done, we change IDL to not allow a NoInterfaceObject ancestor for an interface with an interface object, and close this PR as no longer relevant.  ;)

bzbarsky:
Alright, let's just mutate this PR to do the new hotness: just disallow things that are not NoInterfaceObject inheriting from things that are.  As in, interfaces should not inherit from mixins.

domenic:
LGTM, yay!

heycam:
> Alright, let's just mutate this PR to do the new hotness: just disallow things that are not
> NoInterfaceObject inheriting from things that are. As in, interfaces should not inherit from mixins.

:+1: Like

PR #39, opened 2015-01-19
Fix a typo in editorial note for @@toStringTag

caitp:
All `@@toStringTag` instances, short of `Symbol.toStringTag` (where it's exposed as a public API), are configurable in ES6 --- although not everyone is happy about that. It's likely to stay that way

### DIFF
--- a/index.xml
+++ b/index.xml
@@ -6249,8 +6249,8 @@ interface Person {
         </p>
         <div class='ednote'>
           <p>Should define whether <a>@@toStringTag</a> is writable, enumerable and configurable.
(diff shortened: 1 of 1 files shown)


caitp:
well, it's not `Symbol.@@toStringTag` (which isn't defined), it's just the value property of the `Symbol` function called `toStringTag`, which is a read-only property whose value is the `@@toStringTag` well known symbol --- it seems kind of irrelevant. In terms of `@@toStringTag` uses on builtin object types, it's always configurable.

I believe this is deliberate, and it's unlikely to change now since that ship has pretty much sailed, but I can remove the "Not sure if that is deliberate." if that sounds good

PR #40, opened 2015-01-20
Only use the named properties object as [[Prototype]] if it exists.

Ms2ger:


heycam:
Looks good.  Could you apply the change to v1.xml as well?

PR #45, opened 2015-03-06
Add a platform object [[Set]] internal method so we can have prototype setters show through even if an object has a named property of the same name and can have named creators work correctly on [OverrideBuiltins] objects even if thre is a property with the same name on the prototype.

bzbarsky:
This fixes <https://www.w3.org/Bugs/Public/show_bug.cgi?id=25025> and
<https://www.w3.org/Bugs/Public/show_bug.cgi?id=26521>.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
   <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/W3C-ED" type="text/css" /></head>
 
   <body>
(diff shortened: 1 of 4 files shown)


heycam:
I think I have asked this before, but: what do you think of the idea of not exposing indexed and named properties through [[GetOwnPropertyDescriptor]] and [[DefineProperty]], and instead just exposing them through [[Get]] and [[Set]]? It seems to be a better fit, exposing them at that level. Is it too late for this?

bzbarsky:
Applied the review comments.

As far as adding a [[Get]], see the public-script-coord discussion with Domenic.  I think the overall agreement was that having [[GetOwnProperty]] match [[Get]] is nice, so we don't get weirdness with [[GetOwnProperty]] claiming props don't exist while they are being enumerated and gotten... Basically, try to keep weirdness to a minimum.

PR #47, opened 2015-04-08
Fix up a bunch of cases in maplike/setlike code where we were working with the interface instance when we meant to work with the backing object or returning the backing object when we meant to return the interface instance.

bzbarsky:


heycam:
Otherwise looks good.

bzbarsky:
> If you agree that's the case, can you turn them into assertions that we didn't get an abrupt completion?

Done.

> Typo here. (Well, in the corresponding index.xml bit.)

Fixed.

PR #52, opened 2015-06-22
Fix arrays

heycam:
@bzbarsky can you take a look at this, as a solution to https://www.w3.org/Bugs/Public/show_bug.cgi?id=23682?

### DIFF
--- a/index.xml
+++ b/index.xml
@@ -599,13 +599,13 @@ dictionary <i>identifier</i> {
           <div class='example'>
             <p>Therefore, the following <a class='dfnref' href='#dfn-idl-fragment'>IDL fragment</a> is valid:</p>
             <x:codeblock language='idl'>interface B : A {
-  void f(ArrayOfLongs x);
+  void f(SequenceOfLongs x);
 };
 
 interface A {
 };
 
-typedef long[] ArrayOfLongs;</x:codeblock>
+typedef sequence&lt;long&gt; SequenceOfLongs;</x:codeblock>
           </div>
 
           <div class='example'>
@@ -625,7 +625,7 @@ interface System {
   object createObject(DOMString _interface);
 
   <span class='comment'>// Operation argument identifier: "interface"</span>
-  object[] createObjectArray(DOMString interface);
+  sequence&lt;object&gt; getObjects(DOMString interface);
 
   <span class='comment'>// Operation has no identifier; it declares a getter.</span>
   getter DOMString (DOMString keyName);
@@ -3224,7 +3224,6 @@ partial interface A {
                     <th>callback <br/>function</th>
                     <th>dictionary</th>
                     <th>sequence&lt;<var>T</var>&gt;</th>
-                    <th><var>T</var>[]</th>
                     <th>Date</th>
                     <th>RegExp</th>
                     <th>exception types</th>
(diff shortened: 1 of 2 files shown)


index.xml
```diff
@@ -4040,8 +4042,6 @@ dictionary <i>Derived</i> : <em>Base</em> {
             <li>the type is a dictionary that <a class='dfnref' href='#dfn-dictionary-inherit'>inherits</a> from <var>D</var></li>
             <li>the type is a <a class='dfnref' href='#dfn-nullable'>nullable type</a>
             whose <a class='dfnref' href='#dfn-inner-type'>inner type</a> includes <var>D</var></li>
-            <li>the type is a <a href='#idl-sequence'>sequence type</a> or <a href='#idl-array'>array type</a>
```
bzbarsky: Why can the sequence bits here go away?  Shouldn't they stay but `FrozenArray` get added?

index.xml
```diff
@@ -8593,94 +8243,74 @@ results.numbers.length;       <span class='comment'>// So this still evaluates t
...
-                of type <span class='estype'>unsigned long</span> that exposes the length
-                of the array-like object, in order for the inherited <span class='estype'>Array</span>
-                methods to operate correctly.  Such interfaces would typically also
-                <a class='dfnref' href='#dfn-support-indexed-properties'>support indexed properties</a>,
-                which would provide access to the array elements.
-              </p>
-            </div>
+            <ol class='algorithm'>
+              <li>Let <var>values</var> be the result of
+                <a class='dfnref' href='#dfn-convert-ecmascript-to-idl-value'>converting</a>
```
bzbarsky: This is different from the conversion done in the union type case.  Why?  Seems like we should do the same thing in both cases....

index.xml
```diff
@@ -599,13 +601,13 @@ dictionary <i>identifier</i> {
...
             <x:codeblock language='idl'>interface B : A {
-  void f(ArrayOfLongs x);
+  void f(SequenceOfLongs x);
 };
 
 interface A {
 };
 
-typedef long[] ArrayOfLongs;</x:codeblock>
+typedef sequence&lt;long&gt; ArrayOfLongs;</x:codeblock>
```
jyasskin: s/ArrayOfLongs/SequenceOfLongs/ to match the use above.

index.xml
heycam: They should indeed stay.

index.xml
heycam: Hmm, so in what way are they different, apart from doing the iterable check up front (before doing it again in the "convert to sequence<T>" algorithm)?

Maybe it's better to just call into the "convert to sequence<T>" and "convert to FrozenArray<T>" algorithms directly here from the union conversion algorithm. That would still have some duplicate processing (we'd check for Date/RegExp twice now, rather than checking for iterability twice) but at least would capture any changes we make to those two algorithms without needing to change anything here.

index.xml
heycam: Ah yes, thanks.

index.xml
bzbarsky: Hmm, I guess the point is that we want "create a sequence from iterable" vs "convert to sequence" to in fact avoid doing the is-iterable check twice.  OK, so what you have actually makes sense, then; I just misread it initially.

heycam:
Thanks for reviewing.

PR #58, opened 2015-09-25
Fix the SerializerRest production rule

nox:
Serializers, as everything else, are terminated by a semicolon.


heycam:
Looks good, thanks!

Issue #61, opened 2015-10-07
CheckIterable(V) is not defined in ECMAScript 2015


travisleithead:
So, this leaves [at least] the conversion from an ECMAScript value V to a sequence<T> undefined--not sure exactly what CheckIterable() was supposed to do.

My main question: are array-like objects supported (e.g., NodeLists, or Objects with a length and indexed properties), or must they only have a Symbol.iterator? (Or some other thing.)

http://heycam.github.io/webidl/#es-sequence

I checked with Brian Terlson, and he wasn't sure what CheckIterable() was supposed to do either.

bzbarsky:
CheckIterable got refactored into GetMethod with P === Symbol.iterator, I believe.

> My main question: are array-like objects supported (e.g., NodeLists, or Objects with a length and indexed properties), or must they only have a Symbol.iterator?

The latter, but node that `NodeList` has a `Symbol.iterator`.

PR #63, opened 2015-10-07
Fix https://github.com/heycam/webidl/issues/61 by using GetMethod where we used to use CheckIterable.

bzbarsky:
@heycam Just wanted to double-check the bit about linking to the final HTML ES2015 spec instead of the draft we link to everywhere else here.


heycam:
This looks good.  Thank you for fixing this.

PR #68, opened 2015-10-19
Remove the legacyiterable keyword.

Ms2ger:
Fixes <https://www.w3.org/Bugs/Public/show_bug.cgi?id=26984>.


heycam:
Thanks, looks good.

PR #72, opened 2015-11-12
Fix some problems with iterable declarations.

heycam:
1. iterable<> declarations with a single type ("value iterators") now
   always iterates over the object's indexed properties, and disallows
   its use on objects that don't support indexed properties (and have a
   "length" attribute).

2. The entries and keys properties of interface prototypes for
   interfaces that have value iterators are now defined to have values
   equal to the initial values of Array.prototype.{entries,keys},
   similarly to how @@iterator was already defined when indexed
   properties were present.

3. The definitions of @@iterator and forEach for iterable<> declarations
   with two types ("pair iterators") are fixed.  "Default iterator
   objects" now work only for pair iterators, since we don't support
   prose-defined value iterators now.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
   <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/W3C-ED" type="text/css" /></head>
 
   <body>
-    <div class="head"><div><a href="http://www.w3.org/"><img src="https://www.w3.org/Icons/w3c_home" width="72" height="48" alt="W3C" /></a></div><h1>Web IDL (Second Edition)</h1><h2>W3C Editor’s Draft <em>18 January 2016</em></h2><dl><dt>This Version:</dt><dd><a href="http://heycam.github.io/webidl/">http://heycam.github.io/webidl/</a></dd><dt>Latest Version:</dt><dd><a href="http://www.w3.org/TR/WebIDL/">http://www.w3.org/TR/WebIDL/</a></dd><dt>Previous Versions:</dt><dd><a href="http://www.w3.org/TR/2012/CR-WebIDL-20120419/">http://www.w3.org/TR/2012/CR-WebIDL-20120419/</a></dd><dd><a href="http://www.w3.org/TR/2012/WD-WebIDL-20120207/">http://www.w3.org/TR/2012/WD-WebIDL-20120207/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110927/">http://www.w3.org/TR/2011/WD-WebIDL-20110927/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110712/">http://www.w3.org/TR/2011/WD-WebIDL-20110712/</a></dd><dd><a href="http://www.w3.org/TR/2010/WD-WebIDL-20101021/">http://www.w3.org/TR/2010/WD-WebIDL-20101021/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20081219/">http://www.w3.org/TR/2008/WD-WebIDL-20081219/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20080829/">http://www.w3.org/TR/2008/WD-WebIDL-20080829/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/">http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/</a></dd><dd><a href="http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/">http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/</a></dd><dt>Participate:</dt><dd>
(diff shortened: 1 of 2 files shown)


heycam:
@bzbarsky can you take a look at this?

index.html
```diff
@@ -3617,28 +3617,56 @@ <h2 id="sotd">Status of This Document</h2>
...
+              The value-type of the <a class="dfnref" href="#dfn-value-iterator">value iterator</a>
+              <span class="rfc2119">MUST</span> be the same as the type returned by
+              the <a class="dfnref" href="#dfn-indexed-property-getter">indexed property getter</a>.
+              A <a class="dfnref" href="#dfn-value-iterator">value iterator</a> is implicitly
+              defined to iterate over the object’s indexed properties.
+            </p>
+            <p>
+              A <a class="dfnref" href="#dfn-pair-iterator">pair iterator</a>
+              <span class="rfc2119">MUST NOT</span> be declared on an interface
+              that does <a class="dfnref" href="#dfn-support-indexed-properties">support indexed properties</a>.
```
bzbarsky: "that supports indexed properties" instead of "that does support indexed properties"?

index.html
```diff
@@ -3617,28 +3617,56 @@ <h2 id="sotd">Status of This Document</h2>
...
+                Since <a class="dfnref" href="#dfn-value-iterator">value iterators</a>
+                are currently allowed only on interfaces that
+                <a class="dfnref" href="#dfn-support-indexed-properties">support indexed properties</a>,
+                it makes sense to use an Array-like forEach method.
+                There may be a need for <a class="dfnref" href="#dfn-value-iterator">value iterators</a>
+                (a) on interfaces that do not 
+                <a class="dfnref" href="#dfn-support-indexed-properties">support indexed properties</a>,
+                or (b) with a forEach method that instead invokes its callback like
+                Set.protoype.forEach (where the key is the same as the value).
+                If this is you, please send a request to
```
bzbarsky: Maybe "if you're creating an API that needs such a forEach method, please ...."

index.html
```diff
@@ -12034,7 +12068,12 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
-                If the interface has an <a class="dfnref" href="#dfn-iterable-declaration">iterable declaration</a>,
+                If the interface defines an <a class="dfnref" href="#dfn-indexed-property-getter">indexed property getter</a>,
+                then the <span class="estype">Function</span> object is
+                the initial value of the “forEach” data property of <a class="external" href="http://people.mozilla.org/~jorendorff/es6-draft.html#sec-well-known-intrinsic-objects">%ArrayPrototype%</a> (<a href="#ref-ECMA-262">[ECMA-262]</a>, section 6.1.7.4).
+              </p>
+              <p>
+                If the interface has a <a class="dfnref" href="#dfn-pair-iterator">pair iterator</a>,
                 then the <span class="estype">Function</span> <span class="rfc2119">MUST</span>
                 have the same behavior as one that would exist assuming the interface had
                 this <a class="dfnref" href="#dfn-operation">operation</a> instead of the
```
bzbarsky: `= undefined` is not valid syntax afaict (though this seems to predate this PR).  I thought we were just going to make optional any default to undefined automatically...

index.html
```diff
@@ -12045,15 +12084,23 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
+                    <li>Let <var>pair</var> be the entry in <var>pairs</var> at index <var>i</var>.</li>
+                    <li>Let <var>key</var> be <var>pair</var>’s key.</li>
+                    <li>Let <var>value</var> be <var>pair</var>’s value.</li>
                     <li><a href="#es-invoking-callback-functions">Invoke</a> <var>callback</var> with <var>thisArg</var>
                       as the <a class="dfnref" href="#dfn-callback-this-value">callback this value</a> and
-                      <var>k</var> and <var>value</var> as its arguments.</li>
+                      <var>value</var>, <var>key</var> and <var>O</var> as its arguments.</li>
+                    <li>Update <var>pairs</var> to the current list of <a class="dfnref" href="#dfn-value-pairs-to-iterate-over">value pairs to iterate over</a>.
+                      <div class="note"><div class="noteHeader">Note</div>
+                        <p>Depending on whether prose accompanying the interface defined this to be a snapshot at the time
```
bzbarsky: Hmm.  I don't see how this works.  The list of value pairs to iterate over is just stored in the object, no?  There's nowhere to put the snapshot.  As in, I don't see how prose could define snapshotting behavior here.

bzbarsky:
And a general note: We may want to move towards links to http://www.ecma-international.org/ecma-262/6.0/ instead of the out-of-date draft.

domenic:
You want https://tc39.github.io/ecma262 actually. Lots of bug fixes already.

bzbarsky:
> You want https://tc39.github.io/ecma262 actually. 

Does that promise sufficient link-stability (as in, anchor names won't go dead)?

domenic:
@bterlson can confirm but I am pretty sure that's the intent

bterlson:
tc39.github.io/ecma262 is essentially `master` and thus doesn't make the usual promises WRT stability - there may be significant changes from day-to-day, including changing/removing/refactoring clauses that would impact anchor names. That said, as a general rule, changing anchor names is something to be avoided and, while I make no promises, it seems unlikely to be a problem. So far I've preserved all anchor names from ES6.

I would be more concerned with linking to a document that is not a ratified standard, may have pieces that do not have TC39 consensus, and could really change at any time. The official ES6 spec seems like the best bet (though it's sad its not nearly as usable as the master version).

index.html
heycam: Yeah, that's not going to work.  I suppose we should be more explicit about what's going on here.  But first question: do we want to bother with supporting both snapshotted lists to iterate over here?  Map.prototype.forEach doesn't snapshot, although it will guarantee that an entry added after iteration starts (which is not then deleted) will be iterated before iteration ends, which isn't something we can guarantee by using an index.

bzbarsky: With an index we can't even guarantee that an entry that was there before iteration starts will be iterated, if the right deletions happen.

For example, say our original list consists of the entries [0, 1, 2, 3, 4, 5] each number at itself as index.  We start iterating and in the callback with index == 3 delete entries 0, 1, 2.  At that point length == 3, so iteration is done and no callbacks happen for 4 and 5.

Map and Set go out of their way to avoid this with a specification device that can't be implemented in practice (removal never gives up memory) and some resulting implementation complexity.

Array has the behavior we have now: just use an index.  This is, of course, required because the Array stuff is trying to be generic and hence can't depend on tight cooperation with the thing being iterated.

We _could_ spec the Map/Set behavior for two-type iterables, I think, as follows: spec that the iterator object has a way to notify it on removal from the list (with the index the removal happened at) and if so notified will reduce its index if the removal happened before it.  Similar for insertions before the current index.  Prose for specific iterables would then have to define when (or whether) these notifications are delivered.  Maybe this is worth doing; that way the default behavior will be that there are no notfications so you get the Array iteration behavior but interfaces that want the Map/Set one can opt in to it without too much trouble.  I'm not sure we can make it any _easier_ to opt in to it than that.

I don't think we really want to support fully snapshotting here.  If people want that, they can define a way to produce an iterable snapshot as part of their interface.

heycam: Having notifications to handle the Map/Set behaviour would mean that spec prose would need to interact with the "value pairs to iterate over" by explicitly saying "insert a pair at index i" or "remove a pair at index j", rather than defining more declaratively what the list would be at any given moment.  That's pretty awkward.  And in any case pairs could be removed and re-inserted at later indexes and be iterated over multiple times, which from the ES spec it sounds like Maps/Sets shouldn't do.  One thing that makes it a little easier for Maps/Sets is that new entries go at the end and can't be inserted at arbitrary indexes.

I don't think we need to allow spec authors to choose between iteration behaviour that ensures we get all entries, and one based on simple indexes that means we might skip some.  I can't see why there'd be good reasons to choose differently for different APIs.

An easier way to spec the "ensure we don't skip any entries" iteration behaviour: at each iteration of the loop (for forEach) or invocation of next (on the default iterator object), we scan the list for the first pair whose key isn't present in a set of handled keys that we maintain, then we invoke the callback / return that entry.  As written that's quadratic time, but we could leave it to implementors to work out to handle the common case of the list of value pairs *not* changing, or to inform the iterator about entries that were inserted/removed (using some knowledge of the particular API being implemented), without needing the spec author to specify that.  That might not always be easy, depending on how "dynamic" the definition of the list of pairs is...

heycam: But TBH I'd be willing to break with the Map/Set symmetry and just do something simplistic like Arrays.

bzbarsky: > And in any case pairs could be removed and re-inserted at later indexes and be iterated over multiple times

bzbarsky: > And in any case pairs could be removed and re-inserted at later indexes and be iterated over multiple times

Yep.  I think the same can happen with Map/Set, actually.  For example, this script:

    var s = new Set([1, 2, 3]); 
    var e = s.values(); 
    console.log(e.next()); 
    s.delete(1); 
    s.add(1); 
    console.log(...e)

logs:

    Object { value: 1, done: false }
    2 3 1


> I can't see why there'd be good reasons to choose differently for different APIs.

Well, performance and simplicity, possibly.

But yes, we could do what you suggest.  I don't believe it would match the Map/Set behavior, though, per above.

heycam:
I removed the talk of snapshots and left the Array-like index incrementing pair iterator behaviour.

heycam:
Sorry for letting this slip through the cracks for so long when it just needed landing!

PR #77, opened 2015-12-10
Move the overload resolution algorithm into a separate section.

Ms2ger:


bzbarsky:
Thank you for doing this!

PR #85, opened 2016-01-16
Add NotAllowedError exception

mounirlamouri:
This is needed for the HTML specification, see:
https://github.com/whatwg/html/pull/509

There is unfortunatelly no exception that could be used instead.


jan-ivar:
What about `SecurityError`?

mounirlamouri:
@jan-ivar `SecurityError` is about an "insecure" operation. The operation is rejected for security reasons. In our case, we want to allow UA to expose to pages when they disallow media playback in the background or without user interaction. It's not related to security but more about user experience, bandwidth saving, etc.

jan-ivar:
These errors tend to get reused broadly, don't they? E.g. `getUserMedia` [1] uses `SecurityError` for permission failure (rejection through user-chosen permission settings), which strikes me as similar to this use-case, unless I'm misunderstanding something?

[1] See http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia

mounirlamouri:
A permission rejection is definitely more related to security than audio bothering the user. I do not think these are similar.

jan-ivar:
(Had `NotAllowedError` existed, we probably would have used it for `getUserMedia`.)

foolip:
Ping? I'd like to use this for https://mediasession.spec.whatwg.org/#dom-mediasession-activate as well, as that perfectly matches the "is not allowed by ... the platform in the current context."

heycam:
Sorry for the delay. `NotAllowedError` sounds distinct enough from `SecurityError` to me.

PR #91, opened 2016-02-16
Specify [LegacyUnenumerableNamedProperties] (fixes #82)

### LINKED ISSUE #82: All supported property names should be enumerable or none of them
Currently in the platform, the only interface with [unenumerable](http://heycam.github.io/webidl/#dfn-unenumerable) property names is [NamedNodeMap](https://dom.spec.whatwg.org/#dom-htmlcollection-item). In that interface, they are *all* unenumerable.

WebIDL should reflect that and forbid mixing unenumerable and enumerable property names.

nox:


index.xml
```diff
@@ -11163,10 +11183,8 @@ partial interface Window {
                     <li>Let <var>desc</var> be a newly created <a>Property Descriptor</a> with no fields.</li>
                     <li>Set <var>desc</var>.<span class='prop'>[[Value]]</span> to the result of <a class='dfnref' href='#dfn-convert-idl-to-ecmascript-value'>converting</a>
                       <var>value</var> to an ECMAScript value.</li>
-                    <li>If the named property is defined to be <a class='dfnref' href='#dfn-unenumerable'>unenumerable</a>,
-                      then set <var>desc</var>.<span class='prop'>[[Enumerable]]</span> to <span class='esvalue'>false</span>,
-                      otherwise set it to <span class='esvalue'>true</span>.</li>
-                    <li>Set <var>desc</var>.<span class='prop'>[[Writable]]</span> to <span class='esvalue'>true</span> and
+                    <li>Set <var>desc</var>.<span class='prop'>[[Writable]]</span> to <span class='esvalue'>true</span>,
```
bzbarsky: This is wrong, I think.  Github's broken UI is refusing to show me where this diff is actually applied, but if this is the "Named properties object [[GetOwnProperty]] method" section, as I think it is, that means the `Window` interface in practice, and its properties are currently unenumerable, right?  More to the point, you should be using the `[LegacyUnenumerableNamedProperties]` flag from `A` here, I would think.

index.xml
```diff
@@ -13402,7 +13420,9 @@ C implements A;</x:codeblock>
                     <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>-->, then set
                     <var>desc</var>.<span class='prop'>[[Writable]]</span> to <span class='esvalue'>true</span>, otherwise set it to
                     <span class='esvalue'>false</span>.</li>
-                  <li>If the named property is defined to be <a class='dfnref' href='#dfn-unenumerable'>unenumerable</a>,
+                  <li>If <var>O</var> implements an <a class='dfnref' href='#dfn-interface'>interface</a> with the
+                    <a class='xattr' href='#LegacyUnenumerableNamedProperties'>[[LegacyUnenumerableNamedProperties]]</a>
```
bzbarsky: `[LegacyUnenumerableProperties]`, not `[[LegacyUnenumerableProperties]].

This phrasing is pretty weird, in that it allows `[LegacyUnenumerableProperties]` to appear on one interface but the named getter to appear on another interface.  It probably doesn't matter in practice, of course...

index.xml
```diff
@@ -13749,10 +13769,11 @@ C implements A;</x:codeblock>
               <li>If the object <a class='dfnref' href='#dfn-support-indexed-properties'>supports indexed properties</a>, then
               the object’s <a class='dfnref' href='#dfn-supported-property-indices'>supported property indices</a> are
               enumerated first, in numerical order.</li>
-              <li>If the object <a class='dfnref' href='#dfn-support-named-properties'>supports named properties</a>, then
+              <li>If the object <a class='dfnref' href='#dfn-support-named-properties'>supports named properties</a> and doesn't implement an <a class='dfnref' href='#dfn-interface'>interface</a> with the
+              <a class='xattr' href='#LegacyUnenumerableNamedProperties'>[[LegacyUnenumerableNamedProperties]]</a>
+              <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>, then
```
bzbarsky: How carefully has the enumerability (or not) of this stuff been audited?  I thought it was pretty thorough, but the `Window` thing above makes me worried now...

Again, this has the weirdness where the named getter and the extended attribute could be on different interfaces.  Maybe the definition of the extended attribute should require that it only appears on interfaces with a named getter?

bzbarsky:
Oh, and landing this needs to be coordinated with landing corresponding changes to all the consuming specs (as in, land all at once) so that we don't end up in a state where the specs specify something that doesn't match reality for a bit.

index.xml
nox: What Window thing?

nox:
I don't think it is important to coordinate. No one coordinated `T[]` going away in favour of `FrozenArray<T>`.

index.xml
nox: AFAICT, that is step 2 of "4.8.2. The PlatformObjectGetOwnProperty abstract operation".

nox:
@bzbarsky I added some constraints ine the UnenumerableNamedProperties section.

index.xml
bzbarsky: Well, then setting `[[Enumerable]]` to true unconditionally would be wrong too, right?

But I _think_ this is "4.6.5.1. Named properties object [[GetOwnProperty]] method", not 4.8.2.

bzbarsky:
> No one coordinated `T[]` going away in favour of `FrozenArray<T>`.

I think that was a mistake too.  But a significant difference is that in that case the result was that HTML, say, ended up with invalid IDL, so an implementor implementing it would notice there was a problem.  The result in this case would be that HTML would end up with perfectly correct IDL that simply doesn't mean what it should mean.

index.xml
nox: Oh I see, something weird is going on here indeed.

nox:
You will notice too a difference here because the term "unenumerable" won't exist anymore.

bzbarsky:
That won't be obvious to the person reading the other spec, and being in prose won't be caught by the IDL compiler either....

bzbarsky:
Maybe it's not being clear; let me give a concrete example of how this would play out in practice.

Right now, in Gecko, a class with a named getter implements a function that enables bindings to ask whether a given property name is enumerable.  This function implements the prose definition of which properties are unenumerable.

Say the bindings in Gecko got updated to no longer call this function and instead just use the `[LegacyUnenumerableNamedProperties]` bit from IDL.  Now someone comes along and implements a currently-specced interface that does the prose thing.  They copy the IDL from the spec, implement this "is this name enumerable?" function to return false for all names, and think they're OK.  But of course they're not.  Of course arguably they should have tests that would catch the problem, but I would really rather not rely on that.  Getting the other specs update here shouldn't be terribly complicated; we should just do it.

nox:
> Say the bindings in Gecko got updated to no longer call this function and instead just use the [LegacyUnenumerableNamedProperties] bit from IDL. Now someone comes along and implements a currently-specced interface that does the prose thing. They copy the IDL from the spec, implement this "is this name enumerable?" function to return false for all names, and think they're OK. But of course they're not. Of course arguably they should have tests that would catch the problem, but I would really rather not rely on that. Getting the other specs update here shouldn't be terribly complicated; we should just do it.

How is that different from missing the prose stating that the properties should be unenumerable?

index.xml
nox: Fixed.

bzbarsky:
It's just an additional point of failure.  And in this case an easily avoidable one, so we should just avoid it.

nox:
Fair enough.

annevk:
I'm happy to patch HTML and DOM once this lands. Ping me on IRC.

heycam:
Aside from HTML and DOM, are there any other specs that would need updating?

annevk:
I don't think so, but not a 100% sure. I checked WebRTC, Audio, WebGL, CSSOM.

nox:
@heycam @bzbarsky LGTM?

bzbarsky:
Looks fine in general.  Need to update the changelog at the bottom of the file, right?

nox:
Will do today.

annevk:
It might be worth following https://github.com/whatwg/html/blob/master/TEAM.md in order to have a clean git history from now on, providing you with a changelog for free at the same time (provided you also review the commit messages and such).

nox:
@bzbarsky Done, LGTM?

bzbarsky:
> It might be worth following https://github.com/whatwg/html/blob/master/TEAM.md 

Sure.  Though I'd really rather get out of the editing business entirely...

bzbarsky:
@annevk Please link to the relevant HTML/DOM changes when they happen.

annevk:
DOM: https://github.com/whatwg/dom/commit/742e8c15312ae447580d2fd5498a02d0902d5ff3 (No interface here has supported property names that are enumerable, by the way.)

annevk:
While fixing HTML I wonder if this patch works for the `Window` object. It doesn't quite seem like it.

PR for HTML: https://github.com/whatwg/html/pull/740.

annevk:
Relevant HTML commit: https://github.com/whatwg/html/commit/07f6c7f058c57d582db1aced6741cea1ed63ccf7 (This depends on it applying to derived interfaces too. In particular HTMLCollection has two derived interfaces.)

PR #94, opened 2016-02-28
Clarify [LegacyUnenumerableNamedProperties] with derived interfaces

nox:


nox:
See https://github.com/whatwg/html/pull/740#discussion_r54332414.

heycam:
(I see we already apply [LegacyUnenumerableNamedProperties] to derived interfaces due to the wording in #named-properties-object-getownproperty / #getownproperty-guts about any interface having that extended attribute.)

PR #108, opened 2016-04-12
Link the common typedefs, and fix FrozenArray's lt attribute.

jyasskin:
This uses [Bikeshed's markup](https://github.com/tabatkins/bikeshed/blob/master/docs/dfn-contract.md) like #105.


heycam:
Thank you.

PR #110, opened 2016-04-20
Bikeshed-link fixes

tabatkins:
* The heading-dfns need to have their linking text set explicitly; right now they're including the section number inserted by XSLT.
* The simple errors should be dfn'd, I think.  (Until ES is Bikeshed-friendly, at least.)
* The DOMException types should be "exception" type.  (I misunderstood what they were previously.)


heycam:
Thank you.

PR #111, opened 2016-04-20
Remove support for [Constructor] on dictionaries (fixes #109).

### LINKED ISSUE #109: Remove dictionary constructors
They were added for `DOMPoint` and friends (https://bugzilla.mozilla.org/show_bug.cgi?id=854490), but those turned into interfaces. Unless something else started to depend on it, we should probably drop it.

Ms2ger:


index.xml
```diff
@@ -8390,7 +8382,7 @@ context.setColorClamped(-1, 255, 257);</x:codeblock>
             <p>
               The prose definition of a constructor <span class='rfc2119'>MUST</span>
               either return an IDL value of a type corresponding to the interface
-              or dictionary the <a class='xattr' href='#Constructor'>[Constructor]</a>
+              the <a class='xattr' href='#Constructor'>[Constructor]</a>
               extended attribute appears on, or throw an exception.
             </p>
             <p>
```
heycam: Can you get rid of the conditional in the paragraph that begins on this line (sorry, bit hard to link to) so that it just disallows [NoInterfaceObject] when [Constructor] is used any time.

index.xml
```diff
@@ -8604,7 +8558,6 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
                 implicitly, depending on the type of construct:</p>
                 <dl class='switch'>
                   <dt>interface</dt>
-                  <dt>dictionary</dt>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface
                   or dictionary only contains the
```
heycam: Remove "or dictionary" here.

index.xml
```diff
@@ -8654,23 +8607,17 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
...
               <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>,
               and either:
             </p>
             <ul>
               <li>the <a>relevant settings object</a> for the ECMAScript global object is a <a>secure context</a>; or</li>
-              <li>the interface, interface member or dictionary is not <a class='dfnref' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>.</li>
+              <li>the interface or interface member is not <a class='dfnref' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>.</li>
             </ul>
             <div class='note'>
               <p>Since it is not possible for the <a>relevant settings object</a>
```
heycam: Please remove the mention of dictionary in this note.

index.xml
```diff
@@ -9809,9 +9755,8 @@ counter.value;                                           <span class='comment'>/
...
               appears on an <a class='dfnref' href='#dfn-interface'>interface</a>,
               <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
-              an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a>, or
-              <a class='dfnref' href='#dfn-dictionary'>dictionary</a> with a <a href='#es-dictionary-constructors'>constructor</a>,
-              it indicates that the interface, interface member or <a href='#es-dictionary-constructors'>dictionary constructor</a> is exposed
+              an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a>,
+              it indicates that the interface or interface member is exposed
               only within a
               <a href='https://w3c.github.io/webappsec-secure-contexts/#secure-context'>secure context</a>
               (<a href='#ref-SECURE-CONTEXTS'>[SECURE-CONTEXTS]</a>, section 2).
```
heycam: Two paragraphs below here there is a reference to dictionaries that should be removed.

heycam:
r=me with those changes.

index.xml
Ms2ger: So change

> If the [Constructor] extended attribute is specified on an interface, then…

to

> If the [Constructor] extended attribute is specified, then…

?

heycam: That's OK.  I was thinking: "The [Constructor] extended attribute must not be specified on an interface with [NoInterfaceObject]." or "The [Constructor] and [NoInterfaceObject] extended attributes must not be specified on the same interface." or something like that.

PR #113, opened 2016-05-03
Modernize invoking user code

domenic:
This updates the four algorithms that invoke user code in the following ways:

- Updates them to use HTML's latest framework for managing script settings objects. The previous version referred to concepts, such as the "stack of incumbent scripts", which have not been defined for a long time (i.e. since before whatwg/html#401). Updates in whatwg/html#401 and whatwg/html#1091 have changed the setup here. There is still some ongoing discussion in whatwg/html#473 about whether the current HTML setup is correct for managing _incumbent_ settings objects, but this at least updates the algorithms to correctly manage _entry_ settings objects. The plan implemented here is roughly that discussed in https://github.com/whatwg/html/issues/473#issuecomment-212164549.
- Gives them explicit `<dfn>`s with explicit arguments to be passed.
- Updates them to use some slightly-more-modern ES technology such as Call(), Get(), and Set().

This also updates all ECMAScript references to point to the latest spec.

--

Review appreciated @bzbarsky. As I say above, incumbent might be a bit messed up at the moment, but I am pretty sure this is a strict improvement. I wanted to finish off fixing "entry"; now I need to go off an take care of incumbent.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
   <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/W3C-ED" type="text/css" /></head>
 
   <body>
-    <div class="head"><div><a href="http://www.w3.org/"><img src="https://www.w3.org/Icons/w3c_home" width="72" height="48" alt="W3C" /></a></div><h1>Web IDL (Second Edition)</h1><h2>W3C Editor’s Draft <em>11 May 2016</em></h2><dl><dt>This Version:</dt><dd><a href="http://heycam.github.io/webidl/">http://heycam.github.io/webidl/</a></dd><dt>Latest Version:</dt><dd><a href="http://www.w3.org/TR/WebIDL/">http://www.w3.org/TR/WebIDL/</a></dd><dt>Previous Versions:</dt><dd><a href="http://www.w3.org/TR/2012/CR-WebIDL-20120419/">http://www.w3.org/TR/2012/CR-WebIDL-20120419/</a></dd><dd><a href="http://www.w3.org/TR/2012/WD-WebIDL-20120207/">http://www.w3.org/TR/2012/WD-WebIDL-20120207/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110927/">http://www.w3.org/TR/2011/WD-WebIDL-20110927/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110712/">http://www.w3.org/TR/2011/WD-WebIDL-20110712/</a></dd><dd><a href="http://www.w3.org/TR/2010/WD-WebIDL-20101021/">http://www.w3.org/TR/2010/WD-WebIDL-20101021/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20081219/">http://www.w3.org/TR/2008/WD-WebIDL-20081219/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20080829/">http://www.w3.org/TR/2008/WD-WebIDL-20080829/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/">http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/</a></dd><dd><a href="http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/">http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/</a></dd><dt>Participate:</dt><dd>
+    <div class="head"><div><a href="http://www.w3.org/"><img src="https://www.w3.org/Icons/w3c_home" width="72" height="48" alt="W3C" /></a></div><h1>Web IDL (Second Edition)</h1><h2>W3C Editor’s Draft <em>19 May 2016</em></h2><dl><dt>This Version:</dt><dd><a href="http://heycam.github.io/webidl/">http://heycam.github.io/webidl/</a></dd><dt>Latest Version:</dt><dd><a href="http://www.w3.org/TR/WebIDL/">http://www.w3.org/TR/WebIDL/</a></dd><dt>Previous Versions:</dt><dd><a href="http://www.w3.org/TR/2012/CR-WebIDL-20120419/">http://www.w3.org/TR/2012/CR-WebIDL-20120419/</a></dd><dd><a href="http://www.w3.org/TR/2012/WD-WebIDL-20120207/">http://www.w3.org/TR/2012/WD-WebIDL-20120207/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110927/">http://www.w3.org/TR/2011/WD-WebIDL-20110927/</a></dd><dd><a href="http://www.w3.org/TR/2011/WD-WebIDL-20110712/">http://www.w3.org/TR/2011/WD-WebIDL-20110712/</a></dd><dd><a href="http://www.w3.org/TR/2010/WD-WebIDL-20101021/">http://www.w3.org/TR/2010/WD-WebIDL-20101021/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20081219/">http://www.w3.org/TR/2008/WD-WebIDL-20081219/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-WebIDL-20080829/">http://www.w3.org/TR/2008/WD-WebIDL-20080829/</a></dd><dd><a href="http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/">http://www.w3.org/TR/2008/WD-DOM-Bindings-20080410/</a></dd><dd><a href="http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/">http://www.w3.org/TR/2007/WD-DOM-Bindings-20071017/</a></dd><dt>Participate:</dt><dd>
               Send feedback to <a href="mailto:public-script-coord@w3.org">public-script-coord@w3.org</a> or <a href="https://www.w3.org/Bugs/Public/enter_bug.cgi?product=WebAppsWG&amp;component=WebIDL">file a bug</a> (<a href="https://www.w3.org/Bugs/Public/buglist.cgi?product=WebAppsWG&amp;component=WebIDL&amp;resolution=---">open bugs</a>)
             </dd><dt>Editors:</dt><dd><a href="http://mcc.id.au/">Cameron McCormack</a>, Mozilla Corporation &lt;cam@mcc.id.au&gt;</dd><dd>Boris Zbarsky, Mozilla Corporation &lt;bzbarsky@mit.edu&gt;</dd></dl><p class="copyright"><a href="http://www.w3.org/Consortium/Legal/ipr-notice#Copyright">Copyright</a> &copy; 2016 <a href="http://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a><sup>&reg;</sup> (<a href="http://www.csail.mit.edu/"><abbr title="Massachusetts Institute of Technology">MIT</abbr></a>, <a href="http://www.ercim.eu/"><abbr title="European Research Consortium for Informatics and Mathematics">ERCIM</abbr></a>, <a href="http://www.keio.ac.jp/">Keio</a>, <a href="http://ev.buaa.edu.cn/">Beihang</a>), All Rights Reserved. W3C <a href="http://www.w3.org/Consortium/Legal/ipr-notice#Legal_Disclaimer">liability</a>, <a href="http://www.w3.org/Consortium/Legal/ipr-notice#W3C_Trademarks">trademark</a> and <a href="http://www.w3.org/Consortium/Legal/copyright-documents">document use</a> rules apply.</p></div><hr /><script async="" src="file-bug.js"></script>
 
@@ -72,7 +72,7 @@
         report can be found in the <a href="http://www.w3.org/TR/">W3C technical
           reports index</a> at http://www.w3.org/TR/.
       </em></p><p>
-        This document is the 11 May 2016 <b>Editor’s Draft</b> of the
+        This document is the 19 May 2016 <b>Editor’s Draft</b> of the
         <cite>Web IDL (Second Edition)</cite> specification.
       
       Please send comments about this document to
@@ -176,6 +176,7 @@
             <li>Unicode characters: <span class="char">U+0030 DIGIT ZERO ("0")</span></li>
             <li>Extended attributes: <span class="xattr">[ExampleExtendedAttribute]</span></li>
             <li>Variable names in prose and algorithms: <var>exampleVariableName</var>.</li>
+            <li>Algorithms use <a href="https://tc39.github.io/ecma262/#sec-algorithm-conventions">the conventions of the ECMAScript specification</a>, including the ! and ? notation for unwrapping completion records.</li>
             <li>IDL informal syntax examples:
               <pre class="syntax">interface <i>identifier</i> {
   <em>interface-members…</em>
@@ -6513,6 +6514,17 @@ iframe.appendChild instanceof w.Function;  <span class="comment">// Evaluates to
               for Web pages.
             </p>
           </div>
+
+          <p>
+            Although at the time of this writing the ECMAScript specification does not reflect this,
+            every ECMAScript object must have an <dfn data-export="" id="dfn-associated-realm">associated <a class="external external" href="https://tc39.github.io/ecma262/#sec-code-realms">Realm</a></dfn>. The mechanisms
+            for associating objects with Realms are, for now, underspecified. However, we note that
+            in the case of <a class="dfnref" href="#dfn-platform-object">platform objects</a>, the
+            associated Realm is equal to the object's <a class="external external" href="https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-realm">relevant Realm</a>, and
+            for non-exotic function objects (i.e. not callable proxies, and not bound functions)
+            the associated Realm is equal to the value of the function object's [[Realm]] internal
+            slot.
+          </p>
         </div>
 
         <div id="es-type-mapping" class="section">
@@ -7889,12 +7901,12 @@ alert(b[4]);    <span class="comment">// This would alert "50".</span></code></p
               to an IDL <a class="idltype" href="#idl-promise">Promise&lt;<var>T</var>&gt;</a> value as follows:
             </p>
             <ol class="algorithm">
-              <li>Let <var>resolve</var> be the original value of <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects">%Promise%</a>.resolve.
+              <li>Let <var>resolve</var> be the original value of <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-promise-constructor">%Promise%</a>.resolve.
                 <div class="ednote"><div class="ednoteHeader">Editorial note</div>
                   <p>ECMAScript should grow a %Promise_resolve% well-known intrinsic object that can be referenced here.</p>
                 </div>
               </li>
-              <li>Let <var>promise</var> be the result of calling <var>resolve</var> with <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects">%Promise%</a>
+              <li>Let <var>promise</var> be the result of calling <var>resolve</var> with <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-promise-constructor">%Promise%</a>
                 as the <span class="esvalue">this</span> value and <var>V</var> as the single argument value.</li>
               <li>Return the IDL <a href="#idl-promise">promise type</a> value that is a reference to the
                 same object as <var>promise</var>.</li>
@@ -11860,8 +11872,8 @@ interface
                       <li>If the operation has a <a class="dfnref" href="#dfn-return-type">return type</a>
                         that is a <a href="#idl-promise">promise type</a>, then:
                         <ol>
-                          <li>Let <var>reject</var> be the initial value of <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects">%Promise%</a>.reject.</li>
-                          <li>Return the result of calling <var>reject</var> with <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects">%Promise%</a> as the
+                          <li>Let <var>reject</var> be the initial value of <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-promise-constructor">%Promise%</a>.reject.</li>
+                          <li>Return the result of calling <var>reject</var> with <a class="nocite external" href="https://tc39.github.io/ecma262/#sec-promise-constructor">%Promise%</a> as the
                             <span class="esvalue">this</span> object and the exception as the single
                             argument value.</li>
                         </ol>
@@ -14007,6 +14019,15 @@ C implements A;</code></pre></div></div>
               property name that is that identifier.
             </li>
           </ul>
+
+          <p>
+            Note that ECMAScript objects need not have
+            properties corresponding to <a class="dfnref" href="#dfn-constant">constants</a>
+            on them to be considered as <a class="dfnref" href="#dfn-user-object">user objects</a>
+            implementing <a class="dfnref" href="#dfn-interface">interfaces</a> that happen
+            to have constants declared on them.
+          </p>
+
           <p>
             A <dfn data-export="" id="dfn-single-operation-callback-interface">single operation callback interface</dfn> is
             a <a class="dfnref" href="#dfn-callback-interface">callback interface</a> that:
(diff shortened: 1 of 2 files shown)


Ms2ger:
I think I'd prefer to land #114 first, so this PR only has relevant changes.

Ms2ger:
Rebased at https://github.com/Ms2ger/webidl/tree/modernize-invoke

domenic:
Thanks, reset my branch to yours so this should be mergeable now.

Ms2ger:
Rebased again, since I keep bitrotting you

domenic:
Reset this PR to your commit, thanks.

Ping @bzbarsky to review?

index.html
```diff
@@ -14017,49 +14026,42 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
-            following the algorithm below.  The <dfn data-export="" id="dfn-callback-this-value">callback this value</dfn>
-            is the value to use as the <span class="esvalue">this</span> value
-            when a <a class="external" href="https://tc39.github.io/ecma262/#sec-iscallable">callable</a>
-            object was supplied as the implementation of a
-            <a class="dfnref" href="#dfn-single-operation-callback-interface">single operation callback interface</a>.
-            By default, <span class="esvalue">undefined</span> is used as the <a class="dfnref" href="#dfn-callback-this-value">callback this value</a>,
-            however this <span class="rfc2119">MAY</span> be overridden by other
-            specifications.
+            To <dfn>call a user object's operation</dfn>, given a <a class="dfnref" href="#idl-interface">callback interface type</a> value <var>value</var>, sometimes-optional operation name <var>opName</var>, list of IDL argument <var>idlarg</var><sub>0..<var>n</var>−1</sub>, and optional <dfn data-export="" id="dfn-callback-this-value">callback this value</dfn> <var>thisArg</var>, perform the following steps.
```
bzbarsky: "list of IDL argument" either needs an 's' at the end or the restoration of the "values" that went missing.  Preferably the latter.

index.html
```diff
@@ -14017,49 +14026,42 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
-                <li>If <a class="external" href="https://tc39.github.io/ecma262/#sec-ecmascript-data-types-and-values">Type</a>(<var>X</var>) is not Object, <a href="#ecmascript-throw" class="dfnref external">throw a <span class="estype">TypeError</span></a> exception.</li>
+                <li>If <var>thisArg</var> was not given, let <var>thisArg</var> be <span class="es-value">undefined</span>.</li>
+                <li>Let <var>O</var> be the ECMAScript object corresponding to <var>value</var>.</li>
+                <li>Determine the implementation of the operation, <var>X</var>:
+                  <ol>
+                    <li>If <var>value</var>'s interface is a <a class="dfnref" href="#dfn-single-operation-callback-interface">single operation callback interface</a> and <a class="external" href="https://tc39.github.io/ecma262/#sec-iscallable">IsCallable</a>(<var>O</var>) is true, then set <var>X</var> to <var>O</var>.</li>
+                    <li>Otherwise, <var>opName</var> must be supplied. Set <var>X</var> to <a class="external" href="https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(<var>O</var>, <var>opName</var>).</li>
```
bzbarsky: The handling of abrupt completions here is dodgy, but it was before your changes too, I guess...  We should really fix that up at some point.

index.html
domenic: Fixed and pushed a new commit.

BTW you may find reviewing the XML file slightly more enjoyable than reviewing the generated output.

index.html
```diff
@@ -14017,49 +14026,42 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
                   ECMAScript values, where <var>arg</var><sub><var>i</var></sub> is the result
                   of <a class="dfnref" href="#dfn-convert-idl-to-ecmascript-value">converting</a>
                   <var>idlarg</var><sub><var>i</var></sub> to an ECMAScript value.
                 </li>
-                <li>Let <var>script</var> be the <a class="dfnref" href="#dfn-callback-context">callback context</a> associated with <var>V</var>.</li>
-                <li>Push <var>script</var> on to the <a class="dfnref external" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html">stack of incumbent scripts</a>. <a href="#ref-HTML">[HTML]</a></li>
+                <li>Let <var>execution context</var> be <a class="external" href="https://tc39.github.io/ecma262/#sec-getfunctionrealm">GetFunctionRealm</a>(<var>X</var>)'s <a class="dfnref external" href="https://html.spec.whatwg.org/multipage/webappapis.html#concept-realm-settings-object">settings object</a>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#realm-execution-context">realm execution context</a>.</li>
```
bzbarsky: No, this doesn't work, I think.  The reason it doesn't work is that you're too late: you may already have thrown an exception from the `Get` above, and that needs to report the exception, and that needs the entry settings stuff, because that's where exceptions get reported, right?

I put together two interesting testcases: <http://web.mit.edu/bzbarsky/www/testcases/settings-objects/entry-settings-1.html> and <http://web.mit.edu/bzbarsky/www/testcases/settings-objects/entry-settings-2.html> -- on both you need to click the "click me" text .  The first one uses an element from window A, adds a listener that is an object created from window B, whose `handleEvent` property is a function from window C that throws.  The second one uses an element from window A, adds a listener that is an object created from window B, whose `handleEvent` property has a getter that is a function from window C that throws.

In Firefox, both cases use window B as the place to fire error events, and entry settings object. 

In Chrome, they both seem to use window A as the place to fire error events (because of where the listeners are attached, iirc how this works in Chrome); not sure whether that precisely corresponds to entry settings in Chrome.

In Safari, the behavior is the same as in Chrome (uses window A) for the first testcase and the second testcase doesn't fire any error listeners.  Again, not sure what the actual entry settings thing is.

In Edge, the first testcase doesn't fire error listeners afaict and the second one fires them in window B.   Again, not sure what the actual entry settings thing is.

This spec as written (with your changes) would fire listeners in window C on the first testcase and has undefined behavior on the second one, as far as I can tell.

bzbarsky: Oh, and also, what makes you think `GetFunctionRealm` won't just assert here?  Nothing guarantees that the `Get` returned a callable.

index.html
```diff
@@ -14017,49 +14026,42 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
+                    <li>Set <var>R</var> to <a class="external" href="https://tc39.github.io/ecma262/#sec-call">Call</a>(<var>X</var>, <var>thisArg</var>, <var>arg</var><sub>0..<var>n</var>−1</sub>).</li>
                   </ol>
                   And then, whether or not an exception was thrown:
                   <ol>
-                    <li>Pop <var>script</var> off the <a class="dfnref external" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html">stack of incumbent scripts</a>.</li>
+                    <li>Decrement <var>execution context</var>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#entrance-counter">entrance counter</a> by one.</li>
+                    <li>Remove <var>execution context</var> from the <a class="external" href="https://tc39.github.io/ecma262/#execution-context-stack">JavaScript execution context stack</a>.</li>
                     <li>If an exception was thrown, end these steps, and allow it to propagate.</li>
```
bzbarsky: I'm not sure that works right, though it's preexisting; this goes back to where most invokers want to report the exception if there is one, but the "where to report it" location is ... what exactly?  Once we have propagated out of here, we've taken down our entry settings, which iirc is the place you're supposed to report exceptions.

index.html
```diff
@@ -14082,34 +14084,24 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
-                <li>Let <var>script</var> be the <a class="dfnref" href="#dfn-callback-context">callback context</a> associated with <var>V</var>.</li>
-                <li>Push <var>script</var> on to the <a class="dfnref external" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html">stack of incumbent scripts</a>. <a href="#ref-HTML">[HTML]</a></li>
+                <li>Let <var>O</var> be the ECMAScript object corresponding to <var>object</var>.</li>
+                <li>Let <var>execution context</var> be <var>object</var>'s <a class="dfnref" href="#dfn-callback-context">callback context</a>'s <a class="dfnref external" href="https://html.spec.whatwg.org/multipage/webappapis.html#settings-object">settings object</a>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#realm-execution-context">realm execution context</a>.</li>
+                <li>Increment <var>execution context</var>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#entrance-counter">entrance counter</a> by one.</li>
```
bzbarsky: Not sure this is right either.  The callback context is supposed to be the incumbent settings, but here it's being set up as the entry settings.

index.html
```diff
@@ -14128,27 +14120,22 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
-            <li>Let <var>script</var> be the <a class="dfnref" href="#dfn-callback-context">callback context</a> associated with <var>V</var>.</li>
-            <li>Push <var>script</var> on to the <a class="dfnref external" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html">stack of incumbent scripts</a>. <a href="#ref-HTML">[HTML]</a></li>
+            <li>Let <var>O</var> be the ECMAScript object corresponding to <var>object</var>.</li>
+            <li>Let <var>W</var> be the result of <a class="dfnref" href="#dfn-convert-idl-to-ecmascript-value">converting</a> <var>value</var> to an ECMAScript value.</li>
+            <li>Let <var>execution context</var> be <var>object</var>'s <a class="dfnref" href="#dfn-callback-context">callback context</a>'s <a class="dfnref external" href="https://html.spec.whatwg.org/multipage/webappapis.html#settings-object">settings object</a>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#realm-execution-context">realm execution context</a>.</li>
```
bzbarsky: Similar issue here.

index.html
```diff
@@ -14193,15 +14176,17 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
                   </ol>
                 </li>
                 <li>Truncate <var>values</var> to have length <var>count</var>.</li>
-                <li>Let <var>script</var> be the <a class="dfnref" href="#dfn-callback-context">callback context</a> associated with <var>V</var>.</li>
-                <li>Push <var>script</var> on to the <a class="dfnref external" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html">stack of incumbent scripts</a>. <a href="#ref-HTML">[HTML]</a></li>
+                <li>Let <var>execution context</var> be <a class="external" href="https://tc39.github.io/ecma262/#sec-getfunctionrealm">GetFunctionRealm</a>(<var>F</var>)'s <a class="dfnref external" href="https://html.spec.whatwg.org/multipage/webappapis.html#concept-realm-settings-object">settings object</a>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#realm-execution-context">realm execution context</a>.</li>
+                <li>Increment <var>execution context</var>'s <a class="external" href="https://html.spec.whatwg.org/multipage/webappapis.html#entrance-counter">entrance counter</a> by one.</li>
```
bzbarsky: OK, so this gets the entry settings right, I think.  But not the incumbent...

domenic: Incumbent is meant to be in a follow-up PR. But it sounds like from the rest of the comments there's still a lot of work to do to get entry settings right too... will try to work on it :-/

index.html
domenic: Hmm, I must be missing what the connection is between entry settings object and exception handling? The current spec is very vague on this. EventTarget says "report the exception". And the definition of "report an exception" just says "using the global object specified by the relevant script's settings object as the target." I have no idea what the "relevant script" is. I guess that is https://github.com/whatwg/html/issues/958. I guess you're assuming it's derived from the entry settings object. Maybe that was agreed upon in some earlier thread I forgot or was before my time.

To match Chrome's behavior, I guess we would have event dispatch specifically report errors to the EventTarget object's relevant settings object. Maybe it would be ideal to have every caller of "report an exception" explicitly supply a global object or settings object to dispatch at anyway.

But that of course still doesn't solve the problem of needing to set up the entry settings object before the Get() call.

One solution is to do here what I did below for "get a user object's attribute value": use _value_'s callback context's settings object's realm execution context. I think that works? I guess it would give B in both cases?

Another solution is to acknowledge that all objects have a realm, not just functions, and then use _O_'s realm. No spec acknowledges this currently, but we could just say "_O_'s realm" and add a note stating our assumption here. That would, if I understand correctly, use B in both cases also.

Thoughts?

domenic: Per https://github.com/whatwg/html/issues/958#issuecomment-218615098 it seems more likely to me that "report the exception" should use some concept of "incumbent script", not entry?

domenic: Wait, no, that doesn't work. Ugh.

Check me on this: no matter what, "dispatch an event" wants to report errors for the HTML script corresponding to the event handler function's, or `handleEvent` getter function's, [[ScriptOrModule]]. That script is where it will get the filename/line number/column number.

In both cases in your examples, this is window A: `new Function` uses the currently active script-or-module to fill the created function's [[ScriptOrModule]]. The currently active script-or-module is your main-document `<script>` tag. This makes sense: the actual line of code declaring those functions is inside that window-A `<script>` tag.

(If you had created examples where the function was declared inside window B or window C, instead of using window B or window C's `Function` constructor to create them while inside window A, then it would be a different story. But it would still make sense: a function declared in window B's `srcdoc`'s `<script>` should get errors reported using that script's filename/line number/column number.)

However, the question is which global it dispatches `"error"` events to.

Two approaches:

- It grabs the global of that script to fire `error` events toward. E.g., we define "the script of a callback value or callback interface value" and the "inner invoke" algorithm says "If this throws an exception, report the exception to the script of _callback_."

  This is window A in both your examples. (But, if we modified them to declare the function in window B or C, it would be window B or C.)

- We set up something to be the entry global before doing the Get(). We dispatch the error event to the entry global at the time of the error. E.g., we modify "report an exception" to take separate script and global parameters. Then in the "inner invoke" algorithm we say "If this throws an exception, report the exception given the script of _callback_ and the entry global at the time of invocation."

  If we do this, we need to figure out what entry global to set up (assume every object has a realm? then we can get B).

The first of these approaches sounds better to me, to be honest.

index.html
domenic: So the problem here is I don't know what to set up as the entry settings. There's no obvious function to get a realm from. Should we assume every object has a realm, and use that? Not sure what to do here.

bzbarsky:
> Hmm, I must be missing what the connection is between entry settings object and exception handling?

See https://github.com/whatwg/html/issues/958#issuecomment-218624203

> Maybe that was agreed upon in some earlier thread I forgot or was before my time.

Before your time.

> To match Chrome's behavior, I guess we would have event dispatch specifically report errors to the EventTarget object's relevant settings object.

I don't think we should match Chrome's behavior.  I think Chrome should align with what the spec at least used to say before people started messing with it and what everyone agreed we should do the last time this was discussed.  Specifically, in the simple case of a function callback the error is reporter to the global of that function.  The hard cases can then be figured out, as long as the simple case works correctly.

> One solution is to do here what I did below for "get a user object's attribute value": use value's callback context's settings object's realm execution context.

The "callback context" thing was added to propagate the incumbent, right?  I don't think it's the right thing.

> I guess it would give B in both cases?

No, as far as I can tell it would give D: the window that made the addEventListener call.

> Another solution is to acknowledge that all objects have a realm, not just functions, and then use O's realm. 

That's basically what Firefox does.

> new Function uses the currently active script-or-module to fill the created function's [[ScriptOrModule]]. The currently active script-or-module is your main-document <script> tag. 

Hrmm.  That seems ... weird.  I think people using `new someWindow.Function` would expect that to act just like a function declared in someWindow.  It claims to be instanceof that window's `Function`, for sure.  

But looking at the actual spec at http://www.ecma-international.org/ecma-262/6.0/#sec-function-constructor it gets the "active function object".  This is defined as:

> The value of the Realm component of the running execution context is also called the current Realm. The value of the Function component of the running execution context is also called the active function object.

and in this case it gets set up by http://www.ecma-international.org/ecma-262/6.0/#sec-built-in-function-objects-construct-argumentslist-newtarget which references http://www.ecma-international.org/ecma-262/6.0/#sec-built-in-function-objects-call-thisargument-argumentslist which sets the active function object to the built-in function being called/constructed.    Of course there is no script-or-module in any of that stuff...

> This is window A in both your examples. (But, if we modified them to declare the function in window B or C, it would be window B or C.)

I think this would confuse people a lot.  People expect functions created with `new someWindow.Function` to act like functions defined in that window, from everything I've seen.

domenic:
> Of course there is no script-or-module in any of that stuff...

Right, that's in the modern spec at https://tc39.github.io/ecma262/#sec-functioninitialize

> I think this would confuse people a lot. People expect functions created with new someWindow.Function to act like functions defined in that window, from everything I've seen.

So what filename should be reported when reporting these exceptions?

bzbarsky:
> So what filename should be reported when reporting these exceptions?

Ah, hmm.  Having that be the caller of the Function constructor does make some sense.  OK, but just to check my understanding: if code in window1 does `var f = new window2.Function("stuff")` then `f` has window1 as its script-or-module and window2 as its Realm?

If so, I probably need to go back and think really carefully about all the cases where you propose using script-or-module...

domenic:
> if code in window1 does var f = new window2.Function("stuff") then f has window1 as its script-or-module and window2 as its Realm?

Yes, exactly.

As I said, I see two alternatives here. To rephrase them slightly:

- "report an exception" takes a script. It uses that script's filename, and dispatches error events to that script's global.
- "report an exception" takes a script and a global. It uses that script's filename, but dispatches error events to the supplied global, which may not equal the script's global.

In our example, the first alternative dispatches error events to `window1`, whereas the second allows the possibility of dispatching them to `window2` if we thread everything through correctly.

domenic:
OK, but there's also this tricky case, as you pointed out:

```js
el.addEventListener("click", el.cloneNode); // purposefully not bound
```

When "inner invoke" invokes the IDL callback created from `el.cloneNode`, which of course throws an exception due to invalid `this`, there's no script possible to pass to "report an exception".

I can't imagine a relevant filename for this, unless we wanted to try to use the callback context, and make it capture the "incumbent script" (i.e. the script that did `addEventListener`). Probably it makes more sense to say that there's no filename that's responsible for throwing, since it's really browser code that's doing the `if (isNotCorrectlyBranded(this)) { throw new TypeError(); }`.

So it seems like the approach of having "report an exception" only take a script, and derive the correct global from that script, won't work, unless we capture the incumbent script. Which sounds pretty different from anything anyone is doing, so probably we shouldn't do that...

bzbarsky:
Or more precisely, it has whatever the window1 script was that called the Function ctor as script-or-module.  But the point is that the [[Realm]] of the [[ScriptOrModule]] of the function involved doesn't match the [[Realm]] of the function itself (and this is the only way to produce this situation!), so it really starts to matter how we go about getting to the Realm.

For what it's worth, at least in Firefox in this case the function looks in all ways like it came from window2, except that the "filename" string is something like "urlofWindow1 line N > Function".  Note that this is NOT the same thing as "urlofWindow1", which is what I think you're saying the current setup in the spec does, and is a lot more useful...

Actually, I just went ahead and put up a testcase so we can see what UAs do here: <http://web.mit.edu/bzbarsky/www/testcases/javascript/newFunction-parent.html>.  The output for the top stack frame is the following in various browsers:

* Firefox: `anonymous@http://web.mit.edu/bzbarsky/www/testcases/javascript/newFunction-parent.html line 5 > Function:1:7`
* Chrome: `at eval (eval at onload (http://web.mit.edu/bzbarsky/www/testcases/javascript/newFunction-parent.html:5:13), <anonymous>:2:7)`
* Safari: `anonymous`
* Edge: `at Function code (Function code:1:1)

none of those have `"http://web.mit.edu/bzbarsky/www/testcases/javascript/newFunction-parent.html"` as the filename, and Chrome and Firefox have something much more useful instead.  I don't think we really want to be using the script-or-module of the caller of `new Function` to set the filename if it would require all of them to claim that this error occured in `"http://web.mit.edu/bzbarsky/www/testcases/javascript/newFunction-parent.html"`.

domenic:
Do you have any suggestions for spec language that could usefully specify the filename? Maybe instead of trying to pass through a script object, we should just leave it super vague ("relevant filename, line number and column number" with a note explaining that this is purposefully vague)?

bzbarsky:
> "report an exception" takes a script. It uses that script's filename

Uses it for what?  There is a filename associated with the exception itself, or should be... (in Firefox there isn't one for non-Error exceptions, but that's a bug we should fix).  Note that the filename involved in exceptions is where the exception was thrown/created, not where script was entered anyway, whereas reporting happens at the entry point.

Are we talking about the filename in the error event, basically?

> el.addEventListener("click", el.cloneNode); // purposefully not bound

That will not do the right thing, because event listeners are called with the event target as `this`, so this will clone the event target.  You want something like:

    el.addEventListener("click", document.write);

to make this very simple.  I just tried that, and it reports no useful filename in the console in Safari or Firefox, for what it's worth.  It does report one in Chrome, but that filename doesn't match the `.filename` of the error event that gets fired or the second arg passed to the onerror handler if there is one.  That's set to empty string in Firefox and Chrome.  It's undefined in Safari in this case.  Is this the filename we're talking about here?

> I can't imagine a relevant filename for this

I agree.

I think being vague about precisely what the filename/line/column are is fine for now.  We can revisit it once people get closer to standardizing the .stack of Error instances, since that will involve solving basically this problem....

domenic:
Yeah, I meant the filename (and line and column number) exposed as properties on the error event object. The current spec seems oriented around filling those in, based on the "relevant script".

If we can let those be vague, then I think I know how to address the rest of the issues you raised. Let me finish my workout, and in fortyish minutes I'll post a plan for fixing both this PR and report an exception (plus all callers of report an exception).

domenic:
OK, so here's the plan. Let me know what you think, and I can try to implement it tomorrow.

### Step 1: Fixing this PR

We note that every object (not just platform object, but every object) has an associated Realm. For functions it coincides with the value returned by GetFunctionRealm(); for other objects it's not yet specified.

The four "call user code" operations here all use the corresponding object's Realm to set up the entry stuff. (I.e., get the object's realm's settings object's realm execution context, increment its entrance counter, and push it onto JS execution context stack.)

For the case of "call a user object's operation", which is more complicated since it can potentially invoke user code twice, we potentially set up the entry stuff twice: once for the Get(), using the object, and another time for the Call(), using the function. So if you have an object whose realm is A with a handleEvent property that is a function whose realm is B, invoking the getter will happen with a different entry global than invoking the handleEvent.

(We could change that to use the object in both cases, but that seems weird to me. I'm flexible though.)

This should set up the correct entry settings object per your earlier PR comments.

### Step 2: Fix HTML's "report the exception"

"Report the exception" gets modified to take an exception object, a global object, a filename/line number/column number tuple, and an optional script.

The optional script is necessary to perform muted error checks. (See ["report an error"](https://html.spec.whatwg.org/multipage/webappapis.html#report-the-error).) If it's not present then we assume the error is un-muted. We no longer derive the filename from the script, to allow the better results you mention.

We probably consolidate "report the exception" and "report an error".

### Step 3: Add "guarded" user code invocation to Web IDL

(This would be a follow-up PR performed after step 2, not as part of this step 1 PR.)

We add a new optional flag to all of these four user-code-calling operations: "guarded", default unset. The flag is not allowed to be set if the attribute or operation in question returns a promise type.

If the guarded flag is set, then when an exception is thrown during any of:
- converting arguments
- Get()/Set()/Call()
- converting return values

Then we do "report the exception" on the thrown exception, passing in the exception, the entry global (i.e. the global we went to all this trouble to make the entry global), "some relevant filename/line number/column number" (purposefully underspecified), and the callback's script-or-module.

Note that this means that in our `var f = new window2.Function("stuff")` case, we use the script of this `var` declaration for the muted errors check. This seems correct.

(I'm not sure if we should immediately call out to "report the exception" before un-entryifying the appropriate global, or if we should just save the appropriate global, entryify it, run possibly-throwing steps, unentryify it, and then call out to "report the exception" with the appropriate global. Maybe they are equivalent. I will assume we should do it before un-entryifying since that seems to be the direction your earlier PR comments were hinting at.)

### Step 4: Fix all call sites of "report an exception"

Very few places should directly call "report an exception". Most should switch over to calling one of IDL's invoke-user-code operations with the new "guarded" flag set.

Here are the call sites I know of to "report the exception":

#### HTML

- Custom element callback reactions should switch to using the guarded flag.
- Custom element upgrade reactions will need a bit of care; will have to think on it a bit more.
- Failed module resolution should report to the script's settings object's global, which will be the same as the initiating global for the fetch (except for module workers, where it will be the inner worker global, not the outer global)
- creating a module script will report to the script's settings object's global
- running a classic/module script will report to the entry settings object set during running the script
- EnqueueJob will report to the entry settings object set during job execution
- Timer functions, when supplied a string: their CSP checks will "report the exception" if CSP blocks string compilation. This will need a bit of care to capture the active script at the time setTimeout() et al. are called so it can be used for muted-errors checks later. It already captures the current realm, which it can use for error reporting.
- Timer functions, when supplied a function: these currently call IDL "invoke" but seem to omit "report the exception." That seems like a straight bug. Regardless, it should switch to using the guarded flag.
- Animation frame callbacks should switch to using the guarded flag.
- Custom element errors during parsing will be reported to the document's Window. Might need to thread through the correct script here to perform "muted errors" checks.

#### DOM

- Calling handleEvent should switch to using the guarded flag.
- Calling mutation observer callbacks should switch to using the guarded flag.

#### Pretty much everywhere else

- Should switch to using the guarded flag.

bzbarsky:
> The current spec seems oriented around filling those in, based on the "relevant script".

Ah.  The "relevant script" there is the one throwing the exception (but even then, that's not true; if an Error is created at one place but thrown at another, it's the creation point that gets reflected in the filename in browsers in practice).  It has absolutely nothing to do with incumbent, entry, where the exception is reported, etc.

> (We could change that to use the object in both cases, but that seems weird to me. I'm flexible though.)

Fwiw, Firefox uses the same entry settings for the Get and the call to the gotten thing.  Setting up this stuff is not that cheap, so I would rather not do it twice, honestly, unless there are strong reasons to do so.

> "Report the exception" gets modified to take an exception object, a global object, a filename/line number/column number tuple, and an optional script.

s/exception object/exception value/, right?

> We add a new optional flag to all of these four user-code-calling operations: "guarded", default unset.

It should probably be default set.  See the categorization of the various places in the platform that invoke this stuff as of a few years ago in https://www.w3.org/Bugs/Public/show_bug.cgi?id=17713#c16 -- pretty much everything wants to report the exception.

> I'm not sure if we should immediately call out to "report the exception" before un-entryifying the appropriate global

I'm not sure it matters in practice, if we're explicitly passing in the global anyway.

> Timer functions, when supplied a string: their CSP checks will "report the exception" if CSP blocks string compilation.

I don't think that should be happening.  There should be CSP reporting, sure, but no firing of "error" events in this situation, right?  I haven't tested what UAs do in detail here, but Gecko at least will simply return 0 from the setTimeout call, not schedule anything, and report the CSP violation if the policy asked for reports.

> This will need a bit of care to capture the active script at the time setTimeout() et al. are called so it can be used for muted-errors checks later.

I don't think it's worth worrying about, honestly.  The muted error checks are really important for syntax errors.  For everything else they're mostly not that important.  Certainly I don't think UAs go out of their way to do anything complicated for the string setTimeout case here, and I don't think the spec should either.

The plan sounds good in general!  Thank you for thinking through this.

domenic:
OK, @bzbarsky, this is ready for you to take a look at. https://rawgit.com/domenic/webidl/07ffceee1adb2e0c03ec1fc8a0c8eda94e4da219/index.html#es-user-objects should be a viewable version of the output.

Some notes:

- I went the extra distance and re-did all the abrupt completion stuff, in preparation for fixing all the "report an exception" stuff. This mostly seems better. Although it's sometimes a bit strange that we have ES completion values containing IDL values, I think that's how it has to be.
- I haven't tried tightening the promise language yet (e.g. introducing to ES a "CreateRejectedPromise" operation). We'll leave that for "some day".
- This does not yet fix incumbent settings objects. My current queue on all this mess looks like:
  - Get this PR merged, in order to fix entry settings objects
  - Get https://github.com/whatwg/html/pull/1189 merged, in order to lay the groundwork for fixing incumbent settings objects
  - Do a new Web IDL PR that adds a few steps in order to fix incumbent settings objects
  - Fix https://github.com/whatwg/html/issues/855 to take care of incumbent/entry settings objects in places like StructuredClone
  - Fix "report an exception" as discussed in steps 2-4 above.
  - Go triage and test all uses of incumbent and entry in various specs to see if we can get rid of them in favor of current.

Ms2ger:
ES doesn't use `!` for `IsCallable()`; I think it's better to follow that. (Especially because in this case my first interpretation was that it was a negation.)

domenic:
There's a general movement in ES toward using ! everywhere, though. The current usage conflates completion values and ES values, and we'd like to move away from that. I agree it is easy to misread :(.

bzbarsky:
My personal feeling, from trying to read the ES spec draft, is that the single-char thing is perhaps taking brevity a bit too far (it's easy to miss entirely) and '!' is a particularly poor choice of character.  The right answer might be to change what ES is doing to use more-obvious and less-confusing sigils here.

bzbarsky:
For example, "unbox" might be better than '!' to get the meaning across.  Or something.

domenic:
I'd encourage you to open an issue there to try to change it across the ecosystem, but we've already adopted it pretty wide and far, so it'll be an uphill slog. I think we should stick with the established conventions here, until such a time as such a change is agreed upon.

bzbarsky:
If we stick with it here, we should linkify the '!' at the very least.

domenic:
Nobody else does, but if that's what it takes to get this PR merged, I'll do it.

bzbarsky:
My concern is that the IDL spec needs to be readable/understandable by people designing web APIs.  Using this bizarre syntax with no context does NOT help that...

domenic:
I'm not sure people designing web APIs need to understand the algorithms involved here; they contain a lot of obtuse stuff we've spent several months hashing out. Rather, they need to understand the publicly-exposed "spec API", which is the four `<dfn>`s. _Binding implementers_ need to understand the notation, sure.

bzbarsky:
That's fair, but in practice I've had all sorts of non-binding implementors (both spec authors and people trying to review specs) come to me with questions about the exact behavior some part of IDL produces...

I wonder whether we can do the linkification automagically (find all " ! " and " ? " with the spaces, and replace them with linkified versions) as a stopgap...

Anyway, I filed https://github.com/tc39/ecma262/issues/568

domenic:
I can linkify and add the intro paragraph for ! and ? as I did in #121. Are there other code review comments to be had, or does this LGTY otherwise?

bzbarsky:
I need to read through the details, still.  Probably on Monday morning at this point...

domenic:
Thanks.

Oh, I just realized. Instead of pushing/incrementing and popping/decrementing, I should probably directly call "prepare to run script" and "clean up after running script".

The difference is that "clean up after running script" would run global script clean-up jobs and perform a microtask checkpoint, which seems desirable.

I'll push another commit making that change (and the ?/! linkification). Let me know if I'm wrong though and it should not do that.

index.xml
```diff
@@ -6368,6 +6390,17 @@ iframe.appendChild instanceof w.Function;  <span class='comment'>// Evaluates to
...
+
+          <p>
+            Although at the time of this writing the ECMAScript specification does not reflect this,
+            every ECMAScript object must have an <dfn data-export=''
+            id='dfn-associated-realm'>associated <a class='external'>Realm</a></dfn>. The mechanisms
+            for associating objects with Realms are, for now, underspecified. However, we note that
+            in the case of <a class='dfnref' href='#dfn-platform-object'>platform objects</a>, the
+            associated Realm is equal to the object's <a class='external'>relevant Realm</a>, and
+            for function objects the associated Realm is equal to the Realm returned by <a
+            class='external'>GetFunctionRealm</a>.
```
bzbarsky: I thought about pretty hard, and I am really not happy with the way this ends up working for proxies.

If the intent is to capture "the realm of the code that will actually run" then defining the realm in this way for a callable proxies is pretty suboptimal.  If you just get your hands on any callable from some realm, you can pretend like code from that realm will run while actually running code of your own choosing, by simply implementing the "apply" trap.

I would much rather we ended up handling a callable proxy the same way we handle a function that (possibly) ends up calling another callable inside of itself instead of trying to peek through it like this.

Changing GetFunctionRealm doesn't seem viable given its uses in ArraySpeciesCreate  and GetPrototypeFromConstructor.  But that means maybe we don't want GetFunctionRealm here.

Note that this is primarily a worry for me because I associate certain security guarantees with this stuff.  If none of the uses involve security checks, then it may not matter as much.

index.xml
```diff
@@ -14044,66 +14181,126 @@ C implements A;</x:codeblock>
...
-                  </ol>
-                </li>
-                </ol>
-                </li>
-                <li>If the callback function’s return type is <span class='idltype'>void</span>, return.</li>
-                <li>
-                  Return the result of <a class='dfnref' href='#dfn-convert-ecmascript-to-idl-value'>converting</a>
-                  <var>R</var> to an IDL value of the same type as the callback function’s return type.
-                </li>
+                <li>If the callback function's return type is <span class='idltype'>void</span>,
```
bzbarsky: This was here before, but it might use some explanation in a note...

The only way to end up with a non-callable here is to be dealing with a callback function which came from an attribute setter marked `[TreatNonObjectAsNull]`.  And in that case we just want to no-op the function call.

index.xml
```diff
@@ -14044,66 +14181,126 @@ C implements A;</x:codeblock>
...
+            <li>Let <var>realm</var> be <var>F</var>'s <a class='dfnref'
+            href='#dfn-associated-realm'>associated Realm</a>.</li>
+
+            <li>Let <var>settings</var> be <var>realm</var>'s <a class='dfnref external'
+            href='https://html.spec.whatwg.org/multipage/webappapis.html#concept-realm-settings-object'>settings
+            object</a>.</li>
+
+            <li><a>Prepare to run script</a> with <var>settings</var>.</li>
+
+            <li>If <var>value</var>'s interface is not a <a class='dfnref'
```
bzbarsky: This looks like copy/paste fail.  there is no "`value`'s interface" and no `O` here.  This entire step is not needed; we already set `thisArg` to undefined if not provided, which is the behavior we want for the callback function case.

index.xml
```diff
@@ -14044,66 +14181,126 @@ C implements A;</x:codeblock>
...
+            <li>Let <var>i</var> be 0.</li>
+
+            <li>Let <var>count</var> be 0.</li>
+
+            <li>
+              While <var>i</var> &lt; <var>n</var>:
+
               <ol>
-                <li>If the callback function has a <a class='dfnref' href='#dfn-return-type'>return type</a> that is a <a href='#idl-promise'>promise type</a>, then:
+                <li>If <var>arg</var><sub><var>i</var></sub> is the special value “missing”, then
```
bzbarsky: There is no handling of "missing" in the "call a user object's operation" algorithm above.  That looks like a preexisting bug, but we should fix it while we're here.

bzbarsky:
Thank you for linkifying the '!'/?' and for putting up a generated version.  Made this much more pleasant!

The one remaining concern I have is the microtask checkpoint bit.  I think morally that's the right thing to do (and have been meaning to make that change in Firefox; it's blocked by some other issues in how we handle microtasks), but I'm not sure what UAs do right now.  For example, during event dispatch, do promise callbacks run after every event listener?

I tried creating a testcase for that:

    <body>
    <script>
    function logger(i) {
      console.log("Call for " + i);
      Promise.resolve().then(function() { console.log("Promise callback for " + i) });
    }
    document.body.addEventListener("click", logger.bind(undefined, 1));
    document.body.addEventListener("click", logger.bind(undefined, 2));
    document.documentElement.addEventListener("click", logger.bind(undefined, 3));
    console.log("before click");
    document.body.click();
    console.log("after click");
    </script>

and all of Firefox, Chrome, Safari log:

    before click
    Call for 1
    Call for 2
    Call for 3
    after click
    Promise callback for 1
    Promise callback for 2
    Promise callback for 3

which is not the output I was expecting if any of them performed a microtask checkpoint there... though maybe I misunderstand what microtask checkpoints do.

index.xml
domenic: I'll leave functions as un-specified as normal objects, then, and remove this clause about GetFunctionRealm.

index.xml
domenic: That's very helpful; I had no idea how we could possibly reach this.

index.xml
domenic: I've never seen an algorithm actually use "missing", so I wasn't sure what to make of this, but I can easily sync the two.

index.xml
bzbarsky: Well.  For the honest-to-God Function case there is an obvious answer that everyone agrees on and we should spec that.

For the other cases, we should try to see what various engines do in practice and see if we can get feedback from other implementors as to what they think should happen.

index.xml
bzbarsky: Ah.  "missing" is what you get as the IDL value when `undefined` is passed to an optional argument with no default value.  It's handled in ES-to-IDL conversion for method calls, and for calls into a callback function; just not for the user object operation case.

domenic:
Updated in response to code review comments at https://rawgit.com/domenic/webidl/modernize-invoke/index.html#es-user-objects

Regarding microtasks... hmm. "clean up after running script" says:

> If the JavaScript execution context stack is now empty, perform a microtask checkpoint. (If this runs scripts, these algorithms will be invoked reentrantly.)

I think in your test case the execution context stack is not empty. I think it looks like this at the time the handler executes:

1. **Bottom:** realm execution context (pushed by executing the `<script>`)
2. document.body.click's execution context ??
3. realm execution context (pushed by "invoke")
4. **Top**: function execution context for `logger.bind(undefined, 1)`.

That is, when invoke finishes, the execution context stack still contains 1 and 2. (Or at least 1... I'm having trouble remembering whether host-defined functions have execution contexts, so I'm not sure about 2.)

I created a similar test at http://jsbin.com/gukipafemo/edit?html,console,output where instead of using document.body.click() from inside a script, you manually click the button--so at the time the "invoke" finishes, you should have an empty stack. In this case Chrome works "as expected" with the interleaved calls, although Edge and Firefox do not. This reminds me of the cross-browser investigations in https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/ where Jake concluded that only Chrome was following the spec in these sort of tricky edge cases.

index.xml
domenic: Ah, where "honest-to-God Function" is not a proxy or a bound function. OK, yeah, I'll add back in a note about it matching [[Realm]] for those.

bzbarsky:
Ah, I had missed the "if the stack is now empty" thing.

I think the difference in behavior between the dispatch from click and dispatch from script is pretty weird.  In particular, it means that if a capturing listener grabs the event and redispatches it you get different behavior than if you just allow the event to propagate.... :(

Relying on "stack is empty" also involves other problems if a UA doesn't want `alert` and company to completely block the UI (and yes, I know the spec model of alert is that it blocks everything; this is user-hostile and broken): suddenly your microtasks stop running correctly while the alert is up.

It seems like we need to sit down and figure out how this stuff should actually work in the real world of browsers, which doesn't match the simplistic world of the spec and will keep diverging from it (e.g. iOS Safari recently moved to more-nonblocking alerts), as I understand.

domenic:
Hmm, to me the idea that microtasks run when the stack becomes empty has always been quite intuitive. Sure, it can cause edge case weirdness like the capturing event listener you mention. But it makes sense from a microtask perspective...

As for alert, I agree that figuring out how to modernize those is a good project, but it seems like a separate one.

Is there anything else to do in this PR, though?

bzbarsky:
My point is that while the alert is up the stack is not empty.  This can be the state of things for a _long_ time.

For this PR, I'm not sure there's anything else to do, but it sounds to me like in general the microtask setup is broken and needs fixing.  :(

domenic:
Can we merge this PR then?

bzbarsky:
I think so, but please make sure issues are filed on the microtask bit.  And it might be worth adding a note that that part is being sorted out...

domenic:
Filed https://github.com/whatwg/html/issues/1294.

domenic:
Ping to merge. I think we should wait on the outcome of discussions in whatwg/html#1294 before adding any notes, and if we do, it'd be in HTML not Web IDL.

bzbarsky:
Oh, you don't have commit rights to this repo?  That's moderately dumb.  :(

I agree the note needs to be in HTML, but I think it's pretty important to add ASAP.  Otherwise we get in a cycle of some UAs claiming "but the spec says X, so we must do X" and other UAs saying "but it says this based on bogus premises that aren't actually true except maybe in your particular implementation" and then we end up with things like <https://www.w3.org/Bugs/Public/show_bug.cgi?id=11960>.  Been there, done that, don't want to end up there again.

domenic:
In that case it'd be good to clarify things over in that issue, as right now it sounds like nobody quite agrees with you that there's a problem...

PR #114, opened 2016-05-04
Update links to ES2017.

Ms2ger:
jorendorff's unofficial HTML version already redirects there, so there's no
point in linking to that URL.

This commit only updates the links; more work is needed to ensure all
references are still correct in ES2017.

### DIFF
--- a/index.xml
+++ b/index.xml
@@ -54,91 +54,91 @@
       <links>
         <term name='Unicode scalar value' class='external' href='http://www.unicode.org/glossary/#unicode_scalar_value'/>
         <term name='Unicode scalar values' class='external' href='http://www.unicode.org/glossary/#unicode_scalar_value'/>
-        <term name='NumericLiteral' class='external' href='http://people.mozilla.org/~jorendorff/es6-draft.html#sec-literals-numeric-literals' ref='ECMA-262' section='11.8.3'/>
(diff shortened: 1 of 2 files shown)


annevk:
Note that #113 does this too, to some extent...

index.xml
```diff
@@ -14990,10 +14990,9 @@ d.type = et;
           <dl>
             <dt id='ref-ECMA-262'>[ECMA-262]</dt>
             <dd>
-              <cite><a href="http://wiki.ecmascript.org/doku.php?id=harmony:specification_drafts">ECMAScript Language Specification, 6th Edition</a></cite>,
-              A. Wirfs-Brock, Editor. Ecma International, 8 November 2013 (Working Draft).
-              <span class="print"><br/>Available at http://wiki.ecmascript.org/doku.php?id=harmony:specification_drafts</span>
-              <br/>Unofficial HTML version available at <a href="http://people.mozilla.org/~jorendorff/es6-draft.html">http://people.mozilla.org/~jorendorff/es6-draft.html</a>.
+              <cite><a href="https://tc39.github.io/ecma262/">ECMAScript 2017 Language Specification</a></cite>,
```
domenic: I much prefer the form in #113 as it doesn't need updating every year.

Ms2ger: Fine by me, will let @heycam decide.

heycam: The undated spec reference sounds good to me. (We've already had issues with link anchor stability pointing to the ES6 draft, so I hopefully that problem shouldn't get any worse.)

Ms2ger: Fixed.

Issue #106, opened 2016-04-07
Should specify that all JS functions created by IDL are "built-in functions"


domenic:
That is, they should follow the rules in https://tc39.github.io/ecma262/#sec-built-in-function-objects. Notably:

- No .prototype property
- No [[Construct]] unless otherwise specified
- No arguments or caller properties

According to https://developer.microsoft.com/en-us/microsoft-edge/platform/catalog/, Chrome and Edge violate the arguments and caller prohibition. Not sure about the rest.

/cc @ajklein

Filed https://bugs.chromium.org/p/chromium/issues/detail?id=601272 on Chrome

bzbarsky:
> No [[Construct]] unless otherwise specified

That's not what built-in function objects seem to do in the spec.  And I don't see why we would want it.  Having the default spec [[Construct]] which is equivalent to [[Call]] except for setting newTarget and not initializing this seems correct; we'd just need to make all IDL stuff except interface objects throw when this is not initialized or something.

domenic:
> Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function. 

This is tested in various test262 tests for e.g. the various functions created for during promises. It's observable by using a proxy.

Interfaces could have a throwing [[Construct]], but methods should have no [[Construct]] at all.

bzbarsky:
Ah, I see.  OK, that seems fine.

PR #121, opened 2016-05-11
Add namespaces

domenic:
This adds the concept of namespaces, as discussed in https://github.com/whatwg/console/issues/3 (starting especially around https://github.com/whatwg/console/issues/3#issuecomment-214586757). They can only contain regular operations.

The ES binding for namespaces here is written in a fully modern style, and so differs slightly from similar prose for interfaces' ES binding. It is hoped that it can provide a template for eventually updating interfaces' ES binding to modern ES.

### DIFF
--- a/index.xml
+++ b/index.xml
@@ -769,8 +769,7 @@ interface TextField {
   <i>interface-members…</i>
 };</pre>
           <p>
-            The order that members appear has significance both for <a
-            href='#idl-overloading'>overloading</a> and for property enumeration in the <a
+            The order that members appear in has significance for property enumeration in the <a
             href='#es-interfaces'>ECMAScript binding</a>.
           </p>
           <p>
@@ -4008,15 +4007,14 @@ partial namespace <em>SomeNamespace</em> {
           <div class='note'>
             <p>
               As with partial interface definitions, partial namespace definitions are intended for
-              use as a specification editorial aide, allowing the definition of an interface to be
+              use as a specification editorial aide, allowing the definition of a namespace to be
               separated over more than one section of the document, and sometimes multiple
               documents.
             </p>
           </div>
 
           <p>
-            The order that members appear has significance both for <a
-            href='#idl-overloading'>overloading</a> and for property enumeration in the <a
+            The order that members appear in has significance for property enumeration in the <a
             href='#es-namespaces'>ECMAScript binding</a>.
           </p>
 
@@ -4025,12 +4023,10 @@ partial namespace <em>SomeNamespace</em> {
           </p>
 
           <p>
-            Only the <a class='xattr' href='#Exposed'>[Exposed]</a> and <a class='xattr'
+            Of the extended attributes defined in this specification, only the <a class='xattr'
+            href='#Exposed'>[Exposed]</a> and <a class='xattr'
             href='#SecureContext'>[SecureContext]</a> extended attributes are applicable to
-            namespaces. When declaring a <a class='dfnref' href='#dfn-partial-namespace'>partial
-            namespace</a>, the partial namespace must have the same extended attributes (and
-            extended attribute argument values, if any) specified as the original namespace
-            definition with the same identifier.
+            namespaces.
           </p>
 
           <?productions grammar Partial PartialDefinition Namespace NamespaceMembers NamespaceMember?>
@@ -8671,7 +8667,8 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
               <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>
               appears on an <a class='dfnref' href='#dfn-interface'>interface</a>,
               <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
-              <a class='dfnref' href='#dfn-namespace'>namespace</a>, or
+              <a class='dfnref' href='#dfn-namespace'>namespace</a>,
+              <a class='dfnref' href='#dfn-partial-namespace'>partial namespace</a>, or
               an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a> or
               <a class='dfnref' href='#dfn-namespace-member'>namespace member</a>,
               it indicates that the construct is exposed
@@ -8713,13 +8710,14 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
                 <dl class='switch'>
                   <dt>interface</dt>
                   <dt>namespace</dt>
-                  <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface
-                  only contains the
+                  <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the
+                  interface or namespace only contains the
                   <a class='dfnref' href='#dfn-primary-global-interface'>primary global interface</a>.</dd>
                   <dt>partial interface</dt>
+                  <dt>partial namespace</dt>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the partial
-                  interface is the <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the original
-                  interface definition.</dd>
+                  interface or namespace is the <a class='dfnref' href='#dfn-exposure-set'>exposure
+                  set</a> of the original interface or namespace definition.</dd>
                   <dt>interface member</dt>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface member
                   is the <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface or
@@ -8727,7 +8725,7 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
                   <dt>namespace member</dt>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the
                   namespace member is the <a class='dfnref' href='#dfn-exposure-set'>exposure
-                  set</a> of the namespace the member is declared on.</dd>
+                  set</a> of the namespace or partial namespace the member is declared on.</dd>
                 </dl>
               </li>
             </ul>
@@ -8740,18 +8738,20 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
               The <a class='xattr' href='#Exposed'>[Exposed]</a> extended attribute
               <span class='rfc2119'>MUST NOT</span> be specified on both an interface
               member and a partial interface definition the interface member is declared on.
+              Similarly, the <a class='xattr' href='#Exposed'>[Exposed]</a> extended attribute
+              <span class='rfc2119'>MUST NOT</span> be specified on both a namespace
+              member and a partial namespace definition the namespace member is declared on.
             </p>
             <p>
-              If <a class='xattr' href='#Exposed'>[Exposed]</a> appears on both an interface
-              and one of its interface members, then the interface member's
-              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
-              <span class='rfc2119'>MUST</span> be a subset of the interface's
-              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>.
-              Similarly, if <a class='xattr' href='#Exposed'>[Exposed]</a> appears on both a
-              namespace and one of its namespace members, then the namespace member's
-              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
-              <span class='rfc2119'>MUST</span> be a subset of the namespace's
-              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>.
+              If <a class='xattr' href='#Exposed'>[Exposed]</a> appears an interface member, then
+              the interface member's <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
+              <span class='rfc2119'>MUST</span> be a subset of the <a class='dfnref'
+              href='#dfn-exposure-set'>exposure set</a> of the interface or partial interface it's a
+              member of. Similarly, if <a class='xattr' href='#Exposed'>[Exposed]</a> appears on a
+              namespace member, then the namespace member's <a class='dfnref'
+              href='#dfn-exposure-set'>exposure set</a> <span class='rfc2119'>MUST</span> be a
+              subset of the <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the
+              namespace or partial namespace it's a member of.
             </p>
             <p>
               An interface's <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
@@ -10014,6 +10014,8 @@ counter.value;                                           <span class='comment'>/
               <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>
               appears on an <a class='dfnref' href='#dfn-interface'>interface</a>,
               <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
+              <a class='dfnref' href='#dfn-namespace'>namespace</a>,
+              <a class='dfnref' href='#dfn-partial-namespace'>partial namespace</a>,
               <a class='dfnref' href='#dfn-interface-member'>interface member</a>, or
               <a class='dfnref' href='#dfn-namespace-member'>namespace member</a>,
               it indicates that the construct is exposed
@@ -10032,6 +10034,8 @@ counter.value;                                           <span class='comment'>/
               be used on anything other than an
               <a class='dfnref' href='#dfn-interface'>interface</a>,
               <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
+              <a class='dfnref' href='#dfn-namespace'>namespace</a>,
+              <a class='dfnref' href='#dfn-partial-namespace'>partial namespace</a>,
               <a class='dfnref' href='#dfn-interface-member'>interface member</a>, or
               <a class='dfnref' href='#dfn-namespace-member'>namespace member</a>.
             </p>
@@ -10056,17 +10060,24 @@ counter.value;                                           <span class='comment'>/
                 depends on the type of construct:</p>
                 <dl class='switch'>
                   <dt>interface</dt>
-                  <dd>The interface or dictionary is not
+                  <dt>namespace</dt>
+                  <dd>The interface or namespace is not
                   <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>.</dd>
+
                   <dt>partial interface</dt>
-                  <dd>The partial interface is <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>
-                  if and only if the original interface definition is.</dd>
+                  <dt>partial namespace</dt>
+                  <dd>The partial interface or partial namespace is <a class='dfnre'
+                  href='#dfn-available-only-in-secure-contexts'>available only in secure
+                  contexts</a> if and only if the original interface or namespace definition
+                  is.</dd>
+
                   <dt>interface member</dt>
                   <dd>The interface member is <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>
                   if and only if the interface or partial interface the member is declared on is.</dd>
+
                   <dt>namespace member</dt>
                   <dd>The namespace member is <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>
-                  if and only if the namspace the member is declared on is.</dd>
+                  if and only if the namspace or partial namespace the member is declared on is.</dd>
                 </dl>
               </li>
             </ul>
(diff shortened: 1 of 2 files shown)


zcorpan:
For `CSS` at least I think we need `partial namespace CSS { ... }`, since different specs add stuff to it. Is that supported?

domenic:
No, I was lazy. I can add it...

zcorpan:
👍

Filed https://www.w3.org/Bugs/Public/show_bug.cgi?id=29623 for `CSS`

domenic:
Added partial namespaces and fixed a few things I noticed were missing. Left it as a separate commit in case anyone wants to see the sausage being made.

index.xml
```diff
@@ -3921,6 +3937,102 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+            </p>
+            <x:codeblock language='idl'>namespace VectorUtils {
+  double dotProduct(Vector x, Vector y);
+  Vector crossProduct(Vector x, Vector y);
+};</x:codeblock>
+
+            <p>
+              An ECMAScript implementation would then expose a global property named
+              <code>VectorUtils</code> which was a simple object (with prototype
+              <a>%ObjectPrototype%</a>) with non-enumerable data properties for each declared
```
bzbarsky: This is different from interfaces, where things _are_ enumerable, right?  Why the change?  Changing the behavior of `CSS` and `console` in this way may or may not be web-compatible... (yes, I just checked and all the stuff on `console` is enumerable in at least Safari, Chrome, and Firefox).

index.xml
```diff
@@ -8661,15 +8786,15 @@ interface DedicatedWorkerGlobalScope : WorkerGlobalScope {
 
 <span class='comment'>// MathUtils is available for use in workers and on the main thread.</span>
 [Exposed=(Window,Worker)]
-interface MathUtils {
+namespace MathUtils {
   static double someComplicatedFunction(double x, double y);
 };
 
 <span class='comment'>// WorkerUtils is only available in workers.  Evaluating WorkerUtils
 // in the global scope of a worker would give you its interface object, while
```
bzbarsky: Talking about an "interface object" for a namespace seems a bit odd.

I think you should probably leave these as interfaces, or add interface examples in addition to namespace ones or something.

index.xml
```diff
@@ -14109,6 +14240,155 @@ C implements A;</x:codeblock>
...
+
+            <p>
+              The namespace object for a given <a class='dfnref' href='#dfn-namespace'>namespace</a>
+              <var>namespace</var> and Realm <var>realm</var> is created as follows:
+            </p>
+
+            <ol class='algorithm'>
+              <li>
+                Let <var>namespaceObject</var> be
+                ! <a>ObjectCreate</a>(the <a>%ObjectPrototype%</a> of <var>realm</var>).
```
bzbarsky: This is using syntax that's not defined anywhere in the IDL spec that I can see.  A definition is needed (and ideally the '!' would link to that definition).

index.xml
```diff
@@ -3921,6 +3937,102 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+            <p>
+              As with partial interface definitions, partial namespace definitions are intended for
+              use as a specification editorial aide, allowing the definition of an interface to be
+              separated over more than one section of the document, and sometimes multiple
+              documents.
+            </p>
+          </div>
+
+          <p>
+            The order that members appear in has no significance except in the case of <a
```
bzbarsky: I know you copied this from the text for interfaces, but is this actually true?  I don't see how order matters in the case of overloading, but in the _non-overloading_ case it does matter in practice, though not per spec: it affects the order in which the properties appear in `Object.getOwnPropertyNames(namespaceObj)`.  Per spec the order of properties is not defined, I guess.

index.xml
```diff
@@ -14892,6 +15172,11 @@ d.type = et;
           <prod nt='ExtendedAttributeIdentList'>identifier "=" "(" IdentifierList ")"</prod>
           <prod nt='ExtendedAttributeNamedArgList'>identifier "=" identifier "(" ArgumentList ")"</prod>
 
+          <prod nt='Namespace'>"namespace" identifier "{" NamespaceMembers "}" ";"</prod>
+
+          <prod nt='NamespaceMembers'>ExtendedAttributeList NamespaceMember NamespaceMembers | ε</prod>
+          <prod nt='NamespaceMember'>ReturnType OperationRest</prod>
```
bzbarsky: Why can't a namespace have constants or attributes?  Certainly constants seem like a totally reasonable thing!

index.xml
domenic: I was torn on this. Safari indicated a desire to match ES's namespaces (Math, Reflect, JSON) which are not. And, `Object.keys(console)` currently returns `[]` because of the prototype separation; this preserves that. To me, it seems more likely people are depending on `Object.keys(console)` being empty (so that e.g. `Object.keys(console)` returns any methods they themselves added), than that they are depending on `Object.getOwnPropertyDescriptor(console, "log").enumerable` being true.

I guess maybe they are using `for`-`in` though, which does change.

I should probably flip this.

index.xml
domenic: This is using standard ES syntax. I think we just want to acclimatize people to it, as all ES algorithms across all specs are now written this way. See https://tc39.github.io/ecma262/#sec-algorithm-conventions

index.xml
domenic: Good point, will fix.

index.xml
domenic: Well, nobody needs them yet. We can add them later, but in general constants are avoided these days (in favor of simply using well-known strings). I went for minimal for now.

index.xml
bzbarsky: `for-in` is the case that worries me, compat-wise.

index.xml
bzbarsky: I think this syntax is terrible, as I've pointed out before.  The fact that it's (1) non-obvious and (2) not linkified to point to its definition makes it _really_ hard to get into reading a spec without reading a bunch of front matter that hides the syntax definition quite well.

But also, nothing in Web IDL links to that section of the ES spec, so people would have no idea you're doing that anyway.  This would be a bare minimum requirement for using this syntax, because then you could at least make the (laughable, imo, but technically correct) claim that "if you had read the entire Web IDL spec starting at the very beginning and to this point, and followed all the links from it and read those too, you would know what this syntax means in this spec".  A slight step up would be explicitly saying that '!' and '?" mean what they mean in `#sec-algorithm-conventions`, or that all algorithms are written using those conventions (and hence the reader of Web IDL should read them) or something.

domenic: I'll add an intro paragraph.

domenic:
Updated based on code review comments, including changing to enumerable.

zcorpan:
cc @esprehn

bzbarsky:
@domenic it's hard to tell from the diff; does this PR update http://heycam.github.io/webidl/#dfn-named-definition ?

bzbarsky:
Also, does it update the 

>  Within the set of IDL fragments that a given implementation supports, the identifier of every interface, dictionary, enumeration, callback function and typedef MUST NOT be the same as the identifier of any other interface, dictionary, enumeration, callback function or typedef. 

language in http://heycam.github.io/webidl/#idl-names ?

domenic:
> @domenic it's hard to tell from the diff; does this PR update http://heycam.github.io/webidl/#dfn-named-definition ?

Yes

> Also, does it update the ... language in http://heycam.github.io/webidl/#idl-names ?

Nope, missed that one. Pushed a rebased and squashed commit with that included.

---

The larger question we have before merging and using this for Console is what to do about the fact that console, for legacy reasons, needs an empty [[Prototype]] between it and Object.prototype.

- Do we build that into the definition of namespaces? (Hopefully not)
- Do we create an in-IDL way of expressing that one-off situation, say an extended attribute?
- Do we let the console spec add an exotic binding-level requirement that transcends Web IDL?

If I had some guidance on that I think we'd be ready to merge and use this...

bzbarsky:
In Gecko, I implemented an extended attribute for the console thing.  How we should spec it.... I'm not sure.  :(

domenic:
I guess I lean toward a one-off requirement in the console spec? In which case this should be mergeable...

domenic:
@bzbarsky @heycam ping on merging this? I can rebase if I am assured it will get merged soon, but I'm not eager to do so if it's just going to rot again...

Also same question on #133 and #146.

index.xml
```diff
@@ -3948,6 +3967,101 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+};
+
+partial namespace <em>SomeNamespace</em> {
+  <i>namespace-members…</i>
+};</pre>
+
+          <div class='note'>
+            <p>
+              As with partial interface definitions, partial namespace definitions are intended for
+              use as a specification editorial aide, allowing the definition of an interface to be
```
bzbarsky: "the definition of an interface"

index.xml
```diff
@@ -3948,6 +3967,101 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+              separated over more than one section of the document, and sometimes multiple
+              documents.
+            </p>
+          </div>
+
+          <p>
+            The order that members appear has significance both for <a
+            href='#idl-overloading'>overloading</a> and for property enumeration in the <a
```
bzbarsky: I still don't see how the order matters for overloading, either here or for interfaces.  It shouldn't matter.  I think we should remove mention of overloading from both places.

index.xml
```diff
@@ -753,8 +769,9 @@ interface TextField {
   <i>interface-members…</i>
 };</pre>
           <p>
-            The order that members appear in has no significance except in the
-            case of <a href='#idl-overloading'>overloading</a>.
+            The order that members appear has significance both for <a
```
bzbarsky: "appear in", no?

index.xml
```diff
@@ -3948,6 +3967,101 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+            <p>
+              As with partial interface definitions, partial namespace definitions are intended for
+              use as a specification editorial aide, allowing the definition of an interface to be
+              separated over more than one section of the document, and sometimes multiple
+              documents.
+            </p>
+          </div>
+
+          <p>
+            The order that members appear has significance both for <a
```
bzbarsky: "appear in"

index.xml
```diff
@@ -3948,6 +3967,101 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+            href='#idl-overloading'>overloading</a> and for property enumeration in the <a
+            href='#es-namespaces'>ECMAScript binding</a>.
+          </p>
+
+          <p>
+            Note that unlike interfaces or dictionaries, namespaces do not create types.
+          </p>
+
+          <p>
+            Only the <a class='xattr' href='#Exposed'>[Exposed]</a> and <a class='xattr'
```
bzbarsky: This section should be clearer that it's only talking about extended attributes defined by _this_ specification, because we do want console to define an extended attribute that applies to a namespace.

index.xml
```diff
@@ -3948,6 +3967,101 @@ setlike&lt;<i>type</i>&gt;;</pre>
...
+            href='#SecureContext'>[SecureContext]</a> extended attributes are applicable to
+            namespaces. When declaring a <a class='dfnref' href='#dfn-partial-namespace'>partial
+            namespace</a>, the partial namespace must have the same extended attributes (and
```
bzbarsky: Why?  This seems like an overly onerous requirement to me.  In particular, why would we prohibit this:

    [Exposed=(Window,Worker)]
    namespace foo {
      void func1();
    };
    [Exposed=Worker]
    partial namespace foo {
      void func2();
    };

as a mechanism for extending a namespace with some members that have more restricted exposure?  Similar for [SecureContext].  I don't see a reason to make these extended attributes work differently from the way they work for interfaces, if we're going to support partial namespaces at all.

index.xml
```diff
@@ -8556,9 +8671,11 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
               If the <a class='xattr' href='#Exposed'>[Exposed]</a>
               <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>
               appears on an <a class='dfnref' href='#dfn-interface'>interface</a>,
-              <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>, or
-              an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a>,
-              it indicates that the interface or interface member is exposed
+              <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
+              <a class='dfnref' href='#dfn-namespace'>namespace</a>, or
```
bzbarsky: Need to add partial namespace in here, yes?

index.xml
```diff
@@ -8596,6 +8713,7 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
                 implicitly, depending on the type of construct:</p>
                 <dl class='switch'>
                   <dt>interface</dt>
+                  <dt>namespace</dt>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface
```
bzbarsky: "of the interface or namespace"

bzbarsky: Need to add "partial namespace" to the next branch of this switch, right?

index.xml
```diff
@@ -8607,6 +8725,10 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
                   <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface member
                   is the <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the interface or
                   partial interface the member is declared on.</dd>
+                  <dt>namespace member</dt>
+                  <dd>The <a class='dfnref' href='#dfn-exposure-set'>exposure set</a> of the
+                  namespace member is the <a class='dfnref' href='#dfn-exposure-set'>exposure
+                  set</a> of the namespace the member is declared on.</dd>
```
bzbarsky: "of the namespace or partial namespace"

bzbarsky: Need to clone the "The [Exposed] extended attribute MUST NOT be specified on both an interface member and a partial interface definition the interface member is declared on." thing for namespaces, right?

index.xml
```diff
@@ -8626,6 +8748,11 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
               <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
               <span class='rfc2119'>MUST</span> be a subset of the interface's
               <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>.
+              Similarly, if <a class='xattr' href='#Exposed'>[Exposed]</a> appears on both a
```
bzbarsky: This isn't quite right, just like the existing interface text isn't quite right.  Consider this IDL:

    namespace foo {
      [Exposed=Worker] void func();
    };

This should be disallowed, because the namespace is only exposed on Window.  But the text you're adding does not apply, because the namespace itself doesn't have [Exposed] specified.

I think this text should go like this: "If [Exposed] appears on a namespace member, then the namespace member's exposure set MUST be a subset of the exposure set of the namespace or partial namespace it's a member of." or something along those lines.  And similar for interfaces.

index.xml
```diff
@@ -8626,6 +8748,11 @@ context.setColorEnforcedRange(-1, 255, 256);</x:codeblock>
...
+              namespace and one of its namespace members, then the namespace member's
+              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>
+              <span class='rfc2119'>MUST</span> be a subset of the namespace's
+              <a class='dfnref' href='#dfn-exposure-set'>exposure set</a>.
```
bzbarsky: The spec seems to be missing an obvious requirement for interface that we should add, and add a corresponding one for namespaces: "If [Exposed] appears on a partial interface, then the partial interface's exposure set MUST be a subset of the interface's exposure set."  Otherwise we would be allowing this:

    interface Foo {
    };
    [Exposed=Worker]
    partial interface Foo {
      void func();
    };

index.xml
```diff
@@ -9867,8 +10015,9 @@ counter.value;                                           <span class='comment'>/
               <a class='dfnref' href='#dfn-extended-attribute'>extended attribute</a>
               appears on an <a class='dfnref' href='#dfn-interface'>interface</a>,
               <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
-              an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a>,
-              it indicates that the interface or interface member is exposed
+              <a class='dfnref' href='#dfn-interface-member'>interface member</a>, or
```
bzbarsky: This needs to include namespace and partial namespace, yes?

index.xml
```diff
@@ -9883,8 +10032,9 @@ counter.value;                                           <span class='comment'>/
               extended attribute <span class='rfc2119'>MUST NOT</span>
               be used on anything other than an
               <a class='dfnref' href='#dfn-interface'>interface</a>,
-              <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>, or
-              an individual <a class='dfnref' href='#dfn-interface-member'>interface member</a>.
+              <a class='dfnref' href='#dfn-partial-interface'>partial interface</a>,
+              <a class='dfnref' href='#dfn-interface-member'>interface member</a>, or
```
bzbarsky: Need "namespace or partial namespace" here too.

index.xml
```diff
@@ -9915,6 +10065,9 @@ counter.value;                                           <span class='comment'>/
                   <dt>interface member</dt>
                   <dd>The interface member is <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>
                   if and only if the interface or partial interface the member is declared on is.</dd>
+                  <dt>namespace member</dt>
+                  <dd>The namespace member is <a class='dfnre' href='#dfn-available-only-in-secure-contexts'>available only in secure contexts</a>
+                  if and only if the namspace the member is declared on is.</dd>
```
bzbarsky: Need to add namespace and partial namespace to this list as well, yes?

bzbarsky: s/namspace/namespace/

index.xml
```diff
@@ -9935,7 +10088,8 @@ counter.value;                                           <span class='comment'>/
               The <a class='xattr' href='#SecureContext'>[SecureContext]</a> extended attribute
               <span class='rfc2119'>MUST NOT</span> be specified on both an interface
               member and the interface or partial interface definition the
-              interface member is declared on.
+              interface member is declared within, or on both a namespace member and the namespace
+              definition the namespace member is declared within.
```
bzbarsky: Needs to mention partial namespaces.

But also, this is maybe changing the semantics.  With the new wording, this IDL:

    [SecureContext] interface Foo {};
    partial interface Foo {
      [SecureContext] void func();
    };

is allowed, I think, whereas I _think_ the intent is that it not be allowed or something.  Though it's not clear that the old language prohibited this either; the "declared on" concept is pretty vague.  :(

index.xml
```diff
@@ -14374,6 +14528,155 @@ C implements A;</x:codeblock>
...
+                class='dfnref' href='#dfn-namespace-member'>namespace member</a> of this namespace,
+
+                <ol>
+                  <li>
+                    Let <var>id</var> be <var>op</var>'s <a class='dfnref'
+                    href='#dfn-identifier'>identifier</a>.
+                  </li>
+
+                  <li>
+                    Let <var>steps</var> be the following series of steps, given function argument
```
bzbarsky: Having these "what does the function do?" steps in the middle of the "how to create the namespace object" algorithm makes it really hard to read: it's like a really big plot digression.

Please factor this out into a separate section, and just reference it here.

bzbarsky:
Sorry, it wasn't clear to me what the state of this was.

I guess if the plan is to add an extended attribute in the console spec that changes the behavior then I'm OK with merging this, modulo the comments above...

index.xml
domenic: I think how this currently works for partial interfaces is confusing; it says "Any extended attribute specified on a partial interface definition is considered to appear on the interface itself." Maybe I was attempting to limit the confusion for partial namespaces. But you're right that we probably want to allow this, so it's probably best to just delete this sentence.

index.xml
domenic: I'm not sure how to square that with "Any extended attribute specified on a partial interface definition is considered to appear on the interface itself." It seems like the above snippet somehow makes Foo only exposed in Workers, per that text... (This is similar to my above reply.)

index.xml
bzbarsky: Huh.  Yeah, that language is kinda messed up, since that's clearly not how [Exposed] works... :(

index.xml
domenic: This was somewhat of an attempt to modernize that section to do things like state that methods are built-in functions (#106) and give property descriptors for name and length, and state the realm of the function. It also moves us toward a world where instead of stating what operations MUST be, we give an explicit procedure for creating interface objects and namespace objects. As I said in the OP, "The ES binding for namespaces here is written in a fully modern style, and so differs slightly from similar prose for interfaces' ES binding. It is hoped that it can provide a template for eventually updating interfaces' ES binding to modern ES."

Let me try factoring out this modern way of creating operations and replacing much of the existing #es-operations section with it. Then we can call into it from namespaces, and in the future call into it from interface objects.

bzbarsky: Fwiw, I'm also ok with you landing this as-is but factored out and doing the changes to #es-operations as a followup.  I don't want to hold this up on changes to #es-operations.

domenic:
OK, I think everything is addressed; thanks for the careful review. Clearly partial namespaces were added later and I didn't go through and find all the places they needed to be accounted for. The follow-up issues of https://github.com/heycam/webidl/issues/154 and https://github.com/heycam/webidl/issues/153 are filed, and the "creating an operation function" algorithm is moved into #es-operations with a note about how we'll consolidate soon.

I uploaded as separate commits for easier reviewing/verification, but I can squash if @heycam hasn't enabled squashing-from-GitHub-UI on this repo.

bzbarsky:
OK, I skimmed over the update commits and looks reasonable at first glance.  I wish we had a real review system... ;)

index.xml
zcorpan: `CSS.elementSources` is an attribute. https://github.com/w3c/csswg-drafts/issues/428

zcorpan: It appears we need them now again. https://github.com/w3c/css-houdini-drafts/issues/330#issuecomment-271639727

domenic: Let me file a new issue to track.

PR #133, opened 2016-07-14
Make globals and things they inherit from have immutable prototypes

domenic:
Fixes https://github.com/whatwg/html/issues/1235.

/cc @littledan @jswalden


index.html
```diff
@@ -11422,16 +11427,24 @@ <h4 id="PrimaryGlobal">4.3.6. [Global] and [PrimaryGlobal]</h4>
...
+                When the <span class="prop">[[Delete]]</span> internal method of a <a class="dfnref" href="#dfn-named-properties-object">named properties object</a> is called, the
+                following steps are taken:
               </p>
 
               <ol class="algorithm">
                 <li>Return <span class="esvalue">false</span>.</li>
               </ol>
             </div>
+
+            <div id="named-properties-object-delete" class="section">
```
littledan: Rename this id?

littledan:
Sounds like this applies for workers as well. Also, there's nothing that we want to freeze which isn't a global object. Looks like you covered everything--both the "named properties object" and the prototype chain. LGTM, but let's get feedback from @jswalden that this matches behavior on workers.

index.html
domenic: good catch, thank you -_-

annevk:
@bzbarsky already confirmed in https://github.com/whatwg/html/issues/1235#issuecomment-218756098 that this needs to work for workers.

jswalden:
I'll explicitly concur in the need for this in workers.  Every global object in which scripts can be dynamically loaded requires this treatment.  Workers have `importScripts` that does that, ergo they need immutability treatment.

domenic:
@bzbarsky this is rebased and should be mergeable

bzbarsky:
Seems ok, though there are some stray cleanup changes in there...

PR #146, opened 2016-08-17
Add some IDs to callback-invoking definitions

domenic:
/cc @foolip who was looking at this with me this morning


domenic:
@bzbarsky this is rebased and should be mergeable

PR #155, opened 2016-08-25
Update method creation for operations

### LINKED ISSUE #106: Should specify that all JS functions created by IDL are "built-in functions"

domenic:
This consolidates the spec text for creating the functions for interface operations and namespace operations, using the more modern style seen in namespace operations (introduced in df8c9c4ef7e32298ea3b8b13eed73790957468da) as the base.

Along the way, this fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547 (making all methods [ImplicitThis], and getting rid of [ImplicitThis]), and fixes https://github.com/whatwg/html/issues/643 (a related issue on the HTML side).

This also fixes #106, since we now explicitly use the CreateBuiltinFunction abstract operation.

---

@bzbarsky please take a look! I'm rather pleased that this managed to fix the long-standing [ImplicitThis] issue.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -144,7 +144,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html; spec: HTML
         text: Clean up after running a callback
         text: incumbent settings object
         text: event handler IDL attributes; url: event-handler-attributes
-        text: settings object; url: concept-realm-settings-object            
+        text: settings object; url: concept-realm-settings-object
+        text: global object; url: concept-realm-global; for: Realm
 </pre>
 
 <style>
@@ -1030,7 +1031,6 @@ definition.
 limitations.  The following extended attributes must not
 be specified on partial interface definitions:
 [{{Constructor}}],
-[{{ImplicitThis}}],
 [{{LegacyArrayClass}}],
 [{{NamedConstructor}}],
 [{{NoInterfaceObject}}].
@@ -1057,7 +1057,6 @@ The following extended attributes are applicable to interfaces:
 [{{Constructor}}],
 [{{Exposed}}],
 [{{Global}}],
-[{{ImplicitThis}}],
 [{{LegacyArrayClass}}],
 [{{NamedConstructor}}],
 [{{NoInterfaceObject}}],
@@ -8495,63 +8494,6 @@ for the specific requirements that the use of
 </div>
 
 
-<h4 id="ImplicitThis" extended-attribute lt="ImplicitThis">[ImplicitThis]</h4>
-
-If the [{{ImplicitThis}}]
-[=extended attribute=]
-appears on an [=interface=],
-it indicates that when a <emu-val>Function</emu-val> corresponding
-to one of the interface’s [=operations=]
-is invoked with the <emu-val>null</emu-val> or <emu-val>undefined</emu-val>
-value as the <emu-val>this</emu-val> value, that the ECMAScript global
-object will be used as the <emu-val>this</emu-val> value instead.
-This is regardless of whether the calling code is in strict mode.
-
-The [{{ImplicitThis}}]
-extended attribute must
-[=takes no arguments|take no arguments=].
-
-See [[#es-operations]]
-for the specific requirements that the use of
-[{{ImplicitThis}}] entails.
-
-Note: The [{{ImplicitThis}}] [=extended attribute=]
-is intended for use on the {{Window}} [=interface=].
-
-<div class="example">
-    
-    In the following example, the <code class="idl">Window</code>
-    [=interface=] is defined
-    with the [{{ImplicitThis}}]
-    [=extended attribute=].
-    
-    <pre highlight="webidl">
-        [ImplicitThis]
-        interface Window {
-          // ...
-          attribute DOMString name;
-          void alert(DOMString message);
-        };
-    </pre>
-    
-    Since the <code class="idl">Window</code> object is used as
-    the ECMAScript global object, calls to its functions can be made
-    without an explicit object, even in strict mode:
-    
-    <pre highlight="js">
-        "use strict";
-        
-        window.alert("hello");      // Calling alert with an explicit window object works.
-        alert("hello");             // This also works, even though we are in strict mode.
-        alert.call(null, "hello");  // As does passing null explicitly as the this value.
-        
-        // This does not apply to getters for attributes on the interface, though.
-        // The following will throw a TypeError.
-        Object.getOwnPropertyDescriptor(Window.prototype, "name").get.call(null);
-    </pre>
-</div>
-
-
 <h4 oldids="PrimaryGlobal" id="Global" extended-attribute lt="Global|PrimaryGlobal" dfn>[Global] and [PrimaryGlobal]</h4>
 
 If the [{{Global}}]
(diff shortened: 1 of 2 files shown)


domenic:
Filed https://github.com/w3c/web-platform-tests/issues/3585 to add tests for the built-in method behavior

domenic:
@bzbarsky ping

index.xml
```diff
@@ -12005,12 +11827,51 @@ interface
...
+                        href='#dfn-namespace'>namespace</a>, and <var>op</var> is not a <a
+                        class='dfnref' href='#dfn-static-operation'>static operation</a>:
+
+                        <ol>
+                          <li>
+                            If the <span class='esvalue'>this</span> value is <span
+                            class='esvalue'>null</span> or <span class='esvalue'>undefined</span>,
+                            set <var>O</var> to <var>realm</var>'s <a class='dfnref external'
+                            href='https://html.spec.whatwg.org/multipage/webappapis.html#concept-realm-global'>global
+                            object</a>. (This will subsequently cause a <span
```
bzbarsky: There's no way `target` can be the appropriate global object.  `target` isn't an object at all.

What you probable meant is "if the global does not implement `target`" or something like that?

index.xml
```diff
@@ -12005,12 +11827,51 @@ interface
                     Try running the following steps:
                     <ol>
                       <li>
+                        Let <var>O</var> be <span class='esvalue'>null</span>.
+                      </li>
+
+                      <li>
+                        If <var>target</var> is not a <a class='dfnref'
```
bzbarsky: It's not obvious at this point that `target` can only be a namespace or interface.  So I think it would be clearer to make this check "If `target` is an interface".  And perhaps assert at the top of the algorithm that it's one or the other.

But also, calling this thing `target` is a bit weird.  It's more like "source" or something....

index.xml
```diff
@@ -12005,12 +11827,51 @@ interface
...
+                          value.</li>
+
+                          <li>If <var>O</var> is a <a class='dfnref'
+                          href='#dfn-platform-object'>platform object</a>, then <a class='dfnref'
+                          href='#dfn-perform-a-security-check'>perform a security check</a>,
+                          passing <var>O</var>, <var>realm</var>, <var>id</var>, and "method".</li>
+
+                          <li>If <var>O</var> is not a <a class='dfnref'
+                          href='#dfn-platform-object'>platform object</a> that implements the
+                          interface <var>target</var>, <a href='#ecmascript-throw' class='dfnref
```
bzbarsky: This is not quite right, depending on what it means to "implement" an interface.  The current spec text is careful to ensure that `I` is an actual interface, not a mixin (except for the weirdness around globals, wheere I don't understand what the current spec text is doing and why).  I think we should fix the caller of this algorithm to pass in the "interface whose interface prototype object (or interface object, for a static operation) this property corresponding to the operation appears on" to preserve the current semantics.  At _that_ point calling it `target` would make sense, actually.

index.xml
```diff
@@ -12005,12 +11827,51 @@ interface
...
                         href='#dfn-effective-overload-set'>effective overload set</a> for <a
-                        class='dfnref' href='#dfn-regular-operation'>regular operations</a> with
-                        <a class='dfnref' href='#dfn-identifier'>identifier</a> <var>id</var> on
-                        <a class='dfnref' href='#dfn-namespace'>namespace</a> <var>namespace</var>
-                        and with argument count <var>n</var>.
+                        class='dfnref' href='#dfn-regular-operation'>regular operations</a> (if
+                        <var>op</var> is a regular operation) or for <a class='dfnref'
+                        href='#dfn-static-operation'>static operations</a> (if <var>op</var> is a
+                        static operation) with <a class='dfnref'
+                        href='#dfn-identifier'>identifier</a> <var>id</var> on <var>target</var> and
```
bzbarsky: Right, and here it's _very_ important that `target` be the actual interface we're defining on and not a mixin, because otherwise you don't have the full overload set.  I really wish we'd split mixins and interfaces into two separate concepts...

bzbarsky:
This looks great; all the nits above are minor.

Note that eventually we _may_ want to propagate the this value to static operations on interfaces (e.g. `Promise.resolve`, if it were defined as an IDL thing, would need access to the `this` value to work correctly).  But for now this is preserving the current behavior, which is fine.  This would be a simple tweak anyway.

index.xml
domenic: I think that wording (from the current spec) is just broken for unforgeable operations. I'm trying to figure out a way to phrase this, but despite all the phrases like "supplemental interface" and "consequential interface" I can't actually see a term for "the actual not-mixin interface where this is headed for". Ideas appreciated.

domenic: Also I think that wording actually doesn't work at all. There is no "the" interface whose interface prototype object is `Navigator.prototype`. Navigator, NavigatorID, and NavigatorConcurrentHardware are all interfaces whose interface prototype object is `Navigator.prototype`. Otherwise the previous step, which says "Otherwise, the property exists solely on the interface’s interface prototype object" would not work at all for NavigatorID.

Maybe the solution here is to use a vague phrase like "the interface (or the interface it's being mixed in to, if the interface is actually a mixin)" and have a bug to define this correctly later? If we do that we probably won't fix this until we actually separate out mixins, but that's probably OK given that the current spec seems broken.

bzbarsky: Yeah, I agree this stuff is all broken right now.  :(  Good catch on the unforgeable operations bit.

The reason things are complicated in terms of the current definitions is that the current spec technically allows this, afaict:

    [Constructor] interface Foo {
    };
    [Constructor] interface Bar {
      void something();
    };
    Foo implements Bar;

And then script can do:

    var foo = new Foo(); // proto is Foo.prototype, which has own prop named "something"
    var bar = new Bar(); // proto is Bar.prototype, which has own prop named "something"
    console.log(foo.something == bar.something); // logs false

but the spec's model of this is pretty iffy.  For example, I think in the spec's model there is only one operation here, but it's exposed on two different interfaces and there are two different JS functions involved which do different brand checks.  Or something.  And just talking about "supplemental interface" is not useful, since in my example `Bar` is supplemental for `foo` but not for `bar`.

In practice the right way to address this is to make the instance and interface and interface prototype object setups more imperative.  Then the imperative algorithm will know which non-supplemental interface is causing the property to exist and will be able to pass that information into the "create a function" algorithm...

The other thing that would make all this clearer, of course, is making mixins different syntactically and outlawing my example above; it would have to become something like:

    [Constructor] interface Foo {
    };
    [Constructor] interface Bar {
    };
    mixin FooBarMixin {
      void something();
    };
    Foo implements FooBarMixin;
    Bar implements FooBarMixin;

> There is no "the" interface whose interface prototype object is Navigator.prototype

Well, there is.  It's called `Navigator`.  ;)

> Otherwise the previous step, which says "Otherwise, the property exists solely on the interface’s interface prototype object" would not work at all for NavigatorID.

This wording I agree is broken.  Maybe what we should do in the short term is redefine "interface members" (currently a purely syntactic definition) to be "the things syntactically in this interface, plus the things syntactically in its consequential interfaces".  Then the Navigator interface will have all the bits from NavigatorID in its "interface members" and things will work.  Or, depending on what other things use "interface members" we could define a new term that means the new thing and use it in the http://heycam.github.io/webidl/#es-operations setup to make sure we see all the relevant operations while using Navigator for their brand checks, which is what we want....

bzbarsky: Oh, and I'm certainly ok pushing all this off into a followup.  Please add a note to the effect that this is known broken?

bzbarsky:
@domenic OK, updated bits look good.  Can you please file that followup issue to define that sanely and add a note pointing to it?

My apologies for the lag here and for not being able to look at this again until Tuesday morning... :(

domenic:
Going to hold off on this until the Bikeshed rewrite lands, FYI.

domenic:
This has been rebased and a pointer added to #164; should be ready to merge.

Issue #83, opened 2016-01-05
Check if NodeFilter.length is defined correctly.


Ms2ger:
Seems to be 0 in Gecko/Chrome, not sure if the spec backs that.

bzbarsky:
The spec seems to only define a "length" property for non-callback interfaces, implying that there is no such property on the `NodeFilter` object.  So doing `NodeFilter.length` will return `Function.prototype.length`, which is 0.

That said, `NodeFilter.hasOwnProperty("length")` actually returns true in both Gecko and Blink.

Ms2ger:
I'm not sure where `Function.prototype.length` comes from, given

> Remember that interface objects for callback interfaces only exist if they have constants declared on them; when they do exist, they are not function objects.

bzbarsky:
Oh.  Well, `NodeFilter.__proto__ == Function.prototype` is `true` in both Gecko and Blink.  Yes, that's not what IDL says right now.  It's not clear to me why IDL says what it does and which should change.

nox:
@bzbarsky That condition is false in Safari.

nox:
Mentioning #78 because NodeFilter is the only legacy callback interface we have.

bzbarsky:
@nox What is it in IE?  Seems to me like we should do the simplest thing that requires the fewest UAs to change.

nox:
@bzbarsky No idea, I don't have any Windows box. :(

domenic:
`NodeFilter.__proto__ === Function.prototype` in latest Edge.

ArkadiuszMichalski:
Hmm, more results for:
```
<script>
	console.log(NodeFilter.__proto__ == Function.prototype);
	console.log(typeof NodeFilter);
	console.log(NodeFilter.length);
	console.log(NodeFilter.hasOwnProperty("length"));
</script>
```
IE11: `false, object, undefined, false`
Opera 12.x: `true, function, 0, false`
Firefox and Chrome: `true, function, 0, true`

arronei:
Edge: true, function, 0, false (length is a known Edge bug and should be true)

PR #156, opened 2016-08-29
Callback interface prototype should be Function.prototype

### LINKED ISSUE #96: Object.getPrototypeOf(NodeFilter) should be Function.prototype, not Object.prototype
http://heycam.github.io/webidl/#interface-object

> The internal [[Prototype]] property of an interface object for a callback interface must be the Object.prototype object.

This doesn't match Edge and Gecko, where `Object.getPrototypeOf(NodeFilter) === Function.prototype` holds true. It looks like this could be fixed by using the same language as for non-callback interfaces that don't inherit from any other interface:

> If the interface doesn't inherit from any other interface, the value of [[Prototype]] is %FunctionPrototype%

This was discussed in https://github.com/heycam/webidl/issues/83 and I just noticed it in https://github.com/whatwg/dom/issues/186#issuecomment-193709823

https://github.com/heycam/webidl/issues/78 is also more general issue about the interface objects of callback interface objects.

ayg:
… not Object.prototype.

Fixes #96.  Recent versions of Chrome, Firefox, and Edge all say
Object.getPrototypeOf(NodeFilter) === Function.prototype, whereas the
spec says Object.prototype.  nox reports in #83 that Safari does not
return Function.prototype here, but doesn't say what it does return.  It
seems the spec should match the majority of browsers, unless there's
some good reason not to.

<hr />
[Preview](https://rawgit.com/ayg/webidl/callback-prototype/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Frawgit.com%2Fayg%2Fwebidl%2Fcallback-prototype%2Findex.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10330,7 +10330,7 @@ such an interface object.
 
 The internal \[[Prototype]] property
 of an interface object for a callback interface must be
-the Object.prototype object.
+the Function.prototype object.
 
 Note: Remember that interface objects for callback interfaces only exist if they have
 [=constants=] declared on them;
--- a/index.html
+++ b/index.html
@@ -2296,9 +2296,10 @@ be defined on a <a data-link-type="dfn" href="#dfn-callback-interface" id="ref-f
 };
 </pre>
     <p>to be used like this:</p>
-<pre class="highlight"><span></span><span class="kd">var</span> <span class="nx">a</span> <span class="o">=</span> <span class="nx">getA</span><span class="p">();</span>  <span class="c1">// Get an instance of A.</span>
-
-<span class="nx">a</span><span class="p">.</span><span class="nx">doTask</span><span class="p">(</span><span class="s2">"something"</span><span class="p">,</span> <span class="p">{</span> <span class="nx">option1</span><span class="o">:</span> <span class="s2">"banana"</span><span class="p">,</span> <span class="nx">option3</span><span class="o">:</span> <span class="mi">100</span> <span class="p">});</span></pre>
+<pre class="highlight"><span class="kd">var</span> a <span class="o">=</span> getA<span class="p">(</span><span class="p">)</span><span class="p">;</span>  <span class="c1">// Get an instance of A.
+</span>
+a<span class="p">.</span>doTask<span class="p">(</span><span class="s2">"something"</span><span class="p">,</span> <span class="p">{</span> option1<span class="o">:</span> <span class="s2">"banana"</span><span class="p">,</span> option3<span class="o">:</span> <span class="mi">100</span> <span class="p">}</span><span class="p">)</span><span class="p">;</span>
+</pre>
     <p>instead write the following:</p>
 <pre class="highlight"><span class="kt">dictionary</span> <span class="nv">Options</span> {
   <span class="kt">DOMString</span>? <span class="nv">option1</span>;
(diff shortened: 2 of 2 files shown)


bzbarsky:
> Safari does not return Function.prototype here, but doesn't say what it does return

In Safari, `Object.getPrototypeOf(NodeFilter) === Object.prototype` tests true.

ayg:
Thanks, I updated the commit message.  I think it still makes sense to match the majority of browsers, unless there's some big advantage to the Safari behavior.

bzbarsky:
I strongly suggest talking to @cdumez and making sure Safari is on board with this change before we start thrashing specifications.

cdumez:
I do not feel strongly about this. If no one objects to this change, I can roll out my recent change in WebKit. I am not aware of any real benefit, we merely wanted to match the spec.

cdumez:
Please make sure the web platform tests are also updated though, as this is how we noticed we did not match the spec.

ayg:
@cdumez Personally, when I make changes to conform to the spec/tests, I first check all the other major UAs that I have access to, to make sure the spec is right.  If UAs don't agree with the spec, the spec is probably wrong, and we need to make sure we have buy-in from a majority of UAs before anyone changes to match the spec.  Otherwise implementations are forced to change unnecessarily and we decrease interoperability until they've converged again, which is much more effort for everyone than if we had just changed the spec to start with.

Actually, I filed this bug because I was investigating Gecko wpt failures, found this one, checked against other UAs, and concluded the spec is probably what should change rather than Gecko.

I'll handle the test updates.

cdumez:
@ayg Sure, I did check other browsers and noticed they did not match the specification. However, I went ahead anyway because the change was trivial and the compatibility risk super low. It is not always obvious if the spec needs to be changed or if browsers simply have not been updated yet.

annevk:
@domenic is either behavior better?

domenic:
I can't think of a reason why it should be Object.prototype (when the rest are Function.prototype, e.g. `EventTarget.__proto__ === Function.prototype`). Maybe there was one, even a good one, but reified callback interfaces are so weird that I don't think it really matters; we're not setting an important precedent here.

ayg:
@cdumez Still, if we have majority agreement contrary to the spec, I think it would be best to check with other browsers before changing away from the current consensus.  Often the spec is just a mistake and the person who wrote it didn't test that part in browsers.  It's true that the change is trivial, but a trivial change to the spec and tests is still considerably easier than a trivial change to four separate browser engines.  Especially because some of the browser engines might not actually change for years to come, if ever, and then we don't have interop for that period -- whereas the spec and tests are super-easy to change, much more so than changing even one browser engine.

It's great to hear that someone at WebKit is also looking through and fixing their test failures, though.  Kudos!

ayg:
@bzbarsky Everyone so far seems fine with the change.  If nobody has any further objections, do you mind merging it?

ayg:
I've written the test update, at ayg/testharness.js in the callback branch.  I'll do a PR on that once this is accepted.

ayg:
(I don't know what all the changes in index.html are.  If it helps, I used the web interface instead of a local bikeshed copy to generate it.)

bzbarsky:
Manually merged as https://github.com/heycam/webidl/commit/6b0eb63d6cba3ff1f90f4ef3e733749ff51751a0

PR #165, opened 2016-09-08
Cleanup trailing whitespace

domenic:
This is on top of #155 to avoid extra rebasing after that lands


tobie:
Merged in a3a6aa223762980d639b96190aaf07263a94c1b8.

PR #166, opened 2016-09-08
Remove the "Changes" section

domenic:
It is out of date, incomplete, and a poor substitute for the commit history.


tobie:
@heycam, @bzbarsky: this was on my list too, but do you approve?

bzbarsky:
This is @heycam's call.  I never thought this section was that useful, though maybe it was more so when we had the difference between v1 and tip....

heycam:
That's fine.  It's probably too big to be useful in-document now, and people needing to follow the spec closely can read the commit logs.

PR #167, opened 2016-09-08
Move conventions and conformance to appendices

domenic:
It's nice to get to the good stuff sooner.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -290,119 +290,6 @@ have been written in prose.  In addition, precise language bindings
 for ECMAScript Edition 6 are given.
 
 
-<h3 id="conventions">Typographic conventions</h3>
-
-The following typographic conventions are used in this document:
-
-*   Defining instances of terms: <dfn id="dfn-example-term">example term</dfn>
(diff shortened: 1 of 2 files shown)


tobie:
Agree re conformance. Typographic conventions become less useful at the end imho, so the question becomes whether we need them at all.

marcoscaceres:
>  Typographic conventions become less useful at the end imho, so the question becomes whether we need them at all.

Ditch 'em, IMO.

bzbarsky:
To the extent that we have nontrivial semantics via typography (think the '?' thing) we should have it in the document.  If nothing else so that instances of those semantics can link to the thing that explains what the line noise means.

marcoscaceres:
@bzbarsky, good point; but there is stuff there that is well-known, like what a non-normative example looks like.

bzbarsky:
I dunno how well-known it is.  I recently had someone (a browser engine developer) quoting non-normative example text from the HTML spec at me as justification for something that engine was doing that contradicted the normative text...

marcoscaceres:
Browser engineers are the worst. The. Worst.

Issue #148, opened 2016-08-20
Remove the RegExp type


domenic:
This has no consumers on the platform, at least in Chromium: https://cs.chromium.org/search/?q=RegExp+file:%5C.idl&sq=package:chromium&type=cs

Like the now-dead Date type, if someone actually wanted to use this, they should use `object` plus some prose. But probably nobody should use this.

/cc @nox

bzbarsky:
Fwiw, I agree.  I'm not really sure why we ever added this type, apart from it clearly mapping to one of the "weird" objects in JS; we have not found any use cases for it, and Gecko has no support for it and no uses or prospective uses I'm aware of.

cdumez:
We also have no user or support for RegExp in WebKit bindings.

bzbarsky:
Fwiw, there were web platform tests testing the old spec verbiage.  I'll update them in https://bugzilla.mozilla.org/show_bug.cgi?id=1315135

Issue #145, opened 2016-08-13
Conversion to sequence<T> need not throw on a RegExp object


rniwa:
When [an ECMAScript value V is converted to an IDL sequence<T> value](http://heycam.github.io/webidl/#es-sequence), we don't really need to throw a `TypeError` when _V_ is a native`RegExp` object.

Ordinarily, a `RegExp` object wouldn't have `Symbol.iterator` defined.  If it's defined, e.g. `x = new RegExp; x[Symbol.iterator] = function* () { yield 1; }`, then we should be able to iterate over it.

domenic:
In general I find the way the Web IDL spec treats RegExp and Date objects specially rather strange... Maybe there is historical context I am missing.

bzbarsky:
The historical context is twofold:

1. The intent was to make `Date` and `RegExp` distinguishable from dictionaries and sequences during overload resolution.
2. At one point the intent was that adding an overload should generally not change which overload got invoked for previously-accepted values.  That meant that an overload accepting dictionary or sequence should not accept `RegExp` or `Date`.  The idea was to avoid forward-compat bugs in APIs where behavior would change for existing callers when a new overload was added.

I think we've already deviated that principle in item 2 when we made dictionary and sequence distinguishable while still allowing iterables to represent dictionaries.  And I, personally, don't think it's a big deal to drop that principle entirely.

That said, for the specific case of `RegExp`, I think removing it entirely as https://github.com/heycam/webidl/issues/148 suggests is the right way to go.

PR #180, opened 2016-09-30
Specify open dictionaries.

jyasskin:
Let me know what you think. Preview at https://rawgit.com/jyasskin/webidl/open-dictionaries/index.html.

I tried to position `dictionary<K, V>` as a dictionary type, but we may need to move some things from https://rawgit.com/jyasskin/webidl/open-dictionaries/index.html#idl-dictionaries to https://rawgit.com/jyasskin/webidl/open-dictionaries/index.html#idl-dictionary to really make that work. 

I introduced a ["mappings"](https://rawgit.com/jyasskin/webidl/open-dictionaries/index.html#dictionary-mappings) term for the contents of an open dictionary, but I might be able to re-use "dictionary members".

I used `<div algorithm>` blocks to get better `<var>` checking, but that also results in boxes around the algorithms, which you might or might not like.

@bzbarsky @tobie @annevk

<hr />
[Preview](https://rawgit.com/jyasskin/webidl/open-dictionaries/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Frawgit.com%2Fjyasskin%2Fwebidl%2Fopen-dictionaries%2Findex.html) 
<details><summary>Per Commit Preview &amp; Diffs</summary><code title="Specify open dictionaries&period;">2e084f4</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/2e084f4/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2e084f4%2Findex.html)  _(2016-09-28 23:59)_
<code title="Add the 1-arg dictionary&lt;&gt; form to the grammar&period;">2008d7f</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/2008d7f/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2008d7f%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2e084f4%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2008d7f%2Findex.html) _(2016-09-30 09:38)_
<code title="Make the grammar LL&lpar;1&rpar;&period;">2b5f2a8</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/2b5f2a8/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2b5f2a8%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2008d7f%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2b5f2a8%2Findex.html) _(2016-09-30 23:24)_
<code title="Rename dictionary&lt;&gt; to OpenDictionary&lt;&gt;&period;">3dc8ca9</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/3dc8ca9/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F3dc8ca9%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F2b5f2a8%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F3dc8ca9%2Findex.html) _(2016-09-30 23:29)_
<code title="Remove constraint on OpenDictionary&lt;&gt; value type&period;">1a8ae76</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/1a8ae76/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F1a8ae76%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F3dc8ca9%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F1a8ae76%2Findex.html) _(2016-09-30 23:35)_
<code title="Undo use of &lt;div algorithm&gt;&period;">05252bf</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/05252bf/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F05252bf%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F1a8ae76%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F05252bf%2Findex.html) _(2016-09-30 23:44)_
<code title="Interleave type checks with property retrieval&period;">747ba1e</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/747ba1e/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F747ba1e%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F05252bf%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F747ba1e%2Findex.html) _(2016-09-30 23:58)_
<code title="Convert null and undefined to empty OpenDictionary&lt;&gt;s&period;">6646a17</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/6646a17/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F6646a17%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F747ba1e%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F6646a17%2Findex.html) _(2016-10-01 00:02)_
<code title="Use ObjectCreate instead of &lcub;&rcub;&period;">7c5c25a</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/7c5c25a/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F7c5c25a%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F6646a17%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F7c5c25a%2Findex.html) _(2016-10-01 00:05)_
<code title="Use console&period;assert&lpar;&rpar; instead of assert&lpar;&rpar;&period;">756f55d</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/756f55d/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F756f55d%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F7c5c25a%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F756f55d%2Findex.html) _(2016-10-01 00:06)_
<code title="Improve examples&period;">399fc10</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/399fc10/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F399fc10%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F756f55d%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F399fc10%2Findex.html) _(2016-10-01 00:17)_
<code title="Make open dictionaries act like dictionaries&period;">59c34ac</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/59c34ac/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F59c34ac%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F399fc10%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F59c34ac%2Findex.html) _(2016-10-04 01:18)_
<code title="Use &lpar;k&comma; v&rpar; instead of k &rArr; v&period;">06ca8e2</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/06ca8e2/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F06ca8e2%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F59c34ac%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F06ca8e2%2Findex.html) _(2016-10-04 01:48)_
<code title="Handle lying proxies in OpenDictionary conversion&period;">84d1bc3</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/84d1bc3/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F84d1bc3%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F06ca8e2%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F84d1bc3%2Findex.html) _(2016-10-04 21:08)_
<code title="Require both arguments to OpenDictionary&lt;&gt;&period;">0386c5d</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/0386c5d/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F0386c5d%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F84d1bc3%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F0386c5d%2Findex.html) _(2016-10-04 21:20)_
<code title="Rename &quot;open dictionary&quot; to &quot;record&quot;&period;">fc878b0</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/fc878b0/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffc878b0%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F0386c5d%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffc878b0%2Findex.html) _(2016-10-06 20:30)_
<code title="Interleave enumerability checks with Gets and type conversions&period;">fe44f9c</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/fe44f9c/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffe44f9c%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffc878b0%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffe44f9c%2Findex.html) _(2016-10-14 18:51)_
<code title="Fix a couple &lt;code&gt;s and unescaped &lt;s&period;">665d408</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/665d408/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F665d408%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2Ffe44f9c%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F665d408%2Findex.html) _(2016-10-14 18:56)_
<code title="Woot&excl; Down to nits&excl;">9e22b41</code>: [Preview](https://cdn.rawgit.com/jyasskin/webidl/9e22b41/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F9e22b41%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F665d408%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F9e22b41%2Findex.html) _(2016-10-14 20:07)_
Latest (<code title="Use &laquo;&raquo; to represent a record&apos;s sequence of mappings&period;">20c46cc</code>): [Preview](https://cdn.rawgit.com/jyasskin/webidl/20c46cc/index.html) | [Diff w/ Head](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F20c46cc%2Findex.html)  | [Diff w/ Previous](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F9e22b41%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F20c46cc%2Findex.html) _(2016-10-14 20:29)_</details>


annevk:
I was wondering if the enumeration logic when converting from JavaScript should be closer to step 22, substep 3 of https://html.spec.whatwg.org/multipage/infrastructure.html#structuredclone. (I'm not really sure one way or another, but since they're both copying operations.)

annevk:
Also, should we revert #13 at this point? It was added at the time as some kind of proposed replacement for this, but didn't actually address the dictionary-like problem and has since gone unused.

index.bs
```diff
@@ -7529,6 +7564,82 @@ iterable |iterable| and an iterator getter
...
+IDL {{dictionary}}&lt;|K|, |V|&gt; values are represented by
+ECMAScript <emu-val>Object</emu-val> values.
+
+<div algorithm="convert object to open dictionary" id="es-to-open-dictionary">
+    An ECMAScript value |O| is [=converted to an IDL value|converted=] to an IDL <code>{{dictionary}}<|K|, |V|></code> value as follows:
+
+    1.  If [=Type=](|O|) is not <emu-val>Object</emu-val>,
+        <a lt="es throw">throw a <emu-val>TypeError</emu-val></a>.
+    1.  Let |result| be a new empty instance of <code>{{dictionary}}<|K|, |V|></code>.
+    1.  Let |entries| be [=EnumerableOwnProperties=](|O|, "key+value").
```
jyasskin: @annevk I think step 22, substep 3 of Structured Clone is equivalent to [EnumerableOwnProperties](https://tc39.github.io/ecma262/#sec-enumerableownproperties). There is an interesting question about what order we should type-check vs [[Get]]ing the values. I see at least four options:

1. Get all properties before checking their types (what I have now).
2. Get and check all the keys, and then interleave getting values with checking them.
3. Get all keys, then check a key, get its value, check its value, repeat.
4. Get all keys, then get a key's value, check the key, check the value, repeat.

For sane objects these are the same, but for getters with side-effects they're not.

annevk: Right, I don't have an opinion here, but I think the current approach is different when it comes to getters and I don't know what is better. I just thought I'd raise it to see if anyone else had thoughts on that.

jyasskin: Oh, right, StructuredClone does interleave getting with possible exceptions from the recursive StructuredClone call.

tobie:
> I used `<div algorithm>` blocks to get better `<var>` checking, but that also results in boxes around the algorithms, which you might or might not like.

We can always fix the styling if we don't like it, but we should absolutely put all algorithm in `<div algorithm>` blocks. I'll fix the other algorithms in the spec shortly.

You might also want to wrap your lists in `<ol class="algorithm">` as this gives us the light blue left-hand border which makes visualizing nesting in algorithms much easier (this probably should be automatic within algorithm blocks but isn't for now /cc @tabatkins).

jyasskin:
https://github.com/tabatkins/bikeshed/issues/759 should have made the blue border happen inside algorithm blocks, and it's working in Web Bluetooth. I'm not sure what's different here.

index.bs
```diff
@@ -5171,6 +5175,7 @@ type.
         "DOMException" Null
         BufferRelatedType Null
         "FrozenArray" "&lt;" Type "&gt;" Null
+        "dictionary" "&lt;" Type "," Type "&gt;" Null
```
tobie: You probably want to give it its own non-terminal to account for the first Type being optional:

```html
<pre class="grammar" id="prod-OpenDictionaryType">
    OpenDictionaryType :
        "dictionary" "&lt;" Type "&gt;"
        "dictionary" "&lt;" Type, Type "&gt;"
</pre>
```

jyasskin: Oops, done.

bzbarsky:
I believe this proposed grammar is not LL(1) parseable.  Specifically, `OpenDictionaryType` looks like a textbook example of <https://en.wikipedia.org/wiki/LL_parser#FIRST.2FFIRST_Conflict>.  I'm willing to be proven wrong on this point, of course.  ;)  I think basic refactoring something like this:

    OpenDictionaryType ::
        dictionary < Type OpenDictionaryTypeTail

    OpenDictionaryTypeTail ::
        >
        , Type >

will address that.

That said, I wish we addressed the "first type must be one of these three things" bit in grammar, not prose.  That, though, would cause an irreconcilable FIRST/FIRST conflict, I'm pretty sure.  That part is rather annoying.  I'm open to ideas for how we can have our LL(1) cake and eat our grammar-enforced sanity too.  One obvious way is to make the key type be the second type.  Another obvious way is to use different names for the two things...  Maybe I'm just overthinking this.

Apart from that, I see the following issues:

1. Using `dictionary` here could be somewhat confusing with the totally separate concept of dictionaries, IMO...  We may want to come up with a different name.

2. I'm not sure I follow the "V must be a type allowed for dictionary members." thing: all types are allowed as dictionary members.

3. We need distinguishability information, overload resolution handling, and union conversion handling for this new type.  In practice, it will behave much like "object".

4. Do we want to allow passing null or undefined to mean "empty open dictionary" the way we do for dictionaries?

5. I don't know whether the intent here is to be black-box identical in terms of the enumeration to structured clone step 22 substep 3, but this proposed logic is _not_ black-box identical to that, because it calls MOP operations in a different order.  It also doesn't call the exact same ones; there's no equivalent of the `HasOwnProperty` check from structured clone here, because the different ordering, which interleaves the `[[GetOwnProperty]]` and `[[Get]]`, calls makes it less important to do that.  For what it's worth, what Gecko implements for MozMap matches neither of these algorithms; I think we do structured clone step 22 substep 3, but without the extra check in substep 3 subsubstep 1.  We also do our conversion to an IDL value right where structured clone would do the `StructuredClone` recursive call.

6. "Repeat, for each element [key, value] of entries" is a bit confusing, because each element of entries is an ES Array.  This needs to spec how to get the key and value out of the array (probably via `Get()` calls, not the destructuring that the current language is so suggestive of).  Alternately, taking a more structured clone like approach makes this not be an issue, of course.

index.bs
```diff
@@ -7529,6 +7570,82 @@ iterable |iterable| and an iterator getter
...
+ECMAScript <emu-val>Object</emu-val> values.
+
+<div algorithm="convert object to open dictionary" id="es-to-open-dictionary">
+    An ECMAScript value |O| is [=converted to an IDL value|converted=] to an IDL <code>{{dictionary}}<|K|, |V|></code> value as follows:
+
+    1.  If [=Type=](|O|) is not <emu-val>Object</emu-val>,
+        <a lt="es throw">throw a <emu-val>TypeError</emu-val></a>.
+    1.  Let |result| be a new empty instance of <code>{{dictionary}}<|K|, |V|></code>.
+    1.  Let |entries| be [=EnumerableOwnProperties=](|O|, "key+value").
+    1.  [=ReturnIfAbrupt=](|entries|).
```
domenic: Use ? notation instead of ReturnIfAbrupt.

index.bs
domenic: You should not use EnumerableOwnProperties here, because EnumerableOwnProperties creates an array. This mixes levels between the abstract (list/records/ES spec devices) and the concrete (Arrays), and also makes it unclear whether in subsequent steps when you do [key, value] what the exact semantics are (e.g. is it impacted by changes to Array.prototype or the iteration protocol).

index.bs
```diff
@@ -7529,6 +7570,82 @@ iterable |iterable| and an iterator getter
...
+        1.  Append to |result| a mapping from |typedKey| to |typedValue|.
+    1.  Return |result|.
+</div>
+
+<div algorithm="convert open dictionary to object" id="open-dictionary-to-es">
+    An IDL <code>{{dictionary}}<…></code> value |D| is [=converted to an
+    ECMAScript value|converted=] to an ECMAScript value as follows:
+
+    1.  Let |result| be a new <emu-val>Object</emu-val> instance
+        created as if by the expression <code>{}</code>.
```
domenic: Web IDL may use this phrase elsewhere, but from now on we should be saying "Let result be ! ObjectCreate(%ObjectPrototype%)".

index.bs
```diff
@@ -7529,6 +7570,82 @@ iterable |iterable| and an iterator getter
...
+    Open dictionaries only consider [=own property|own=] [=enumerable=] properties, so given an IDL operation
+    <code>dictionary&lt;long> identity(dictionary&lt;long> arg)</code> which
+    returns its argument, the following code passes its assertions:
+
+    <pre highlight="js">
+        let proto = {a: 3, b: 4};
+        let obj = {__proto__: proto, d: 5, c: 6}
+        Object.defineProperty(obj, "e", {value: 7, enumerable: false});
+        let result = identity(obj);
+        assert(result.a === undefined);
```
domenic: Nit: I like using `console.assert` in these examples, since it is an actual function that exists in browsers.

index.bs
```diff
@@ -7529,6 +7570,82 @@ iterable |iterable| and an iterator getter
...
+        let entries = Object.entries(result);
+        assert(entries[0][0] === "d");
+        assert(entries[0][1] === "5");
+        assert(entries[1][0] === "c");
+        assert(entries[1][1] === "6");
+    </pre>
+
+    Dictionary keys and values can be constrained, although keys can only be
+    constrained among the three string types. The following conversions all
+    throw {{TypeError}}s:
```
domenic: Is this true? I thought the conversion behavior was generally to do things like replacing surrogate pairs or returning 0. In particular I am almost certain that converting an object to a long does not throw a TypeError.

index.bs
```diff
@@ -7529,6 +7570,82 @@ iterable |iterable| and an iterator getter
...
+    <table class="data">
+        <thead><th>Value</th><th>Passed to type</th></thead>
+        <tr><td><code>{"😞": 1}</code></td><td><code>{{dictionary}}&lt;ByteString, long></code></td></tr>
```
domenic: Consider never using `long` in examples, per https://w3ctag.github.io/design-principles/#numeric-types. (`long long` would be more appropriate in this case.)

domenic:
IMO we should omit class=algorithm from this PR and, if people think it's desirable, do it in a separate pass all at once.

bzbarsky:
> because EnumerableOwnProperties creates an array

If only it did. What it returns is a `List` of `Array`s, when used in "key+value" mode. ;)

index.bs
bzbarsky: Conversion to ByteString will throw if one of the 16-bit code units in the string is larger than 255.

But yes, the other two examples here would not throw: USVString just sticks in a replacement char and converting a random object to an integer type will produce 0.

index.bs
jyasskin: I've switched to EnumerableOwnProperties("key"), which returns a List.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
jyasskin: Good point. I've changed the examples to have a "result" column which shows the coercion.

index.bs
jyasskin: I used `double` since I don't need the integer conversion.

jyasskin:
I've done everything except the distinguishability information, which I'll work on now.

It looks impossible to enforce that the key is a string in an LL(1) grammar, so I've stayed with the semantic check. It would be possible if we put the key second, but I'm really reluctant to use that confusing syntax. I would be willing to make the key type required.

index.bs
```diff
@@ -7529,6 +7575,95 @@ iterable |iterable| and an iterator getter
...
+ECMAScript <emu-val>Object</emu-val> values.
+
+<p id="es-to-open-dictionary">
+    An ECMAScript value |O| is [=converted to an IDL value|converted=] to an IDL <code>{{OpenDictionary}}<|K|, |V|></code> value as follows:
+</p>
+
+<ol class="algorithm">
+    1.  Let |result| be a new empty instance of <code>{{OpenDictionary}}<|K|, |V|></code>.
+    1.  If [=Type=](|O|) is <emu-val>Undefined</emu-val> or <emu-val>Null</emu-val>,
+        return |result|.
```
domenic: This should throw a TypeErro for consistency with closed dictionaries.

index.bs
```diff
@@ -7529,6 +7575,95 @@ iterable |iterable| and an iterator getter
...
+    1.  If [=Type=](|O|) is not <emu-val>Object</emu-val>,
+        <a lt="es throw">throw a <emu-val>TypeError</emu-val></a>.
```
domenic: This check should not be there, for consistency with closed dictionaries.

index.bs
```diff
@@ -7529,6 +7575,95 @@ iterable |iterable| and an iterator getter
...
+    1.  Let |keys| be [=?=] [=EnumerableOwnProperties=](|O|, "key").
+    1.  Repeat, for each element |key| of |keys| in [=List=] order:
+        1.  If [=!=] [=HasOwnProperty=](|O|, |key|) is <code>true</code>, then:
```
domenic: Interesting. So this comes in to play if enumerating/getting causes side effects that delete the property. I don't think that's great semantics though. I think instead you want to inline the EnumerableOwnProperties algorithm here yourself, instead of separating the enumeration pass from the conversion/Get pass.

index.bs
```diff
@@ -7529,6 +7575,95 @@ iterable |iterable| and an iterator getter
...
+        <thead><th>Value</th><th>Passed to type</th><th>Result</th></thead>
+        <tr>
+            <td><code>{"😞": 1}</code></td>
+            <td><code>{{OpenDictionary}}&lt;ByteString, double></code></td>
+            <td><emu-val>TypeError</emu-val></td>
+        </tr>
+        <tr>
+            <td><code>{"\uD83D": 1}</code></td>
+            <td><code>{{OpenDictionary}}&lt;USVString, double></code></td>
+            <td>[ "\uFFFD" ⇒ 1 ]</td>
```
domenic: It seems you are inconsistent about notation for the mappings; here you use an arrow, whereas above you use "Repeat, for each mapping (key, value) in D". Both seem reasonable, but it would be good to pick one and maybe mention it in the section that introduces the concept of mappings.

index.bs
jyasskin: This was in response to @bzbarsky's «Do we want to allow passing null or undefined to mean "empty open dictionary" the way we do for dictionaries?», and he's right that closed dictionaries [treat null and undefined as empty](https://heycam.github.io/webidl/#es-to-dictionary).

index.bs
jyasskin: Again, closed dictionaries do throw a `TypeError` for non-Object,Null,Undefined arguments: https://heycam.github.io/webidl/#es-to-dictionary

index.bs
jyasskin: I'd like you, @annevk, and @bzbarsky to come to some conclusion about this. So far, @annevk [wondered](https://github.com/heycam/webidl/pull/180#issuecomment-250673458) if we should match [StructuredClone()](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredclone:structuredclone-4), and @bzbarsky [said](https://github.com/heycam/webidl/pull/180#issuecomment-250836913) that MozMap is like StructuredClone but omits the HasOwnProperty() check (so, I think we'd include the key in the map but initialize it from `undefined`). I don't care which we do.

index.bs
jyasskin: Yeah. I don't want to make WebIDL's users type a non-ascii character in order to iterate over a map, so (key, value) is better for that. I can change to that here too.

index.bs
domenic: Uh, what are you reading? Can you quote? I see there a very clear step 1 that says "If Type(V) is not Undefined, Null or Object, then throw a TypeError."

index.bs
domenic: Again, can you quote? There's no such step that I can see; only for RegExp, which we are planning to remove (#148 #145).

index.bs
domenic: Fair enough. @annevk, @bzbarsky, what do you think of my suggestion? I'm not sure why we do what we do in StrucuredClone, and I suspect legacy constraints.

index.bs
domenic: Oh, wow, I totally misread the sentence that I quoted. Sorry. This seems good then.

index.bs
domenic: As above, my bad; disregard this.

tobie:
@jyasskin [wrote](#issuecomment-250880371): 
> It looks impossible to enforce that the key is a string in an LL(1) grammar, so I've stayed with the semantic check. It would be possible if we put the key second, but I'm really reluctant to use that confusing syntax. I would be willing to make the key type required.

So I know just enough about grammars to be able to read them and defer to those more knowledgeable on the topic (pointers to good resources are welcomed, btw). I concur that putting the key second is a really bad idea and would suggest making both keys required. It's slightly more verbose, but also much clearer. I imagine we'd go for something like this?

```
OpenDictionaryType ::
    dictionary < OpenDictionaryKeyType , Type >
    
OpenDictionaryKeyType ::
    "DOMString"
    "USVString"
    "ByteString"
```

index.bs
jyasskin: Done.

index.bs
```diff
@@ -3249,7 +3256,6 @@ extended attribute’s [=takes a named argument list|named argument list=].
 
 Two types are <dfn id="dfn-distinguishable" export>distinguishable</dfn> if
 at most one of the two [=includes a nullable type=]
-or is a [=dictionary type=],
```
jyasskin: This seems to be redundant with the table.

index.bs
```diff
@@ -3280,7 +3286,7 @@ and at least one of the following three conditions is true:
                     <span>callback function</span>
             </div></th>
                 <th><div>
-                    <span>dictionary</span>
+                    <span>[open] dictionary</span>
```
jyasskin: Is this notation ok to say that open dictionaries are distinguishable like dictionaries?

index.bs
```diff
@@ -5219,6 +5226,17 @@ type.
         "Promise" "&lt;" ReturnType "&gt;"
 </pre>
 
+<pre class="grammar" id="prod-OpenDictionaryType">
+    OpenDictionaryType :
+        "OpenDictionary" "&lt;" Type OpenDictionaryTypeMaybeValue
```
jyasskin: Before switching this to require both arguments, I'd like a couple more people to agree with @tobie.

index.bs
annevk: Maybe @ajklein can tell us? Would rather we make an informed decision here.

index.bs
annevk: It seems fine to me if you can also typedef the thing. We don't have lots of these and first argument being optional is weird.

index.bs
tobie: So others might have different opinions here, but overall, I feel like this section is rather complicated and I'm not sure this actually makes it more readable.

bzbarsky:
As I said above, I'm not convinced that we want to call this thing "dictionary" when it has a totally different processing model from the things we already call "dictionaries"....

index.bs
tobie: The table doesn't need to be satisfied if you meet either one of the following two conditions (I'm not too sure they apply frankly, still reading up on union types specifics).

annevk:
You mean in terms of syntax? Is openDictionary<> much better?

bzbarsky:
I mean in terms of not having the word "dictionary" in it at all.

I'm not sure why the existing thing is named "dictionary" instead of "optionsobject" or something, by the way.  But the new thing is not really meant to be used for options objects, right?  It's meant to provide a key/value store with non-repeating keys.  And I assume we're not using a Map because those a bit of a PITA to create.

Call the new thing keyvalueset?  Having your sets be ordered is not exactly new in ES either....

annevk:
Hmm. @heycam?

tobie:
So I agree with @bzbarsky, here. As I [noted above](#discussion_r81783877), it's unclear to me from reading the updated spec what precisely this new construct shares with regular dictionaries. I found that rather confusing and I'm concerned I won't be alone doing so.

@jyasskin also mentioned [possible future extensions](https://www.w3.org/Bugs/Public/show_bug.cgi?id=20158#c18) with half-open dictionaries in mind. What's the plan with that? As future plans to tie the two together (or the absence of such) might help us name these things properly. My feeling is we could easily add openness to `dictionary`, but combining the type requirements of this new construct won't be trivial (and there actual use cases for it?).

index.bs
bzbarsky: I doubt there is much in the way of legacy stuff around structured clone here, honestly.

I _think_ structured clone was created back before there were proxies in ES (but I could totally be wrong about that!).  If that were the case, then the [[Get]] calls in that algorithm would have been the only thing that could have changed whether props exist, so it would have made sense to double-check that a previos [[Get]] has not removed the property or something.

In the proxy world, where HasOwnProperty can itself lie and remove the property or whatnot, and where [[GetOwnProperty]] can do likewise, it's not clear to me that there is any sort of sanity that can be provided in the face of a "malicious" input object that just tries to mess with you.  So the only question is whether there are sane cases which would benefit from the extra double-checking.  

In EnumerableOwnProperties(), the double-checking that exists is just a side-effect of checking for enumerability.  That _happens_ to also check that the property is still own, because it uses [[GetOwnProperty]].  But that looks to me like an accidental byproduct of the enumerability check.

Note that structured clone, as specified, is buggy: if one of those [[GetOwnProperty]] calls in step 22.3.2.1.1 returns undefined, what happens?

Also, it's not clear to me whether UAs actually match the structured clone spec.  Might be worth checking that...

I guess all of that is to say that I see no obvious problem with removing the HasOwnProperty check in structured clone, with the understanding that this may cause the clone to have own props where the clonee doesn't have any by the time cloning is done... but that's possible now anyway, afaict.

jyasskin:
The new construct is a value object like dictionaries and sequences, rather than a reference like objects, FrozenArrays, Maps, and Sets.

Half-open-ness might, of course, never happen, but I think we have to do it by initializing the dictionary members, and then iterating over all keys to initialize the mappings, skipping the ones that match dictionary members.

tobie:
@jyasskin it seems we have a number of different requirements here:

- specify members vs. allow openness,
- type coerce keys vs. don't,
- type coerce all values identically vs. allow per member coercion.

Are these things reconcilable in a single data structure?

i.e. should we have half-openness down the road, would we want the types on dictionary to apply to all members (including the ones specified) or just to unspecified (open) members, etc.

bzbarsky:
> The new construct is a value object like dictionaries and sequences, rather than a reference like objects, FrozenArrays, Maps, and Sets.

Sure, but my point is that we could have a pass-by-value "ordered key-value store" thing whose ES representation is a Map object.  The only reason not to do that is presumably ergonomics.

jyasskin:
> Sure, but my point is that we could have a pass-by-value "ordered key-value store" thing whose ES representation is a Map object. The only reason not to do that is presumably ergonomics.

Sorry, yes, it's only not a `Map` because it's easier to enter and read a plain JS object.

We could allow a `Map` IDL argument to be initialized from a plain JS object when the key type is appropriate. However, that would be worse for Web Bluetooth's use case until we add an algorithm for allowing non-string keys, because I'd have to write `Map<DOMString, DataFilter>` instead of `Map<unsigned short, DataFilter>`.

annevk:
The way I see it:

* Dictionaries: map constrained by specific key members and corresponding value types
* Open dictionaries: map constrained by key type and value type
* Half-open dictionaries: map constrained by key type, fallback value type, and specific value type for specific key members

jyasskin:
+1 to @annevk's description.
There are differences in the order that fields are retrieved from the JS object, but this only matters in obfuscated-javascript competitions, so it shouldn't prevent us from naming the new types intuitively.

bzbarsky:
It's not just differences in the order; it's also differences in which properties are considered at all.  The "open" part only considers own props, but the "closed" part looks at the proto chain...

bzbarsky:
One more thing worth thinking about: a how would a JS library implement this sort of object?  Probably via for-in (hence considering proto props) , or via Object.getOwnPropertyNames.  Would it check for enumerability in the latter case?

The other thing I find myself coming back to is what the actual use cases are in practice.  We have `Headers`, which really does have an open-ended list of things.  We have the animation stuff, which really has a fixed set of possible props, but typically sets a few out of hundreds and doesn't want the performance impact of the dictionary processing model.  We have the worklet things, right?

Would typical consumers have basically an object literal here, or an object that comes from "somewhere" or a thing like a Map that they have to adapter into looking like an object for purposes of these APIs?  Would it make any sense to design the API surface such that you can pass in either a Map or a random other object, and then extract values from the passed-in thing via `.entries()` (possibly the initial value) and `Object.entries` respectively?  I guess that would have the drawback that we would no longer be guaranteed key uniqueness and would have to handle repeating keys... but we have to do that anyway if we use [[OwnPropertyKeys]], because nothing enforces that [[OwnPropertyKeys]] returns a list with no repeats in it, as far as I can tell.

tobie:
FWIW, repeated keys are actually kind of useful for `Headers` and `UrlParams`.

bzbarsky:
Headers currently supports explicitly repeated headers via the `sequence<sequence<ByteString>>` thing.  Which means you can already hand it any iterator you want and it will do the right thing with it.  The question is whether it makes sense to have a thing where input objects and maps get auto-coerced into iterators.  For Headers this seems totally fine.  What I'm less clear on is other use cases.

index.bs
jyasskin: Ok, I've made it require both arguments.

annevk:
@tobie and I were wondering what the reason is for prototype properties to not be considered for open dictionaries. I'm not sure @jyasskin motivated the difference with dictionaries in this thread and it might not have been intentional.

> We have the animation stuff, which really has a fixed set of possible props,

This is not true. As far as I can tell https://w3c.github.io/web-animations/#processing-a-keyframes-argument supports custom properties, so it's open-ended.

It does seem that algorithm although first treating things like dictionaries, then also doesn't consider prototype properties. What is the motivation for that, @birtles?

index.bs
annevk: I think what you're missing is that structured clone fails on proxies in step 19.

domenic:
I think we should leave this Map business for a separate follow-up. In general I think using `sequence<sequence<>>` is the way to do that, maybe with some IDL-provided machinery for converting it into an IDL type or something.

> One more thing worth thinking about: a how would a JS library implement this sort of object? Probably via for-in (hence considering proto props) , or via Object.getOwnPropertyNames. Would it check for enumerability in the latter case?

(/cc @wycats since this touches on things he's had opinions on in recent TC39. For context, we are discussing a new "open dictionary" type to be used for things like HTTP headers, and trying to figure out how to enumerate its properties.)

I think old JS libraries would use for-in; old libraries that followed advice in certain books would use for-in + a hasOwnProperty guard; new ones would use `Object.keys` or `Object.assign`. Future ones might use `{ ...objectSpread }` (see [stage 3 proposal](https://github.com/sebmarkbage/ecmascript-rest-spread)).

In general TC39 has recognized this issue and settled on "own enumerable" as the protocol for these sorts of "record" types. We had a pretty extensive discussion at last week's TC39 about a very related problem when trying to decide the semantics of object spread. We framed it as "records vs. classes". Object spread is designed specifically around the "record" use case, and previous decisions (Object.keys/values/entries, Object.assign, probably some more I am forgetting) have settled on "own enumerable" as the way of determining the keys when using objects as records.

Trying to use class instances as "records" falls down pretty hard, throughout the language, due to this decision. The example we discussed last week was refactoring from `{ x, y }` to an instance of `class Point { get x() { ... } get y() { ... } }`; the refactoring will destroy the value's "recordness". Our conclusion was "that's sad, but it's what we've got for now", plus speculative talk of in the future providing a specific opt-in "record fields protocol" so that classes could bridge the gap and "become" records by declaring how to convert themselves to a record. Then various parts of the language might start automatically respecting that protocol.

In the end there's always going to be a fundamental mismatch between this own-enumerable design and the fact that the natural property access in JavaScript, of `obj.prop`, will consult non-enumerable and non-own properties. I don't think there's much to do there. But I do agree with @jyasskin's intuition that this mostly matters in obfuscated JavaScript contests. The 95% case of `myAPI({ prop1: val1, prop2: val2 })` is not affected.

So in conclusion, I am in favor of sticking with enumerable-own for open dictionaries too.

birtles:
Yes, the custom property support is a recent addition to the keyframe processing algorithm so it's now open-ended (although we only read properties that meet certain criteria&mdash;being recognized animatable properties of having a name that matches custom property syntax, so I guess that makes it not quite open-ended).

For the specific enumeration strategy, I mailed public-script-coord and [was told to use [[Enumerate]]](https://lists.w3.org/Archives/Public/public-script-coord/2014AprJun/0037.html). Then when I went to make the changes I received feedback from @bzbarsky and @heycam that I should probably just use `[[EnumerableOwnNames]]` (see discussion starting https://github.com/w3c/web-animations/issues/107#issuecomment-148103916).

jyasskin:
I knew I was ignoring the prototype chain, but I don't have any reason for doing that differently from dictionaries. If @domenic has convinced people that I got it right, then I'll leave it this way. If folks can convince him I got it wrong, I'm happy to change it. I have absolutely no opinions on the name. If we want to call this `record<K, V>` that's fine with me.

annevk:
I like record given the different semantics and upcoming JS features. However, not sure how to tackle a future half-open. Should that first do the dictionary thing and then enumerate and ignore the keys already handled? That is, what semantics should web animations keep/get.

domenic:
> However, not sure how to tackle a future half-open. Should that first do the dictionary thing and then enumerate and ignore the keys already handled? That is, what semantics should web animations keep/get.

That's a great question. We basically have two options:

1. enumerate and copy, then pluck properties
2. pluck properties, then enumerate and copy, ignoring already-plucked properties

I really don't think it matters much. If I had to pick one I'd go for (1).

index.bs
jyasskin: FWIW, #183 will simplify this too: once it lands, I'll put OpenDictionary/record in the dictionary-like category.

tobie:
Recapping here, as although the contentious parts of this design are actually quite small, they directly affect Web-developer-facing APIs and I want to make sure we do the right thing.

### Context

Currently, the platform let’s you specify passing an options bag to operations using plain JS objects, e.g.:

```webidl
interface Foo {
    void doStuff(DoStuffOptions);
};

dictionary DoStuffOptions {
    boolean prop;
    boolean anotherProp;
    DOMString? name;
};
```

This is then used in JS like so:

```js
foo.doStuff({ prop: true, anotherProp: false, name: "blah" });
```

Note that as this essentially calls the [Get abstract operation](https://tc39.github.io/ecma262/#sec-get-o-p) of the option bag, it reaches into the prototype chain. This includes getters and setters on the object’s prototype, the common way for attributes to be exposed on the platform. This enables neat use cases such as using the attributes exposed on a previous request to set the options of a new one:

```js
var r = new Request(url, currentRequest);
// where currentRequest is itself an instance of Request.
```

### The Case for Open Dictonaries

For cases where you don't know beforehand all the possible names dictionary members can take (maybe because this could be application specific, e.g.: URL parameters, or specified in a different layer of the platform, e.g.: HTTP headers), you wan't to be able to specify an open (or semi-open dictionary).

The exact syntax by which this will be done in WebIDL is not super important, but currently, here’s what's considered:

```webidl
interface Foo {
    void doStuff(record<DOMString,DOMString>);
};
```

or:

```webidl
interface Foo {
    void doStuff(dictionary<DOMString,DOMString>);
};
```

The above simply means that you can pass the `.doStuff()` method a plain JS object with strings as keys and strings as values. The WebIDL won't tell you anything about what keys are acceptable though prose around it can. Only strings are supported as key types for now (there’s three types of strings in WebIDL, hence the need to specify the key type), and values could take any types, including unions of types.

### Developer Expectations

**From a web developer’s perspective, the precise syntax of this WebIDL construct has little importance; Web developers encounter such APIs essentially through examples** along the lines of:

```js
bar.doSomething({ k: true, y: false, z: 123 });
```

And so how these APIs are actually specified doesn’t (and shouldn’t) affect Web developers. For them, they all look the same and should thus behave the same.

There’s a catch, here, though, which may lead to subtle bugs in application-level code.

### Difference in Design between Open and Regular Dirctionaries

While properties of the current dictionary type reach into the prototype chain, the proposed design for open dictionaries (the `record<KeyType,ValueType>` or `dictionary<KeyType,ValueType>` syntax suggested above) has a similar design to structured cloning as it relies on [EnumerableOwnProperties](https://tc39.github.io/ecma262/#sec-enumerableownproperties) and therefore **only looks at the properties found on the object itself, not those on the prototype chain**.

So for example, doing something like this wouldn’t yield the expected outcome at all:

```js
var defaultOptions = {
    color: "blue",
    textSize: "16px",
    fontFamily: "Times New Roman"
};

var options = Object.create(defaultOptions);
options.textDecoration = "underline";
CSS.toCSSText(options);
```

which would yield:

```js
"text-decoration: underline;"
```

if specified as:

```webidl
namespace CSS {
  DOMSTring toCSSText(record<DOMString,DOMString> styles);
};
```

but:

```js
"text-decoration: underline; color: blue; text-size: 16px; font-family: Times New Roman"
```
if specified like so:

```webidl
namespace CSS {
  DOMSTring toCSSText(StyleRules styles);
};

dictionary StyleRules {
  DOMString color;
  DOMString textSize;
  DOMString fontFamily;
  DOMString textDecoration;
};
```

### Further Concerns with Half-Open Dictionaries

To make matters worse, the situation would get even more confusing with half-open dictionaries (pseudo syntax—as these are still in the works):

```webidl
namespace CSS {
  DOMSTring toCSSText(StyleRules styles);
};

dictionary<DOMString,DOMString> StyleRules {
  DOMString color;
  DOMString textSize;
};
```

Which in our above configuration would yield (missing `fontFamily`):

```js
"text-decoration: underline; color: blue; text-size: 16px;"
```

### Work-Around

Arguably, the following would give the desired/expected outcome in all above cases:

```js
var options = Object.assign({ textDecoration: "underline" }, defaultOptions);
CSS.toCSSText(options);
```

### Other Problematic Example

But the following wouldn’t:

```js
function DefaultOptions(options) {
    for (k in options)
         this[k] = options[k];
}

DefaultOptions.prototype.fontSize = "16px";
// etc…

var options = new DefaultOptions({ textDecoration: "underline" });
CSS.toCSSText(options);
```

### Different Design Solutions and Their Respective Issues

#### 1. Leave `dictionary` As Is, Add `dictionary<T,T>` which doesn't reach into the Prototype Chain

This is the current design. As described above, its main issue is that it reaches into the prototype chain for specified members but not for unspecified ones. It's main advantage is that it maps the newly clarified "records vs. classes" semantics TC-39 agreed upon as described in @domenic's comment above, see https://github.com/heycam/webidl/pull/180#issuecomment-251669022.

#### 2. Leave `dictionary` As Is, Add `record<T,T>` which doesn't reach into the Prototype Chain

This is a variant of the current design. It has the same issues and benefits as the previous ones, but makes it clear to API designers and implementers that the behavior is different. This doesn't clarify the issue for web developers, though.

#### 3. Leave `dictionary` As Is, Add `dictionary<T,T>` which also reaches into the Prototype Chain

This solves the issues raised above but has the dis-advantage of not matching the way structured cloning works or the "records vs. classes" semantics TC-39 seems to have settled on with the platform.

#### 4. create a new `record<T,T>` type, that covers open, half-open, and closed dictionaries and that never reaches into the prototype chain. Deprecate existing dictionaries and migrate APIs to this new type.

As dictionaries are passed by value (and not by reference), this can be explained pretty much the same way structured cloning is.

**Benefits:** consistency across the different open, half-open and closed dictionary types, structured cloning and the way TC-39 envisions records vs. classes.

**Issues:** It is possible that not reaching into the prototypes of dictionaries _at all_ would stump developers even more than the possible discrepancy between types in the current proposal. This would require migrating existing APIs currently using dictionaries to this new type. This is arguably a lot of work and might cause subtle breakage of the Web. It would also require being explicit about things we get implicitly with dictionaries such as in the `new Request(url, currentRequest)` example given above. So the [Request [[Constructor]]](https://fetch.spec.whatwg.org/#request-class) would need to be rewritten as:

```webidl
typedef (RequestInit or Request) RequestOptions;

[Constructor(RequestInfo input, optional RequestOptions init), Exposed=(Window,Worker)]
interface Request {
  // ...
};
```

### TL;DR

I'm concerned about the Web developer experience of having APIs that look the same but behave slightly differently. I also think that, as much as possible, we should avoid having different types for things that look exactly the same on the Web developer side. I'd love to have a few more eyeballs on this issue, especially folks with a Webdev background. I might also be overthinking this, and maybe @jyasskin's intuition that this is a non-issue is the right one.

domenic:
One point I think is missing in that analysis is that the records vs. classes perspective TC39 has arrived at is explicitly derived from developer expectations and common coding patterns, as I tried to point out. Developers, in their own code, work with open-ended records via methods like `Object.keys` or `Object.assign`. They do not use for-in loops (at least not without a hasOwnProperty guard). In other words, this isn't a TC39 vs. developers dichotomy; I think TC39 and developers are fairly aligned on this.

So I think options 3 and 4 would be surprising to developers, as it would cause platform code to behave in unexpected ways that are not natural for normal JavaScript libraries.

tobie:
> So I think options 3 and 4 would be surprising to developers, as it would cause platform code to behave in unexpected ways that are not natural for normal JavaScript libraries.

So option 3 would be surprising because developers don't expect prototypes to be enumerated and option 4 because developers expect `options.foo` to reach into the prototype chain, right?

domenic:
Yeah, exactly.

tobie:
I guess what bothers me about option 3 being surprising, is developer's won't know they're using an open dictionary vs. a closed one unless they actually look at the WebIDL (which in practice they never do).

Whether the editor chooses to rely on an open dictionary or a closed one can be an editorial decision in lots of cases. For example, for HTTP headers, every existing HTTP header could be specified as a dictionary member (thus reaching into the prototype chain) and unspecified ones wouldn't (and thus would be enumerated instead. But  I coud equally imagine this spec'ed as a pure open dictionary (thus never reaching into the prototype chain at all). The reasons which would push an editor to choose one over the other option wouldn't be API semantics at all, but other concerns such as spec maintainability or the stability of references. _This_ is precisely what bothers me with the current proposal.

bzbarsky:
Note https://github.com/whatwg/fetch/issues/164#issuecomment-252752477 about how some current impls do this sort of thing right now...

In particular, for the `Headers` case Firefox does Tobie's option 2, Chrome does Tobie's option 3, and the fetch polyfill does something that considers only own props, but includes non-enumerable ones....

jussi-kalliokoski:
My personal expectation as a developer is that these work as they would were they implemented with simple JS, so (I think) that means following destructuring semantics:

```
{knownKey} = record
{...headers} = dict
```

annevk:
All things considered I'm still happy with 2.

tobie:
Thanks for the post-summary feedback folks. Having [toyed a bit more with Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Cstage-0&experimental=false&loose=false&spec=false&code=var%20o%20%3D%20Object.create(%7B%20x%3A%20%22x%22%2C%20y%3A%20%22y%22%20%7D)%0AObject.defineProperties(o%2C%20%7B%0A%20%20a%3A%20%7Benumerable%3A%20true%2C%20value%3A%20'a'%7D%2C%0A%20%20b%3A%20%7Benumerable%3A%20true%2C%20value%3A%20'b'%7D%2C%0A%20%20c%3A%20%7Benumerable%3A%20false%2C%20value%3A%20'c'%7D%2C%0A%20%20d%3A%20%7Benumerable%3A%20false%2C%20value%3A%20'd'%7D%0A%7D)%3B%0A%0Alet%20%7Ba%2C%20c%2C%20x%2C%20...rest%7D%20%3D%20o%3B%0Aconsole.log(a)%3B%0Aconsole.log(c)%3B%0Aconsole.log(x)%3B%0Aconsole.log(rest.b)%3B%20%0Aconsole.log(rest.d)%3B%0Aconsole.log(rest.y)%3B%0A%0A%0A) overnight, and following @domenic's and @jussi-kalliokoski's comments above, I'm now convinced option #1 (so @jyasskin's original proposal) is the right choice.

Domenic's right, this is the expected behavior of records; they reach into the prototype chain when using `[[Get]]` and don't when `spread`. Choosing a different name for open and closed dictionaries doesn't help here. And it will only make things more confusing once we have half-open dictionaries.

That said, we might want to add a note at the intention of editors saying open dictionaries match the semantics of `spread`, and should only be used in places where the available options are truly open-ended.

bzbarsky:
> Choosing a different name for open and closed dictionaries doesn't help here.

The question is whether it helps with specification clarity.  It's possible that using the same name for the two things is ok if it makes the spec simpler, or that it makes things more complicated because now whenever you mention a dictionary you have to make it clear which one.

It's also _really_ confusing, in my opinion, that one of these is just a type and the other is a named construct that involves inheritance, etc.

So if we're serious about the half-open bit: what are we thinking of as the syntax for those?  Because it's worth thinking about that up front a bit if that's the real end goal, so we don't end up with weird contortions later with named type definitions inheriting from unnamed types and whatnot, like some of the proposals I've seen floating around.

tobie:
@bzbarsky: agreed. If we're going with half-open down the road, we may want to converge towards a single model that can have 0.._n_ members and may be open-ended, e.g.:

```webidl
// opended, no members
dictionary CSSStyles {
  openended<DOMString, DOMString>;
}

// openended, some defined members
dictionary PaintOptions {
  DOMString? fillPattern = "black";
  DOMString? strokePattern = null;
  Point position;
  openended<DOMString, any>;
};
```

bzbarsky:
That or:

    record<T1, T2>
    record<DictionaryType, T1, T2>

to indicate an open thing and a half-open thing using DictionaryType as the closed part.

That avoids issues about what happens if `openended<>` is defined on both a dictionary and another dictionary that inherits from it, or on a dictionary but _not_ a dictionary that inherits from it, and so forth.

tobie:
Trying to recap here. Seems we have agreement on:

*   difference between dictionaries, record and half-open-records ([quoting](#issuecomment-251471612) @annevk):
     *   **Dictionaries:** map constrained by specific key members and corresponding value types
     *   **Records** (open dictionaries): map constrained by key type and value type
     *   **Dictionary-based records** (Half-open dictionaries): map constrained by key type, fallback value type, and specific value type for specific key members.
*   syntax:

    ```webidl
    record<StringType, Type>; // open-ended dictionaries
    record<DictionaryType, StringType, Type>; // half-open
    ```
*   high-level behavior (hit the prototype chain for dictionary members, use `spread` for record mappings).

Still on the todo list:
- review and merge @jyasskin's PR to simplify the distinguishability table (#183),
- decide whether we need to tackle half-open dictionaries as part of this PR or if we can defer,
- figure out how to properly name half-open dictionaries,
- anything else?

annevk:
Half-open seems like a follow-up. Perhaps just dictionary, record, and record with dictionary as types?

domenic:
Agreed on record being separate.

Now that we've settled the high-level prototype-chain-vs.-not, I think there's a matter of what exact algorithm to use that was discussed earlier. https://github.com/heycam/webidl/pull/180#discussion_r81291068 is what I can find although I thought there was more.

The question is whether we should interleave Get()s while enumerating the properties and checking their enumerability, or enumerate and collect all the properties first, then Get() them. I would prefer an algorithm like [`Object.assign`](https://tc39.github.io/ecma262/#sec-object.assign) / [spread syntax](https://sebmarkbage.github.io/ecmascript-rest-spread/#AbstractOperations-CopyDataProperties) which interleaves the Get()s with the [[GetOwnProperty]]()s.

bzbarsky:
If we interleave the Get()s we should interleave the to-IDL-value conversions too so we don't have to build up a temporary list of values.  Imo.

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+        return |result|.
+    1.  If [=Type=](|O|) is not <emu-val>Object</emu-val>,
+        <a lt="es throw">throw a <emu-val>TypeError</emu-val></a>.
+    1.  Let |keys| be [=?=] |O|.[=[[OwnPropertyKeys]]=]().
+    1.  Repeat, for each element |key| of |keys| in [=List=] order:
+        1.  Let |desc| be [=?=] |O|.[=[[GetOwnProperty]]=](|key|).
+        1.  If |desc| is not <emu-val>undefined</emu-val>
+            and |desc|.[=[[Enumerable]]=] is <emu-val>true</emu-val>:
+            1.  Let |typedKey| be |key| [=converted to an IDL value=] of type |K|.
+            1.  Let |value| be [=?=] [=Get=](|O|, |key|).
```
jyasskin: @domenic Done. Note that it's a little weird to call `Get()` when we already have the descriptor, but that's what `Object.assign()` does so oh well.

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+IDL {{record}}&lt;|K|, |V|&gt; values are represented by
+ECMAScript <emu-val>Object</emu-val> values.
+
+<p id="es-to-record">
+    An ECMAScript value |O| is [=converted to an IDL value|converted=] to an IDL <code>{{record}}&lt;|K|, |V|></code> value as follows:
+</p>
+
+<ol class="algorithm">
+    1.  Let |result| be a new empty instance of <code>{{record}}&lt;|K|, |V|></code>.
+    1.  If [=Type=](|O|) is <emu-val>Undefined</emu-val> or <emu-val>Null</emu-val>,
```
domenic: Nit: types (Undefined, Null, and Object) are not emu-vals. That is reserved for language values, not spec values.

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+        1.  Let |desc| be [=?=] |O|.[=[[GetOwnProperty]]=](|key|).
+        1.  If |desc| is not <emu-val>undefined</emu-val>
+            and |desc|.[=[[Enumerable]]=] is <emu-val>true</emu-val>:
+            1.  Let |typedKey| be |key| [=converted to an IDL value=] of type |K|.
+            1.  Let |value| be [=?=] [=Get=](|O|, |key|).
+            1.  Let |typedValue| be |value| [=converted to an IDL value=] of type |V|.
+            1.  If |typedKey| is already a key in |result|, set its value to |typedValue|.
+
+                Note: This can happen when |O| is a proxy object.
+            1.  Otherwise, append to |result| a mapping from |typedKey| to |typedValue|.
```
domenic: Link mapping? And use (|typedKey|, |typedValue|) notation?

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+    1.  Repeat, for each [=record/mapping=] (|key|, |value|) in |D|:
+        1.  Let |esKey| be |key| [=converted to an ECMAScript value=].
+        1.  Let |esValue| be |value| [=converted to an ECMAScript value=].
+        1.  Let |created| be [=!=] [=CreateDataProperty=](|result|, |esKey|, |esValue|).
+        1.  Assert: |created| is <emu-val>true</emu-val>.
+    1.  Return |result|.
+</ol>
+
+<div class="example" id="example-es-record">
+    Passing the ECMAScript value <code>{b: 3, a: 4}</code> to a
```
domenic: s/to/as/

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+    <code>{{record}}&lt;DOMString, double></code> argument
+    would result in the IDL value <code>[ ("b", 3), ("a", 4) ]</code>.
```
domenic: I don't know about this new array-ish notation. I'd instead say something like "a record containing the two mappings ("b", 3) and ("a", 4), in that order." Hmm, I guess it makes an appearance later in the table too, where it's more necessary for space reasons... not sure what to do. At least, remove the `<code>` from this one, IMO.

I think this also contradicts the statement "There is no way to represent a constant record value in IDL."

index.bs
```diff
@@ -7695,6 +7837,9 @@ represented by ECMAScript values that correspond to the union’s
         1.  If |types| includes a [=dictionary type=], then return the
             result of [=converted to an IDL value|converting=]
             |V| to that dictionary type.
+        1.  If |types| includes a [=record type=], then return the
+            result of [=converted to an IDL value|converting=]
+            |V| to that dictionary type.
```
domenic: s/dictionary/record

index.bs
```diff
@@ -7529,6 +7576,101 @@ iterable |iterable| and an iterator getter
...
+    An ECMAScript value |O| is [=converted to an IDL value|converted=] to an IDL <code>{{record}}&lt;|K|, |V|></code> value as follows:
+</p>
+
+<ol class="algorithm">
+    1.  Let |result| be a new empty instance of <code>{{record}}&lt;|K|, |V|></code>.
+    1.  If [=Type=](|O|) is <emu-val>Undefined</emu-val> or <emu-val>Null</emu-val>,
+        return |result|.
+    1.  If [=Type=](|O|) is not <emu-val>Object</emu-val>,
+        <a lt="es throw">throw a <emu-val>TypeError</emu-val></a>.
+    1.  Let |keys| be [=?=] |O|.[=[[OwnPropertyKeys]]=]().
```
domenic: IMO the links for the internal methods and [[enumerable]] are not good to add here. If we want to do that more generally, we can, but we shouldn't do it just in this section.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
domenic: https://bterlson.github.io/ecmarkup/#editorial-conventions I guess, although that's not the most obvious place to look given that we're basically only using emu-val.

index.bs
jyasskin: I think
> There is no way to represent a constant record value in IDL.

means that we don't have syntax for it in the WebIDL grammar. The phrase,
> a record containing the two mappings ("b", 3) and ("a", 4), in that order.

would still "represent a constant record value", but I'd like to have a more concise way to do it.

index.bs
jyasskin: Whoops, done.

index.bs
jyasskin: Perfect, thanks!

bzbarsky:
@tobie, @jyasskin, @domenic Are we sure we want the "is already present" behavior as checked in?  That seems a bit annoying to implement in terms of performance, and it's not really clear to me why we're doing it at all.

jyasskin:
@bzbarsky I don't see "is already present" in the spec. Do you mean the "If _typedKey_ is already a key in _result_," wording? That's the same operation as in [`Map.prototype.set`](https://tc39.github.io/ecma262/#sec-map.prototype.set), which I believe we implement efficiently. I do think we want to avoid the possibility of a `record` winding up with a duplicate key: it can only happen with a proxy object input, and spec authors are unlikely to consider it and handle it well.

bzbarsky:
> Do you mean the "If typedKey is already a key in result," wording?

Yes.

> That's the same operation as in Map.prototype.set, which I believe we implement efficiently.

It's possible to do.  It just requires some performance/memory/complexity tradeoffs, which it's not clear to me are valuable here.

> I do think we want to avoid the possibility of a record winding up with a duplicate key

I guess the question is whether it's better to have implementation complexity/overhead to handle the rare proxy case or to have specification complexity instead.

https://dev.w3.org/html5/html-design-principles/#priority-of-constituencies would normally err on the side of specification complexity, but this is an odd case where we are likely to have more specifiers affected than implementors, so maybe this is the right tradeoff...

In any case, I just wanted to make sure we are explicitly and deliberately making this choice and understand that it's likely to cause implementation delays, in general.

jyasskin:
We could also throw a `TypeError` in the case of a duplicated key, which would allow the implementation to just keep a set of used keys, instead of needing to be able to find the value associated with a previously-used key.

bzbarsky:
That makes things a bit simpler, but not really _that_ much simpler.  You still need some sort of dual-purpose data structure (O(1) lookup and fixed-order iteration).

annevk:
The specification complexity would be making that duplicate key check and then throwing? Seems better handled by IDL, or would specifications handle it in other ways?

bzbarsky:
> The specification complexity would be making that duplicate key check and then throwing?

Or whatever they want to do in the duplicate case.  @jyasskin is probably right that spec authors won't even realize they should handle it, so it probably does make sense to handle in IDL.  It just makes me a bit sad that ES proxies don't enforce this themselves per spec.  :(

domenic:
It looks like there is talk of proxies enforcing that in https://github.com/tc39/ecma262/issues/461. I am not sure whether there was consensus to change though...

Issue #59, opened 2015-09-27
"it is not possible for a single platform object to implement both interfaces"


domenic:
Apparently two IDL interface types are distinguishable if all of the following hold
- The two identified interfaces are not the same
- It is not possible for a single platform object to implement both interfaces
- It is not the case that both are callback interfaces.

This second condition is very confusing. I can think of two possible interpretations:
### Interpretation A

```
[NoInterfaceObject] interface Foo {};
[NoInterfaceObject] interface Bar {};

interface Baz {};
Baz implements Foo;
Baz implements Bar;
```

Baz implements both Foo and Bar, so Foo and Bar are not distinguishable. (Contrapositive of the given statement.)
### Interpretation B

```
[NoInterfaceObject] interface Foo { readonly attribute x; };
[NoInterfaceObject] interface Bar { attribute x; };

interface Baz {};
Baz implements Foo;
Baz implements Bar; // not allowed
```

It is not possible to implement both Foo and Bar at the same time, so they are distinguishable.
### Discussion

Both of these interpretations kind of suck.

Interpretation A is a very nonlocal definition. To determine if A and B are distinguishable, you have to look at all interfaces in the web platform, and find out if any of them implement both A and B. (I think something that extends A and implements B would also work, or vice-versa.) This means A and B's distinguishability can change at any time.

Interpretation B is very implicit. It comes down to finding all places in the spec that prohibit implementing a given interface. In this case I found the condition

> For a given interface, there MUST NOT be any member defined on any of its consequential interfaces whose identifier is the same as any other member defined on any of those consequential interfaces or on the original interface itself. 

but there are others, e.g. the no-cycles condition.

Which is it? And, is this actually implementable, or implemented by existing binding layers?

bzbarsky:
The point here is that if two types are distinguishable, then given any object you have to be able to tell which of the two types it is.

This means that if there is an object which can claim to be of both types, they need to not be distinguishable.  Therefore interpretation A is the correct one and we should adjust phrasing as needed to make that clear.

> To determine if A and B are distinguishable, you have to look at all interfaces in the web platform

Yes, you do.  That's necessary to achieve the "given any object" goal.  I agree that this is rather annoying, but I don't see how the goal can be achieved otherwise.

> And, is this actually implementable, or implemented by existing binding layers?

What Gecko's binding layer implements is the following:

[...]

PR #183, opened 2016-10-04
Simplify the distinguishable table.

### LINKED ISSUE #50: Distinguishability of buffers source types needs to be defined better
Right now the spec says that "buffer source types" are not distinguishable from each other.  But this is clearly wrong, since if that were the case the `BufferSource` typedef would not be valid.

I believe the right behavior here is that `ArrayBuffer` is distinguishable from everything that's not `ArrayBuffer`, `ArrayBufferView` is distinguishable from everything that's not `ArrayBufferView` or a typed array type, and a typed array type is distinguishable from everything except `ArrayBufferView` and itself.

I'm not sure how best to put this in the form of this table, because it would make the table huge...

Perhaps the simplest thing to do is to define that distinguishability works as if these were all interface types with the typed array interfaces considered as inheriting the `ArrayBufferView` interface.

jyasskin:
Primarily by grouping types into larger categories, but also by recategorizing callback interfaces as more like dictionaries than normal interfaces.

Fixes #50 and #59.

***
[Preview](https://cdn.rawgit.com/jyasskin/webidl/6a5406a/index.html) | [Diff w/ current ED](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F6a5406a%2Findex.html) | [Diff w/ base](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2Ffef2cda%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fjyasskin%2Fwebidl%2F6a5406a%2Findex.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -203,6 +203,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         #distinguishable-table {
           font-size: 80%;
           border-collapse: collapse;
+          width: auto;
         }
 
         #distinguishable-table th {
@@ -3315,11 +3316,24 @@ the following algorithm returns <i>true</i>.
     with the non-union type,
     or <i>false</i> otherwise.
 1.  If the two types (taking their [=inner types=]
-    if they are [=nullable types=]) appear
+    if they are [=nullable types=]) appear or are in categories appearing
     in the following table and there is a “●” mark in the corresponding entry
     or there is a letter in the corresponding entry and the designated additional
     requirement below the table is satisfied, then return <i>true</i>.
     Otherwise return <i>false</i>.
+
+    Categories:
+    <dl>
+        <dt>interface-like</dt>
+        <dd>
+            * [=interfaces=]
+            * [=exception types=]
+            * [=buffer source types=]
+        <dt>sequence-like</dt>
+        <dd>
+            * <code>[=sequence=]&lt;T></code>
+            * <code>{{FrozenArray}}&lt;T></code>
+    </dl>
 
     <table id="distinguishable-table" class="matrix data complex">
         <tr>
@@ -3334,11 +3348,11 @@ the following algorithm returns <i>true</i>.
                     <span>string types</span>
             </div></th>
                 <th><div>
-                    <span>interface</span>
-            </div></th>
-                <th><div>
                     <span>object</span>
             </div></th>
+                <th><div>
+                    <span>interface-like</span>
+                </div></th>
                 <th><div>
                     <span>callback function</span>
             </div></th>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -3481,9 +3391,9 @@ and at least one of the following three conditions is true:
     </table>
 
     <ol type="a">
-        1.  The two identified [=interfaces=] are
-            not the same, it is not possible for a single [=platform object=]
-            to implement both [=interfaces=],
+        1.  The two identified interface-like types are
+            not the same, no single [=platform object=] implements both
+            interface-like types,
             and it is not the case that both are [=callback interfaces=].
```
jyasskin: I suspect we can make this easier to read by splitting callback interfaces out of interface-like, and maybe moving them into a dictionary-like category. I didn't do that yet in order to make this change easier to review, but I could if y'all want.

tobie:
You might want to add a note saying that types that appear in the same category are undistinguishable.

tobie:
While you're at it, what about replacing:

> Two types are distinguishable if at most one of the two includes a nullable type…

by:

> Two types are distinguishable if they are not both (a) nullable (type)…

jyasskin:
> You might want to add a note saying that types that appear in the same category are undistinguishable.

The normative part of that is covered by the lack of a ● in all foo/foo cells except for interface/interface. Are you saying that's not clear enough, so I should clarify it with a note?

> While you're at it, what about replacing:
>> Two types are distinguishable if at most one of the two includes a nullable type…
>
> by:
>
>> Two types are distinguishable if they are not both (a) nullable (type)…

Is your point that the union-decomposing steps already do the work of ["includes a nullable type"](https://rawgit.com/jyasskin/webidl/simplify-distinguishable/index.html#dfn-includes-a-nullable-type), so we can use the more primitive notion in the header?

tobie:
>> You might want to add a note saying that types that appear in the same category are undistinguishable.

>The normative part of that is covered by the lack of a ● in all foo/foo cells except for interface/interface. Are you saying that's not clear enough, so I should clarify it with a note?

You're right. This is actually quite clear. Never mind my comment.

> Is your point that the union-decomposing steps already do the work of "includes a nullable type", so we can use the more primitive notion in the header?

No. It's the "at most one" that bothers me. It suggests we're comparing _n_ types rather than 2. I should have suggested the following instead:

> Two types are distinguishable if they do not both [includes a nullable type|include a nullable type]…

tobie:
Other than that, this LGTM. I'd love to have someone else's approval, here, though. /cc @heycam, @bzbarsky.

bzbarsky:
I can try to take a look tomorrow...

jyasskin:
This is now updated for record types.

tobie:
Looking at this more, I'm unsure as to whether it actually helps understanding the algorithm or creates an extra, artificial layer of abstraction in the middle.

I think I'd be more tempted to find it helpful if these new categories had broader use cases throughout the spec. e.g. be able to replace all cases of "dictionary or record" by "dictionary-like".

bzbarsky:
For what it's worth, the updated version looks ok to me.  I think there's some value to splitting "callback interfaces" and "non-callback interfaces" into separate columns, at the very least, which this change does.

jyasskin:
I've now rebased this to HEAD. Is it still a good idea?

tobie:
Let's do it. Can you force-push it?

index.bs
```diff
@@ -3550,9 +3492,9 @@ the following algorithm returns <i>true</i>.
         </pre>
 
         <code>CBIface</code> is distinguishable from <code>Iface</code>
-        because the pair satisfies note (a),
+        because there's a ● at the intersection of dictionary-like and interface-like,
```
jyasskin: I realized today that I also had to update this example now that we aren't using the notes to handle callback interfaces.

jyasskin:
@tobie Done.

PR #209, opened 2016-10-26
Prevent [[SetPrototypeOf]] on all globals

domenic:
HTML already prohibits it for WindowProxy; see https://html.spec.whatwg.org/multipage/browsers.html#windowproxy. However, this does not cover the other globals, which need to have their prototype chain locked down. See #133 for previous work.

/cc @littledan @jswalden

***
[Preview](https://cdn.rawgit.com/heycam/webidl/9a39bab/index.html) | [Diff w/ current ED](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F9a39bab%2Findex.html) | [Diff w/ base](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F7cb900a%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F9a39bab%2Findex.html)


littledan:
LGTM

annevk:
We should add a note here that user agents do not need to implement it for the `Window` object (or make that object opt-out normatively somehow). We went to some lengths to ensure the `Window` object does not need to be an exotic object and I don't think you can write a test that proves it isn't.

annevk:
Hmm, it's even worse. While writing a patch for HTML to export WindowProxy I noticed that HTML explicitly refers to the Window object as an ordinary object. This would make that somewhat contradictory. Shall we just embrace the fact that IDL needs to know about the Window object? (We have some Window<>WindowProxy bridging still to be done I think.)

domenic:
Sure, so omit the [PrimaryGlobal] clause I guess?

annevk:
Well, I think we also want to get rid of [PrimaryGlobal] at some point (and require explicit [Exposed] usage everywhere) so I think it's best to just call out `Window` directly.

bzbarsky:
What about other things on the global's proto chain?  Are those already handled somewhere?

bzbarsky:
Also, is there a reason we're not using https://tc39.github.io/ecma262/#sec-immutable-prototype-exotic-objects here instead of defining a manual override?

bzbarsky:
And in particular a manual override with different behavior?

bzbarsky:
Oh, I guess the proto chain was handled in #133 and in fact used <https://tc39.github.io/ecma262/#sec-immutable-prototype-exotic-objects> (though in a slightly odd way in the case of the named properties object; I don't know why we didn't just make it an immutable prototype exotic object too).

domenic:
An immutable prototype exotic object is, as far as I can tell, an object whose *only* overridden internal method is [[SetPrototypeOf]].

bzbarsky:
@domenic Where does it say that?  It says that an immutable prototype exotic object must have that internal method overridden, but doesn't say anything about the other methods.

domenic:
@bzbarsky: https://tc39.github.io/ecma262/#sec-built-in-exotic-object-internal-methods-and-slots

> The following exotic objects use the ordinary object internal methods except where it is explicitly specified otherwise below:

bzbarsky:
Oh, I see.  Gotta love non-local effects in specs.  :(

annevk:
(So #133 doesn't change the ordinary object status of the Window object since it only affects its prototype object.)

bzbarsky:
OK.  So taking a step back for a second: from my point of view as an implementor, Window is not special in terms of wanting it to be "ordinary".  I want that for all globals.

Which means that my definition of "ordinary" includes immutable prototype exotic objects, given that they only have [[SetPrototypeOf]] modified.  In practice we implement that via a flag on such objects and checking that flag in the default [[SetPrototypeOf]]...

What this should translate to in spec terms, I'm not sure.  But my point was that there current ordinariness of Window is not really sacred, or is no more sacred than the ordinariness of any other global object.

domenic:
So good point on the wrong behavior. Let me push a commit fixing that.

I'm not really concerned about making Window a perfectly ordinary object. I think the note suffices.

annevk:
Will you change HTML too then?

domenic:
I don't really see the need. Window is still an ordinary object in practice since this is unobservable.

domenic:
What is missing to get this merged?

index.bs
```diff
@@ -12261,6 +12252,20 @@ defined in section [[#platformobjectset]].
     1.  Perform steps 3-11 of the [=default [[Set]] internal method=].
 </div>
 
+<h4 id="platformobjectsetprototypeof">Platform object \[[SetPrototypeOf]] method</h4>
+
+<div algorithm="to invoke the internal [[SetPrototypeOf]] method of platform objects">
+
+    The internal \[[SetPrototypeOf]] method of every [=platform object=]
+    that implements an [=interface=] with the [{{Global}}] or [{{PrimaryGlobal}}]
+    [=extended attribute=] must execute the same algorithm must be executed as is defined for the
```
bzbarsky: "must execute the same algorithm must be executed" has too many "musts".  Probably should be:

  ... must execute the same algorithm as is defined for ...

domenic: Fixed, thanks

Issue #212, opened 2016-10-27
Make [SameObject] beneficial


annevk:
I believe the contract with [SameObject] is supposed to be that whatever is returned the first time cannot ever change, right?

Currently specifications have to maintain that invariant and add the [SameObject] extended attribute as documentation.

However, what IDL could do instead is to allocate an internal slot when such an attribute is used. IDL's algorithm for the getter could then be to return the value from the internal slot when it's set, and when it's unset run the specification algorithm to get a value, set the internal slot to that value, and return that value.

That way the specification does not have to define how the value is stored, just how it's computed/retrieved.

annevk:
This might be a duplicate of https://www.w3.org/Bugs/Public/show_bug.cgi?id=26524.

annevk:
Note that changing this will affect specifications. @bzbarsky collected a list that will need to be updated after the change: https://github.com/heycam/webidl/pull/211#issuecomment-259230251. Seems doable.

annevk:
Unlike [SameObject] I think we should allow this new [Cached] (or whatever we call it) extended attribute to be used for getters that return promises. That a non-cached value is returned when this is wrong seems fine.

bzbarsky:
Yes, I agree.

inexorabletash:
+1 to [Cached] - it's basically what we do in Blink, and end up writing `[SameObject, SaveSameObject]` to enable autogenerated caching.

Aside: I'd love to put [SameObject] on IDBObjectStore and IDBIndex's `keyPath` attributes, but they can produce string or Array-of-string results. So if we can drop the restriction requiring the type to be `object` or an Interface, I'd like that too.

bzbarsky:
Yeah, so what Gecko has is:

* `[Cached]` to indicate that autogenerated caching should happen.  In addition to that, attributes marked `[Cached]` are allowed to have sequence, dictionary, and record types, because the reason for not allowing attributes with those types (creation of a new object on each get) does not apply in the `[Cached]` case.  Also, we check that `[NewObject]` and `[Cached]` are not used together, because that's nonsensical.
* `[SameObject]` as defined in the spec, but with the additional implication that there are no side-effects from the get.
* `[Constant]` which is just like `[SameObject]` but without type restrictions, so you can use it on any return type.  Again, implies no side-effects from the get.
[...]

PR #217, opened 2016-11-01
Update attribute setter/getters in various ways

### LINKED ISSUE #186: Should promise-returning attributes reject in case of wrong this value?
When a promise-returning attribute getter is called on a wrong this value, it is expected to throw a type error. Promise-returning operations on the other side will return a rejected promise.

I am wondering why attribute getters are not also rejecting.
Although not WebIDL based, https://streams.spec.whatwg.org/#default-reader-closed is doing so for instance.

domenic:
- Fixes getters on the global when no explicit `this` is passed, similarly to what was already done for operations in a2b4599631fe0fa254a7a4993dd7f00b74c8ad20. Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=29421.
- Make promise getters reject in case of an error. Disallow promise types on attributes that will generate setters. Fixes #186 and fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=25048.
- Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=27905 by more clearly specifying the target (as was done previously for operations).
- Updated to use modern ES Get/Set abstract operations instead of [[Get]] and [[Put]].
- Also made minor fixes to the operations definition, e.g. wrapping it in `<div algorithm>` and remembering to actually return F.

***
[Preview](https://cdn.rawgit.com/heycam/webidl/9962a71/index.html) | [Diff w/ current ED](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F9962a71%2Findex.html) | [Diff w/ base](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F0c5f48b%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F9962a71%2Findex.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1377,9 +1377,9 @@ ignored or an exception is thrown.
     };
 </pre>
 
-The type of an attribute that is not [=read only=] must not be a [=promise type=]. Additionally,
-the type of an attribute with any of the [=extended attributes=] [{{LenientThis}}],
-[{{PutForwards}}], [{{Replaceable}}], or [{{SameObject}}] must not be a [=promise type=].
+Attributes whose type is a [=promise type=] must be [=read only=]. Additionally, they cannot have
+any of the [=extended attributes=] [{{LenientThis}}], [{{PutForwards}}], [{{Replaceable}}], or
+[{{SameObject}}].
 
 A [=regular attribute=]
 that is not [=read only=]
@@ -10537,7 +10537,7 @@ The characteristics of this property are as follows:
     1.  Let |steps| be the following series of steps:
         1.  Try running the following steps:
             1.  Let |O| be <emu-val>null</emu-val>.
-            1.  If |attribute| is not a [=static attribute=]:
+            1.  If |attribute| is a [=regular attribute=]:
                 1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
                     <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
                     (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
@@ -10586,11 +10586,11 @@ The characteristics of this property are as follows:
         <emu-val>undefined</emu-val>; there is no [=attribute setter=] function.
     1.  Assert: |attribute|'s type is not a [=promise type=].
     1.  Let |steps| be the following series of steps:
-        1.  If no arguments were passed, then
+        1.  If no arguments were passed, then [=ECMAScript/throw=] a <emu-val>TypeError</emu-val>.
         1.  Let |V| be the value of the first argument passed.
         1.  Let |id| be |attribute|'s [=identifier=].
         1.  Let |O| be <emu-val>null</emu-val>.
-        1.  If |attribute| is not a [=static attribute=]:
+        1.  If |attribute| is a [=regular attribute=]:
             1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
                 <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
                 (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
@@ -10600,16 +10600,14 @@ The characteristics of this property are as follows:
             1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
             1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
                 |id|, and "setter".
-            1.  Let |validThis| be <emu-val>true</emu-val> if |O| is a [=platform object=] that
-                implements the interface |target|, or <emu-val>false</emu-val> otherwise.
-            1.  If |validThis| is <emu-val>false</emu-val> and |attribute| was not specified
-                with the [{{LenientThis}}] [=extended attribute=], then [=ECMAScript/throw=] a
-                <emu-val>TypeError</emu-val>.
+            1.  Let |validThis| be true if |O| is a [=platform object=] that implements the
+                interface |target|, or false otherwise.
+            1.  If |validThis| is false and |attribute| was not specified with the [{{LenientThis}}]
+                [=extended attribute=], then [=ECMAScript/throw=] a <emu-val>TypeError</emu-val>.
             1.  If |attribute| is declared with the [{{Replaceable}}] extended attribute, then:
                 1.  Perform [=?=] [=CreateDataProperty=](|O|, |id|, |V|).
                 1.  Return <emu-val>undefined</emu-val>.
-            1.  If |validThis| is <emu-val>false</emu-val>, then return
-                <emu-val>undefined</emu-val>.
+            1.  If |validThis| is false, then return <emu-val>undefined</emu-val>.
             1.  If |attribute| is declared with a [{{LenientSetter}}] extended attribute, then
                 return <emu-val>undefined</emu-val>.
             1.  If |attribute| is declared with a [{{PutForwards}}] extended attribute, then:
--- a/index.bs
+++ b/index.bs
@@ -1377,6 +1377,10 @@ ignored or an exception is thrown.
     };
 </pre>
 
+The type of an attribute that is not [=read only=] must not be a [=promise type=]. Additionally,
+the type of an attribute with any of the [=extended attributes=] [{{LenientThis}}],
+[{{PutForwards}}], [{{Replaceable}}], or [{{SameObject}}] must not be a [=promise type=].
+
 A [=regular attribute=]
 that is not [=read only=]
 can be declared to <dfn id="dfn-inherit-getter" export>inherit its getter</dfn>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -10724,130 +10728,141 @@ The characteristics of this property are as follows:
...
+            1.  If |attribute| is not a [=static attribute=]:
+                1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
+                    <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
+                    (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
+                    the global object does not implement |target| and [{{LenientThis}}] is not
+                    specified.)
+                    <!-- https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547#c9 -->
+                1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
+                1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
+                    |realm|, |id|, and "method".
```
bzbarsky: "getter", not "method", right?

index.bs
```diff
@@ -10724,130 +10728,141 @@ The characteristics of this property are as follows:
...
+                of the inherited attribute, if this attribute is declared to
+                [=inherit its getter=]), on |O| if |O| is not <emu-val>null</emu-val>.
+            1.  Return the result of [=converted to an ECMAScript value|converting=] |R| to an
+                ECMAScript value of the type |attribute| is declared as.
+
+        And then, if an exception was thrown:
+
+        1.  If |attribute|'s type is a [=promise type=], then:
+            1.  Let |reject| be the initial value of [=%Promise%=].reject.
+            1.  Return the result of calling |reject| with [=%Promise%=] as the
```
bzbarsky: Did we ever decide how this should interact with [SameObject]?  Seems to me like we should just disallow [SameObject] on Promise-returning attributes, or change the definition of [SameObject] or something; the current definition of [SameObject] just would not match this new behavior.

I'm still not 100% sure we want to make this behavior change, for what it's worth.  What do UAs do in practice?

index.bs
```diff
@@ -10724,130 +10728,141 @@ The characteristics of this property are as follows:
...
+        1.  If |attribute| is not a [=static attribute=]:
+            1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
+                <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
+                (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
+                the global object does not implement |target| and [{{LenientThis}}] is not
+                specified.)
+                <!-- https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547#c9 -->
+            1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
+            1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
+                |realm|, |id|, and "method".
```
bzbarsky: "setter", not "method".

index.bs
```diff
@@ -10902,53 +10917,56 @@ For [=namespaces=], the properties corresponding to each declared operation are
...
-    |steps|, the [=%FunctionPrototype%=] of |realm|).
-1.  Perform [=!=] [=SetFunctionName=](|F|, |id|).
-1.  Let |S| be the [=effective overload set=] for [=regular operations=] (if |op| is a regular
-    operation) or for [=static operations=] (if |op| is a static operation) with [=identifier=] |id|
-    on |target| and with argument count 0.
-1.  Let |length| be the length of the shortest argument list in the entries in |S|.
-1.  Perform [=!=] [=DefinePropertyOrThrow=](|F|, "length",
-    PropertyDescriptor<span class="prop-desc">{\[[Value]]: |length|, \[[Writable]]: <emu-val>false</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val>}</span>).
+<div algorithm>
+    To <dfn id="dfn-create-operation-function" lt="creating an operation function">create an operation function</dfn>,
```
bzbarsky: Wish this reindenting were a separate commit....

index.bs
```diff
@@ -1385,6 +1385,10 @@ ignored or an exception is thrown.
     };
 </pre>
 
+The type of the attribute that is not [=read only=] must not be a [=promise type=]. Additionally,
```
domenic: Fixed

index.bs
domenic: Fixed

index.bs
domenic: Fixed [SameObject] to reflect this.

In practice UAs are waiting for us to make a decision here; they throw the error per the current spec but are uncomfortable with doing so.

index.bs
domenic: Separated out into #232, but this PR for now is on top of that so it's included here.

index.bs
bzbarsky: > Fixed [SameObject] to reflect this.

I would really rather we just disallowed using [SameObject] on Promise-valued attributes if we take this route, given the weird behavior.  If we do allow it, UAs have to start doing something weird themselves, because they can no longer trust [SameObject] annotations...

In particular, it's not obvious that these changes to [SameObject] are compatible with #212

> they throw the error per the current spec but are uncomfortable with doing so.

I don't think Gecko is uncomfortable.  It sounds like Blink is?  Do you have references for the other UAs, by any chance?

domenic: > I would really rather we just disallowed using [SameObject] on Promise-valued attributes if we take this route, given the weird behavior.

I really don't think it's that weird. It's about the exception case. You don't require the same exception objects be thrown each time, so similarly, you shouldn't require the same promise object be returned each time.

> Do you have references for the other UAs, by any chance?

WebKit: https://github.com/heycam/webidl/issues/186

bzbarsky: > I really don't think it's that weird. It's about the exception case.

The problem is that unlike sync functions, which have a return value and a possible exception which are exposed to the caller in different ways, Promise-returning functions use a return value to represent both things.  This makes [SameObject] a pretty confusing thing to start with: is it referring to the Promise object returned, or to its resolution value in the non-exception case, for example?  If we really reason by analogy to sync functions, it should be referring to the resolution value...  but in practice people and the spec use it to mean the actual Promise return value.  Which doesn't even have an explicit equivalent in the sync function case in JS; it would be a continuation in languages that have such things.  So any reasoning by analogy here is broken right now anyway.

It's worth taking a step back for a second and asking what the point of [SameObject] really is and why we have it at all, then trying to figure out what it means to map that concept to promise-returning getters.  Assuming that's really possible.

> WebKit: #186

Thank you.  And Edge?

domenic: > It's worth taking a step back for a second and asking what the point of [SameObject] really is and why we have it at all, then trying to figure out what it means to map that concept to promise-returning getters. Assuming that's really possible.

Right. I perhaps am not the best person for this, since I don't like these documentation-only extended attributes. But I tried to put myself in the shoes of someone who does, and an interpretation of "returns the same object in non-exceptional cases" seemed like a good thing to document.

> Thank you. And Edge?

I haven't seen any preferences from them.

bzbarsky: OK.  Well, as another point of view, with my JIT implementor hat on, after this change all Promise-returning cases would be non-exceptional cases (because from a JIT's point of view they are), so my JIT would have to know that [SameObject] doesn't mean "the same object" in Promise-returning cases and hence can't be optimized the same way as other [SameObject] annotations...

index.bs
```diff
@@ -1374,6 +1374,10 @@ ignored or an exception is thrown.
     };
 </pre>
 
+The type of an attribute that is not [=read only=] must not be a [=promise type=]. Additionally,
+the type of an attribute with any of the [=extended attributes=] [{{LenientThis}}],
+[{{PutForwards}}], [{{Replaceable}}], or [{{SameObject}}] must not be a [=promise type=].
```
tobie: I thought there were requirements for [{{SameObject}}] promise type attributes. E.g. for https://drafts.csswg.org/css-font-loading/#font-face-set-ready, no?

domenic: Yes, but @bzbarsky prefers we don't use them, as discussed upthread.

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
-        1.  Let |V| be the result of [=converted to an ECMAScript value|converting=]
-            |idlValue| to an ECMAScript value.
-        1.  Return |V|.
-    </div>
+    *   |G| is the [=attribute getter=] created given the attribute, the interface (or the interface
+        it's being mixed in to, if the interface is actually a mixin), and the [=relevant Realm=] of
+        the object that is the location of the property; and
+    *   |S| is the [=attribute setter=] created given the attribute, the interface (or the interface
+        it's being mixed in to, if the interface is actually a mixin), and the [=relevant Realm=] of
+        the object that is the location of the property.
```
tobie: I'm aware this is purposefully copied from the text in the operations binding, but: s/the object that is the location of the property/the object that contains the property/ would read better.

Maybe change it everywhere? Or not at all for now also works. :)

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+    1. Return |F|.
+</div>
 
 Note: Although there is only a single property for an IDL attribute, since
 accessor property getters and setters are passed a <emu-val>this</emu-val>
 value for the object on which property corresponding to the IDL attribute is
 accessed, they are able to expose instance-specific data.
 
-Note: Note that attempting to assign to a property corresponding to a
+Note: Attempting to assign to a property corresponding to a
```
tobie: Can't/shouldn't we express that normatively?

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+                    |attribute|'s [=identifier=], and "getter".
+                1.  If |O| is not a [=platform object=] that implements the interface |target|,
+                    then:
+                    1.  If |attribute| was specified with the [{{LenientThis}}]
+                        [=extended attribute=], then return <emu-val>undefined</emu-val>.
+                    1. Otherwise, [=ECMAScript/throw=] a <emu-val>TypeError</emu-val>.
+            1.  Let |R| be the result of performing the actions listed in the
+                description of |attribute| that occur on getting (or those listed in the description
+                of the inherited attribute, if this attribute is declared to
+                [=inherit its getter=]), on |O| if |O| is not <emu-val>null</emu-val>.
```
tobie: I find it hard to parse the meaning of the hanging conditional. I assume it means you carry out the getting in both case, but once it's carried with the |O| as `this` and the other one with `undefined` as `this`?

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+    The <dfn id="dfn-attribute-setter" export>attribute setter</dfn> is created as follows, given an
+    [=attribute=] |attribute|, an [=interface=] |target|, and a [=Realm=] |realm|:
+
+    1.  If |attribute| is [=read only=] and does not have a
+        [{{LenientThis}}], [{{PutForwards}}] or [{{Replaceable}}] [=extended attribute=], return
+        <emu-val>undefined</emu-val>; there is no [=attribute setter=] function.
+    1.  Assert: |attribute|'s type is not a [=promise type=].
+    1.  Let |steps| be the following series of steps:
+        1.  If no arguments were passed, then
+        1.  Let |V| be the value of the first argument passed.
```
tobie: There's something unclear in the above two steps (either there's no argument or there is one).

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+        1.  Let |id| be |attribute|'s [=identifier=].
+        1.  Let |O| be <emu-val>null</emu-val>.
+        1.  If |attribute| is not a [=static attribute=]:
```
tobie: `If |attribute| is a [=regular attribute=]:`

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+            1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
```
tobie: If find all of the substeps of this step 5 super difficult to parse. I don't doubt that the algorithm is correct, but I really struggle to understand its intent. I can't imagine modifying something in here without accidentally breaking stuff.

Here's a proposed alternative (which is slightly less DRY, but hopefully a bit clearer). I added an optional note (prefixed with ?) to clarify the exit point of the substep. Maybe that helps (or is over the top—I'm not sure).

```
1.  If |attribute| is a [=regular attribute=]:
    1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
        <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
        (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
        the global object does not implement |target| and [{{LenientThis}}] is not
        specified.)
        <!-- https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547#c9 -->
    1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
    1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
        |id|, and "setter".
     1.  Let |validThis| be `true` if |O| is a [=platform object=] that
         implements the interface |target|, or `false` otherwise.
     1.  If |validThis| is `true`, then:
         1.  If |attribute| is declared with a [{{LenientSetter}}] extended attribute, then:
             1.  Return <emu-val>undefined</emu-val>.
         1.  If |attribute| is declared with a [{{PutForwards}}] extended attribute, then:
             1.  Let |Q| be [=?=] [=Get=](|O|, |id|).
             1.  If [=Type=](|Q|) is not Object, then [=ECMAScript/throw=] a
                 <emu-val>TypeError</emu-val>.
             1.  Let |forwardId| be the identifier argument of the [{{PutForwards}}] extended
                 attribute.
             1.  Perform [=?=] [=Set=](|Q|, |forwardId|, |V|).
             1.  Return <emu-val>undefined</emu-val>.
         1.  If |attribute| is declared with the [{{Replaceable}}] extended attribute, then:
             1.  Perform [=?=] [=CreateDataProperty=](|O|, |id|, |V|).
             1.  Return <emu-val>undefined</emu-val>.
        
        ? Note: setters who are not special-cased here will
        ? continue being defined in the parent steps
        ? along with static attributes.
     1.  Otherwise if |attribute| was specified with the [{{LenientThis}}] [=extended attribute=],
         1.  If |attribute| is declared with the [{{Replaceable}}] extended attribute,
             then perform [=?=] [=CreateDataProperty=](|O|, |id|, |V|).
         1. Return <emu-val>undefined</emu-val>.
     1.  Otherwise, [=ECMAScript/throw=] a <emu-val>TypeError</emu-val>.
```

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+                <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
+                (This will subsequently cause a <emu-val>TypeError</emu-val> in a few steps, if
+                the global object does not implement |target| and [{{LenientThis}}] is not
+                specified.)
+                <!-- https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547#c9 -->
+            1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
+            1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
+                |id|, and "setter".
+            1.  Let |validThis| be <emu-val>true</emu-val> if |O| is a [=platform object=] that
```
tobie: Shouldn't the booleans here really be just \`true\` ans \`false\`? These are not JS values, they're infra booleans, no?

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
-            with |O| as the object and |idlValue| as the value.
-        1.  Otherwise, the attribute is a [=static attribute=].
-            Perform the actions listed in the description of the attribute that occur when setting with |idlValue| as the value.
-        1.  Return <emu-val>undefined</emu-val>.
-    </div>
+        1.  Let |idlValue| be determined as follows:
 
-    The value of the <emu-val>Function</emu-val> object’s “length”
-    property is the <emu-val>Number</emu-val> value <emu-val>1</emu-val>.
+            <dl class="switch">
```
tobie: Why are we using a switch here rather than a regular if..otherwise clause?

index.bs
```diff
@@ -10518,126 +10522,140 @@ The characteristics of this property are as follows:
...
+                    |idlValue| is the result of [=converted to an IDL value|converting=] |V| to an
+                    IDL value of |attribute|'s type.
+                </dd>
+            </dl>
 
-    The value of the <emu-val>Function</emu-val> object’s “name”
-    property is a <emu-val>String</emu-val> whose value is the
-    concatenation of “set ” and the identifier of the attribute.
+        1.  Perform the actions listed in the description of |attribute| that occur on setting, on
+            |O| if |O| is not <emu-val>null</emu-val>.
```
tobie: Same comments as for the hanging conditional in the getter, here.

tobie:
A few nits and a few questions. LGTM otherwise.

index.bs
domenic: I'll go with not changing it for now as this area is related to the mixin mess and will get straightened out when that is.

index.bs
domenic: Well, there's no `this` formally (that's https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301). The idea is not to provide anything at all; providing `undefined` isn't really correct.

index.bs
domenic: Hmm. I don't really find that clearer; I like how you just check the conditions one after another in the existing version. Let's leave it for a follow-up, as we'd also want to change the analogous steps for operations.

index.bs
domenic: That's probably more correct I guess. I was thinking this is more like an ES spec algorithm where booleans are emu-val'ed, but this seems like a better place to lean toward web spec conventions than ES.

index.bs
domenic: Not sure. We do it a lot, it seems. I guess because it's determining a value?

index.bs
domenic: I still think this is pretty clear as-is, but we could do a different version perhaps as a follow-up if you have a proposal.

index.bs
domenic: I'm pretty sure it is. This was just a drive-by fix to not repeat "Note" twice.

tobie: For my own education, where is that behavior defined normatively?

bzbarsky:
I'm really sorry for the horrible lag here.  I finally had a chance to read through these changes, and they look great.  There is one preexisting issue that I filed as https://github.com/heycam/webidl/issues/271

bzbarsky:
Oh, and filed https://bugs.chromium.org/p/chromium/issues/detail?id=683310 for Blink not doing PutForwards per spec.

bzbarsky:
And https://bugzilla.mozilla.org/show_bug.cgi?id=1332713 on updating Gecko to these changes.

bzbarsky:
Also filed spec bugs on existing [SameObject] promises:

* https://github.com/w3c/presentation-api/issues/407
* https://github.com/w3c/ServiceWorker/issues/1060

bzbarsky:
And https://github.com/w3c/testharness.js/pull/232 on idlharness no longer matching the spec now.  And filed https://github.com/w3c/testharness.js/issues/233 on what's left to do there.

annevk:
It seems https://fetch.spec.whatwg.org/ was missed. It uses [SameObject] for trailers, a feature that isn't implemented yet.

annevk:
Filed https://github.com/whatwg/fetch/issues/473.

bzbarsky:
Right, I only filed bugs on the things I found by disallowing [SameObject] in Gecko's IDL and seeing what failed...

Issue #240, opened 2016-11-15
No seemingly consistent Shepherd story


annevk:
The way Bikeshed annotations are used is rather weird.

Some types are exported (enumeration types), some are not (dictionary types, callback function types).

Some types are labeled dfn, some are interface, some are dictionary. E.g., sequence is an interface, but record is a dictionary.

Some types are co-located with their values (records, it seems like), others are in different places (dictionaries).

What's the strategy here?

@tabatkins any advice?

tobie:
Whatever we agree on should be documented in the README.

tabatkins:
> Some types are exported (enumeration types), some are not (dictionary types, callback function types).

Recall that nothing needs to be explicitly exported except for "dfn" type definitions. The rest export implicitly unless you use `noexport`.

> Some types are labeled dfn, some are interface, some are dictionary. E.g., sequence is an interface, but record is a dictionary.

This is a little hand-wavey. Unless there's a better definition type (like `dictionary`), I prefer everything that defines a type name to be `interface`, just for regularity. But "record" and "sequence" don't really do that; they're not really IDL names, they're closer to IDL keywords, like the "interface" in `interface Foo`. You wouldn't ever write `{{record}}`, either. I'd probably mark up "record" and "sequence" as `dfn`, same as the keyword "interface" is.

annevk:
@tabatkins dictionaries, sequences, and interfaces are all types though. I don't think you can say one is a name and another is syntax.

tobie:
Yeah--I've found the mapping to be exceptionally challenging when doing the conversion to Bikeshed. Happy to help fix this once we settle on something that seems clearer.

tabatkins:
> dictionaries, sequences, and interfaces are all types though. I don't think you can say one is a name and another is syntax. (Not really sure what an IDL name even is.)

By "names", I'm referring to actual types.  `record` isn't a type, `record<Foo>` is; `record` is a syntax keyword used to *construct* a type. (In Haskell terms, it would literally be a type constructor, which is a first-class value of a higher kind than types.) So the `record` keyword should be marked up as a "dfn" type definition, same as the `interface` keyword is.

> Yeah--I've found the mapping to be exceptionally challenging when doing the conversion to Bikeshed. 

Yeah, this is a particularly meta situation, it's not surprising that confusion abounds. ^_^

annevk:
I'm not sure that's really helping.

[...]

Issue #233, opened 2016-11-10
Shepherd: difference between "record" and "record type"


annevk:
I don't understand https://heycam.github.io/webidl/#idl-record. The heading is a dictionary definition (not exported). Then we have record type (exported). Then we have mappings (not exported, for=record).

What is the difference between record and record type? Why is mappings for=record and not for="record type"?

And can we export mappings?

I was trying to clean up some of this, but got
```
Traceback (most recent call last):
  File "/usr/local/bin/bikeshed", line 9, in <module>
    load_entry_point('Bikeshed==0.0.0', 'console_scripts', 'bikeshed')()
  File "/Users/annevk/GH/bikeshed/bikeshed/__init__.py", line 196, in main
    doc.preprocess()
  File "/Users/annevk/GH/bikeshed/bikeshed/__init__.py", line 590, in preprocess
    highlight.addSyntaxHighlighting(self)
  File "/Users/annevk/GH/bikeshed/bikeshed/highlight.py", line 85, in addSyntaxHighlighting
    highlight(el, translateLang(lang))
  File "/Users/annevk/GH/bikeshed/bikeshed/highlight.py", line 27, in highlight
    widl = parser.Parser(text, IDLUI())
NameError: global name 'parser' is not defined
```

annevk:
This blocks https://github.com/whatwg/fetch/pull/412 by the way.

tobie:
Bikeshed's bleeding edge is broken right now, see: https://github.com/tabatkins/bikeshed/issues/867.

tobie:
> What is the difference between record and record type.

I don't know. We seem to have this for some types but not all. Unless there's a good reason to have both, we should probably clean it up.

annevk:
@jyasskin was this copypasta or intentional?

bzbarsky:
A record is a thing of that a key-value mapping.

A record type is a type whose values are records.

The difference is the same as the difference between "long" and "5".

annevk:
I don't really get the analogy.

5 is an instance of a long, sure.

But "record" is not an instance of a record type. A specific record, sure, but the word record by itself is not.

Anyway, I'm going to submit a PR to at least export mappings since that's blocking Fetch from using records.

tobie:
Closed via #239 and https://github.com/heycam/webidl/commit/274117f761b5366a8ac88187ddb00089f8a63cd3.

annevk:
[...]

PR #243, opened 2016-11-22
Meta: export record and sequence

annevk:
And also stop marking them up as Bikeshed IDL types since they’re not.

See also #233 and #240 though I don’t think this commit addresses
either of them fully.

***
[Preview](https://cdn.rawgit.com/heycam/webidl/daeea0b/index.html) | [Diff w/ current ED](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2Fdaeea0b%2Findex.html) | [Diff w/ base](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2Fa0b2c0c%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2Fdaeea0b%2Findex.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3277,7 +3277,7 @@ the following algorithm returns <i>true</i>.
         <ul>
             <li>
                 <code>{{double}}?</code> and
-                <code>{{record}}&lt;{{DOMString}}, {{DOMString}}></code>
+                <code><a>record</a>&lt;{{DOMString}}, {{DOMString}}></code>
             </li>
             <li>
                 <code>(Interface1 or {{long}})?</code> and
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -3277,7 +3277,7 @@ the following algorithm returns <i>true</i>.
         <ul>
             <li>
                 <code>{{double}}?</code> and
-                <code>{{record}}&lt;{{DOMString}}, {{DOMString}}></code>
+                <code><a>record</a>&lt;{{DOMString}}, {{DOMString}}></code>
```
tobie: Think you could use the [=record=] "shorthand" here for consistency with the rest of the spec?

annevk: Sure, added a commit. Didn't change the output so I guess that went okay.

tobie: Just saw the above now (it was folded in in the UI). Sorry for asking on irc.

PR #244, opened 2016-11-23
Editorial: markup legacy caller correctly

annevk:
***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/annevk/legacy-caller.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/9af88fa...e6e34e3.html)


annevk:
I'm not a 100% sure about the change here. It seems wrong to export "legacy caller" as "legacy", but it might also be wrong to change the links to no longer point to the "Legacy caller" section. It's not clear to me what is intended here.

annevk:
Perhaps it would be better if the "Special operation" column in https://heycam.github.io/webidl/#idl-special-operations were not marked up as definition?

Not sure where the existing "getters" and "setters" reference would have to go. But it seems "deleters" is referenced from nowhere currently, "legacy caller" (well, "legacy") isn't either, and "stringifiers" and "serializers" references could instead reference their corresponding heading.

Thoughts?

tobie:
:( evidently screwed-up the force push. fixing.

tobie:
The pull request is against the gh-pages branch. Duh. :-/

New PR in the works.

PR #272, opened 2017-01-20
Fix incorrect [LenientThis] that should be [LenientSetter]

### LINKED ISSUE #271: readonly attribute setter synthesis should happen for LenientSetter, not LenientThis
The spec says, at <https://heycam.github.io/webidl/#dfn-attribute-setter>:

> If attribute is read only and does not have a [LenientThis], [PutForwards] or [Replaceable] extended attribute, return undefined; there is no attribute setter function.

That should be [LenientSetter], not [LenientThis].

@domenic, @tobie

domenic:
Closes #271.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1378,7 +1378,7 @@ ignored or an exception is thrown.
 </pre>
 
 Attributes whose type is a [=promise type=] must be [=read only=]. Additionally, they cannot have
-any of the [=extended attributes=] [{{LenientThis}}], [{{PutForwards}}], [{{Replaceable}}], or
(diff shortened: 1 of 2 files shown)


bzbarsky:
Looks great, thank you!

bzbarsky:
I just got mail that says:

>The page build failed with the following error:
>
>  Page build failed.

presumably as the result of merging that....  I have no idea what that's about.

domenic:
I guess that refers to gh-pages and the way GitHub is supposed to automatically deploy to https://heycam.github.io/webidl/. It does indeed seem the change is not reflected there yet... @heycam could re-trigger a deploy from the settings panel, or we could hope that it's temporary and that the next commit will deploy fine.

PR #286, opened 2017-01-25
Allow extended attributes to apply to types

domenic:
This implements the plan discussed in https://lists.w3.org/Archives/Public/public-script-coord/2017JanMar/0003.html and sets the stage for using extended attributes for SharedArrayBuffer-related purposes.

@bzbarsky, I could use some close review on this. Some points for discussion:

- Is `[Clamp] long` being a distinct type from `long` a good way of doing this? I based that on how `long?` is different from `long`, and it seems like a good mental model.
- Is it sufficiently clear that extended attributes propagate through typedefs? (A point you called out specifically in your email.) If not, any suggestions on how to make this clear?
- We have separate TypeWithAttributes and Type mainly so that ReturnType does not get attributes. Does that seem good?
- I didn't allow (gramatically) specifying an extended attribute over an entire union. I forget why, but an hour ago it seemed like a bad idea... if you have any thoughts to help guide me one way or another on this that would be appreciated.
- This will invalidate a lot of existing IDL which will need to get updated. E.g. `[TreatNullAs=EmptyString] attribute DOMString data;` must become `attribute [TreatNullAs=EmptyString] DOMString data;`. I think this is the right way to go, as it's clearer how things work and doesn't require any propagation of extended attributes from attribute/argument declarations into their types, which sounds annoying to spec and would cause inconsistency as sometimes people would do that and sometimes not. I also think this would be a good test case for future more-disruptive changes.
- TreatNullAs questions:
  - [Searching Chromium](https://cs.chromium.org/search/?q=TreatNullAs%3DEmptyString+file:.idl&sq=package:chromium&type=cs) I find no instances of this being used on callback interfaces. Can we nuke that part of the spec?
  - While we're here, since this is going to invalidate existing IDL anyway, can we just get rid of the silly identifier argument and make this `[TreatNullAsEmptyString]`?

I hope I didn't miss anything major; this seemed almost too easy...

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/extended-attrs-on-types.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/83d33a6...77cbdcb.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6009,7 +6009,11 @@ For any [=type=], the [=extended attributes associated with=] it must only conta
 
 The [=type name=] of a type associated with [=extended attributes=] is the concatenation of the
 type name of the original type with the set of strings corresponding to the [=identifiers=] of each
-[=extended attribute associated with=] the type, in order.
+[=extended attribute associated with=] the type, sorted in lexicographic order.
+
+<div class="example">
+    The [=type name=] for a type of the form <code>[B, A] long?</code> is "LongOrNullAB".
+</div>
 
 <h4 oldids="dom-Error" id="idl-Error" interface>Error</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -3329,9 +3329,10 @@ the following algorithm returns <i>true</i>.
     [=member type=] of the union type is distinguishable
     with the non-union type,
     or <i>false</i> otherwise.
-1.  If the two types (taking their [=nullable types/inner type=]
-    if they are [=nullable types=]) appear or are in categories appearing
-    in the following table and there is a “●” mark in the corresponding entry
+1.  Consider the two "innermost" types derived by taking each type's [=annotated types/inner type=]
+    if it is an [=annotated type=], and then taking its [=nullable types/inner type=] inner type
+    if the result is a [=nullable type=]. If these two innermost types appear or are in categories
+    appearing in the following table and there is a “●” mark in the corresponding entry
     or there is a letter in the corresponding entry and the designated additional
     requirement below the table is satisfied, then return <i>true</i>.
     Otherwise return <i>false</i>.
--- a/index.bs
+++ b/index.bs
@@ -1454,12 +1454,9 @@ interface will be stringified to the value of the attribute.  See
 
 The following [=extended attributes=]
 are applicable to regular and static attributes:
-[{{Clamp}}],
-[{{EnforceRange}}],
 [{{Exposed}}],
 [{{SameObject}}],
-[{{SecureContext}}],
-[{{TreatNullAs}}].
+[{{SecureContext}}].
 
 The following [=extended attributes=]
 are applicable only to regular attributes:
@@ -1489,7 +1486,7 @@ are applicable only to regular attributes:
 
 <pre class="grammar" id="prod-AttributeRest">
     AttributeRest :
-        "attribute" Type AttributeName ";"
+        "attribute" TypeWithExtendedAttributes AttributeName ";"
 </pre>
 
 <pre class="grammar" id="prod-AttributeName">
@@ -1891,13 +1888,7 @@ The following extended attributes are applicable to operations:
 [{{Exposed}}],
 [{{NewObject}}],
 [{{SecureContext}}],
-[{{TreatNullAs}}],
 [{{Unforgeable}}].
-
-The following extended attributes are applicable to operation arguments:
-[{{Clamp}}],
-[{{EnforceRange}}],
-[{{TreatNullAs}}].
 
 <pre class="grammar" id="prod-DefaultValue">
     DefaultValue :
@@ -1956,13 +1947,8 @@ The following extended attributes are applicable to operation arguments:
 
 <pre class="grammar" id="prod-Argument">
     Argument :
-        ExtendedAttributeList OptionalOrRequiredArgument
-</pre>
-
-<pre class="grammar" id="prod-OptionalOrRequiredArgument">
-    OptionalOrRequiredArgument :
-        "optional" Type ArgumentName Default
-        Type Ellipsis ArgumentName
+        ExtendedAttributeList "optional" TypeWithExtendedAttributes ArgumentName Default
+        ExtendedAttributeList Type Ellipsis ArgumentName
 </pre>
 
 <pre class="grammar" id="prod-ArgumentName">
@@ -3768,12 +3754,12 @@ The following extended attributes are applicable to [=iterable declarations=]:
 
 <pre class="grammar" id="prod-Iterable">
     Iterable :
-        "iterable" "&lt;" Type OptionalType "&gt;" ";"
+        "iterable" "&lt;" TypeWithExtendedAttributes OptionalType "&gt;" ";"
 </pre>
 
 <pre class="grammar" id="prod-OptionalType">
     OptionalType :
-        "," Type
+        "," TypeWithExtendedAttributes
         ε
 </pre>
 
@@ -3857,7 +3843,7 @@ interfaces must not have an
 
 <pre class="grammar" id="prod-MaplikeRest">
     MaplikeRest :
-        "maplike" "&lt;" Type "," Type "&gt;" ";"
+        "maplike" "&lt;" TypeWithExtendedAttributes "," TypeWithExtendedAttributes "&gt;" ";"
 </pre>
 
 No [=extended attributes=]
@@ -3946,7 +3932,7 @@ interfaces must not have an
 
 <pre class="grammar" id="prod-SetlikeRest">
     SetlikeRest :
-        "setlike" "&lt;" Type "&gt;" ";"
+        "setlike" "&lt;" TypeWithExtendedAttributes "&gt;" ";"
 </pre>
 
 No [=extended attributes=]
@@ -4316,10 +4302,6 @@ The following extended attributes are applicable to dictionaries:
 [{{Exposed}}],
 [{{SecureContext}}],
 
-The following extended attributes are applicable to dictionary members:
-[{{Clamp}}],
-[{{EnforceRange}}].
-
 <div data-fill-with="grammar-Partial"></div>
 
 <div data-fill-with="grammar-PartialDefinition"></div>
@@ -4331,19 +4313,14 @@ The following extended attributes are applicable to dictionary members:
 
 <pre class="grammar" id="prod-DictionaryMembers">
     DictionaryMembers :
-        ExtendedAttributeList DictionaryMember DictionaryMembers
+        DictionaryMember DictionaryMembers
         ε
 </pre>
 
 <pre class="grammar" id="prod-DictionaryMember">
     DictionaryMember :
-        Required Type identifier Default ";"
-</pre>
-
-<pre class="grammar" id="prod-Required">
-    Required :
-        "required"
-        ε
+        ExtendedAttributeList "required" TypeWithExtendedAttributes identifier Default ";"
+        ExtendedAttributeList Type identifier Default ";"
 </pre>
 
 <pre class="grammar" id="prod-PartialDictionary">
@@ -4867,7 +4844,7 @@ the type in the IDL.
 <pre highlight="webidl" class="syntax">typedef type identifier;</pre>
 
 The type being given a new name is specified after the <emu-t>typedef</emu-t>
-keyword (matching <emu-nt><a href="#prod-Type">Type</a></emu-nt>), and the
+keyword (matching <emu-nt><a href="#prod-Type">TypeWithExtendedAttributes</a></emu-nt>), and the
 <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> token following the
 type gives the name.
 
@@ -4879,7 +4856,7 @@ defined in this specification are applicable to [=typedefs=].
 
 <pre class="grammar" id="prod-Typedef">
     Typedef :
-        "typedef" Type identifier ";"
+        "typedef" TypeWithExtendedAttributes identifier ";"
 </pre>
 
 <div class="example">
@@ -5192,6 +5169,12 @@ type.
         UnionType Null
 </pre>
 
+<pre class="grammar" id="prod-TypeWithExtendedAttributes">
+    TypeWithExtendedAttributes :
+        ExtendedAttributeList SingleType
+        ExtendedAttributeList UnionType Null
+</pre>
+
 <pre class="grammar" id="prod-SingleType">
     SingleType :
         NonAnyType
@@ -5205,7 +5188,7 @@ type.
 
 <pre class="grammar" id="prod-UnionMemberType">
     UnionMemberType :
-        NonAnyType
+        ExtendedAttributeList NonAnyType
         UnionType Null
 </pre>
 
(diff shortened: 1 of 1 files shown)


bzbarsky:
I'm pretty swamped with reviews for various stuff right now, so it will likely take me a few days to get to this.  I'm sorry.  I simply have a huge backlog, both in terms of specs and in terms of code.  :(

You should probably look at https://lists.w3.org/Archives/Public/public-script-coord/2017JanMar/0009.html in case it affects your approach here.

One question I do have up front: what is the behavior you're actually aiming for with the patches applied?  Knowing that will let me evaluate the behavior, as well as whether the actual spec changes implement that behavior correctly.  Ideally, that sort of thing would live in the commit message...  I ask this because you say you're implementing the plan discussed in https://lists.w3.org/Archives/Public/public-script-coord/2017JanMar/0003.html but I actually outlined a few different options in that mail, and I'm not sure which ones you decided to take.

Some responses to questions without having read the changes very carefully yet.

> Is [Clamp] long being a distinct type from long a good way of doing this? 

That seems reasonable, at first glance, yes.

Will take some thought to check whether expressing interactions like "not both [Clamp] and [EnforceRange]" works right in that setup.

> Is it sufficiently clear that extended attributes propagate through typedefs?

I don't think it is.  In particular, it's not obvious to me _how_ they propagate, exactly, based on searching the diff for "typedef".  Like what the actual result is.  I don't quite recall what model the spec has of typedefs in general, nor what implementations do, exactly...  :(

> We have separate TypeWithAttributes and Type

That's fine, I think.   ReturnType definitely shouldn't have attributes.  Then again, ReturnType is already separate from Type, so we don't really care about this part.

There are other places where types probably shouldn't have attributes (at least not the attributes we're defining).  For example, callback argument types, right?

> I didn't allow (gramatically) specifying an extended attribute over an entire union.

I'm not sure what you mean.  Your grammar has "ExtendedAttributeList UnionType Null" as a possible thing for TypeWithAttributes, right?

> This will invalidate a lot of existing IDL which will need to get updated.

Hmm.  So we _could_ support the old syntax, in theory, for attributes.  As you note, it would be complicated.

In practice, TreatNullAs is ony used in a few specs: HTML, CSSOM, DOM.  Updating them would not be a huge deal.  I see ~40 places in Gecko's IDL using this annotation, which is not so bad to just update en-masse.

> Can we nuke that part of the spec?

I think so.  I see no consumers in Gecko either, and we really shouldn't add new ones, imo.

> can we just get rid of the silly identifier argument

That's probably fine, yes.

domenic:
No problem on the delay.

> One question I do have up front: what is the behavior you're actually aiming for with the patches applied.

(Yes, the commit message needs work.)

The basic idea is that the three extended attributes [Clamp], [ExtendRange], and [TreatNullAs] can now only be used on types, instead of in their previous positions as annotations on method arguments, attributes, and dictionary members. And they should only be usable in positions where they affect conversions from ES to Web IDL. (Hmm, now that I put it that way, maybe we should grammatically disallow them on non-readonly attributes.)

This adds the new concept of "types annotated with extended attributes". That is, `[Clamp] long` is a new type, consisting of `long` "annotated with" `[Clamp]`.

In the spec, we say that the sections on the types like `long` also govern versions generated by extended attributes, and then the algorithm for converting ES to `long` says "if the type is annotated with [Clamp]".

I couldn't really find a formal model for "is annotated with"; it's used elsewhere without any real precision. (E.g., there is no set of annotations on a given grammar production or construct that we consult. I guess this is what leads to problems like "what happens if you specify `[SecureContext]` on a mixin or partial interface?") Maybe I could achieve something more precise by studying the nullable types more.

> Will take some thought to check whether expressing interactions like "not both [Clamp] and [EnforceRange]" works right in that setup.

This again hinges on the "annotated with" concept. We just say "a type cannot be annotated by both [Clamp] and [EnforceRange]". If we are comfortable with leaving the definition of "annotated with" vague, we're good to go, but are we? For example, it's not exactly clear how "annotated with" propagates through typedefs. Again maybe I can study how nullable propagates through typedefs.

> That's fine, I think. ReturnType definitely shouldn't have attributes. Then again, ReturnType is already separate from Type, so we don't really care about this part.

What I meant is, I could have just changed Type, except then ReturnType would inherit the changes, and we don't want that. So that's where there are distinct grammar productions.

> There are other places where types probably shouldn't have attributes (at least not the attributes we're defining). For example, callback argument types, right?

Good catch, thanks. Will fix.

> I'm not sure what you mean. Your grammar has "ExtendedAttributeList UnionType Null" as a possible thing for TypeWithAttributes, right?

Oh, oops. I guess I meant I didn't allow specifying annotations on individual types inside a union, e.g. `(Node or [Clamp] long)`. Probably I should...

> Hmm. So we could support the old syntax, in theory, for attributes. As you note, it would be complicated.
>
> In practice, TreatNullAs is ony used in a few specs: HTML, CSSOM, DOM. Updating them would not be a huge deal. I see ~40 places in Gecko's IDL using this annotation, which is not so bad to just update en-masse.

This is the same issue as https://lists.w3.org/Archives/Public/public-script-coord/2017JanMar/0009.html, correct?

And yeah, I think we should bite the bullet. En-masse update for TreatNullAs, Clamp, and EnforceRange.

>> Can we nuke that part of the spec?
>
> I think so. I see no consumers in Gecko either, and we really shouldn't add new ones, imo.

Great. Maybe I'll do that as a separate PR first.

bzbarsky:
> I guess I meant I didn't allow specifying annotations on individual types inside a union

Ah.  It seems like we do want to allow that.  On the other hand, specifying [Clamp] or [EnforceRange] or [TreatNullAs] on a union as a whole doesn't really make sense.  But in terms of grammar, I don't see a great way to disallow it (because the union could be hidden behind a typedef, say)...  We can just enforce it via the "can't be specified except on these types"  validation mechanism.

> This is the same issue as https://lists.w3.org/Archives/Public/public-script-coord/2017JanMar/0009.html, correct?

I'm not sure what you mean.  I mean, we do want to keep allowing extended attributes on attributes in general.  _And_ allow them on the type of the attribute.  That's not a problem because the two locations (start of attribute, start of attribute type) are not adjacent, so the grammar is unambiguous.  The only place where they _are_ adjacent are non-required dictionary members and non-optional operation arguments...

domenic:
> I'm not sure what you mean. I mean, we do want to keep allowing extended attributes on attributes in general. And allow them on the type of the attribute. That's not a problem because the two locations (start of attribute, start of attribute type) are not adjacent, so the grammar is unambiguous. The only place where they are adjacent are non-required dictionary members and non-optional operation arguments...

Ah oops, I was indeed confusing two issues.

So the question is whether we have a use case for specifying extended attributes on operation arguments and dictionary members, instead of specifying them on their types. The spec does not, but Gecko does, and Gecko's reasons seem reasonably widely applicable. So I guess we should figure out how that works...

bzbarsky:
I think the simplest way to spec it is that non-optional operation arguments have a Type, not a TypeWithAttributes.  But for certain extended attributes, if they're specified on the operation argument they get applied to its type.  Similar for non-required dictionary members.

It's a bit annoying on the spec side, but pretty clear in terms of actual usage.

index.bs
```diff
@@ -1480,7 +1477,7 @@ are applicable only to regular attributes:
 
 <pre class="grammar" id="prod-AttributeRest">
     AttributeRest :
-        "attribute" Type AttributeName ";"
+        "attribute" TypeWithAttributes AttributeName ";"
```
tobie: I agree `TypeWithExtendedAttributes` is long, but I think it would help make things clearer, here.

index.bs
```diff
@@ -5145,6 +5123,16 @@ is a string, not necessarily unique, that identifies the type.
...
+Additional types can be created from existing ones by specifying certain
+[=extended attributes=] on the existing types. For example,
+<code>[Clamp] long</code> defines a new type whose behavior is based on
+that of the {{long}} type, but modified as specified by the [{{Clamp}}]
+extended attribute.
+
+The [=type name=] of a type annotated with an [=extended attribute=] is
+the concatenation of the type name of the annotated type and a string
+corresponding to the extended attribute's identifier.
+
```
tobie: An example would be nice here.

I image what you want is ClampLong, not LongClamp. So the sentence above would imply that more clearly if it was reversed:

```
The [=type name=] of a type annotated with an [=extended attribute=] is
the concatenation of the string corresponding to the extended attribute's identifier
and the type name of the annotated type.
```

domenic: LongClamp seems fine.

tobie: Oh, you had LongClamp in mind. I see. I feel like that's counterintuitive, given the syntax is `[Clamp] long`.

tobie: But that's a minor issue, really.

index.bs
```diff
@@ -9423,9 +9385,8 @@ The [{{TreatNullAs}}]
 extended attribute must [=takes an identifier|take the identifier=]
 <code>EmptyString</code>.
 
-The [{{TreatNullAs}}] extended attribute
-must not be specified on an operation argument,
-attribute or operation return value whose type is not {{DOMString}}.
+The [{{TreatNullAs}}] extended attribute must not be specified on a [=type=] that is not
+{{DOMString}}.
```
tobie: While you're at it, get rid of the double negation? E.g.:
```
The [{{TreatNullAs}}] extended attribute can only be specified on the {{DOMString}} [=type=].
```

domenic: "can" is not a normative word, so I don't think that really works.

index.bs
domenic: In general I think it's not worth fussing over this or creating an example given that nobody knows what type names are used for.

bzbarsky: Fwiw, Gecko uses type names to auto-generate union types and their accessors.  So for every type in the union, there is an IsFoo() accessor on the union, where Foo is the type name.

It ends up being pretty ugly in practice.  ;)

domenic:
OK, this is ready for another review. I think it meets our goals now.

annevk:
Does this also mean we're going to change USVString back to [EnsureUTF16] DOMString as extended attributes are no longer problematic? Is there any other goal here other than avoiding doubling the number of typed-array-related types?

bzbarsky:
> Does this also mean we're going to change USVString back to [EnsureUTF16] DOMString

We _could_ do that if we wanted to, yes.  Whether we want to is less clear.

> Is there any other goal here other than avoiding doubling the number of typed-array-related types?

Or quadrupling...

But note that this also fixes longstanding issues wherein you couldn't have a [Clamp] integer in a union or a dictionary, for example.

maritza1979:
So funny my husband fucking around 6025927374

Sent from Yahoo Mail on Android 
 
  On Tue, Feb 7, 2017 at 8:02 AM, Boris Zbarsky<notifications@github.com> wrote:   

Does this also mean we're going to change USVString back to [EnsureUTF16] DOMString


We could do that if we wanted to, yes. Whether we want to is less clear.


Is there any other goal here other than avoiding doubling the number of typed-array-related types?


Or quadrupling...

But note that this also fixes longstanding issues wherein you couldn't have a [Clamp] integer in a union or a dictionary, for example.

domenic:
Ping @bzbarsky

dontcallmedom:
for sake of book keeping, this would also address https://www.w3.org/Bugs/Public/show_bug.cgi?id=28834 (which is something that is also needed in https://github.com/w3c/mediacapture-main/issues/384

index.bs
```diff
@@ -8555,6 +8553,66 @@ entails.
 </div>
 
 
+<h4 id="LegacyTreatNullAsEmptyString" extended-attribute lt="LegacyTreatNullAsEmptyString" oldids="TreatNullAs">[LegacyTreatNullAsEmptyString]</h4>
```
tobie: Wondering whether the name change belongs in a different PR. I guess this depends on what makes it easier to file and track this in specs and implementations. Thoughts?

index.bs
```diff
@@ -5953,6 +5935,54 @@ and joining them with the string “Or”.
...
+    1.  If |type| is a [=typedef=], add the [=extended attributes associated with=] the type being
+        given a new name to |extended attributes|.
+    1. Return |extended attributes|.
+</div>
+
+For any [=type=], the [=extended attributes associated with=] it must only contain
+[=extended attributes=] that are
+<a href="#extended-attributes-applicable-to-types">applicable to types</a>.
+
+The [=type name=] of a type associated with [=extended attributes=] is the concatenation of the
```
domenic: It is specified, since the type is `Foo?` and the new type is `[Clamped] Foo?`. I'll add an example though.

domenic: OK, I can't add an example since it's actually impossible to use any of the current type-applicable extended attributes with nullable types, or use them together.

tobie: Heh!

index.bs
```diff
@@ -6505,6 +6536,9 @@ when passed to a [=platform object=] expecting that type, and how IDL values
 of that type are <dfn id="dfn-convert-idl-to-ecmascript-value" export lt="converted to an ECMAScript value|converted to ECMAScript values">converted to ECMAScript values</dfn>
 when returned from a platform object.
 
+Note that the sub-sections and algorithms below also apply to the related types created by applying
```
domenic: I don't think we want to do that in every conversion algorithm.

domenic:
OK, ready for another pass...

domenic:
Great catches again!! Ready for another pass.

dontcallmedom:
Looking at the grammar, I see 3 places where `Type` is used rather than `TypeWithExtendedAttributes`:
* `ReturnType`
* `ExtendedAttributeList Type identifier Default` in `DictionaryMember`
* `ExtendedAttributeList Type Ellipsis ArgumentName` in `Argument`

I think I understand the reason for `ReturnType` (no point in WebIDL-constraining a returned value), but I don't understand the reasoning for `DictionaryMember` or `Argument`.

dontcallmedom:
I've figured it out - without `optional` and `required`, extended attributes that apply to the member or the argument vs the type are not distinguishable.

tobie:
@domenic: It would be impossible for the parse to distinguish whether the extended attribute belongs to the dictionary member (or the argument) in such cases. This is thus [clarified in prose](https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with) on a per extended attribute basis. It remains transparent to editors, though.

For example in the below sample, [Clamp] will be associated with the `long` type, while [CustomAttr] with the `bar` dictionary member:

```webidl
dictionary Foo {
  [Clamp] long foo;
  [CustomAttr] DOMString bar;
};
```

tobie:
Also closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28834.

cdumez:
At https://heycam.github.io/webidl/#TreatNullAs , the example says:
[TreatNullAs=EmptyString] attribute DOMString owner;

Should this be:
attribute [TreatNullAs=EmptyString] DOMString owner;

now?

Or did you not do this part of the change?

domenic:
That's a bug, yeah. Will fix!

Manishearth:
Updated the File and WebCrypto APIs https://github.com/w3c/FileAPI/pull/116, https://github.com/w3c/webcrypto/pull/218

PR #289, opened 2017-01-31
Adjust wording in the legacy callers section.

rakuco:
Say "interfaces that also support", not "interfaces that also supports".


tobie:
Thanks for the PR!

PR #299, opened 2017-02-03
Ban nullable records in dictionary members and operation arguments.

### LINKED ISSUE #295: Where does it say that a nullable record<> cannot be the type of a dictionary member?
There is a note in https://heycam.github.io/webidl/#idl-nullable-type that says:

> Note: Although dictionary and record types can in general be nullable, they cannot when used as the type of an operation argument or a dictionary member.

but I see nothing in the spec supporting the claim that a nullable record type cannot be the type of a dictionary member.  The nullable-dictionary case is covered by this language in <https://heycam.github.io/webidl/#idl-dictionaries>:

> If the Type is an identifier followed by ?, then the identifier must identify an interface, enumeration, callback function or typedef.

(But even that doesn't handle a typedef of a dictionary!)  But nothing there says anything about records.

/cc @jyasskin @domenic @tobie @annevk

### LINKED ISSUE #297: Where does it say that a nullable record<> cannot be the type of an operation argument?
This is like https://github.com/heycam/webidl/issues/295 but for operation arguments.  https://heycam.github.io/webidl/#idl-operations says:

> If the Type of an operation argument is an identifier followed by ?, then the identifier must identify an interface, enumeration, callback function or typedef.

Again, nothing about records.

/cc @jyasskin @domenic @tobie

jyasskin:
Fixes #295 and fixes #297.

@bzbarsky How's this?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1630,6 +1630,10 @@ If the operation argument type is an [=identifier=]
 not followed by <emu-t>?</emu-t>, then the identifier must
 identify any one of those definitions or a [=dictionary=].
 
+If the operation argument type, after resolving typedefs,
+is a [=nullable type=],
+its [=inner type=] must not be a [=record type=] or [=dictionary type=].
+
 <pre highlight="webidl" class="syntax">
     interface interface_identifier {
       return_type identifier(type identifier, type identifier /* , ... */);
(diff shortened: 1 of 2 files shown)


bzbarsky:
Looks pretty good.  I think this is what Gecko implements in practice, by the way.

PR #300, opened 2017-02-06
Clarify that exceptions are always thrown in the current Realm

### LINKED ISSUE #292: Environment stuff for exception throwing is nonsense
Was just pointed to https://heycam.github.io/webidl/#dfn-current-global-environment today.  Why do we think there is a Function on "the ECMAScript call stack" at all?  What if we're in https://html.spec.whatwg.org/multipage/browsers.html#windowproxy-getownproperty invoked at toplevel like so:

    <script>crossOriginIframe.contentWindow.history</script>

?

Once we get past that, why would this Function be the `this` of anything?  That part doesn't make sense.

I think this whole section should be replaced by a reference to https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object or so.

@domenic

domenic:
Closes #292, in particular by removing the nonsensical "current global environment" concept and switching the example to reflect the desired semantics instead of the opposite.

This commit also modernizes the section and fixes a few small bugs, such as the omission of the message for the DOMException case, or the attempt to call the DOMException constructor as a function, which in most implementations will throw.

***
[Preview](https://cdn.rawgit.com/heycam/webidl/977f043/index.html) | [Diff w/ current ED](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fheycam.github.io%2Fwebidl%2F&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F977f043%2Findex.html) | [Diff w/ base](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F662b008%2Findex.html&doc2=https%3A%2F%2Fcdn.rawgit.com%2Fheycam%2Fwebidl%2F977f043%2Findex.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -116,6 +116,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         url: sec-code-realms
             text: Realm
             text: ECMAScript global environment
+        text: current Realm; url: current-realm
         text: Completion; url: sec-completion-record-specification-type
         text: ObjectCreate; url: sec-objectcreate
         text: CreateBuiltinFunction; url: sec-createbuiltinfunction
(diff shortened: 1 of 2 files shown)


bzbarsky:
Looks great, thanks!

PR #304, opened 2017-02-08
Allow records to be omitted

annevk:
Fixes https://github.com/whatwg/fetch/issues/479.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/annevk/optional-records.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/8c11c7e..annevk/optional-records:10d27c8.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1772,16 +1772,12 @@ corresponding argument omitted.
     conversion of <emu-val>undefined</emu-val> to be used (i.e., <emu-val>false</emu-val>).
 </p>
 
-If the type of an argument is a [=dictionary type=] or [=record type=]
-or a [=union type=] that has a
-dictionary or record type as one of its [=flattened member types=],
-and that dictionary type and its ancestors have no [=required dictionary member|required members=],
-and the argument is either the final argument or is followed only by
-[=optional arguments=], then
-the argument must be specified as optional.
-Such arguments are always considered to have a
-[=optional argument/default value=] of an empty dictionary or record, as appropriate,
-unless otherwise specified.
+If the type of an argument is a [=dictionary type=] or a [=union type=] that has a dictionary as one
+of its [=flattened member types=], and that dictionary type and its ancestors have no
+[=required dictionary member|required members=], and the argument is either the final argument or is
+followed only by [=optional arguments=], then the argument must be specified as optional. Such
+arguments are always considered to have a [=optional argument/default value=] of an empty
+dictionary, unless otherwise specified.
 
 <div class="note">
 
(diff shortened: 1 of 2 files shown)


annevk:
Note that if we do this a nice follow-up would be to define `{}` so you can provide a default value and not have to deal with "present" checking in prose.

tobie:
Trying to wrap my head around this. A couple of questions:

1. Does this affect the ["distinguishability" algorithm](https://heycam.github.io/webidl/#dfn-distinguishable)? (And in particular its first step.)
2. Records have [restrictions](https://heycam.github.io/webidl/#idl-nullable-type) as to when they can be nullable similar to those of dictionaries. Do these still hold?

bzbarsky:
Yeah, we need some more changes here.  A minimal list (there might be others):

1. We need to change distinguishability to make record distinguishable from nullable things.
2. We need to remove the "optional record argument gets null as default value" bit, because that would now just cause it to throw.
3. We need to re-allow nullable records as dictionary members. See the "If the type of the dictionary member, after resolving typedefs, is a nullable type" wording in https://heycam.github.io/webidl/#dfn-dictionary

annevk:
I can't find 2 of your list and I don't understand 3. Why should we allow records to be nullable?

annevk:
Anyway, it's probably best if @jyasskin or you or @domenic do this instead. Not really my area.

bzbarsky:
Actually, add:

4. Need to re-allow nullable record operation arguments.

> I can't find 2 of your list

Hmm.  I guess we effectively only had this for trailing optional records, not all records, and your PR changes that part.

> Why should we allow records to be nullable?

Why shouldn't we?  Note that they are _already_ allowed to be nullable everywhere except dictionary members and operation arguments.  The only reason they're not allowed there is because `null` used to be a valid record value, and the idea was that if null is passed that means "record produce from `null`", not "null".  But if `null` is _not_ a valid record value, then a nullable record makes perfect sense.

> Anyway, it's probably best if @jyasskin or you or @domenic do this instead.

Sure.  I'm pretty swamped with other things, but if @jyasskin or @domenic don't get to this by the time I clear other things off my plate and am seriously blocked on this to implement `record` in Gecko I will do it.

bzbarsky:
Anne, I put up https://github.com/heycam/webidl/pull/305 for what I think we should do here.  It's basically your changes, with a few more restrictions on nullable records removed and one typo fix where an "or" went away that shouldn't have.

annevk:
Thanks, I thought I had responded to your question here, but the it seems that didn't get through. The main reason I thought we didn't want nullable records was because you should use undefined for that scenario. Null is essentially a value that isn't really needed here. (Now I understand that for return values we don't really allow undefined or object at the moment, which is a problem that is tracked somewhere I believe.)

bzbarsky:
> was because you should use undefined for that scenario.

Is there a record-specific argument for that?  Why would the same argument not apply to "object"?

annevk:
I don't think there is a record-specific argument, other than not letting "bad practice" spread.

domenic:
Superceded by https://github.com/heycam/webidl/pull/305

PR #305, opened 2017-02-11
Allow records to be nullable in all cases and omitted as optional operation arguments.

bzbarsky:
Basically, make record<> more like sequence<> than like dictionary in terms of things that can be converted to it (without the requirement for iterability, of course)

***
[Preview](https://s3.amazonaws.com/pr-preview/bzbarsky/webidl/record-not-dictionarylike.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/8c11c7e..bzbarsky:record-not-dictionarylike:c3530c7.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1633,7 +1633,7 @@ identify any one of those definitions or a [=dictionary=].
 
 If the operation argument type, after resolving typedefs,
 is a [=nullable type=],
-its [=inner type=] must not be a [=record type=] or [=dictionary type=].
+its [=inner type=] must not be a [=dictionary type=].
 
 <pre highlight="webidl" class="syntax">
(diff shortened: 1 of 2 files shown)


bzbarsky:
Fixes whatwg/fetch#479 and is an updated version of https://github.com/heycam/webidl/pull/304

bzbarsky:
> But, should the distinguishability categories also be updated to move record from "dictionary-like" to "sequence-like"?

Good question!  I stopped on that too, when reading this.  The answer is "no".  That's because for distinguishability categories "sequence-like" means "iterable thing" and "dictionary-like" means "some random object".  I mean, "callback interfaces" are "dictionary-like"....

PR #328, opened 2017-03-23
Add the ability to construct a callback function

domenic:
This helps move some of the heavy lifting into IDL, to avoid issues like https://github.com/whatwg/html/issues/2381.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/construct-helper.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/fe42994...d445e3f.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12169,7 +12169,7 @@ a [=callback interface=] that:
 A <dfn>Web IDL arguments list</dfn> is a [=list=] of values each of which is either an IDL value or
 the special value “missing”, which represents a missing optional argument.
 
-<div algorithm="convert an IDL arguments list to an ECMAScript arguments list">
+<div algorithm>
     To <dfn lt="converting" for="Web IDL arguments list">convert a Web IDL arguments list to an
     ECMAScript arguments list</dfn>, given a [=Web IDL arguments list=] |args|, perform the
     following steps:
@@ -12186,8 +12186,7 @@ the special value “missing”, which represents a missing optional argument.
             1.  [=list/Append=] |convertResult| to |esArgs|.
             1.  Set |count| to |i| + 1.
         1.  Set |i| to |i| + 1.
-    1.  [=list/Remove=] all items from |esArgs| except the first |count|, so that |esArgs|'s
-        [=list/size=] is now |count|.
+    1.  Truncate |esArgs| to contain |count| items.
     1.  Return |esArgs|.
 </div>
 
@@ -13002,8 +13001,6 @@ The following typographic conventions are used in this document:
 *   Unicode characters: <span class="char">U+0030 DIGIT ZERO ("0")</span>
 *   Extended attributes: [<code class="idl">ExampleExtendedAttribute</code>]
 *   Variable names in prose and algorithms: <var ignore>exampleVariableName</var>.
-*   Algorithms use the [=conventions of the ECMAScript specification=],
-    including the ! and ? notation for unwrapping completion records.
 *   IDL informal syntax examples:
     <pre highlight="webidl" class="syntax">
         interface identifier {
@@ -13042,6 +13039,17 @@ The following typographic conventions are used in this document:
         // This is an ECMAScript code block.
         window.onload = function() { window.alert("loaded"); };
     </pre>
+
+The following conventions are used in the algorithms in this document:
+
+*   Algorithms use the [=conventions of the ECMAScript specification=],
+    including the ! and ? notation for unwrapping completion records.
+*   Algorithms sometimes treat returning/throwing values and returning completion records
+    interchangeably. That is, an algorithm that uses return/throw terminology may be treated as
+    returning a completion record, while one that returns a completion record may be treated as
+    returning a value or throwing an exception.
+*   Completion records are extended by allowing them to contain values that are not ECMAScript
+    values, such as Web IDL values.
 
 
 <h2 id="conformance" class="no-num" oldids="conformant-algorithms">Conformance</h2>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12374,6 +12374,56 @@ described in the previous section).
...
+<div algorithm>
+
+    To <dfn id="construct-a-callback-function" export>construct</dfn> a
+    [=callback function type=] value |callable| with
+    a list of arguments |arg|<sub>0..|n|−1</sub>, each of which is either
+    an IDL value or the special value “missing” (representing a missing optional argument),
+    perform the following steps.
+    These steps will either return an IDL value or throw an exception.
+
+    1.  Let |completion| be an uninitialized variable.
```
bzbarsky: So this algorithm is pretty similar to <https://heycam.github.io/webidl/#invoke-a-callback-function>.  I wonder whether it's worth trying to factor parts of it out...  I guess the main similarity is the preparation (after the basic sanity-check, which is different) and the argument conversions, right?

Maybe that's not enough to worry about it.

index.bs
```diff
@@ -12374,6 +12374,56 @@ described in the previous section).
...
+    1.  If |callResult| is an abrupt completion, set |completion| to
+        |callResult| and jump to the step labeled <a href="#construct-return"><i>return</i></a>.
+    1.  Set |completion| to the result of [=converted to an IDL value|converting=]
+        |callResult|.\[[Value]] to an IDL value of the same type as the operation’s
+        return type.
+    1.  <i id="construct-return">Return:</i> at this
+        point |completion| will be set to an ECMAScript completion value.
+        1.  [=Clean up after running a callback=] with |stored settings|.
+        1.  [=Clean up after running script=] with |relevant settings|.
+        1.  Return |completion|.
```
bzbarsky: Up front we said "These steps will either return an IDL value or throw an exception."  But we're actually returning an ES completion value... I'm not sure how to reconcile those.

Of course I'm also not sure in what sense "converting callResult.[[Value]] to an IDL value" returns an ES completion value...

index.bs
domenic: I factored out the arguments conversion, which I think factors the most cleanly. The rest varies a bit too much IMO.

index.bs
domenic: This is a preexisting "problem" that ideally I'd like to not fix in this PR.

Personally I think it is OK for specs to treat returning completion values and traditional return/throw flow the same way, without an explicit conversion step. (And similarly, treating return/flow throw as returning completion values.) It's a bit strange that the ES completion record holds a Web IDL value, but IMO fine.

domenic: We could mention this equivalence and our expanded use of ES completion values in the conventions section.

index.bs
```diff
@@ -12166,14 +12166,37 @@ a [=callback interface=] that:
...
+    1.  While |i| &lt; |args|'s [=list/size=]:
+        1.  If |args|[|i|] is the special value “missing”, then [=list/append=]
+            <emu-val>undefined</emu-val> to |esArgs|.
+        1.  Otherwise, |args|[|i|] is an IDL value:
+            1.  Let |convertResult| be the result of [=converted to an ECMAScript value|converting=]
+                |args|[|i|] to an ECMAScript value. Rethrow any exceptions.
+            1.  [=list/Append=] |convertResult| to |esArgs|.
+            1.  Set |count| to |i| + 1.
+        1.  Set |i| to |i| + 1.
+    1.  [=list/Remove=] all items from |esArgs| except the first |count|, so that |esArgs|'s
```
tobie: "Truncate" is a lot clearer to read, even though it's not defined in infra. Should it be defined there? (see https://github.com/whatwg/infra/issues/100)

index.bs
```diff
@@ -12166,14 +12166,37 @@ a [=callback interface=] that:
...
+    1.  Let |count| be 0.
+    1.  While |i| &lt; |args|'s [=list/size=]:
+        1.  If |args|[|i|] is the special value “missing”, then [=list/append=]
+            <emu-val>undefined</emu-val> to |esArgs|.
+        1.  Otherwise, |args|[|i|] is an IDL value:
+            1.  Let |convertResult| be the result of [=converted to an ECMAScript value|converting=]
+                |args|[|i|] to an ECMAScript value. Rethrow any exceptions.
+            1.  [=list/Append=] |convertResult| to |esArgs|.
+            1.  Set |count| to |i| + 1.
+        1.  Set |i| to |i| + 1.
```
tobie: Increment? (see whatwg/infra#99).

index.bs
```diff
@@ -12374,6 +12373,42 @@ described in the previous section).
...
+    1.  Let |stored settings| be |callable|'s [=callback context=].
+    1.  [=Prepare to run script=] with |relevant settings|.
+    1.  [=Prepare to run a callback=] with |stored settings|.
+    1.  Let |esArgs| be the result of [=Web IDL arguments list/converting=] |args| to an ECMAScript
+        arguments list. If this throws an exception, set |completion| to the completion value
+        representing the thrown exception and jump to the step labeled
+        <a href="#construct-return"><i>return</i></a>.
+    1.  Let |callResult| be [=Construct=](|F|, |esArgs|).
+    1.  If |callResult| is an abrupt completion, set |completion| to
+        |callResult| and jump to the step labeled <a href="#construct-return"><i>return</i></a>.
```
tobie: I know we use steps labelled "return" all over the place. I still find that icky. Of course we shouldn't change it here.

index.bs
tobie: Yeah, we seem to mix that up all over the place. I guess we should define it more clearly. Maybe that's yet another thing for the infra standard?

bzbarsky: I think having an explicit equivalence here is fine, fwiw.  We should just do that.

PR #329, opened 2017-03-23
Note what "present" means in ES close to its definition

domenic:
Also change how it's linked to a little bit.

See https://github.com/whatwg/fetch/issues/513 for the motivation.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/present.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/fe42994...4442344.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4102,12 +4102,17 @@ to the dictionary members defined on |D| and on any of |D|’s
 On a given dictionary value, the presence of each dictionary member
 is optional, unless that member is specified as required.
 When specified in the dictionary value, a dictionary member is said to be
-<dfn id="dfn-present" export>present</dfn>, otherwise it is [=present|not present=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4102,13 +4102,18 @@ to the dictionary members defined on |D| and on any of |D|’s
 On a given dictionary value, the presence of each dictionary member
 is optional, unless that member is specified as required.
 When specified in the dictionary value, a dictionary member is said to be
-<dfn id="dfn-present" export>present</dfn>, otherwise it is [=present|not present=].
+<dfn id="dfn-present" export lt="present|not present" for="dictionary member">present</dfn>, otherwise it is [=not present=].
```
annevk: Should this be for=dictionary/member? See https://github.com/tabatkins/bikeshed/issues/895.

domenic: Hmm maybe, but we already have default value for="dictionary member", so I went for consistency.

Issue #129, opened 2016-06-07
Should probably spec a hasInstance for DOM interface objects


bzbarsky:
Right now we have a [[HasInstance]] thing at http://heycam.github.io/webidl/#es-interface-hasinstance but that's gone in ES6.

Proposed prose with similar semantics is something like:
1. Let `F` be the `this` value.
2. Let `V` be the first argument we were passed.
3. If `F` is not an object, return `false`.
4. If `F` is not an interface object, return `false` (or delegate to `OrdinaryHasInstance`?)
5. If `V` is not an object, return `false`.
6. If `V` is an object that implements the interface for which `F` is  the interface object (defined in terms of brands, whatever), return `true`.
7. Return `OrdinaryHasInstance(F, V)`.

@domenic, thoughts?

An open question is whether `Node[Symbol.hasInstance] == Element[Symbol.hasInstance]` should test true or false.  We could go either way, given the above prose, since they would implement the same algorithm...

bzbarsky:
> If F is not an interface object, return false (or delegate to OrdinaryHasInstance?)

Fwiw, the behavior I'm going to do in Firefox for now is to delegate.

domenic:
Can you clarify the intent of the existing [[HasInstance]]? I'm unable to understand in what ways it is different from OrdinaryHasInstance.

Is the test "is a platform object that implements the interface for which O is the interface prototype object" meant to work cross-realm? It seems like youre revised version might be or might not be, depending on whether brands are cross-realm? If that's the desired semantics, how many browsers implement that?

Web IDL's current algorithm performs an observable Get of the `"prototype"` property, as does OrdinaryHasInstance. Both also do observable proto-walks. Your algorithm omits these. Is changing that a good idea?

bzbarsky:
> Is the test "is a platform object that implements the interface for which O is the interface prototype object" meant to work cross-realm?

Yes.  This is in fact the entire intent of the existing [[HasInstance]].

> It seems like youre revised version might be or might not be, depending on whether brands are cross-realm?

They better be, because DOM functions can be `.call`-ed cross-realm, yes?

> If that's the desired semantics, how many browsers implement that?

I believe Firefox is the only one that does.  That said, this is the behavior people agreed on when this was discussed on es-discuss several years ago...  I can't help it if other browsers didn't make implementing Web IDL a very high priority.  ;)

I agree that this does cause problems in terms of potential compat issues for whoever switches behavior.

[...]

PR #339, opened 2017-04-07
Define 'CancelationError'.

mikewest:
In order to support the 'CancelationController' concept defined in
whatwg/dom#434, this patch specifies a new 'DOMException' named
'CancelationError' in order to ensure that APIs that consume this
new concept have a standard way of responding to its signals.

***
[Preview](https://s3.amazonaws.com/pr-preview/mikewest/webidl/cancelationerror.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/faaaaa9...mikewest:1bb6c7b.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -142,6 +142,11 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: @@unscopables; url: sec-well-known-symbols
         text: NewTarget; url: sec-built-in-function-objects
         text: Number Type; url: sec-ecmascript-language-types-number-type
+
+<!-- Remove once DOM takes https://github.com/whatwg/dom/pull/434 -->
+urlPrefix: https://github.com/whatwg/dom/pull/434; spec: DOM
+    type: interface
+        text: CancelationController
 </pre>
 
 <style>
@@ -4666,7 +4671,7 @@ Note: If an error name is not listed here, please file a bug as indicated at the
         </tr>
         <tr>
             <td>"<dfn id="cancelationerror" exception export><code>CancelationError</code></dfn>"</td>
-            <td>The operation was canceled before completion.</td>
+            <td>The operation was canceled via a {{CancelationController}}.</td>
             <td>—</td>
         </tr>
     </tbody>
--- a/index.bs
+++ b/index.bs
@@ -4664,6 +4664,11 @@ Note: If an error name is not listed here, please file a bug as indicated at the
             <td>The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.</td>
             <td>—</td>
         </tr>
+        <tr>
+            <td>"<dfn id="cancelationerror" exception export><code>CancelationError</code></dfn>"</td>
+            <td>The operation was canceled before completion.</td>
+            <td>—</td>
+        </tr>
     </tbody>
 </table>
 


mikewest:
WDYT, @tobie, @annevk, @jakearchibald, and anyone else who cares?

annevk:
We should note that this is exclusively for users of `CancelationController` objects. And maybe wait a bit to give @ljharb a chance to explain why this should be cross-realm safe while so many other things are not.

annevk:
I should add that it would be quite easy to provide cross-realm safety checks for all DOMException objects now or at any point in the future. So even if we don't do it now it should be easy to retrofit if it becomes a problem.

tobie:
I'm sure there's a good reason not to reuse `AbortError`. It would be good to document it (here).

mikewest:
In https://github.com/whatwg/dom/pull/434#issuecomment-292471825, @annevk noted:

> (As an aside, I don't think we should reuse "AbortError" DOMException even if we decided not to care about cross-realm branding, since that is already used by APIs and it's not clear they all use it with the semantics we are going for here.)

I agree with that.

bzbarsky:
So my first reaction is actually that the name is extremely error-prone, because both "Cancelation" and "Cancellation" are accepted spellings.  It's worth thinking about renaming this to avoid that problem.  :(

As evidence, note that the code and title of https://github.com/whatwg/dom/pull/434 don't even agree on the spelling, and they were written by the same person.

jakearchibald:
FWIW, as a Brit, I don't have issues with `color` in CSS, I just learned the spelling for that property.

ljharb:
(My explanations for wanting "no collisions", "cross-realm", and "not forgeable" are in https://github.com/whatwg/dom/pull/434#issuecomment-292640413)

bzbarsky:
> FWIW, as a Brit, I don't have issues with color in CSS, I just learned the spelling for that property.

Yes, but that's a somewhat different situation, in that there are at least clear boundaries about which spelling is accepted where.  For "cancellation" there are just two accepted spellings period, as far as I can tell.

domenic:
I question the need for this at all as opposed to just using "AbortError". I don't see a need for an API to distinguish how it was canceled/aborted. That should be transparent to the user of the API (who sees this exception). Only the producer should care about CancelationController and friends.

I think it would be better to go with the platform's long precedent of using "AbortError" for these types of things. Otherwise it's very strange that e.g. XHR uses "AbortError" but fetch would use "CancelationError".

ljharb:
If it's an error, I might want to give the user an error message and a "retry" button. If it's a cancellation, I might instead prefer to reset the UI.

domenic:
That seems irrelevant to my point. "AbortError" is used to represent cancelation/abortion all over the platform today, where you would reset the UI. We should keep using it that way.

ljharb:
Ah - I don't have any opinion on what you call the thing that means "it was cancelled", or whether it's reused - if you want to use an already existing AbortError, that's fine! but I'd still want some or all of https://github.com/heycam/webidl/pull/339#issuecomment-292667566 to apply to AbortError, which seems possible in an additive way.

annevk:
If we WONTFIX this we'll need to revert https://github.com/w3c/testharness.js/pull/260.

tobie:
OK, so it seems we should:

- [ ] revert w3c/testharness.js#260,
- [ ] file a new issue against WebIDL for the requirements exposed by @ljharb in https://github.com/heycam/webidl/pull/339#issuecomment-292667566,
- [ ] close this.

Right?

annevk:
@tobie I think it's fine to open an issue on branding, but as per https://github.com/heycam/webidl/issues/129#issuecomment-292522793 it's really something the JavaScript folks should embrace first as a thing that deserves exposure.

Agreed on the other points.

domenic:
Yes, I would not open an issue on branding.

PR #347, opened 2017-04-19
Introduce the CSSOMString type, impls choice of DOMString or USVString

SimonSapin:
The observable behavior difference is whether surrogate code units are preserved or replaced with U+FFFD REPLACEMENT CHARACTER.

CSSWG resolved to allow either in CSSOM: https://github.com/w3c/csswg-drafts/issues/1217#issuecomment-295053842

***
[Preview](https://s3.amazonaws.com/pr-preview/SimonSapin/webidl/cssomstring.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/9d039f6...SimonSapin:74c0019.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2248,7 +2248,7 @@ can also be placed on an [=attribute=].
 In this case, the string to convert the object to is the
 value of the attribute.  The <emu-t>stringifier</emu-t> keyword
 must not be placed on an attribute unless
-it is declared to be of type {{DOMString}} or {{USVString}}.
+it is declared to be of type {{DOMString}}, {{USVString}}, or {{CSSOMString}}.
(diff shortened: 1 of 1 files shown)


annevk:
Can't this just be a typedef in CSS? It seems weird to allow two code paths, but okay, hope we can align on one or the other long term.

SimonSapin:
Yes, that’s better. I didn’t know `typedef` was a thing in WebIDL, thanks!

tabatkins:
Yeah, it's weird to have two code paths; browsers do intend to align long-term on being Unicode-clean, but right now Servo will have perf issues using DOMString, and other browsers will have perf issues using USVString, so :shrug:

Issue #352, opened 2017-04-28
Throwing when converting detached ArrayBuffers is not what browsers do


domenic:
Test: https://jsbin.com/qusumat/edit?html,console

I will remove this in my upcoming [AllowShared] PR. I was already finding it weird that we threw for detached buffers, but not for detached typed arrays surrounding detached buffers.

annevk:
Is JavaScript now aligned with implementations on ArrayBuffer around this? I believe some folks still tried to make them throw for more things.

annevk:
Also note that for the APIs you tested browsers might not use IDL's copying infrastructure since that's not the option that yields the best performance.

domenic:
> Is JavaScript now aligned with implementations on ArrayBuffer around this? I believe some folks still tried to make them throw for more things.

No. I thought we'd best remove any aspirational stuff at least for now, especially given the inconsistency.

> Also note that for the APIs you tested browsers might not use IDL's copying infrastructure since that's not the option that yields the best performance.

It's independent of the copy/reference; it should occur whenever converting a buffer source type to IDL, which happens before any algorithm steps.

annevk:
So what happens instead, we'll just treat it as an empty buffer?

I guess what I'm wondering about is whether we should leave some kind of issue marker in for the aspirational stuff or whether everyone is down with just giving up.

domenic:
Fixed by https://github.com/heycam/webidl/commit/c59bdcbcf2039c89eb7ea9a35e1e2dba621519ec

PR #353, opened 2017-04-28
Add [AllowShared] to allow accepting SharedArrayBuffers

domenic:
Previous to this commit, any API that accepted any of the buffer source types would also accept a SharedArrayBuffer. This was not the approach we want; instead, we want APIs to explicitly opt-in. Enabled by previous work creating annotated types, this adds the new [AllowShared] extended attribute to allow APIs to opt in to accepting SharedArrayBuffers.

While here, this removes an unimplemented check for detached array buffers when converting from ES values to IDL types, fixing #352.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/allowshared.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/9d039f6...67710b2.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -41,6 +41,10 @@ urlPrefix: http://www.unicode.org/glossary/; spec: UNICODE
     type: dfn
         text: Unicode scalar value; url: unicode_scalar_value
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
+    type: interface; for: ECMAScript
+        text: ArrayBuffer; url: sec-arraybuffer-objects
+        text: DataView; url: sec-dataview-objects
+        text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
     type: dfn
         text: NumericLiteral; url: sec-literals-numeric-literals
         text: ECMAScript error objects; url: sec-error-objects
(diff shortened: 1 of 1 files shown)


annevk:
So this is primarily for WebGL at this point right? Do we have some follow-up filed there?

domenic:
Yes, although I am told it might expand over time, e.g. for web audio or similar.

We don't have a file-up filed on Web GL, because to my knowledge their specs are developed in private with no public issue tracker or "editor's draft". I am not sure what the procedure will be. Maybe we'll need to maintain a moneypatch somewhere.

I'll try to poke https://github.com/tc39/ecmascript_sharedmem/issues/168 about that issue.

bzbarsky:
The WebGL issue tracker is at https://github.com/KhronosGroup/WebGL/issues

domenic:
Ah OK yeah, I just found that myself; clearly my info was out of date.

bzbarsky:
And https://www.khronos.org/registry/webgl/specs/latest/ links to the editor's drafts of WebGL1 and WebGL2, fwiw.

index.bs
```diff
@@ -7945,6 +7966,38 @@ This section defines a number of
...
+If the [{{AllowShared}}] [=extended attribute=] appears on one of the [=buffer source types=], it
+creates a new IDL type that allows the buffer source type to be backed by an ECMAScript
+{{ECMAScript/SharedArrayBuffer}}, instead of only by a non-shared {{ECMAScript/ArrayBuffer}}.
+
+The [{{AllowShared}}] extended attribute must [=takes no arguments|take no arguments=].
+
+A type that is not a [=buffer source type=] must not be
+[=extended attributes associated with|associated with=] the [{{AllowShared}}] extended attribute.
+
+See the rules for converting ECMAScript values to IDL [=buffer source types=] in
```
tobie: I admit I had to look "entails" up. Maybe instead write:

> […] for the specific requirements that using [{{AllowShared}}] creates.

domenic: This is already used for every other extended attribute, so I'd rather be consistent.

domenic:
Does anyone else want to review before we merge? Ideally speak up within the next day-ish.

PR #356, opened 2017-05-01
Remove custom [[HasInstance]] behavior from interface objects

### LINKED ISSUE #129: Should probably spec a hasInstance for DOM interface objects

domenic:
This was only ever implemented in Firefox. It was added in 5982ce7e323cffd7a673cb9af973c1c4f1e76fbe. after being removed as a result of https://www.w3.org/Bugs/Public/show_bug.cgi?id=12295.

Closes #129.

---

This ensures that authors can maintain the invariant they are used to where `x instanceof Y` means that `Y.prototype`'s methods are available on `x` (which is not true in Firefox). It also aligns with normal author-created or built-in ECMAScript objects, where e.g. `{} instanceof otherWindow.Object` returns false.

As pointed out in #129, only @travisleithead from a non-Firefox browser expressed support for the currently-specced behavior, and they have not implemented it in the intervening 4 years; Travis, I'd be curious if your team's feelings have changed. Chrome strongly objects to the currently-specced behavior. /cc @cdumez (EDIT: I guess @samweinig is probably the right person for Web IDL) to hear about WebKit's opinion.

As such, we'd like to remove this from the spec, as it seems that it never got consensus for being added, although people may not have voiced their objections in a timely manner when it was originally proposed.

---

If there is a desire for adding a brand-checking API to the platform, we'd like that to be proposed separately. But simply as a matter of matching reality and improving interop, we think it's important to get the spec fixed on this matter sooner, and not block on such an API addition.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/hasinstance-non-exotic.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/9d039f6...0881b62.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6383,8 +6383,9 @@ more of the following are redefined in accordance with the rules for exotic obje
 \[[Call]],
 \[[Set]],
 \[[DefineOwnProperty]],
-\[[GetOwnProperty]], and
-\[[Delete]].
+\[[GetOwnProperty]],
+\[[Delete]] and
+\[[HasInstance]].
 
 Other specifications may override the definitions
 of any internal method of a [=platform object=]
--- a/index.bs
+++ b/index.bs
@@ -6383,9 +6383,8 @@ more of the following are redefined in accordance with the rules for exotic obje
 \[[Call]],
 \[[Set]],
 \[[DefineOwnProperty]],
-\[[GetOwnProperty]],
-\[[Delete]] and
-\[[HasInstance]].
+\[[GetOwnProperty]], and
+\[[Delete]].
 
 Other specifications may override the definitions
 of any internal method of a [=platform object=]
@@ -10283,16 +10282,6 @@ when applied to a [=legacy callback interface object=].
     1.  Perform [=!=] [=SetFunctionName=](|F|, |id|).
     1.  Perform [=!=] [=DefinePropertyOrThrow=](|F|, "length",
         PropertyDescriptor{\[[Value]]: 0, \[[Writable]]: <emu-val>false</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val>}).
-</div>
-
-<h5 id="es-legacy-callback-interface-object-hasinstance">\[[HasInstance]]</h5>
-
-<div algorithm="to invoke the internal [[HasInstance]] method of legacy callback interface objects">
-
-    The internal \[[HasInstance]] method of a [=legacy callback interface object=]
-    must behave as follows:
-
-    1.  [=ECMAScript/Throw=] a <emu-val>TypeError</emu-val> exception.
 </div>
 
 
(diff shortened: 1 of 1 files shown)


cdumez:
I support this change. I think it is best to align with EcmaScript here.

annevk:
Do you want me to submit a WPT for https://github.com/heycam/webidl/issues/129#issuecomment-292522793 with the opposite result? Or are you taking care of it?

index.bs
```diff
@@ -6384,9 +6383,8 @@ more of the following are redefined in accordance with the rules for exotic obje
 \[[Call]],
 \[[Set]],
 \[[DefineOwnProperty]],
-\[[GetOwnProperty]],
-\[[Delete]] and
-\[[HasInstance]].
```
tobie: We're still [modifying [[HasInstance]]](https://heycam.github.io/webidl/#es-legacy-callback-interface-object-hasinstance) for legacy callback interfaces, so you might want to keep it listed here.

domenic:
I can take care of the tests; I wanted to make sure this would be merged first.

domenic:
It appears nobody implements the throwing [[HasInstance]] for NodeFilter. We should probably just remove that.

tobie:
> It appears nobody implements the throwing [[HasInstance]] for NodeFilter. We should probably just remove that.

Why is it even specified?

domenic:
Unclear. Maybe because it's not really a sensible thing to be testing. But per spec it will throw anyway in most cases because these objects have no `.prototype`; see forthcoming tests.

domenic:
@bzbarsky OK to merge this?

bzbarsky:
Yes.  Note that Firefox is probably not going to be able change its behavior for a while; certainly not until we're able to offer a different branding-detection solution to consumers of the current behavior...

annevk:
@bzbarsky maybe we could experiment with a chrome-only-branding API? Presumably we don't need it for web sites...

bzbarsky:
We could try.  It'd be a bunch of work and compat risk that I'm not willing to spend time on any time in the next 6 months, personally, so just trying to set realistic expectations here.

domenic:
Bugs commented on:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1360715#c8
- https://bugs.chromium.org/p/chromium/issues/detail?id=717016#c5

Bugs for legacy callback interface objects (not directly related to this PR, but caught while writing tests in https://github.com/w3c/web-platform-tests/pull/5762):

- https://bugs.chromium.org/p/chromium/issues/detail?id=569301#c46 
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11894121/

If someone could check http://w3c-test.org/WebIDL/ecmascript-binding/legacy-callback-interface-object.html in Safari Tech Preview and file a bug if necessary, I'd much appreciate it, as I'm away from my testing Mac for the next week.

annevk:
Safari TP passes all 8 tests.

PR #358, opened 2017-05-02
WIP: add localizable dictionary

marcoscaceres:
Bringing this over from https://github.com/w3c/browser-payment-api/pull/455 where [it was suggested](https://github.com/w3c/browser-payment-api/pull/455#issuecomment-286342385) there that this be uplifted here.

I omitted the xrefs because: 
 1. like to get the prose right first.
 1. I don't know how to do it in BS yet.

***
[Preview](https://s3.amazonaws.com/pr-preview/marcoscaceres/webidl/localizable.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/9d039f6...marcoscaceres:e78d5f5.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12750,43 +12750,68 @@ The {{VoidFunction}} [=callback function=]
 type is used for representing function values that take no arguments and do not
 return any value.
 
-<h3>Locatlizable dictionary</h3>
 
-A localizable member is a dictionary member that represents a bidirectional algorithm paragraph when displayed, as defined by the bidirectional algorithm’s rules P1, P2, and P3, including, for instance, supporting the paragraph-breaking behavior of U+000A LINE FEED (LF) characters.
-
-A user agent is expected to honor the Unicode semantics of localizable members.
-
-The name of a localizable member is left to the defining specification (e.g., "title").
-
-Dictionaries that specify a localizable member must inherit from the {{Localizable}} dictionary.
-
-Specification authors must specify in prose which member(s) of the inheriting dictionary serve as localizable members. It is RECOMMENDED that specification authors limit localizable member in the prototype chain of inherited dictionaries (ideally to one).
+<h3 id="TextDirection" enum>TextDirection</h3>
 
 <pre class="idl">
-enum TextDirection {
-  "auto",
-  "ltr",
-  "rtl"
-};
+    enum TextDirection {
+      "auto",
+      "ltr",
+      "rtl"
+    };
+</pre>
 
-dictionary Localizable {
-  DOMString lang;
-  TextDirection dir = "auto";
-};</pre>
+The {{TextDirection}} [=enumerable=] is used to represent text directionality as described in [[!BIDI]].
 
-The {{lang}} member specifies the primary language for the localizable members in the prototype chain. Its value is a string. The empty string indicates that the primary language is unknown.
-Any other string must be interpreted as a language tag. Validity or well-formedness are not enforced. [[!LANG]]
 
-The {{dir}} member provides the higher-level override of rules P2 and P3 if has a value other than "auto". [[!BIDI]]
+<h3 id="localizable" dictionary>Localizable</h3>
+
+A <dfn export>localizable member</dfn> is a [=dictionary member=] that
+represents a bidirectional algorithm paragraph when displayed,
+as defined by the bidirectional algorithm’s rules P1, P2, and P3,
+including, for instance, supporting the paragraph-breaking behavior
+of <span class="char">U+000A LINE FEED (LF)</span> characters. [[!BIDI]]
+
+A user agent is expected to honor the Unicode semantics of [=localizable members=].
+
+The [=identifier=] of a [=localizable member=] is left to the defining specification (e.g., "title").
+
+Dictionaries that specify a [=localizable member=] must inherit from the {{Localizable}} dictionary.
+
+Specification authors must specify in prose which [=dictionary members|member(s)=]
+of the inheriting dictionary serve as [=localizable members=].
+
+It is recommended that specification authors limit the number of [=dictionary members=]
+that are [=localizable member=], ideally to one.
+
+<pre class="idl">
+    dictionary Localizable {
+      DOMString lang = "";
+      TextDirection dir = "auto";
+    };
+</pre>
+
+The {{Localizable/lang}} member specifies the primary language for the [=dictionary members=]
+which are [=localizable members=].
+Its value is a {{DOMString}}.
+The empty string indicates that the primary language is unknown.
+Any other string must be interpreted as a language tag.
+Validity or well-formedness are not enforced. [[!BCP47]]
+
+The {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
+if it has a value other than <a for=TextDirection enum-value>"auto"</a>. [[!BIDI]]
 
 <div class=example>
-dictionary PaymentItem : Localizable {
-    // "label" is a localizable member
-    required DOMString             label;
-    required PaymentCurrencyAmount amount;
-             boolean               pending = false;
-};
+   <pre highlight="webidl">
+       dictionary PaymentItem : Localizable {
+           // "label" is a localizable member
+           required DOMString label;
+           required PaymentCurrencyAmount amount;
+           boolean pending = false;
+       };
+    </pre>
 </div>
+
 
 <h2 id="extensibility">Extensibility</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -12766,46 +12766,38 @@ The {{TextDirection}} [=enumerable=] is used to represent text directionality as
 
 <h3 id="localizable" dictionary>Localizable</h3>
 
-A <dfn export>localizable member</dfn> is a [=dictionary member=] that
-represents one or more bidirectional algorithm paragraphs when displayed,
+A <dfn export>localizable member</dfn> is a dictionary member whose type is {{LocalizableString}}.
+Such members represent one or more bidirectional algorithm paragraphs when displayed,
 as defined by the bidirectional algorithm’s rules P1, P2, and P3,
 including, for instance, supporting the paragraph-breaking behavior
 of <span class="char">U+000A LINE FEED (LF)</span> characters. [[!BIDI]]
 
 A user agent is expected to honor the Unicode semantics of [=localizable members=].
 
-The [=identifier=] of a [=localizable member=] is left to the defining specification (e.g., "title").
-
-Dictionaries that specify a [=localizable member=] must inherit from the {{Localizable}} dictionary.
-
-Specification authors must specify in prose which [=dictionary members|member(s)=]
-of the inheriting dictionary serve as [=localizable members=].
-
-Specification authors should limit the number of [=localizable members=] to one per dictionary,
-including in [=inherited dictionaries=] (if any).
-
 <pre class="idl">
     dictionary Localizable {
       DOMString lang = "";
+      DOMString value = "";
       TextDirection dir = "auto";
     };
+    typedef (DOMString or LocalizedValue) LocalizableString;
 </pre>
 
-The {{Localizable/lang}} member specifies the primary language for the [=dictionary members=]
-which are [=localizable members=].
+The {{Localizable/lang}} member specifies the primary language for the {{Localizable/value}} member.
 Its value is a {{DOMString}}.
 The empty string indicates that the primary language is unknown.
 Any other string must be interpreted as a language tag.
 Validity or well-formedness are not enforced. [[!BCP47]]
 
-For the [=localizable members=], the {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
+The {{Localizable/value}} is a {{DOMString}} to be localized.
+
+For the {{Localizable/value}}, the {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
 if it has a value other than <a for=TextDirection enum-value>"auto"</a>. [[!BIDI]]
 
 <div class=example>
    <pre highlight="webidl">
-       dictionary PaymentItem : Localizable {
-           // "label" is a localizable member
-           required DOMString label;
+       dictionary PaymentItem {
+           required LocalizableString label;
            required PaymentCurrencyAmount amount;
            boolean pending = false;
        };
--- a/index.bs
+++ b/index.bs
@@ -12767,7 +12767,7 @@ The {{TextDirection}} [=enumerable=] is used to represent text directionality as
 <h3 id="localizable" dictionary>Localizable</h3>
 
 A <dfn export>localizable member</dfn> is a [=dictionary member=] that
-represents a bidirectional algorithm paragraph when displayed,
+represents one or more bidirectional algorithm paragraphs when displayed,
 as defined by the bidirectional algorithm’s rules P1, P2, and P3,
 including, for instance, supporting the paragraph-breaking behavior
 of <span class="char">U+000A LINE FEED (LF)</span> characters. [[!BIDI]]
--- a/index.bs
+++ b/index.bs
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
 type is used for representing function values that take no arguments and do not
 return any value.
 
+<h3>Locatlizable dictionary</h3>
+
+A localizable member is a dictionary member that represents a bidirectional algorithm paragraph when displayed, as defined by the bidirectional algorithm’s rules P1, P2, and P3, including, for instance, supporting the paragraph-breaking behavior of U+000A LINE FEED (LF) characters.
+
+A user agent is expected to honor the Unicode semantics of localizable members.
+
+The name of a localizable member is left to the defining specification (e.g., "title").
+
+Dictionaries that specify a localizable member must inherit from the {{Localizable}} dictionary.
+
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@annevk, I copy/pasta'ed a bunch from the Notifications spec, as suggested. But likely wrong. Can you please TAL.

index.bs
```diff
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
 type is used for representing function values that take no arguments and do not
 return any value.
 
+<h3>Locatlizable dictionary</h3>
```
annevk: Typo

index.bs
```diff
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
...
+
+A localizable member is a dictionary member that represents a bidirectional algorithm paragraph when displayed, as defined by the bidirectional algorithm’s rules P1, P2, and P3, including, for instance, supporting the paragraph-breaking behavior of U+000A LINE FEED (LF) characters.
```
annevk: This should probably be an exported dfn, no?

annevk: Also, this should reference BIDI, right?

index.bs
```diff
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
...
+
+<pre class="idl">
+enum TextDirection {
+  "auto",
+  "ltr",
+  "rtl"
+};
+
+dictionary Localizable {
+  DOMString lang;
```
annevk: Why doesn't lang default to the empty string?

index.bs
```diff
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
...
+  TextDirection dir = "auto";
+};</pre>
+
+The {{lang}} member specifies the primary language for the localizable members in the prototype chain. Its value is a string. The empty string indicates that the primary language is unknown.
+Any other string must be interpreted as a language tag. Validity or well-formedness are not enforced. [[!LANG]]
+
+The {{dir}} member provides the higher-level override of rules P2 and P3 if has a value other than "auto". [[!BIDI]]
```
annevk: if has?

index.bs
```diff
@@ -12750,6 +12750,43 @@ The {{VoidFunction}} [=callback function=]
...
+A user agent is expected to honor the Unicode semantics of localizable members.
+
+The name of a localizable member is left to the defining specification (e.g., "title").
+
+Dictionaries that specify a localizable member must inherit from the {{Localizable}} dictionary.
+
+Specification authors must specify in prose which member(s) of the inheriting dictionary serve as localizable members. It is RECOMMENDED that specification authors limit localizable member in the prototype chain of inherited dictionaries (ideally to one).
+
+<pre class="idl">
+enum TextDirection {
```
annevk: Maybe this should be its own chapter if we're going to have this here? This might be useful on its own as well.

annevk: Canvas apparently has CanvasDirection, which uses "inherit" rather than "auto"... Good times.

index.bs
tobie: Should spell "Lolcatlizable"

tobie:
Sent you a whole bunch of fixes in https://github.com/marcoscaceres/webidl/pull/1.

Didn't would to force push them to your branch, hence the pull request.

index.bs
tobie: Fixed in https://github.com/marcoscaceres/webidl/pull/1

index.bs
tobie: Should we add "inherit" to the enum and mark it as legacy?

index.bs
tobie: Fixed in https://github.com/marcoscaceres/webidl/pull/1.

index.bs
tobie: Fixed in https://github.com/marcoscaceres/webidl/pull/1.

annevk:
> Should we add "inherit" to the enum and mark it as legacy?

Dunno, probably not worth the churn.

index.bs
```diff
@@ -12751,6 +12751,68 @@ type is used for representing function values that take no arguments and do not
...
+A user agent is expected to honor the Unicode semantics of [=localizable members=].
+
+The [=identifier=] of a [=localizable member=] is left to the defining specification (e.g., "title").
+
+Dictionaries that specify a [=localizable member=] must inherit from the {{Localizable}} dictionary.
+
+Specification authors must specify in prose which [=dictionary members|member(s)=]
+of the inheriting dictionary serve as [=localizable members=].
+
+It is recommended that specification authors limit the number of [=dictionary members=]
```
annevk: "Specification authors should limit ..." is a little clearer than this. Per Infra we should avoid using "recommended".

index.bs
```diff
@@ -12751,6 +12751,68 @@ type is used for representing function values that take no arguments and do not
...
+
+The {{Localizable/lang}} member specifies the primary language for the [=dictionary members=]
+which are [=localizable members=].
+Its value is a {{DOMString}}.
+The empty string indicates that the primary language is unknown.
+Any other string must be interpreted as a language tag.
+Validity or well-formedness are not enforced. [[!BCP47]]
+
+The {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
+if it has a value other than <a for=TextDirection enum-value>"auto"</a>. [[!BIDI]]
```
annevk: This should be more clearly made to apply to the localizable members.

index.bs
```diff
@@ -12751,6 +12751,68 @@ type is used for representing function values that take no arguments and do not
...
+
+The [=identifier=] of a [=localizable member=] is left to the defining specification (e.g., "title").
+
+Dictionaries that specify a [=localizable member=] must inherit from the {{Localizable}} dictionary.
+
+Specification authors must specify in prose which [=dictionary members|member(s)=]
+of the inheriting dictionary serve as [=localizable members=].
+
+Specification authors should limit the number of [=localizable members=] to one per dictionary,
+including in [=inherited dictionaries=] (if any).
```
marcoscaceres: Is this bit about inherited dictionaries ok?

marcoscaceres: > It also fails to account for multiple paragraphs of a single member.

Said it's one or more. Does empty string count as one or zero?

index.bs
```diff
@@ -12798,7 +12798,7 @@ The empty string indicates that the primary language is unknown.
 Any other string must be interpreted as a language tag.
 Validity or well-formedness are not enforced. [[!BCP47]]
 
-The {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
+For the [=localizable member=], the {{Localizable/dir}} member provides the higher-level override of rules P2 and P3
```
annevk: This needs to be plural. It's not just one.

index.bs
annevk: I'm not sure. You'd have to check the bidi algorithm I'm afraid.

marcoscaceres: np, checking

marcoscaceres: If I'm reading correctly, a paragraph is denoted by the presence of a paragraph separator. However, it says that we can also define what a paragraph means: "Paragraphs may also be determined by higher-level protocols: for example, the text in two different cells of a table will be in different paragraphs".

Like we could say that a member's value constitutes a paragraph? Dunno. Opinions?

annevk: I think we should just leave it as one or more (per member).

tobie: I'm struggling to understand what we're trying to describe here. I'm clearly missing context, so please bear with me.

bzbarsky:
So just to check, this does not add any actual requirements on user agents in terms of the algorithms defined in the IDL spec, right?  This is just some infrastructure to allow other specs to not have to keep redefining Localizable?

annevk:
Yes.

index.bs
marcoscaceres: @tobie, basically, spec authors *should* avoid doing the following: 

```JS
dictionary Foo : Bar {
    DOMString label; // localizable member
    DOMString title; // localizable member
}

dictionary Bar : localizable {
   DOMString thingToShow; // localizable member
}
```

Because generally you want dir/lang to only influence one member - because, say, "title" might be in French (e.g., name of a Restaurant) - but the thingToShow might be in Spanish.  

However, we do have instances in the platform where dictionaries already have multiple localizable members (e.g., notifications) - but authors should generally try to avoid that kind of design. Instead, you might have (ignore the crap naming):

```JS
dictionary ThingLocalizedTitle : Localizable {
   DOMString title;
}
dictionary ThingLocalizedLabel : Localizable {
   DOMString label;
}

dictionary Thing {
   ThingLocalizedLabel label;
   ThingLocalizedLabel ThingTitle;
}
```

Or possibly something like the following - which would give allow localization of members on a case by case basis:

```JS
dictionary Thing {
   (DOMString or ThingLocalizedLabel) label;
   (DOMString or ThingLocalizedTitle) title;
}
```

domenic: Maybe we should go a step further and define

```webidl
dictionary LocalizedValue : Localizable {
  DOMString value;
}

typedef (DOMString or LocalizedValue) LocalizableString;
```

and then encourage

```webidl
dictionary Thing {
  LocalizableString label;
  LocalizableString value;
}
```

? That seems better than each consumer having to define its own dictionary.

domenic: In fact I'm pretty skeptical now of the web payments/notifications design of

```webidl
dictionary Thing : Localizable {
  DOMString title;
  // other members, hopefully not DOMStrings
}
```

The advantage is that authors have slightly less typing:

```js
const thing = {
  title: "foo",
  dir: "rtl"
};

// instead of

const thing = {
  title: { value: "foo", dir: "rtl" }
};
```

but the disadvantage is it means you can never add new string members to the dictionary without  running afoul of bad practice.

As such I think web payments should use my `LocalizableString` proposal instead of its dictionaries inheriting from `Localizable`. (In fact we can remove `Localizable` entirely and just have `LocalizedValue`.) We can then say Notifications' design is old-school.

marcoscaceres: > ? That seems better than each consumer having to define its own dictionary.

Yeah, I quite the typedef and the flexibility it gives. Do you want to add it? 

> As such I think web payments should use my LocalizableString proposal instead of its dictionaries inheriting from Localizable. (In fact we can remove Localizable entirely and just have LocalizedValue.) We can then say Notifications' design is old-school. 

Would be great!

marcoscaceres: @domenic, we can probably merge `LocalizedValue` into `Localizable`, or is there some reason to have it inherit from `Localizable` that I'm not seeing?

marcoscaceres: This is what Payment Request would look like if using `LocalizableString` (😍):
https://github.com/w3c/browser-payment-api/pull/455/files

Also, `LocalizableString` means we can get rid of the "should" restriction, because now all localizable members now carry around their own `lang`, `dir`, and `value` 🎉.

domenic: > Do you want to add it?

As in, do I think it's a good idea? Yes. Do I want to do the work myself? Well, I will if I have to, but not this week; I was hoping you could :)

> we can probably merge LocalizedValue into Localizable, or is there some reason to have it inherit from Localizable that I'm not seeing?

Nope, let's merge them.

marcoscaceres: > As in, do I think it's a good idea? Yes. Do I want to do the work myself? Well, I will if I have to, but not this week; I was hoping you could :)

Happy to!

marcoscaceres:
Ok, new version is up with LocalizableString added.

annevk:
I'm not sure I agree with your discussion. We explicitly decided against making each thing localizable on its own for Notifications. It's also not a feature HTML offers for UI-exposed attributes. It's quite a lot of complexity for edge cases nobody is really convinced about.

marcoscaceres:
@annevk, which part are you not convinced about? Direction obviously doesn't affect us westerners much. But as a screen reader user, it's super frustrating when (web and native) apps don't label the language correctly for text. This causes foreign words to get read back in the system default, which is usually English (or worst, the OS just guesses - which is funny at first, but gets old pretty quick). 

I can provide you with example recordings if you would like.

annevk:
I'm not convinced developers will go through the trouble of getting the language of each of these correctly. And even if you went to those length it still wouldn't be perfect as multiple paragraphs in a single member can each have their own language or multiple languages, etc. So as with all engineering you need to make a tradeoff and what HTML and Notifications set the precedent for hasn't resulted in complaints.

marcoscaceres:
> I'm not convinced developers will go through the trouble of getting the language of each of these correctly. And even if you went to those length it still wouldn't be perfect as multiple paragraphs in a single member can each have their own language or multiple languages, etc.

It's likely, but at least we would be giving developers the opportunity to try to do the right thing - even if not perfect. We've done a pretty good job in advocating for accessibility on the Web - we can do a better job at advocating that people add the appropriate members where it makes sense (specially if we can get them to experience how lousy things are when they don't!).   

> Notifications set the precedent for hasn't resulted in complaints.

The number of people who this affects are underrepresented, so it's unfair to say that "we have not heard complaints". And end-users generally wouldn't know where to complain (do we expect them to make Bugzilla accounts?). Additionally, developers don't use a screen reader, so they are probably woefully ignorant of the lived experience of users who rely on them. 

Consider how bad the situation is just trying to get Safari to read The Intercept on iOS - you will hear it switch through numerous languages:

[l10n_fail.m4a.zip](https://github.com/heycam/webidl/files/975279/l10n_fail.m4a.zip)

Users (including me) don't need more of that in their life. My motivation is to use `LocalizedStrings` in the Web Payments front-end of Firefox... but yes, it would depend again on developers doing the right thing. 

> HTML 

Does a good job with elements, because you can be quite explicit by setting `dir`, `lang`. For content attributes, like `title`, it's definitely not great. 

> and Notifications set the precedent for hasn't resulted in complaints. 

My interest in making sure we add this is a result of me complaining and trying to make sure we fix it where we can in the platform. Again, I'm trying to fix it because I'm personally affected (as a developer and a user).

annevk:
What's the use case for multiple localized members each in their own language/direction where you don't have individual words in those localized members that would also need to be marked up to get it right?

Again, this was discussed in depth in the past and has been discussed for HTML too (the attributes are equivalent with the case we're discussing here), including with members of the RTL community.

marcoscaceres:
Use cases are fairly limited, so I'm ok with dropping back to the simpler model: where dir/lang apply to "localizable members" identified in prose.

tobie:
I know extended attributes aren't designed for this purpose, but:

 ```webidl
dictionary PaymentItem : Localizable {
  [localizable] required DOMString label;
  required PaymentCurrencyAmount amount;
  boolean pending = false;
};
```

**Comment by marcoscaceres** [comment]
@tobie, something like that would be nice and would address @domenic's concerns about distinguishing between members that use DOMStrings but are not intended to be localized.

**Comment by tobie** [comment]
Would like to hear the thoughts of @bzbarsky, @annevk, and @domenic on the below non-standard use of extended attributes so we can move ahead with the PR:

 ```webidl
dictionary PaymentItem : Localizable {
  [localizable] required DOMString label;
  required PaymentCurrencyAmount amount;
  boolean pending = false;
};
````

**Comment by annevk** [comment]
I don't really understand @domenic's concern. You have to define in prose which members are the localizable members so there cannot be confusion among multiple DOMString members. Moving that coupling from prose to IDL seems fine (with an uppercase L in the extended attribute), but not an absolute necessity.

**Comment by domenic** [comment]
I'd rather not use extended attributes for this.

I guess we need to decide whether we want to support per-member localization like the i18n group suggests, or not, like @annevk suggests. That governs the shape of the solution.

**Comment by annevk** [comment]
Their position changed since https://lists.w3.org/Archives/Public/public-web-notification/2012Jul/0039.html? Pointer appreciated.

**Comment by tobie** [comment]
> I'd rather not use extended attributes for this.

Noted.

> I guess we need to decide whether we want to support per-member localization like the i18n group suggests, or not, like @annevk suggests. That governs the shape of the solution.

Can't we somehow have both? E.g. have dictionary members whose `LocalizableString` is a DOMString take the values of the `lang` and `dir`  member of the dictionary:

```webidl
dictionary FooBar : Localizable {
  LocalizableString foo;
  LocalizableString bar;
};
```
For example:

```js
{ lang: "fr-CH", foo: { value: "hello", lang: "en-US" }, bar: "bonjour" };
```

where `bar` takes its `lang` value from the top-level `lang` attribute and `foo` from `foo.lang`.

annevk:
We can, but it's not clear to me the complexity is justified.

marcoscaceres:
> Their position changed since https://lists.w3.org/Archives/Public/public-web-notification/2012Jul/0039.html? Pointer appreciated.

@annevk, I'm not sure if this represents the position of the i18n wg, but this was the motivating comment: https://github.com/w3c/browser-payment-api/issues/327#issuecomment-287125209

annevk:
Okay, we should probably first figure out implementer interest, especially if we turn this into something generic. Note that not providing it doesn't make the direction case impossible. You can still use Unicode code points for that.

marcoscaceres:
@annevk, I guess it depends on what you mean by implementer interest. If you mean that these members would be used in the UIs they are needed in, then, at least for Web Payments, I can confirm interest from Mozilla. The UI components we are building to enable payments are just HTML, so it's trivial to add these and they should "just work"(tm). I'm implementing the UI - so I will add them if this lands.  

In Web notifications (particularly as it pertains to Gecko hooking into OS-level notifications): I guess you investigated if these values can influence the rendering of notifications in various OSs? If yes, then that strengthens the case for standardization of these.  

If by implementation you mean doing something special at the Web IDL binding layer, I guess @bzbarsky is best positioned to answer that.

annevk:
@marcoscaceres what's the status around this now?

marcoscaceres:
Let's close.

PR #364, opened 2017-05-16
Add [LegacyWindowAlias] extended attribute for WebKitCSSMatrix et al

### LINKED ISSUE #362: How to do aliases of legacy interfaces (HTMLDocument, WebKitCSSMatrix...)
So we have this in HTML:

> For historical reasons, Window objects must also have a writable, configurable, non-enumerable property named HTMLDocument whose value is the Document interface object.

https://html.spec.whatwg.org/#htmldocument

The intent here is to expose it in the same way as an actual interface is exposed per WebIDL rules.

This method of aliasing legacy interfaces is also used in the Geometry interfaces spec.

> For historical reasons, Window objects must also have writable, configurable, non-enumerable properties named SVGMatrix and WebKitCSSMatrix whose value is the DOMMatrix interface object.

https://drafts.fxtf.org/geometry/#webkitcssmatrix

In implementing this it appears tempting to try to do so via an IDL attribute on `Window`, but that does not have correct behavior (it will have a getter instead of value, it's enumerable, etc). c.f. https://codereview.chromium.org/2709763004/#msg53

Should we add dedicated IDL syntax for this?

zcorpan:
Fixes #362.

***
[Preview](https://s3.amazonaws.com/pr-preview/zcorpan/webidl/zcorpan/legacywindowalias.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/4dbd7d2...zcorpan:d346eb0.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -940,6 +940,7 @@ limitations.  The following extended attributes must not
 be specified on partial interface definitions:
 [{{Constructor}}],
 [{{LegacyArrayClass}}],
+[{{LegacyWindowAlias}}],
 [{{NamedConstructor}}],
 [{{NoInterfaceObject}}].
 
@@ -965,6 +966,7 @@ The following extended attributes are applicable to interfaces:
 [{{Exposed}}],
 [{{Global}}],
 [{{LegacyArrayClass}}],
+[{{LegacyWindowAlias}}],
 [{{NamedConstructor}}],
 [{{NoInterfaceObject}}],
 [{{OverrideBuiltins}}],
@@ -8865,6 +8867,34 @@ is to be implemented.
         Example.prototype.y;
     </pre>
 </div>
+
+
+<h4 id="LegacyWindowAlias" extended-attribute lt="LegacyWindowAlias">[LegacyWindowAlias]</h4>
+
+The [{{LegacyWindowAlias}}]
+[=extended attribute=]
+must either
+[=takes an identifier|take an identifier=] or
+[=takes an identifier list|take an identifier list=].
+
+If the [{{LegacyWindowAlias}}] extended attribute was specified on an interface,
+then for each identifier |id| mentioned in the extended attribute,
+the [=primary global interface=] must
+have a property named |id|,
+which is the <dfn id="dfn-legacy-window-alias" export>legacy window alias</dfn>,
+with attributes
+{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
+whose value is a reference to the [=interface object=] for the interface.
+
+The [{{LegacyWindowAlias}}] and [{{NoInterfaceObject}}]
+extended attributes must not be specified on the same interface.
+
+<p class="advisement">
+    [{{LegacyWindowAlias}}] extended attributes are an undesirable feature.
+    They exist only so that legacy Web platform features can be specified.
+    They should not be used in specifications
+    unless required to specify the behavior of legacy APIs.
+</p>
 
 
 <h4 id="NamedConstructor" extended-attribute lt="NamedConstructor">[NamedConstructor]</h4>
--- a/index.bs
+++ b/index.bs
@@ -8750,7 +8750,7 @@ on a [=callback interface=].
 
 An interface must not have more than one [{{LegacyWindowAlias}}] extended attributes specified.
 
-See [[#legacy-window-alias]] for details on how legacy window aliases
+See [[#es-interfaces]] for details on how legacy window aliases
 are to be implemented.
 
 <p class="advisement">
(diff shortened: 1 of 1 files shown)


tobie:
First of all, thanks for doing the work!

A couple of high-level nits:

* Would be great to have an example.
* Extended attributes are sorted by alphabetical order, so this should go between `[LegacyUnenumerableNamedProperties]` and `[LenientSetter]`.
* The weird line break you'll see around the spec is the result of the XML -> Bikeshed conversion, not something to emulate. Because we can't agree on what to use, however, feel free to use semantic line-breaks or wrap at 100 chars.

Seems it would be useful to include this in [§3.6 Interfaces](https://heycam.github.io/webidl/#es-interfaces) too, no?

zcorpan:
Thanks! Will fix.

Some more todos:

* Ban [LegacyWindowAlias] for interfaces that do not include the primary interface object in its exposure set.
* Ban [LegacyWindowAlias] for callback interfaces.

zcorpan:
OK, please take another look.

index.bs
```diff
@@ -10202,6 +10269,21 @@ the <code>typeof</code> operator will return "function" when applied to an inter
 </div>
 
 
+<h4 id="legacy-window-alias">Legacy window alias</h4>
```
tobie: I'd move that whole section up in to the introductory section of §3.6 either as second or third ¶ and turn it into something like this:

```
If the [{{LegacyWindowAlias}}] extended attribute was specified on an interface,
then for each [=identifier=] mentioned in the extended attribute,
there must be a corresponding property on the [=primary global interface=],
known as the <dfn id="dfn-legacy-window-alias" export>legacy window alias</dfn>.
The name of the property is the [=legacy window alias=]'s [=identifier=],
and its value is a reference to the [=interface object=] for the [=interface=].
The property has attributes
{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
```

zcorpan: Done. Only changed "The property has attributes" to "The property has the attributes" for consistency with the surrounding paragraphs.

index.bs
```diff
@@ -10119,6 +10186,15 @@ and its value is an object called the <dfn id="dfn-interface-object" export>inte
 The property has the attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
 The characteristics of an interface object are described in [[#interface-object]].
 
+If the [{{LegacyWindowAlias}}] extended attribute was specified on an interface,
+then for each [=identifier=] mentioned in the extended attribute,
+there must be a corresponding property on the [=primary global interface=],
+known as the <dfn id="dfn-legacy-window-alias" export>legacy window alias</dfn>.
```
tobie: I know I wrote that, but this reads poorly: "known as the <dfn id="dfn-legacy-window-alias" export>legacy window alias</dfn>". It's not super clear whether we're naming the property or the object it references, which already has another name. Maybe not a useful concept to export?

index.bs
```diff
@@ -10119,6 +10186,15 @@ and its value is an object called the <dfn id="dfn-interface-object" export>inte
...
+The name of the property is the [=legacy window alias=]'s [=identifier=],
```
tobie: What we did for NamedConstructor is define a "NamedConstructor identifier" in the [NamedConstructor] section above. Maybe you'd want something similar here, especially if we get rid of defining the [=legacy window alias=]?

index.bs
```diff
@@ -10119,6 +10186,15 @@ and its value is an object called the <dfn id="dfn-interface-object" export>inte
 The property has the attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
 The characteristics of an interface object are described in [[#interface-object]].
 
+If the [{{LegacyWindowAlias}}] extended attribute was specified on an interface,
```
tobie: Right now this doesn't specify clearly whether the LegacyWindowAlias should be exposed if its related interface isn't (e.g. because of SecureContext). NamedConstructor is more explicit here: "for every [{{NamedConstructor}}] extended attribute on an [=exposed=] interface, […]". Maybe copy that text?

zcorpan: Can't copy it directly (since here it's a just single extended attribute), but certainly makes sense to add [=exposed=].

zcorpan:
OK addressed latest round of feedback.

tobie:
@domenic think you could kindly take a quick look at this? I tuned the language a bit and I'm not sure it's to the pull request's advantage. :(

Outside of that, we're good to go per #362, so I'll just merge it once you give it the thumbs up.

tobie:
Sounds like @zcorpan's fine with the changes. Merging.

PR #375, opened 2017-06-23
Add dedicated URL for filing bugs with WebKit

tobie:
/cc @cdumez: does the default here sound reasonable to file WebIDL-related issued?


domenic:
Probably @samweinig is a good person to check too

cdumez:
Yes, the URL is good. Thanks for checking.

PR #378, opened 2017-06-28
Re-align DOMException objects with what is implemented

### LINKED ISSUE #55: DOMException function definition makes no sense
@domenic, how does http://heycam.github.io/webidl/#es-DOMException-call actually produce a DOMException instance?  It's delegating to Construct() on its proto, which is %Error%, which certainly won't produce the right sort of object if the intent is to produce a "platform object".  They need the right branding, etc.

domenic:
This fixes #55. As discussed there, the definition of DOMException does not match any implementations. https://github.com/heycam/webidl/commit/137856d678f67b6f79f0c3ef4bd7d233629a685f exacerbated this, but it was already incorrect (e.g. defining own properties). Instead, implementations seem to implement DOMException as an interface with a small amount of custom bindings. This updates the specification to reflect that.

As part of this, it is no longer a distinct type, but instead an interface type.

---

Tests: https://github.com/w3c/web-platform-tests/issues/6361

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/fix-domexception.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/4269cc0...4dd161f.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12653,8 +12653,9 @@ fragment:
 ]
 interface DOMException { // but see below note about ECMAScript binding
   readonly attribute DOMString name;
-  stringifier readonly attribute DOMString message;
+  readonly attribute DOMString message;
   readonly attribute unsigned short code;
+  stringifier;
 
   const unsigned short INDEX_SIZE_ERR = 1;
   const unsigned short DOMSTRING_SIZE_ERR = 2;
@@ -12707,6 +12708,10 @@ The <dfn attribute for="DOMException"><code>message</code></dfn> attribute's get
 The <dfn attribute for="DOMException"><code>code</code></dfn> attribute's getter must return the
 legacy code indicated in the [=error names table=] for this {{DOMException}} object's
 [=DOMException/name=], or 0 if no such entry exists in the table.
+
+The <dfn dfn for="DOMException">stringification behavior</dfn> is to return the concatenation of
+this {{DOMException}} object's [=DOMException/name=], ": ", and this {{DOMException}} object's
+[=DOMException/message=].
 
 
 <h3 id="DOMTimeStamp" typedef oldids="common-DOMTimeStamp">DOMTimeStamp</h3>
--- a/index.bs
+++ b/index.bs
@@ -5788,19 +5788,6 @@ The [=type name=] of the
 {{Error!!interface}} type is “Error”.
 
 
-<h4 oldids="dfn-DOMException" id="idl-DOMException" interface>DOMException</h4>
-
-The {{DOMException}} type corresponds to the
-set of all possible non-null references to objects
-representing [=exceptions=] that are not [=simple exceptions=].
-
-There is no way to represent a constant {{DOMException}}
-value in IDL.
-
-The [=type name=] of the
-{{DOMException}} type is “DOMException”.
-
-
 <h4 id="idl-buffer-source-types">Buffer source types</h4>
 
 There are a number of types that correspond to sets of all possible non-null
@@ -6260,8 +6247,6 @@ and comprise the following:
 *   <a href="#es-stringifier">the <emu-val>Function</emu-val> objects that correspond to stringifiers</a>
 *   <a href="#es-iterators">the <emu-val>Function</emu-val> objects that correspond to iterators</a>
 *   [=map size getters=]
-*   [=DOMException constructor objects=]
-*   [=DOMException prototype objects=]
 
 Each [=Realms|ECMAScript global environment=]
 must have its own unique set of each of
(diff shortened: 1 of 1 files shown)


domenic:
There are a few easy fixes we could do that IMO would make DOMException better without much burden:

- Make `DOMException.__proto__ === Error`
- Make the stringification something more like Safari's by including "(DOMException)"

But I'm aligning with the majority for now, after having gotten bitten already in trying to make DOMException nicer in the past.

bzbarsky:
Some comments based on comparing to Gecko's IDL:

1. Gecko has `[LenientThis]` on the `message` and `name` attributes, with a comment about how otherwise doing `DOMException.prototype.toString()` will throw, because `Error.prototype.toString` accesses `.message` and `.name` on `this`.  Then again, looks like other browsers throw on `DOMException.prototype.toString()` (or more likely code like `alert(obj)` when `obj` happens to be `DOMException.prototype`) anyway...  

2. In Gecko DOMException does not implement a stringifier; it picks up the default Error.prototype.toString.  So if you do `String(new DOMException("a", "b"))` you get `"b: a"`.  Looks like you just changed the stringifier to do that, but I guess I'm saying we don't really need a custom stringifier at all, unless we want to do the Safari thing.

3. Minor nit: missing space after `=` for NO_MODIFICATION_ALLOWED_ERR.

4. Gecko documents some of the error constants as "historical": DOMSTRING_SIZE_ERR, NO_DATA_ALLOWED_ERR, INUSE_ATTRIBUTE_ERR, VALIDATION_ERR, TYPE_MISMATCH_ERR.  Not sure what that means in practice.

Thank you for doing this!

domenic:
@tabatkins can you help with the following Bikeshed error?

```
[1;33mLINK ERROR:[0m Multiple possible 'dfn' local refs for 'stringification behavior'.
Arbitrarily chose the one with type 'dfn' and for ''.
<a data-link-type="dfn" data-lt="stringification behavior">stringification behavior</a>
```

I think I'm doing everything right:

```html
The <dfn dfn for="DOMException">stringification behavior</dfn> is to return the concatenation of
this {{DOMException}} object's [=DOMException/name=], ": ", and this {{DOMException}} object's
[=DOMException/message=].
```

domenic:
> Gecko has `[LenientThis]` on the `message` and `name` attributes

I'll be sure to include a test case, but it looks like we can leave that out.

> In Gecko DOMException does not implement a stringifier; it picks up the default Error.prototype.toString.

Good idea, let's do that instead. In that case the stringifier _will_ be influenced by shadowing the properties; I'll be sure to test that too.

In that case, @tabatkins, nevermind, no need for Bikeshed help :)

> Gecko documents some of the error constants as "historical"

This is probably related to the grayed-out rows in https://heycam.github.io/webidl/#dfn-error-names-table.

index.bs
```diff
@@ -12471,112 +12456,27 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+In the ECMAScript binding, the {{DOMException}} type has some additional requirements:
 
-For every legacy code listed in the <a href="#dfn-error-names-table">error names table</a>,
-there must be a property on the DOMException prototype object
-whose name and value are as indicated in the table.  The property has
-attributes { \[[Writable]]: <emu-val>false</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>false</emu-val> }.
+*   Unlike normal [=interface types=], the [=interface prototype object=] for {{DOMException}} must
+    have as its \[[Prototype]] the intrinsic object [=%ErrorPrototype%=].
 
+*   If an implementation gives native <emu-val>Error</emu-val> objects special powers or
```
rakuco: FWIW, I've sent an [email to v8-dev](https://groups.google.com/forum/#!topic/v8-dev/nQVOwXFqVnE) to discuss what we'd need to do in Chromium and there are some questions over how this would interact with https://github.com/tc39/proposal-error-stacks.

tabatkins:
> can you help with the following Bikeshed error?

I'm not getting that error when I run Bikeshed on the file in this commit. Can you confirm that this is indeed the file causing the problem?

domenic:
The Bikeshed error went away when I removed the stringifier in a subsequent commit, sorry :). All good now.

domenic:
@bzbarsky would you mind reviewing so we can merge?

domenic:
Bugs:

- https://bugs.webkit.org/show_bug.cgi?id=174216
- https://bugzilla.mozilla.org/show_bug.cgi?id=1378985
- https://twitter.com/domenic/status/883069403109236736 (#EdgeBug... we'll see how that goes)
- https://bugs.chromium.org/p/chromium/issues/detail?id=737497

PR #385, opened 2017-07-19
Remove DOMException terminal from the grammar

domenic:
This is a fixup of 5c8eb310ef7a6bcb3c4813e9791212ee1624f2d3, which forgot to remove the DOMException terminal from the grammar when giving it a new definition.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/domexception-fixup.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f3b6a2a...25c1044.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4911,7 +4911,6 @@ type.
         "sequence" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         "object" Null
         "Error" Null
-        "DOMException" Null
         BufferRelatedType Null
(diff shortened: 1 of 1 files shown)


tobie:
Do we need to file issues downstream for this?

domenic:
Good question. Not on browsers, I don't think, but maybe on the tools? I will check if they implement this part of the grammar and report back with filed issues if so.

domenic:
- https://github.com/plinss/widlparser/issues/26
- webidl2.js does not appear to have any special handling for DOMException

Issue #366, opened 2017-05-31
[[Set]] algorithm for legacy platform objects does not seem to be compatible with the Storage interface


samweinig:
By my reading of the [[Set]] algorithm for legacy platform objects (https://heycam.github.io/webidl/#legacy-platform-object-set), it seems to state that an interface that doesn't support indexed properties, but does support named properties, will not allow setting indexed properties (by way of conversion to DOMString). (See step 1.2).

At the same time, the Storage interface (https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface) only supports named properties, so as per my understanding, something like:

`window.localStorate[7] = "foo";`

should not be allowed. But, browsers (I tested Safari and Firefox) do allow this.

domenic:
@bzbarsky, can you confirm we are not missing something here? Otherwise I think the correct fix is to just remove "P is not an array index property name," from step 1.2 of that algorithm.

domenic:
OK, we were missing something. It will fall through to the default [[Set]], which will call [[DefineOwnProperty]], which will handle the named property stuff.

This raises the question of why [[Set]] has shortcut steps defined for named/indexed properties, instead of always falling back to [[DefineOwnProperty]]. There are undoubtedly edge cases where this matters, and I even have vague memories of discussing them in the past. We should figure out what they are, and document them in the spec for future people.

bzbarsky:
Yes, agreed.  I'm not quite sure what those edge cases are either, at the moment, nor whether they still exist now that setters are always creators as well.  It's possible that the current setup was created back when we could have setters that were not creators...

samweinig:
Ok, I see how my original concern is not accurate, as it will fall back to [[DefineOwnProperty]], but now it seems odd that the special cases in step 1 don't match the behavior of [[DefineOwnProperty]]. Specifically, they don't take into account unforgeable properties or [OverrideBuiltins].

bzbarsky:
That does seem like a problem, good catch.  This is relevant only for named setters, as far as I can tell, so in practice only for Storage and DOMStringMap.  Neither one has unforgeable attributes, so we're ok there, a bit by accident.

[...]

Issue #175, opened 2016-09-09
Various named property stuff seems confused in the face of symbols


bzbarsky:
Consider this simple testcase:

```
var x = document.documentElement.dataset;
var s = Symbol("foo");
Object.defineProperty(x, s, { value: "hey" });
```

What should happen?  We land in http://heycam.github.io/webidl/#defineownproperty and the interface supports named properties.  `s` is not a supported property name, so _creating_ ends up true.  There is a named property setter, so we're supposed to invoke the named property setter... but _P_ is a symbol and named property setters expect strings.

Seems to me like all the named property stuff should be skipped for symbol names or something.

domenic:
> Seems to me like all the named property stuff should be skipped for symbol names or something.

Agreed, FWIW. Just never reach into that machinery when a symbol comes along.

TimothyGu:
Am I correct in understanding that only a simple fastpath to OrdinaryDefineOwnProperty for symbols w/o any additional processing is needed?

domenic:
Yes, I believe that's the right thing to do. I wouldn't quite call it a fastpath, since the current state probably leads to spec bugs where the named property setter algorithm in various specs does bad things when given a symbol. (E.g., perhaps adding an attribute to the DOM element whose name is given by a symbol.)

In practice this probably means adding "P is not a symbol" to the first two bullets. We do also want to flip the configurability for symbols, so step 3 still needs to happen.

bzbarsky:
@tobie, @TimothyGu are there web platform tests that got added for this?  Issues filed on browsers?

tobie:
@bzbarsky filed https://github.com/w3c/web-platform-tests/issues/6912 not to forget about it.

Do we want to file issues against browsers that fail the tests once those are written, or should we assume failing tests in wpt will be handled somehow?

bzbarsky:
> Do we want to file issues against browsers that fail the tests once those are written

That would be ideal, yes!  There are literally tens of thousands of failing WPTs in browsers (I just checked in Firefox and it's over 50k) and many newly added tests fail due to browser or test bugs, so depending on people noticing a new failing test is pretty unreliable, unfortunately.  :(  I wish we were closer to that being something that sets off alarms...

bzbarsky:
@tobie Do you know what the status is of those tests and browser bugs?  It's not great having this situation where the spec says X but we have no tests and browsers have no idea the spec says X...

tobie:
I agree. I need to look into it.

[...]

PR #402, opened 2017-08-07
Define [[OwnPropertyKeys]] of legacy platform objects

### LINKED ISSUE #400: Define a [[OwnPropertyKeys]] for legacy platform objects
Consider the following JS snippet:

```js
(() => {
  const element = document.createElement("div");
  element.setAttribute("length", "len_val");
  element.setAttribute("unrestricted", "unr_val");
  element.attributes.ownProp = 'own prop';
  console.log(Reflect.ownKeys(element.attributes));
  console.log(Object.keys(element.attributes));
  console.log(element.attributes.hasOwnProperty("unrestricted"));
  console.log(element.attributes.unrestricted);
})();
```

On Chrome this emits:

```js
["0", "1", "ownProp"]
["0", "1", "ownProp"]
true
unrestricted=​"unr_val"
```

while on Firefox it shows:

```js
[ "0", "1", "unrestricted", "ownProp" ]
[ "0", "1", "ownProp" ]
true 
unrestricted="unr_val"
```

Though it is not defined as an [invariant](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods) in ECMAScript, it would still seem a good idea to bring Firefox's behavior into the standard, as one would expect an own property according to `hasOwnProperty` to be in the `ownKeys` of that object.

Web IDL does already define [property enumeration](https://heycam.github.io/webidl/#legacy-platform-object-property-enumeration) orders for enumerable properties, but this specific issue concerns *all* keys, not only those that are enumerable. And the way to specify that is by imperatively defining an [[OwnPropertyKeys]] internal method, like ES's built-in [OrdinaryOwnPropertyKeys](https://tc39.github.io/ecma262/#sec-ordinaryownpropertykeys) abstract op.

FWIW, both Chrome and Firefox's `Object.keys` return the correct properties according to the spec. `element.attributes`'s type `NamedNodeMap` is defined with [`LegacyUnenumerableNamedProperties`], so none of those named properties are enumerable.

TimothyGu:
Fixes #400.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/legacy-keys.html#legacy-platform-object-ownpropertykeys) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/d6c3ddb...TimothyGu:ef6832d.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8516,7 +8516,7 @@ If the [{{LegacyUnenumerableNamedProperties}}]
 extended attribute is specified on an interface, then it applies to all its derived interfaces and
 must not be specified on any of them.
 
-See [[#legacy-platform-object-property-enumeration]]
+See [[#legacy-platform-object-getownproperty]]
 for the specific requirements that the use of
 [{{LegacyUnenumerableNamedProperties}}]
 entails.
@@ -12005,7 +12005,7 @@ and [[#legacy-platform-object-set]].
 </div>
 
 
-<h4 id="legacy-platform-object-property-enumeration" oldids="property-enumeration">Property enumeration</h4>
+<h4 id="legacy-platform-object-ownpropertykeys" oldids="property-enumeration,legacy-platform-object-property-enumeration">\[[OwnPropertyKeys]]</h4>
 
 This document does not define a complete property enumeration order
 for [=platform objects=] implementing [=interfaces=]
@@ -12014,19 +12014,26 @@ However, for [=legacy platform objects=],
 properties on the object must be
 enumerated in the following order:
 
-1.  If the object [=support indexed properties|supports indexed properties=], then
-    the object’s [=supported property indices=] are
-    enumerated first, in numerical order.
-1.  If the object [=support named properties|supports named properties=] and doesn't implement an [=interface=] with the
-    [{{LegacyUnenumerableNamedProperties}}]
-    [=extended attribute=], then
-    the object’s [=supported property names=] that
-    are visible according to the [=named property visibility algorithm=]
-    are enumerated next, in the order given in the definition of the set of supported property names.
-1.  Finally, any enumerable own properties or properties from the object’s prototype chain are then enumerated,
-    in no defined order.
+<div algorithm="to invoke the internal [[OwnPropertyKeys]] method of legacy platform objects">
 
-Note: Future versions of the ECMAScript specification may define a total order for property enumeration.
+    When the \[[OwnPropertyKeys]] internal method of a [=legacy platform object=] |O| is called,
+    the following steps are taken:
+
+    1.  Let |keys| be a new empty [=list=] of ECMAScript String and Symbol values.
+    1.  If |O| [=support indexed properties|supports indexed properties=], then [=list|for each=] |index| of |O|’s
+        [=supported property indices=], in ascending numerical order, [=list|append=] [=!=] [=ToString=](|index|) to
+        |keys|.
+    1.  If |O| [=support named properties|supports named properties=], then [=list|for each=] |P| of |O|’s
+        [=supported property names=] that is visible according to the [=named property visibility algorithm=],
+        [=list|append=] |P| to |keys|.
+    1.  [=list|For each=] |P| of |O|’s own property keys that is a String, in ascending chronological order of
+        property creation, [=list|append=] |P| to |keys|.
+    1.  [=list|For each=] |P| of |O|’s own property keys that is a Symbol, in ascending chronological order of
+        property creation, [=list|append=] |P| to |keys|.
+    1.  Assert: |keys| has no duplicate items.
+    1.  Return |keys|.
+
+</div>
 
 <h4 id="legacy-platform-object-abstract-ops">Abstract operations</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -12010,9 +12010,8 @@ and [[#legacy-platform-object-set]].
 This document does not define a complete property enumeration order
 for [=platform objects=] implementing [=interfaces=]
 (or for <a href="#es-exception-objects">platform objects representing exceptions</a>).
-However, for [=legacy platform objects=],
-properties on the object must be
-enumerated in the following order:
+However, it does for [=legacy platform objects=] by defining the \[[OwnPropertyKeys]]
+internal method as follows.
 
 <div algorithm="to invoke the internal [[OwnPropertyKeys]] method of legacy platform objects">
 


index.bs
```diff
@@ -12014,19 +12014,26 @@ However, for [=legacy platform objects=],
...
+        [=supported property indices=], in ascending numerical order, [=list|append=] [=!=] [=ToString=](|index|) to
+        |keys|.
+    1.  If |O| [=support named properties|supports named properties=], then [=list|for each=] |P| of |O|’s
+        [=supported property names=] that is visible according to the [=named property visibility algorithm=],
+        [=list|append=] |P| to |keys|.
+    1.  [=list|For each=] |P| of |O|’s own property keys that is a String, in ascending chronological order of
+        property creation, [=list|append=] |P| to |keys|.
+    1.  [=list|For each=] |P| of |O|’s own property keys that is a Symbol, in ascending chronological order of
+        property creation, [=list|append=] |P| to |keys|.
+    1.  Assert: |keys| has no duplicate items.
```
tobie: Why not used an ordered list instead (as you initially) suggested?

index.bs
```diff
@@ -12014,19 +12014,26 @@ However, for [=legacy platform objects=],
 properties on the object must be
 enumerated in the following order:
 
-1.  If the object [=support indexed properties|supports indexed properties=], then
-    the object’s [=supported property indices=] are
-    enumerated first, in numerical order.
-1.  If the object [=support named properties|supports named properties=] and doesn't implement an [=interface=] with the
-    [{{LegacyUnenumerableNamedProperties}}]
```
tobie: The effects of the [{{LegacyUnenumerableNamedProperties}}] extended attribute seem totally gone from the updated algorithm. Is that on purpose?

index.bs
```diff
@@ -12014,19 +12014,26 @@ However, for [=legacy platform objects=],
...
+    When the \[[OwnPropertyKeys]] internal method of a [=legacy platform object=] |O| is called,
+    the following steps are taken:
+
+    1.  Let |keys| be a new empty [=list=] of ECMAScript String and Symbol values.
+    1.  If |O| [=support indexed properties|supports indexed properties=], then [=list|for each=] |index| of |O|’s
+        [=supported property indices=], in ascending numerical order, [=list|append=] [=!=] [=ToString=](|index|) to
+        |keys|.
+    1.  If |O| [=support named properties|supports named properties=], then [=list|for each=] |P| of |O|’s
+        [=supported property names=] that is visible according to the [=named property visibility algorithm=],
+        [=list|append=] |P| to |keys|.
```
tobie: Are these also in ascending chronological order?

index.bs
```diff
@@ -12014,19 +12014,26 @@ However, for [=legacy platform objects=],
 properties on the object must be
```
tobie: Would be great to add a small blurb here that this is done by the [[OwnPropertyKeys]] method, e.g: 

```
This document does not define a complete property enumeration order
for [=platform objects=] implementing [=interfaces=]
(or for <a href="#es-exception-objects">platform objects representing exceptions</a>).
However, it does for [=legacy platform objects=] by defining the \[[OwnPropertyKeys]] method as follows:
```

index.bs
TimothyGu: Yes, `LegacyUnenumerableNamedProperties` controls the `enumerable` bit in the descriptor but not the existence of the key. See [[GetOwnProperty]]. The original pseudo-algorithm defines the "enumeration order", basically the order of for-in loops, which accounts for both enumerability and key existence.

index.bs
TimothyGu: The order is defined by the specification for the specific interface.

index.bs
TimothyGu: (Assuming you meant ordered set...) ES uses a plain List, and it would actually be a specification error if there are duplicates, since the named property visibility algorithm should filter out any supported property names that are also own properties on the object (or in its prototype chain).

index.bs
TimothyGu: I seem to have forgotten to update that blurb… will do.

index.bs
tobie: Oh, right. That's the order in which they're spec'ed. In one of the [toJSON-related algorithms](https://heycam.github.io/webidl/#collect-attribute-values) we write this as:

```
If a toJSON operation with a [Default] extended attribute is declared on |I|,
then for each exposed regular attribute |attr| that is an interface member of |I|,
in order:
```
Maybe just add `in order, ` before `[=list|append=] |P| to |keys|`?

index.bs
tobie: > (Assuming you meant ordered set...)

I did. :)

> ES uses a plain List, and it would actually be a specification error if there are duplicates, since the named property visibility algorithm should filter out any supported property names that are also own properties on the object (or in its prototype chain).

That makes sense.

index.bs
annevk: "for each" from Infra already defines the order so you don't have to add that kind of thing. (At least, iirc that's the idea.)

domenic:
First of all, this PR is awesome.

Second, I am concerned about the extent to which it matches implementations, especially with regard to ordering. In particular, it's possible implementations don't have a good order except in enumeration cases, if they don't map onto the recent ES spec updates which derive enumeration from own property keys.

So, we'll need some good tests of the various permutations here to be sure.

index.bs
tobie: @annevk, you're right. I'd just like to be a little more explicit in both those cases that enumeration order follows the order in which the attributes are specified in the interface.

tobie:
> So, we'll need some good tests of the various permutations here to be sure.

@domenic how do you want to go about this? Land this, then add the tests, then given the test results, file issues with rendering engines? Or do you want to wait until we have test results before we land this?

TimothyGu:
> I am concerned about the extent to which it matches implementations, especially with regard to ordering

While I did write this PR to match what implementations are doing, I would agree that there are a lot of corner cases, like index prop names that are not array index prop names ([2^32, 2^53)), assigning array index prop names to objects that don't support indexed properties (and vice versa), and symbols. To that end I'd definitely agree that there should be tests.

> if they don't map onto the recent ES spec updates which derive enumeration from own property keys

Can you point out the update you are referring to?

> So, we'll need some good tests of the various permutations here to be sure.

Is there a specific section in WPT for Web IDL compliance or should I look into adding these tests to the section of the tested interface (e.g. `HTMLCollection`)?

domenic:
> While I did write this PR to match what implementations are doing

Ah OK, that's great. I thought it might have been just a translation of the previous enumeration section, in which case there was more cause for concern. But if you have validated this roughly matches browsers, then I would be comfortable merging before in-depth tests. @tobie

> Can you point out the update you are referring to?

https://github.com/tc39/ecma262/pull/367

> Is there a specific section in WPT for Web IDL compliance or should I look into adding these tests to the section of the tested interface (e.g. HTMLCollection)?

https://github.com/w3c/web-platform-tests/tree/master/WebIDL, almost always under ecmascript-binding. As you can see from some tests there, we generally pick representative objects. https://github.com/w3c/web-platform-tests/blob/255d54144a82ce76d8e50a4aa8de284151119f8b/WebIDL/ecmascript-binding/legacy-platform-object.html is probably the file to work on, although I think nobody would mind having a separate file for the OwnPropertyKeys sub-cases as that file is so generically named it could grow without bound.

TimothyGu:
I've completed a more comprehensive audit of what browsers currently do. And the results are, well, inconsistently inconsistent. The test is live at https://timothygu.me/webidl-keys-test/, and the files available in https://github.com/TimothyGu/webidl-keys-test.

I specifically looked at three interfaces: [NamedNodeMap](https://dom.spec.whatwg.org/#interface-namednodemap) (used for `element.attributes`), [NodeList](https://dom.spec.whatwg.org/#nodelist) (through `node.childNodes`), and [Storage](https://html.spec.whatwg.org/multipage/webstorage.html#storage-2) (through `window.sessionStorage`). For browsers, I checked Chrome 60 (stable) and 61 (beta), Firefox 55 beta, and Edge 15. I couldn't get used to Safari's developer console, but at a glance its behavior is fairly similar to Chrome's.

----

From results from `NamedNodeMap` and `NodeList`:

Chrome and Edge both exhibit the original issue I opened #400 for, namely they do not include named properties on an interface annotated with `[LegacyUnenumerableNamedProperties]` in `Reflect.ownKeys()`, while Firefox does. This PR follows the Firefox behavior, as `[LegacyUnenumerableNamedProperties]` should only set the `enumerable` bit rather than hide it from getting all property names.

While Chrome and Firefox (and this proposal) consistently put own property symbols at the end for legacy platform objects in the same way as ES operates, Edge puts symbols in the front of all own properties.

Not directly related to this proposal, Chrome 60 and Edge follow the current [[Set]] semantics in that for interfaces not supporting indexed properties, it still sets any indexed properties as own properties. However, Chrome 61 and Firefox will throw when trying to do so.

Also not directly related to this proposal, Edge has a bug where its `Reflect.ownKeys()` returns duplicated property names when an object has both an own property and a supported indexed property [link](https://github.com/TimothyGu/webidl-keys-test/blob/480437a2a1152dbc5b705376cdcc5458a9632444/test-named-node-map.js#L82). In that situation, when getting the duplicated key it gets the own property instead. Chrome 60 does it correctly, but the enumeration order is a bit off. Chrome 61 and Firefox did not get to this test because it threw an error when trying to set the own property earlier.

----

As for `Storage`, [all hell broke loose](https://github.com/TimothyGu/webidl-keys-test/blob/480437a2a1152dbc5b705376cdcc5458a9632444/test-storage.js). No pairs of browsers do the same, and no browsers implement the spec for [[Set]]. It does reveal an [issue](https://github.com/TimothyGu/webidl-keys-test/blob/480437a2a1152dbc5b705376cdcc5458a9632444/test-storage.js#L56-L61) in [[Set]] that we can fix fairly easily, but other than that there is no agreement. It also reveals browsers have subtly different [[Set]] semantics re. [forbidden property names (prototype method shadowing)](https://github.com/TimothyGu/webidl-keys-test/blob/480437a2a1152dbc5b705376cdcc5458a9632444/test-storage.js#L14-L27) and [symbol properties](https://github.com/TimothyGu/webidl-keys-test/blob/480437a2a1152dbc5b705376cdcc5458a9632444/test-storage.js#L29-L38). As such, it's impossible to properly derive a pattern for enumeration from the `Storage` interface we can reference in [[OwnPropertyKeys]].

----

From the results, I see an even stronger need for standardization around [[OwnPropertyKeys]]. From the fact that both Chrome and Edge hide named properties from `Reflect.ownKeys()` for `[LegacyUnenumerableNamedProperties]` while Firefox and this proposal do not (i.e. the original #400), I'm open for some discussion regarding that. Other than this, I don't see anything that should be changed in this PR.

bzbarsky:
The comments in the `Storage` test don't quite make sense to me.  `[[Set]]` there should land in https://heycam.github.io/webidl/#legacy-platform-object-set which will call the named setter regardless of what the named property algorithm says (this is purposeful, fwiw).  It will not add a normal own property.  What it will do is add a thing named "getItem" to the storage.  But getting the list of own properties _will_ check the visibility algorithm and nor claim a "getItem" property.

Hence `typeof sessionStorage.getItem` should be `"function"`, `Reflect.ownKeys(sessionStorage)` should not include `"getItem"`, etc.

The bit about whether an index should be treated as a name for Storage seems to be https://github.com/heycam/webidl/issues/366 which has stalled for some reason.  :(

I agree compat around that stuff is poor.

TimothyGu:
@bzbarsky Thanks! I've updated the comments in the file, and it does make things clearer. To summarize, for `Storage`:

- Firefox and Edge's setters incorrectly ignore symbols
- Chrome's getter incorrectly gets the shadowed prototype property even when it should not be visible
- Edge's getter correctly handles shadowed properties, but still makes them visible through `Reflect.ownKeys()`.
- Chrome enumerates symbols in chronological order instead of last.
- Chrome and Firefox put integer index properties first, even when it is really a named property and should be enumerated chronologically. *Within* integer index properties, Chrome then sorts *them* numerically. Firefox sorts them chronologically.

Do you have any comments regarding this PR itself though?

index.bs
tobie: I totally mixed up a bunch of things here. Apologies for muddling the conversation. Please ignore this comment altogether.

tobie:
So we avoid doing duplicate work, heads-up that I've started refactoring legacy-platform-object.html and turning @TimothyGu's tests into wpt in a [dedicated branch](https://github.com/w3c/web-platform-tests/compare/webidl-legacy-platform-object).

bzbarsky:
> Firefox and Edge's setters incorrectly ignore symbols

The behavior for symbols is actually not defined right now; the spec's text for them makes no sense and isn't implementable.  In particular, [[Set]] in the spec ends up calling OrdinarySet for symbols, which eventually lands in the webidl-defined [[DefineOwnProperty]], which is broken in the face of symbols; see https://github.com/heycam/webidl/issues/175.  Once that issue is resolved, I'll update Firefox to whatever that resolution is.

> Chrome and Firefox put integer index properties first even when it is really a named property and should be enumerated chronologically

If we're still talking about Storage, then in the case of Firefox the order is rather random.  In particular, it's just iterating the keys of a non-ordered hashtable, so the precise order will depend on how the keys collided, where they landed in the hashtable, etc.  As in, we just don't implement the "chronological order" bit from the spec at all.

As far as the test goes, it might also be worth checking that while `sessionStorage.getItem` keeps returning the function, `sessionStorage.getItem("getItem")` returns `42`.

> Do you have any comments regarding this PR itself though?

Yes, I do.  It's totally awesome.  ;)   I was worried about duplication in the list, but the fix we did for named props not shadowing existing own props even on [OverridBuiltins] interfaces makes everything work out great.

Thank you for doing this!

TimothyGu:
> In particular, it's just iterating the keys of a non-ordered hashtable

Ah, thanks for clarifying then.

Glad I could help! ❤️

PR #405, opened 2017-08-13
Use <emu-val> only for constant values

### LINKED ISSUE #404: Decide whether to use emu-val for ES types
In ES spec, "Object" "Function" "Symbol" etc. are never wrapped in emu-val (which is reserved for special constants like **undefined**, **true** and **false**). In Web IDL though it's common to see them bolded through emu-val, but also unbolded in other places. Example of a section that uses both styles simultaneously: https://heycam.github.io/webidl/#es-object

We should decide which style to use and then apply it to the entire document.

TimothyGu:
- Change <code>&lt;emu-val>*Type*&lt;/emu-val></code> where *Type* is one of the [fundamental ES language types](https://tc39.github.io/ecma262/#sec-ecmascript-language-types) to <code>*Type*</code>.
- Change <code>&lt;emu-val><var>Class</var>&lt;/emu-val></code> where *Class* is `Set`, `Map`, `SyntaxError`, or `Error` to <code>{{ECMAScript/<var>Class</var>}}</code> with appropriate links to ES spec.
- Change <code>&lt;emu-val><var>Exception</var>&lt;/emu-val></code> where *Exception* is one of the IDL simple exceptions to <code>{{<var>Exception</var>}}</code> with appropriate section links.
- Change `<emu-val>Function</emu-val>` to [function object](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects), linked.

Fixes #404.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/emu-val.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/14792cb...TimothyGu:ca6a07a.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7028,9 +7028,8 @@ values.
 
 <h4 id="es-callback-function">Callback function types</h4>
 
-IDL [=callback function types=] are represented by ECMAScript function
-objects, except in the [{{TreatNonObjectAsNull}}] case,
-when they can be any object.
+IDL [=callback function types=] are represented by ECMAScript [=function objects=], except in the
+[{{TreatNonObjectAsNull}}] case, when they can be any object.
 
 <div id="es-to-callback-function" algorithm="convert an ECMAScript value to callback function">
 
@@ -10984,7 +10983,7 @@ The value of the [=@@iterator=] [=function object=]’s “name” property
 is the String value “entries”
 if the interface has a [=pair iterator=] or a [=maplike declaration=]
 and the String “values”
-if the interface has a [=setlike declaration=] or defines an [=indexed property getter=].
+if the interface has a [=setlike declaration=].
 
 
 <h5 id="es-forEach">forEach</h5>
@@ -11736,8 +11735,8 @@ each property that corresponds to one of the
 [=setlike declarations=]
 that exist on all of the interface prototype objects of |A|’s
 [=consequential interfaces=].
-For operations, where the property is a data property with a function
-object value, each copy of the property must have
+For operations, where the property is a data property with a [=function object=]
+value, each copy of the property must have
 distinct [=function objects=].  For attributes, each
 copy of the accessor property must have
 distinct [=function objects=] for their getters,
--- a/index.bs
+++ b/index.bs
@@ -44,7 +44,13 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
     type: interface; for: ECMAScript
         text: ArrayBuffer; url: sec-arraybuffer-objects
         text: DataView; url: sec-dataview-objects
+        text: Map; url: sec-map-objects
+        text: Promise; url: sec-promise-objects
+        text: Set; url: sec-set-objects
         text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
+    type: exception; for: ECMAScript
+        text: Error; url: sec-error-objects
+        text: SyntaxError; url: sec-native-error-types-used-in-this-standard-syntaxerror
     type: dfn
         text: NumericLiteral; url: sec-literals-numeric-literals
         text: ECMAScript error objects; url: sec-error-objects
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6668,7 +6674,7 @@ In effect, where <var ignore>x</var> is a <emu-val>Number</emu-val> value,
 
     1.  Let |x| be [=?=] [=ToNumber=](|V|).
     1.  If |x| is <emu-val>NaN</emu-val>, +∞, or −∞,
-        then [=ECMAScript/throw=] a <emu-val>TypeError</emu-val>.
+        then [=ECMAScript/throw=] a {{TypeError}}.
```
TimothyGu: Yes, this was intentional. TypeError [maps](https://heycam.github.io/webidl/#es-exception-objects) to ES anyway, while SyntaxError is a special case as it's an example of an exception type that doesn't map to IDL.

tobie: Well, yes it maps, but what you're effectively throwing here is the ES TypeError, not the WebIDL one. So the conversion is implied, not stated. I think we're loosing clarity by doing this (that is, consistently returning ES values from IDL to ES conversion lowers the cognitive load, imho).

TimothyGu: Speaking of that, I'm not too clear on the difference between [ES/throw](https://heycam.github.io/webidl/#ecmascript-throw) and [ES binding throw](https://heycam.github.io/webidl/#es-creating-throwing-exceptions) -- they seem to be equivalent IMO.

index.bs
```diff
@@ -139,7 +145,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: sections 9.1; url: sec-ordinary-object-internal-methods-and-internal-slots
         text: 9.3.1; url: sec-built-in-function-objects-call-thisargument-argumentslist
         text: ECMA-262 section 9.3; url: sec-built-in-function-objects
-        text: function object; url: sec-built-in-function-objects
```
TimothyGu: The only place that links "function object" is [here](https://heycam.github.io/webidl/#ecmascript-binding), which already has another reference to that section:

> If an object is defined to be a function object, then it has characteristics as described in [ECMA-262 section 9.3].

index.bs
TimothyGu: Okay makes sense, I'll change that.

tobie: > Speaking of that, I'm not too clear on the difference between ES/throw and ES binding throw -- they seem to be equivalent IMO.

The first one is just a proxy to the one in the ES spec. The second one deals with DOMExceptions too.

TimothyGu:
@tobie Besides the first one, other comments are all addressed.

tobie:
@domenic, @bzbarsky: thoughts on linking to function objects vs built-in function objects here? Are pretty much all function objects here built-ins (outside of legacy callback interfaces)? If so, should we either clarify the terminology or say something to this effect?

domenic:
It appears every time we say "function object", saying and linking to "built-in function object" would be clearer. Except legacy callback interfaces, as you said, since those are exotic objects.

(Note however we often say "**Function** object" which is used a bit differently...)

TimothyGu:
Okay. I've pushed a commit that distinguishes between "function objects"/"**Function** objects" and "built-in function objects"/"function objects" better, please take a look.

index.bs
```diff
@@ -10970,14 +10977,14 @@ then the <emu-val>Function</emu-val> object is [=%ArrayProto_values%=].
...
 
-The value of the [=@@iterator=] <emu-val>Function</emu-val> object’s “name” property
-is the <emu-val>String</emu-val> value “entries”
+The value of the [=@@iterator=] [=function object=]’s “name” property
+is the String value “entries”
 if the interface has a [=pair iterator=] or a [=maplike declaration=]
-and the <emu-val>String</emu-val> “values”
-if the interface has a [=setlike declaration=].
+and the String “values”
+if the interface has a [=setlike declaration=] or defines an [=indexed property getter=].
```
tobie: This seem like an orthogonal issue that would be better left in a separate PR, no?

index.bs
```diff
@@ -1144,8 +1145,8 @@ defined on the same interface.
 The identifier also must not
 be “length”, “name” or “prototype”.
 
-Note: These three names are the names of properties that exist on all
-[=function objects=].
+Note: These three names are the names of properties that may exist on
```
TimothyGu: Yeah, another remnant from ES5.

bzbarsky:
Yeah, all the functions IDL defines (getters, setters, operations on non-callback interfaces, interface objects, etc) should all be built-in function objects.

PR #412, opened 2017-08-16
Remove legacy callers

### LINKED ISSUE #407: [[Construct]] for legacy platform objects?
On a page with `#title`:

```js
new document.all('title')
```

In Chrome the snippet returns the same result as `document.all('title')`, while on Firefox the following error is thrown:

```
TypeError: document.all is not a constructor
```

[[Construct]] is technically undefined by this spec right now, so either behavior is strictly speaking acceptable. However, the Chrome behavior doesn't make much sense to me, so I'd like to propose a dedicated section for [[Construct]] saying that it does not exist for legacy platform objects.

### LINKED ISSUE #408: [[Call]] for objects w/o legacycaller
For legacy platform objects w/o any legacy callers, the [[Call]] internal method [immediately](https://heycam.github.io/webidl/#legacy-platform-object-call) throws a TypeError. Unfortunately, the mere fact that [[Call]] is defined can cause certain issues, like the object being treated as [callable](https://tc39.github.io/ecma262/#sec-iscallable) by ES, and [`typeof` operator](https://tc39.github.io/ecma262/#sec-typeof-operator) returning "function".

TimothyGu:
Turns out there is only one interface in the Platform using legacy callers (`HTMLAllCollection`). Define the exotic behaviors there instead. PR for that is pending.

Fixes #407.
Fixes #408.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/legacycaller.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f97da39...TimothyGu:5ee55c8.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -568,7 +568,6 @@ underscore.
         "inherit"
         "interface"
         "iterable"
-        "legacycaller"
         "maplike"
         "namespace"
         "partial"
@@ -1950,7 +1949,6 @@ The following extended attributes are applicable to operations:
         "getter"
         "setter"
         "deleter"
-        "legacycaller"
 </pre>
 
 <pre class="grammar" id="prod-OperationRest">
@@ -2142,11 +2140,6 @@ is used to declare it and what the purpose of the special operation is:
         <td>Defines behavior for when an object is indexed for property deletion.</td>
     </tr>
     <tr>
-        <td><dfn id="dfn-legacy-caller" export lt="legacy caller">Legacy callers</dfn></td>
-        <td><emu-t>legacycaller</emu-t></td>
-        <td>Defines behavior for when an object is called as if it were a function.</td>
-    </tr>
-    <tr>
         <td><dfn id="dfn-stringifier" export lt="stringifier">Stringifiers</dfn></td>
         <td><emu-t>stringifier</emu-t></td>
         <td>Defines how an object is converted into a {{DOMString}}.</td>
@@ -2231,8 +2224,6 @@ there must exist at most one
 stringifier, at most one
 [=named property deleter=],
 and at most one of each variety of getter and setter.
-Multiple legacy callers can exist on an interface
-to specify overloaded calling behavior.
 
 If an interface has a setter of a given variety,
 then it must also have a getter of that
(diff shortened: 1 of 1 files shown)


domenic:
@bzbarsky, are you OK with this? I believe last time we discussed document.all we found out Firefox is the only one using actual IDL to fully define its behavior, so for other browsers I think they will be fine with custom bindings in the HTML spec. But I want to check with you from Firefox's perspective.

I think this is a good change from a spec point of view, myself.

annevk:
(I think @domenic knows, but we probably need to ping again in two weeks unless bz manages to not miss a single email once he gets back.)

bzbarsky:
In practice this probably doesn't matter too much; Gecko would just keep supporting legacycaller and using IDL for this.  If no one else is using IDL here, then clearly there's no implementation change involved, so this is just spec bookkeeping....

samweinig:
> In practice this probably doesn't matter too much; Gecko would just keep supporting legacycaller and using IDL for this. If no one else is using IDL here, then clearly there's no implementation change involved, so this is just spec bookkeeping....

WebKit also implements and uses legacycaller for its HTMLAllCollection implementation.

domenic:
@samweinig are you OK with removing legacycaller from IDL and just having a one-off [[Call]] defined in HTML, like in https://github.com/whatwg/html/pull/2932 ? I think it'd be best to leave features used in only one, legacy place out of IDL.

samweinig:
@domenic yeah, I'm not sure it makes much difference. I'll either just keep our legacycaller stuff (or perhaps simplify it since it supports things like overloading) or change the syntax to an extended attribute.

I'm not 100% I see the value of removing it. I get that we don't want more specs using it, but it seems like having as much of this legacy stuff in the same place has some value.  I also have some concern that the more we specify exotic behavior (like a custom [[Call]]) the more it might encourage other specs to do that, kind of moving the problem around.

That said, it seems ok.

annevk:
@samweinig note that we already have legacy stuff in HTML for `WindowProxy` and `Location`. And we'll have to add a special slot to `HTMLAllCollection` that influences the type of object it is as far as the ECMAScript standard goes as well. So legacy is already HTML + IDL.

samweinig:
@annevk that's a good point, and ultimately why it seems fine.

domenic:
@TimothyGu mind rebasing?

TimothyGu:
@domenic Done.

TimothyGu:
Freshly rebased.

annevk:
I went ahead and merged this to avoid another merge conflict. Hope that's okay.

PR #417, opened 2017-08-18
Update and link ES exception handling

### LINKED ISSUE #414: "Handling exceptions" section in ES contains outdated information
https://heycam.github.io/webidl/#es-handling-exceptions says:

> None of the algorithms or processing requirements in the ECMAScript language binding catch ECMAScript exceptions.

But it is not so in [attribute getter](https://heycam.github.io/webidl/#dfn-attribute-getter)'s definition:

> ...
> And then, if an exception was thrown:
>
> 1. If op has a return type that is a promise type, then:
>     ...
> 2. Otherwise, end these steps and allow the exception to propagate.

TimothyGu:
Fixes #414.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/exception-handling-update.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/48f1e62...TimothyGu:c6752b0.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -106,7 +106,6 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: internal slot; url: sec-object-internal-methods-and-internal-slots
         text: JavaScript execution context stack; url: execution-context-stack
         text: running JavaScript execution context; url: running-execution-context
-        text: ReturnIfAbrupt; url: sec-returnifabrupt
         text: GetIterator; url: sec-getiterator
         text: IteratorStep; url: sec-iteratorstep
         text: NormalCompletion; url: sec-normalcompletion
@@ -7132,9 +7131,7 @@ ECMAScript Array values.
 
     1.  If [=Type=](|V|) is not Object,
         [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  Let |method| be the result of
-        [=GetMethod=](|V|, [=@@iterator=]).
-    1.  [=ReturnIfAbrupt=](|method|).
+    1.  Let |method| be [=?=] [=GetMethod=](|V|, [=@@iterator=]).
     1.  If |method| is <emu-val>undefined</emu-val>,
         [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
     1.  Return the result of [=creating a sequence from an iterable|creating a sequence=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -13110,12 +13123,14 @@ The following typographic conventions are used in this document:
 The following conventions are used in the algorithms in this document:
 
 *   Algorithms use the [=conventions of the ECMAScript specification=],
-    including the ! and ? notation for unwrapping completion records.
-*   Algorithms sometimes treat returning/throwing values and returning completion records
+    including the ! and ? notation for unwrapping [=Completion Records=].
```
tobie: Would have been nice to link directly to the Completion Record dfn, here, if you're so inclined to file a bug with the TC-39 spec.

index.bs
```diff
@@ -13110,12 +13123,14 @@ The following typographic conventions are used in this document:
...
+*   Algorithms sometimes treat returning/throwing values and returning [=Completion Records=]
     interchangeably. That is, an algorithm that uses return/throw terminology may be treated as
-    returning a completion record, while one that returns a completion record may be treated as
-    returning a value or throwing an exception.
-*   Completion records are extended by allowing them to contain values that are not ECMAScript
+    returning a [=Completion Record=], while one that returns a [=Completion Record=] may be treated as
+    returning a value or throwing an exception. Similarly, to catch exceptions, defining the
+    behavior to adopt [=if an exception was thrown=] and checking if the [=Completion Record=]’s
+    \[[Type]] field is “throw” are equivalent as well.
```
tobie: "as well" is a bit redundant, here.

index.bs
```diff
@@ -12626,14 +12629,17 @@ A {{DOMException}} is represented by a
...
-ends due to an exception being thrown, that exception
+Unless specified otherwise,
+whenever ECMAScript runtime semantics are invoked due
+to requirements in this document and
+end due to an exception being thrown, that exception
 must propagate to the caller, and if
 not caught there, to its caller, and so on.
 
+Per [[#conventions]], an algorithm specified in this document may intercept thrown exceptions, either by specifying
+the exact steps to take <dfn id="dfn-if-an-exception-was-thrown">if an exception was thrown</dfn>, or by explicitly
```
tobie: If feels weird to integrate the conditional ("if") within the DFN. I'd be more confortable with:

```
<dfn id="dfn-an-exception-was-thrown">an exception was thrown</dfn>
```

index.bs
tobie: Nice, but not a show stopper.

index.bs
domenic: I think the id is not necessary either; we include it a lot elsewhere to preserve pre-Bikeshed IDs, but we can use the autogenerated one when introducing new terms.

TimothyGu: Ah didn't know the id could be autogenerated. Will fix.

index.bs
TimothyGu: The link does go to Completion Record directly: https://tc39.github.io/ecma262/#sec-completion-record-specification-type

index.bs
```diff
@@ -75,10 +75,11 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: abs; url: eqn-abs
         text: modulo; url: eqn-modulo
         url: sec-algorithm-conventions
-            text: !
-            text: ?
             text: ECMA-262 section 5.2
             text: conventions of the ECMAScript specification
+        url: sec-returnifabrupt-shorthands
+            text: !
+            text: ?
```
tobie: Oh, being able to link directly to an explanation of what the shorthand does is amazing!

Issue #419, opened 2017-08-20
Class string of default iterator object seems to have been accidentally changed?


TimothyGu:
Commit d433e0a6ae4f81b35364e0890694a8b094a5f669 (#198) has a commit message of the following:

```git
commit d433e0a6ae4f81b35364e0890694a8b094a5f669
Author:     Tobie Langel <tobie@codespeaks.com>
AuthorDate: Wed Oct 26 09:13:35 2016 +0200
Commit:     GitHub <noreply@github.com>
CommitDate: Wed Oct 26 09:13:35 2016 +0200

    Use Bikeshed’s algorithm construct. (#198)
    
    * Marks up all algorithms with <div algorithm>.
    * Gives a name to algorithms which don't include a <dfn>.
    * Fix algorithms that generate Bikeshed warnings, notably by:
        * Replacing prose calls to [[Call]] method by Call(args).
        * Fixing incorrectly named variables.
        * Adding |T|, |U|, |A|, |map|, |object|, |I| and |id| to Ignored Vars.
```

However, it contains the following normative change (white space differences ignored):

```diff
 When a [=default iterator object=] is first created,
 its index is set to 0.
 
 The [=class string=] of a [=default iterator object=] for a given [=interface=]
 is the result of concatenting the [=identifier=] of the [=interface=]
-and the string “ Iterator”.
+and the string “Iterator”.
 
 
 <h5 id="es-iterator-prototype-object">Iterator prototype object</h5>
```

It seems to have been accidental, but there are already some implementations using the new changed spec (admittedly, most of which were written by me).

/cc @tobie

tobie:
That's bad! No idea how it happened. Needs to be fixed.

annevk:
Maybe best to replace with concatenating X, U+0020 SPACE, and Y rather than X and " Y" for clarity.

tobie:
(Mis)read your message before morning coffee and thought *I* had introduced the whitespace. Sorry.

This was a deliberate fix, as no other class string contained whitespace (including the class string of the [iterator prototype object](https://cdn.rawgit.com/heycam/webidl/dd77b1907e6af7ec27ff67edfb38f892f2fde3a8/index.html#dfn-iterator-prototype-object)), and I assumed it was a typo.

I think we should leave it as such unless there's a good reason not to.

tobie:
[...]

PR #421, opened 2017-08-22
Update SessionManager example

TimothyGu:
Would be happy to update the class string once https://github.com/heycam/webidl/issues/419 is resolved, but submitting this for review first.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/values-to-iterate-over.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/48f1e62...TimothyGu:369c747.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3414,14 +3414,13 @@ interfaces that have interface members with these names.
 <div class="example">
 
     Consider the following interface <code class="idl">SessionManager</code>, which allows access to
-    a number of <code class="idl">Session</code> objects:
+    a number of <code class="idl">Session</code> objects keyed by username:
 
     <pre highlight="webidl">
         interface SessionManager {
           Session getSessionForUser(DOMString username);
-          readonly attribute unsigned long sessionCount;
 
-          iterable&lt;Session&gt;;
+          iterable&lt;DOMString, Session&gt;;
         };
 
         interface Session {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3460,7 +3457,9 @@ interfaces that have interface members with these names.
 
         typeof SessionManager.prototype.values;            // Evaluates to "function"
         var it = sm.values();                              // values() returns an iterator object
-        String(it);                                        // Evaluates to "[object SessionManager Iterator]"
+        String(it);                                        // Evaluates to "[object SessionManagerIterator]"
```
tobie: I don't think we should change this now.

index.bs
```diff
@@ -3434,24 +3433,22 @@ interfaces that have interface members with these names.
...
     In the ECMAScript language binding, the [=interface prototype object=]
     for the <code class="idl">SessionManager</code> [=interface=]
     has a <code>values</code> method that is a function, which, when invoked,
     returns an iterator object that itself has a <code>next</code> method that returns the
-    next value to be iterated over.  It has <code>values</code> and <code>entries</code>
-    methods that iterate over the indexes of the list of session objects
-    and [index, session object] pairs, respectively.  It also has
+    next value to be iterated over.  It has <code>keys</code> and <code>entries</code>
+    methods that iterate over the usernames of session objects
+    and [username, Session object] pairs, respectively.  It also has
```
tobie: Can you use the tuple/pair notation from infra here (either `(username, Session object)` or `username/Session object`)?

index.bs
tobie: I don't think you should change this now. Maybe comment out the whole thing?

index.bs
```diff
@@ -3460,7 +3457,8 @@ interfaces that have interface members with these names.
 
         typeof SessionManager.prototype.values;            // Evaluates to "function"
         var it = sm.values();                              // values() returns an iterator object
-        String(it);                                        // Evaluates to "[object SessionManager Iterator]"
+<!--    String(it);                                        // Evaluates to "[object SessionManagerIterator]"
+                                                           // TODO: https://github.com/heycam/webidl/issues/419 -->
         typeof it.next;                                    // Evaluates to "function"
 
         // This loop will alert "anna" and then "brian".
```
tobie: Sorry I didn't notice that before, can you change this to "log" stuff rather than "alert" it?

So `s/This loop will also alert/This loop will also log/` and `s/window.alert/console.log/`.

PR #425, opened 2017-08-23
Add [LegacyNamespace] to allow defining an interface in a namespace

littledan:
This is necessary to specify certain APIs such as WebAssembly,
which includes several constructors as properties of the root
WebAssembly object. See a WebAssembly draft specification where
this is in use at https://github.com/littledan/spec/blob/new-js.bs/document/JS.bs

***
[Preview](https://s3.amazonaws.com/pr-preview/littledan/webidl/namespace-interface.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f869be8...littledan:7f1899b.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8448,50 +8448,6 @@ corresponding to [=interface members=].
     </pre>
 </div>
 
-
-<h4 id="LegacyNamespace" extended-attribute lt="LegacyNamespace">[LegacyNamespace]</h4>
-
-If the [{{LegacyNamespace}}] [=extended attribute=] appears on an [=interface=], it indicates that
-the [=interface object=] for this interface will not be created as a property of the global
-object, but rather as a property of the [=namespace=] identified by the argument to the extended
-attribute.
-
-The [{{LegacyNamespace}}] extended attribute [=takes an identifier|take an identifier=].
-This identifier must be the identifier of a namespace.
-
-The [{{LegacyNamespace}}] and [{{NoInterfaceObject}}]
-extended attributes must not be specified on the same interface.
-
-The [{{LegacyNamespace}}] extended attribute
-must not be used on a [=callback interface=].
-
-See [[#namespace-object]] for details on how an interface is exposed on a namespace.
-
-<p class="advisement">
-    LegacyNamespace is discouraged for new specifications. Instead, interface names
-    can be formed with a naming convention of starting with a particular prefix for
-    a set of interfaces, as part of the identifier, rather than using a namespace.
-</p>
-
-<div class="example">
-    The following [=IDL fragment=] defines a [=namespace=] and an [=interface=] which
-    uses [{{LegacyNamespace}}] to be defined inside of it.
-
-    <pre highlight="webidl">
-        namespace Foo { };
-
-        [LegacyNamespace=Foo, Constructor]
-        interface Bar { };
-    </pre>
-
-    In an ECMAScript implementation of the above namespace and interface, the
-    constructor Bar can be accessed as follows:
-
-    <pre highlight="js">
-        var instance = new Foo.Bar();
-    </pre>
-</div>
-
 <h4 id="LegacyArrayClass" extended-attribute lt="LegacyArrayClass">[LegacyArrayClass]</h4>
 
 <div class="advisement">
@@ -8585,6 +8541,58 @@ entails.
     succeed on objects implementing <code class="idl">ImmutableItemList</code>.
     The exact behavior depends on the definition of the [=Array methods=] themselves.
 
+</div>
+
+
+<h4 id="LegacyNamespace" extended-attribute lt="LegacyNamespace">[LegacyNamespace]</h4>
+
+<p class="advisement">
+    The [{{LegacyArrayClass}}] [=extended attribute=] is an undesirable feature.
+    It exists only so that legacy Web platform features can be specified.
+    It should not be used in specifications
+    unless required to specify the behavior of legacy APIs,
+    or for consistency with these APIs.
+    Instead, interface names
+    can be formed with a naming convention of starting with a particular prefix for
+    a set of interfaces, as part of the identifier, rather than using a namespace.
+    Editors who wish to use this feature are strongly advised to discuss this
+    by <a href="https://github.com/heycam/webidl/issues/new?title=Intent%20to%20use%20[LegacyNamespace]">filing an issue</a>
+    before proceeding.
+</p>
+
+If the [{{LegacyNamespace}}] [=extended attribute=] appears on an [=interface=], it indicates that
+the [=interface object=] for this interface will not be created as a property of the global
+object, but rather as a property of the [=namespace=] identified by the argument to the extended
+attribute.
+
+The [{{LegacyNamespace}}] extended attribute [=takes an identifier|take an identifier=].
+This identifier must be the identifier of a namespace.
+
+The [{{LegacyNamespace}}] and [{{NoInterfaceObject}}]
+extended attributes must not be specified on the same interface.
+
+The [{{LegacyNamespace}}] extended attribute
+must not be used on a [=callback interface=].
+
+See [[#namespace-object]] for details on how an interface is exposed on a namespace.
+
+<div class="example">
+    The following [=IDL fragment=] defines a [=namespace=] and an [=interface=] which
+    uses [{{LegacyNamespace}}] to be defined inside of it.
+
+    <pre highlight="webidl">
+        namespace Foo { };
+
+        [LegacyNamespace=Foo, Constructor]
+        interface Bar { };
+    </pre>
+
+    In an ECMAScript implementation of the above namespace and interface, the
+    constructor Bar can be accessed as follows:
+
+    <pre highlight="js">
+        var instance = new Foo.Bar();
+    </pre>
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -8449,6 +8449,48 @@ corresponding to [=interface members=].
 </div>
 
 
+<h4 id="InNamespace" extended-attribute lt="InNamespace">[InNamespace]</h4>
+
+If the [{{InNamespace}}] [=extended attribute=] appears on an [=interface=], it indicates that
+the [=interface object=] for this interface will not be created as a property of the global
+object, but rather the namespace which is the argument to the extended attribute.
+
+The [{{InNamespace}}] extended attribute [=takes an identifier|take an identifier=].
+This identifier must be the identifier of a namespace.
+
+The [{{InNamespace}}] and [{{NoInterfaceObject}}]
+extended attributes must not be specified on the same interface.
+
+The [{{InNamespace}}] extended attribute
+must not be used on a [=callback interface=].
+
+The [=class string=] of an interface with the [{{InNamespace}}] extended
+attribute is the concatenation of the class string of the namespace, <code>"."</code>,
+and the class string that the interface would otherwise have without this
+extended attribute.
+
+See [[#namespace-object]] for details on how an interface is exposed on a namespace.
+
+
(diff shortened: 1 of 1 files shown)


littledan:
This is just a first pass; I'd be interested in any feedback you might have. cc @domenic .

index.bs
```diff
@@ -8440,6 +8440,48 @@ corresponding to [=interface members=].
 </div>
 
 
+<h4 id="InNamespace" extended-attribute lt="InNamespace">[InNamespace]</h4>
+
+If the [{{InNamespace}}] [=extended attribute=] appears on an [=interface=], it indicates that
+the [=interface object=] for this interface will not be created as a property of the global
+object, but rather the namespace which is the argument to the extended attribute.
```
domenic: "but rather as a property of" is clearer, I think

"namespace identified by the argument", I think

`[=namespace=]`

index.bs
```diff
@@ -8440,6 +8440,48 @@ corresponding to [=interface members=].
...
+The [{{InNamespace}}] and [{{NoInterfaceObject}}]
+extended attributes must not be specified on the same interface.
+
+The [{{InNamespace}}] extended attribute
+must not be used on a [=callback interface=].
+
+The [=class string=] of an interface with the [{{InNamespace}}] extended
+attribute is the concatenation of the class string of the namespace, <code>"."</code>,
+and the class string that the interface would otherwise have without this
+extended attribute.
```
domenic: I think this should not be located here but instead be a modification to

> The class string of a platform object that implements one or more interfaces must be the identifier of the primary interface of the platform object.

Also, no `<code>` around strings in Web IDL, it appears.

index.bs
```diff
@@ -12516,6 +12558,10 @@ The characteristics of a namespace object are described in [[#namespace-object]]
         1.  Let |F| be the result of [=creating an operation function|creating an operation function=]
             given |op|, |namespace|,  and |realm|.
         1.  Perform [=!=] [=CreateDataProperty=](|namespaceObject|, |op|'s [=identifier=], |F|).
+    1.  For each [=exposed=] [=interface=] |intr| which has the [{{InNamespace}}] extended attribute, with the identifier of this namespace as its argument,
```
domenic: Wrap to 100 chars here too

`[=identifier=]`

Just _interface_ instead of _intr_ is nicer, I think.

No comma before "with", I think.

index.bs
```diff
@@ -12516,6 +12558,10 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+        1.  Let |I| be the [=interface object=] for |intr|.
+        1.  Let |newDesc| be the PropertyDescriptor{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val>, \[[Value]]: |I| }.
```
domenic: It looks like at this point we don't put spaces after { and before } in Web IDL.

tobie:
Thanks for the PR!

A comment and some bikeshedding.

My understanding was the wanted to avoid such constructs in favor of waiting for modules instead, but maybe things have changed?

If we do go this route though, shouldn't we go for dedicated syntax here instead of (ab)using extended attributes? E.g.: something like:

```webidl
namespace Foo { };

interface Foo.Bar {
};

interface Foo.Baz : Foo.Bar {
};
```
And if we do choose extended attributes, maybe `[Namespace=Foo]` reads better than `[InNamespace=Foo]`.

tobie:
> I'm on the fence as to whether we should add some sort of usage note, along the lines of "most places in the platform don't do this; maybe check first, and don't go crazy using it everywhere."

I know @rwaldron is going to jump on this.

TimothyGu:
I would definitely prefer @tobie's syntax.

domenic:
My thinking for advising @littledan to go with this instead of new syntax is that we already use extended attributes to control what gets exposed for a given interface (viz. [NamedConstructor], [NoInterfaceObject], and [Exposed]). Plus, it's always a bonus when we don't need to update every IDL toolchain to add a new feature.

TimothyGu:
The main issue for me is that this would disallow the following fragment since the interface name is the same (or at least I think, I cannot find the clause that disallows this but it seems reasonably logical):

```webidl
[Exposed=Window]
interface Foo {};

[Exposed=Window]
namespace bar {};
[InNamespace=bar]
interface Foo {};
```

... and namespacing was kind of the point of namespaces.

On the other hand, even if we do allow that, it can be difficult (for a human) to immediately recognize `Foo` was not indeed exposed globally but through a namespace.

I agree that a syntax change can cause churn, no doubt about that, but I still believe it's the best way forward.

annevk:
Whether more dedicated syntax makes sense very much depends on whether we want this kind of namespacing going forward or use some kind of builtin modules.

Does Wasm really need this?

tobie:
> My thinking for advising @littledan to go with this instead of new syntax is that we already use extended attributes to control what gets exposed for a given interface (viz. [NamedConstructor], [NoInterfaceObject], and [Exposed]).

If we could end-up in a situation where those extended attributes are mostly there for legacy purposes, I think that would be better. We're adding new syntax for mixins soon, so [NoInterfaceObject] will be a legacy construct.

> Plus, it's always a bonus when we don't need to update every IDL toolchain to add a new feature.

Well, for namespaces to be at all useful, we'd be breaking the invariant that identifiers are unique, so we're in for a world of pain anyway (e.g. idlharness assumes interface identifiers are unique).

tobie:
> Whether more dedicated syntax makes sense very much depends on whether we want this kind of namespacing going forward or use some kind of builtin modules.
>
> Does Wasm really need this?

I agree with @annevk's sentiment here. It seems weird to suddenly allow this for a part of the platform for implementor/editor convenience when we've been refusing it for the rest of the platform despite developers requiring it for a long time (e.g. for grouping all sensors under a single namespace, see https://github.com/w3c/sensors/issues/127).

littledan:
Historically, WebAssembly seems to have somehow skipped the WebIDL formalization and TAG review before shipping ([blink-dev Intent to Ship thread](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/6iqiskBdDQE)), so it recently got out of the door with these namespaces. It has now be shipping in Chrome for around six months, and has also shipped in Firefox. At this point, it could be a risk to web compatibility to change the API to be anything but namespaced.

The issue that @TimothyGu raises sort of comes up in practice already: Between WebAssembly and HTML, there are two interfaces called `Table`. Bikeshed seems to be able to disambiguate just fine [in the draft WebAssembly/JS spec](https://littledan.github.io/spec/document/JS.html), but I'm not sure if this could create some sort of problem in the future. @tobie 's syntax could get around this issue.

annevk:
HTML doesn't define an interface called `Table` as far as I know.

littledan:
Sorry, not sure why I got confused about that; I believe in some past version there were some funny cross-references generated, but I don't see them now.

tobie:
> HTML doesn't define an interface called Table as far as I know.

So if we agree this was a design mistake and that we keep it for Web compat reasons, we can restrict it to the existing set of interfaces defined by WebAssembly and keep the interface identifier uniqueness invariant as a requirement. In that case, we should call it [LegacyNamespace] and add an advisement to it such as we do for [[LegacyArrayClass]](https://heycam.github.io/webidl/#LegacyArrayClass).

littledan:
OK, LegacyNamespace seems to be the way to go for now, though I won't be surprised if, in the future, others come in wanting interfaces in namespaces ("WebAssembly has it, so why can't my API?").

index.bs
littledan: Done

index.bs
littledan: Done

index.bs
littledan: Done

index.bs
littledan: Done

tobie:
> though I won't be surprised if, in the future, others come in wanting interfaces in namespaces

I won't either. Which is why I suggested dedicated syntax. If this opens up to other use cases, the unique identifier invariant won't hold long, and we'll need to address this in the spec.

littledan:
Would it be reasonable to start with the LegacyNamespace form, and consider adding dedicated syntax later (and going back and updating the WebAssembly spec) if it turns out to be needed for other specifications?

tobie:
> Would it be reasonable to start with the LegacyNamespace form, and consider adding dedicated syntax later (and going back and updating the WebAssembly spec) if it turns out to be needed for other specifications?

I'm fine with that. Can you add an advisement in the same style as those I just added for all the [LegacX] stuff?

littledan:
@tobie I believe this PR should be in that same style.

tobie:
I just merged some more consistent text earlier for all of these advisement (always at the top, with a link to a file an "intent to use" GH issue before using it, etc. ), think you could follow that format?

littledan:
Done

index.bs
```diff
@@ -8545,6 +8544,58 @@ entails.
 </div>
 
 
+<h4 id="LegacyNamespace" extended-attribute lt="LegacyNamespace">[LegacyNamespace]</h4>
+
+<p class="advisement">
+    The [{{LegacyArrayClass}}] [=extended attribute=] is an undesirable feature.
```
tobie: s/LegacyArrayClass/LegacyNamespace/

index.bs
```diff
@@ -10135,7 +10186,7 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
 For every non-callback [=interface=] that is [=exposed=] in
 a given ECMAScript global environment and that is not declared with
-the [{{NoInterfaceObject}}] [=extended attribute=],
+the [{{NoInterfaceObject}}] or [{{LegacyNamespace}}] [=extended attribute|extended attributes=],
```
tobie: s/[=extended attribute|extended attributes=]/[=extended attributes=]/ (Bikeshed should recognize the plurals.

littledan:
Fixed those nits.

tobie:
Just discussed this with @bzbarsky. He pointed out that Mozilla's Wasm implementation wasn't relying on WebIDL at all so far, which creates significant differences with what a WebIDL-based implementation would look like. Is Chrome's implementation WebIDL-based? @lukewagner, what's Mozilla's plan here? Are implementations going to converge on a WebIDL-based spec?

lukewagner:
Yes, we currently hand-code our wasm JS API bindings in the JS engine.  However, we should be able to precisely match the order of coercions and property descriptor values to match a Web IDL spec without much trouble.  The plan is also to write plentiful wpts for all the observable differences.

tobie:
Thanks, @lukewagner. So if the plan is for all to move to WebIDL semantics, we should just go-ahead with this.

tobie:
That said, are we sure that the backwards compatibility issues that will be created by moving to a WebIDL-based specification aren't the occasion to eschew [LegacyNamespace] altogether and move to the more standard `NamespaceInterface` format seen elsewhere?

lukewagner:
From the changes I anticipate (order of coercions, enumerability of properties), I don't think breakage will actually occur in the wild at this point by simply moving to WebIDL (and there's probably already divergence between browsers in these minute details).

littledan:
V8's implementation is not WebIDL-based either, but I plan to work on those wpts. Thanks for the merge!

annevk:
As far as I know the differences are fairly minor: https://github.com/heycam/webidl/issues/226. But they are a source of annoyance. I still wonder whether we should offer some kind of opt-in mechanism to get JavaScript-style classes rather than IDL-style classes. Then we could also potentially describe other things with IDL, such as Promise and ReadableStream, if we wanted to.

lukewagner:
@annevk That sounds even better :)

PR #426, opened 2017-08-25
Add a section on overloads vs. union/optional

### LINKED ISSUE #307: Add an informative section on overloads vs. unions/optional
After I understand this better---see https://github.com/jsdom/webidl2js/issues/29#issuecomment-278347717 --- I'd like to contribute a non-normative section talking about the differences between using overloads for your APIs versus using unions/optional. Potential points to cover:

- Examples of cases that cannot be expressed in one but can in the other (or, if no such cases exist, examples where it is much easier to express in one)
- Examples of non-obvious semantic differences between an interface written using overloads vs. one written using unions/optional.
- An explanation of how the prose algorithms for the two different styles are intended to be written---i.e., overloads you should have a separate prose algorithm per overload, whereas unions/optional should have a single algorithm. It's worth adding a comment here about how overloads map well to C++ via the binding layer, and should be thought of that way; they don't as obviously map to JavaScript.
- Some small encouragement to use unions/optional instead of overloads when possible, since the spec machinery involved is much less complicated and is easier to reason about.

I'm hoping for some help on the examples; see the conversation linked to above.

TimothyGu:
Fixes #307.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/overloads-vs-union-optional.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f869be8...TimothyGu:778d44d.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3347,14 +3347,14 @@ be the same.
 
 <div class="non-normative">
 
-For specifications defining IDL [=operations=], it may seem that [=overloaded|overloads=] and a
+For specifications defining IDL [=operations=], it might seem that [=overloaded|overloads=] and a
 combination of [=union types=] and [=optional arguments=] have some feature overlap.
 
-It is first important to note that [=overloaded|overloads=] may have different behaviors than
-[=union types=] or [=optional arguments=], and one <em>cannot</em> be fully defined using the other
-(unless, of course, additional prose is provided, which can defeat the purpose of the Web IDL type
-system). For example, consider the {{CanvasDrawPath/stroke()}} operations defined on the
-{{CanvasDrawPath}} interface [[HTML]]:
+It is first important to note that [=overloaded|overloads=] have different behaviors than [=union
+types=] or [=optional arguments=], and one <em>cannot</em> be fully defined using the other (unless,
+of course, additional prose is provided, which can defeat the purpose of the Web IDL type system).
+For example, consider the {{CanvasDrawPath/stroke()}} operations defined on the {{CanvasDrawPath}}
+interface [[HTML]]:
 
 <pre highlight="webidl">
     interface CanvasDrawPathExcerpt {
@@ -3394,10 +3394,10 @@ appearing in legacy APIs or in specialized circumstances.
 That being said, we offer the following recommendations and examples in case of difficulties to
 determine what Web IDL language feature to use:
 
-*   In the unusual case where the operation must return values of different types for different
-    argument types, [=overloaded|overloading=] should be used. <span class="note">This is almost
-    never appropriate API design, and separate operations with distinct names should be used for
-    such cases.</span>
+*   In the unusual case where the operation needs to return values of different types for different
+    argument types, [=overloaded|overloading=] will result in more expressive IDL fragments. <span
+    class="note">This is almost never appropriate API design, and separate operations with distinct
+    names usually are a better choice for such cases.</span>
 
     Suppose there is an operation <code class="idl">calculate()</code> that accepts a {{long}},
     {{DOMString}}, or <code class="idl">CalculatableInterface</code> (an [=interface type=]) as its
@@ -3426,7 +3426,7 @@ determine what Web IDL language feature to use:
 
     The problem is exacerbated when one of the overloads has a return type of {{void}}, since
     [=union types=] cannot even contain {{void}} as a [=member type=]. In that case, a return type
-    of {{any}} must be used with appropriate prose defining the return type, further decreasing
+    of {{any}} needs to be used with appropriate prose defining the return type, further decreasing
     expressiveness.
 
     If the specified <code class="idl">calculate()</code> is a new API and does not have any
@@ -3474,32 +3474,32 @@ determine what Web IDL language feature to use:
 
     Another consideration is that the prose for [=overloaded=] operations can be specified in
     separate blocks, which can aid in both reading and writing specifications. This is not the case
-    for [=optional arguments=]. This means that in the first case the specification author may write
+    for [=optional arguments=]. This means that in the first case the specification author can write
     the prose definition of the operations as:
 
     <div algorithm="execute supports() with overloads">
 
         The <code class="idl">supports(<var ignore>property</var>, <var ignore>value</var>)</code>
-        method, when called, must run these steps:
+        method, when called, takes the following steps:
 
         1.  …
 
         ----
 
         The <code class="idl">supports(<var ignore>conditionText</var>)</code> method, when called,
-        must run these steps:
+        takes the following steps:
 
         1.  …
 
     </div>
 
-    Yet using <var ignore>value</var> as an [=optional argument=], the specification author must use
-    more boilerplate-style text to effectively replicate the [=overload resolution algorithm=].
+    Yet using <var ignore>value</var> as an [=optional argument=], the specification author has to
+    use more boilerplate-style text to effectively replicate the [=overload resolution algorithm=].
 
     <div algorithm="execute supports() with optional argument">
 
         The <code class="idl">supports(|propertyOrConditionText|, |value|)</code> method, when
-        called, must run these steps:
+        called, takes the following steps:
 
         1.  If |value| is given, then:
             1.  Let <var ignore>property</var> be |propertyOrConditionText|.
@@ -3557,9 +3557,10 @@ determine what Web IDL language feature to use:
     Using the ECMAScript language binding, calling <code>foo(undefined)</code> and
     <code>foo(null)</code> would both run the steps corresponding to the <code
     class="idl">foo(|arg|)</code> operation, with |arg| set to null, while <code>foo()</code> alone
-    would go to the first overload. This may be a surprising behavior for many API users. Instead,
-    it is recommended to use an optional argument, which would categorize both <code>foo()</code>
-    and <code>foo(undefined)</code> as "|arg| is [=not present=]".
+    would go to the first overload. This can be a surprising behavior for many API users. Instead,
+    specification authors are encouraged to use an [=optional argument=], which would categorize
+    both <code>foo()</code> and <code>foo(undefined)</code> as "|arg| is [=not
+    present=]".
 
     <pre highlight="webidl">
         interface A {
@@ -3576,7 +3577,7 @@ intended behavior. However, the definition and <a href="#es-to-union">conversion
 [=union types=] and [=optional arguments=] are simpler to implement and reason about than [=overload
 resolution algorithm|those=] of [=overloaded|overloads=], and usually result in more idiomatic APIs
 in the ECMAScript language binding. Thus, unless any other considerations apply, [=union types=]
-(and/or [=optional arguments=]) should be the default choice.
+(and/or [=optional arguments=]) are the default choice.
 
 Specifications are also free to mix and match union types and overloads, if the author finds it
 appropriate and convenient.
--- a/index.bs
+++ b/index.bs
@@ -3341,6 +3341,207 @@ be the same.
 </div>
 
 
+<h5 id="idl-overloading-vs-union">Overloading vs. union types</h5>
+
+<i>This section is informative.</i>
+
+For specifications defining IDL [=operations=], it may seem that [=overloaded|overloads=] and a
+combination of [=union types=] and [=optional argument=] have some feature overlap.
+
+It is first important to note that [=overloaded|overloads=] may have different behaviors than
+[=union types=] or [=optional arguments=], and one <em>cannot</em> be fully defined using the
+other (unless, of course, additional prose is provided, which can defeat the purpose of the Web IDL
+type system). For example, given the {{CanvasDrawPath/stroke()}} operations defined on the
+{{CanvasDrawPath}} interface [[HTML]]:
+
+<pre highlight="webidl">
+    interface CanvasDrawPath {
+      void stroke();
+      void stroke(Path2D path);
+    };
+</pre>
+
+Per the ECMAScript language binding, calling <code>stroke(undefined)</code> on a object implementing
+{{CanvasDrawPath}} would throw a {{TypeError}}. However, if the operations were instead defined with
+[=optional arguments=] and merged into one:
+
+<pre highlight="webidl">
+    interface CanvasDrawPath {
+      void stroke(optional Path2D path);
+    };
+</pre>
+
+The [=overload resolution algorithm=] would treat the |path| argument as not present given the same
+ECMAScript language code <code>stroke(undefined)</code>, and not throw any exceptions.
+
+Additionally, there are semantic differences as well. [=Overloaded=] operations are designed to map
+well to C++ overloading at the language binding, and are usually a better fit for operations with
+more substantial differences in what they do given arguments of different types. [=Union types=], in
+contrast, are usually used in the sense that "any of the types would work in about the same way".
+
+That being said, we offer the following recommendations and examples in case of difficulties to
+determine what Web IDL language feature to use.
+
+*   When the operation must return values of different types for different argument types,
+    [=overloaded|overloading=] is almost always a better choice.
+
+    Suppose there is an operation <code class="idl">calculateWithType()</code> that accepts a single
+    argument of either {{long}}, {{DOMString}}, or <code class="idl">CalculatableInterface</code>
+    (an interface type) types, and returns a value of the same type as its argument. It would seem
+    to be clearer to write the IDL fragment using [=overloaded=] operations as
+
+    <pre highlight="webidl">
+        interface A {
+          long calculateWithType(long input);
+          DOMString calculateWithType(DOMString input);
+          CalculatableInterface calculateWithType(CalculatableInterface input);
+        };
+    </pre>
+
+    than using a [=union type=] with a [=typedef=] as
+
+    <pre highlight="webidl">
+        typedef (long or DOMString or CalculatableInterface) Calculatable;
+        interface A {
+          Calculatable calculateWithType(Calculatable input);
+        };
+    </pre>
+
+    which does not convey the fact that the return value is always of the same type as |input|.
+
+    The problem is exacerbated when one of the overloads has a return type of {{void}}, since
+    [=union types=] cannot even contain {{void}} as a [=member type=]. In that case, a return type
+    of {{any}} must be used with appropriate prose defining the return type, further decreasing
+    expressiveness.
+
+*   When the operation has significantly different semantics for different argument types or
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+<i>This section is informative.</i>
+
+For specifications defining IDL [=operations=], it may seem that [=overloaded|overloads=] and a
+combination of [=union types=] and [=optional argument=] have some feature overlap.
+
+It is first important to note that [=overloaded|overloads=] may have different behaviors than
+[=union types=] or [=optional arguments=], and one <em>cannot</em> be fully defined using the
+other (unless, of course, additional prose is provided, which can defeat the purpose of the Web IDL
+type system). For example, given the {{CanvasDrawPath/stroke()}} operations defined on the
```
domenic: s/given/consider, as otherwise it sounds like the first part of a sentence that is then left hanging.

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
 </div>
 
 
+<h5 id="idl-overloading-vs-union">Overloading vs. union types</h5>
+
+<i>This section is informative.</i>
```
domenic: There's some way of wrapping a section to say everything inside it is non-normative so that autolinks like `{{CanvasDrawPath}}` don't generate normative references, and so you get warnings when you try to use normative keywords. It appears to be undocumented. @tabatkins?

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+combination of [=union types=] and [=optional argument=] have some feature overlap.
+
+It is first important to note that [=overloaded|overloads=] may have different behaviors than
+[=union types=] or [=optional arguments=], and one <em>cannot</em> be fully defined using the
+other (unless, of course, additional prose is provided, which can defeat the purpose of the Web IDL
+type system). For example, given the {{CanvasDrawPath/stroke()}} operations defined on the
+{{CanvasDrawPath}} interface [[HTML]]:
+
+<pre highlight="webidl">
+    interface CanvasDrawPath {
```
domenic: Maybe call it `CanvasDrawPathExcerpt` or something? Especially because the below version is not CanvasDrawPath at all.

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+      void stroke();
+      void stroke(Path2D path);
+    };
+</pre>
+
+Per the ECMAScript language binding, calling <code>stroke(undefined)</code> on a object implementing
+{{CanvasDrawPath}} would throw a {{TypeError}}. However, if the operations were instead defined with
+[=optional arguments=] and merged into one:
```
domenic: comma here instead of colon, I think

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+<pre highlight="webidl">
+    interface CanvasDrawPath {
+      void stroke(optional Path2D path);
+    };
+</pre>
+
+The [=overload resolution algorithm=] would treat the |path| argument as not present given the same
```
domenic: lowercase "the" since it's continuing a sentence

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+<pre highlight="webidl">
+    interface CanvasDrawPath {
+      void stroke();
+      void stroke(Path2D path);
+    };
+</pre>
+
+Per the ECMAScript language binding, calling <code>stroke(undefined)</code> on a object implementing
+{{CanvasDrawPath}} would throw a {{TypeError}}. However, if the operations were instead defined with
```
domenic: "would attempt to call the second overload, yielding a TypeError since undefined cannot be converted to a Path2D".

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+    interface CanvasDrawPath {
+      void stroke(optional Path2D path);
+    };
+</pre>
+
+The [=overload resolution algorithm=] would treat the |path| argument as not present given the same
+ECMAScript language code <code>stroke(undefined)</code>, and not throw any exceptions.
+
+Additionally, there are semantic differences as well. [=Overloaded=] operations are designed to map
+well to C++ overloading at the language binding, and are usually a better fit for operations with
```
domenic: Not sure this is "at the language binding", but the sentence is true and this is good to point out...

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+more substantial differences in what they do given arguments of different types. [=Union types=], in
+contrast, are usually used in the sense that "any of the types would work in about the same way".
+
+That being said, we offer the following recommendations and examples in case of difficulties to
+determine what Web IDL language feature to use.
```
domenic: Colon here

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+*   When the operation must return values of different types for different argument types,
+    [=overloaded|overloading=] is almost always a better choice.
+
+    Suppose there is an operation <code class="idl">calculateWithType()</code> that accepts a single
+    argument of either {{long}}, {{DOMString}}, or <code class="idl">CalculatableInterface</code>
+    (an interface type) types, and returns a value of the same type as its argument. It would seem
+    to be clearer to write the IDL fragment using [=overloaded=] operations as
```
domenic: "it would be clearer", I think

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+    Using [=optional arguments=] one can rewrite the IDL fragment as follows:
+
+    <pre highlight="webidl">
+        partial interface CSS {
+          static boolean supports(CSSOMString propertyOrConditionText, optional CSSOMString value);
+        };
+    </pre>
+
+    It can be observed that, even though the IDL is shorter, two distinctively different concepts
```
domenic: Remove "It can be observed that" clause

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+    are conflated in the first argument. This makes the second version remarkably less readable than
+    the first.
+
+    Another consideration is that many specification preprocessors allow linking to two different
```
domenic: I think this paragraph mentions specification preprocessors when that's not the primary issue. The real issue is clarity of speccing and duplicating the overload resolution logic inside your steps. I'd just say something about how specifications are easier to write and read in this case.

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+        The <code class="idl">supports(<var ignore>conditionText</var>)</code> method, when called,
+        must run these steps:
+
+        1.  …
+
+    </div>
+
+    Yet using optional argument, the specification author must use more boilerplate-style text to
+    effectively replicate the overload logic. If the two overloads have little to no shared parts,
+    it is better to leave overload resolution to the IDL mechanism.
```
domenic: Let's move the conclusion sentence after this next algorithm.

Should "overload logic" link to the algorithm?

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+    class="idl">foo(|arg|)</code> operation, with |arg| set to null, while <code>foo()</code> alone
+    would go to the first overload. This may be a surprising behavior for many API users. Instead,
+    it is recommended to use an optional argument, which would categorize both <code>foo()</code>
+    and <code>foo(undefined)</code> as "|arg| is not present".
+
+    <pre highlight="webidl">
+        interface A {
+          void foo(optional Node? arg);
+        };
+    </pre>
```
domenic: Maybe a conclusion of "In general, optionality is best expressed using the optional keyword, and not using overloads".

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+
+When the case fits none of the categories above, it is up to the specification author to choose the
+style, since it is most likely that either style would sufficiently and conveniently describe the
+intended behavior. However, it should be noted that the definition and conversion algorithms of
```
domenic: remove "it should be noted"

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+[=union types=] are simpler to implement and reason about, and can discourage implementation
+mistakes.
```
domenic: maybe add "and so should be your default choice"

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
...
+    <pre highlight="webidl">
+        interface A {
+          void foo(optional Node? arg);
+        };
+    </pre>
+
+When the case fits none of the categories above, it is up to the specification author to choose the
+style, since it is most likely that either style would sufficiently and conveniently describe the
+intended behavior. However, it should be noted that the definition and conversion algorithms of
+[=union types=] are simpler to implement and reason about, and can discourage implementation
```
domenic: "simpler ... than overloads, and can"

index.bs
```diff
@@ -3331,6 +3331,207 @@ be the same.
 </div>
 
 
+<h5 id="idl-overloading-vs-union">Overloading vs. union types</h5>
+
+<i>This section is informative.</i>
+
+For specifications defining IDL [=operations=], it may seem that [=overloaded|overloads=] and a
+combination of [=union types=] and [=optional argument=] have some feature overlap.
```
domenic: optional argument_s_

index.bs
TimothyGu: Seems to be https://github.com/tabatkins/bikeshed/issues/1069.

TimothyGu:
@domenic Done!

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+Additionally, there are semantic differences as well. [=Overloaded=] operations are designed to map
+well to language features such as C++ overloading, and are usually a better fit for operations with
+more substantial differences in what they do given arguments of different types. [=Union types=], in
+contrast, are usually used in the sense that "any of the types would work in about the same way".
+
+That being said, we offer the following recommendations and examples in case of difficulties to
+determine what Web IDL language feature to use:
+
+*   When the operation must return values of different types for different argument types,
+    [=overloaded|overloading=] is almost always a better choice.
```
annevk: I think we should recommend separate operations instead. I don't think overloads are generally a good idea as they lead to weird edge cases around optional arguments.

annevk: And different return value types also doesn't seem like a good pattern necessarily.

index.bs
tabatkins: Yeah, if you put `class=non-normative` on an ancestor, the links will be non-normative. I don't currently look at headings, because it's an expensive check to do, but now that I'm memoizing I should see if it can be done relatively cheaply.

domenic:
@annevk fair point. I was addressing this more from the perspective of "assuming you've made up your mind on an API design, here is how Web IDL's tools work best in helping you accomplish that". But I think we should also add another layer of advice explaining that for most of the scenarios where overloads are better, the API would be better designed with separate methods.

I'd be OK doing that either in this PR or as a follow-up.

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+
+<i>This section is informative.</i>
+
+<div class="non-normative">
+
+For specifications defining IDL [=operations=], it may seem that [=overloaded|overloads=] and a
+combination of [=union types=] and [=optional arguments=] have some feature overlap.
+
+It is first important to note that [=overloaded|overloads=] may have different behaviors than
+[=union types=] or [=optional arguments=], and one <em>cannot</em be fully defined using the other
```
domenic: Missing `>` kills much of the output here

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+
+<pre highlight="webidl">
+    interface CanvasDrawPathExcerpt {
+      void stroke();
+      void stroke(Path2D path);
+    };
+</pre>
+
+Per the ECMAScript language binding, calling <code>stroke(undefined)</code> on a object implementing
+{{CanvasDrawPath}} would attempt to call the second overload, yielding a {{TypeError}} since
```
domenic: Implementing CanvasDrawPathExcerpt?

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+<emu-val>undefined</emu-val> cannot be <a href="#es-to-interface">converted</a> to a {{Path2D}}.
+However, if the operations were instead defined with [=optional arguments=] and merged into one,
+
+<pre highlight="webidl">
+    interface CanvasDrawPathExcerpt {
```
domenic: CanvasDrawPathExcerpt2?

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+      void stroke(optional Path2D path);
+    };
+</pre>
+
+the [=overload resolution algorithm=] would treat the <var ignore>path</var> argument as not present
+given the same ECMAScript language code <code>stroke(undefined)</code>, and not throw any
```
domenic: Maybe "not present, given the same call `stroke(undefined)`,"

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+exceptions.
```
domenic: Maybe we should note this latter behavior is actually more like what web developers expect, and that if CanvasDrawPath were defined today, this would use optional arguments instead.

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+
+Additionally, there are semantic differences as well. [=Overloaded=] operations are designed to map
+well to language features such as C++ overloading, and are usually a better fit for operations with
+more substantial differences in what they do given arguments of different types. [=Union types=], in
+contrast, are usually used in the sense that "any of the types would work in about the same way".
```
domenic: This may be a good place to insert something like

> In most cases, operations with such substantial differences are best off with different names. As such, overloads are rarely appropriate, often appearing in legacy APIs or in specialized circumstances.

index.bs
domenic: Yeah. So maybe "In the unusual case where an operation must return values... This is almost never appropriate API design, and separate operations with distinct names should be used instead. But if we assume you have such an API, overloading is almost always a better choice."

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+    which does not convey the fact that the return value is always of the same type as <var
+    ignore>input</var>.
+
+    The problem is exacerbated when one of the overloads has a return type of {{void}}, since
+    [=union types=] cannot even contain {{void}} as a [=member type=]. In that case, a return type
+    of {{any}} must be used with appropriate prose defining the return type, further decreasing
+    expressiveness.
+
+*   When the operation has significantly different semantics for different argument types or
+    lengths, [=overloaded|overloading=] is preferred.
```
domenic: Maybe: "Again, in such scenarios, it is usually better to create separate operations with distinct names, but legacy APIs sometimes follow this pattern."

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+        partial interface CSS {
+          static boolean supports(CSSOMString propertyOrConditionText, optional CSSOMString value);
+        };
+    </pre>
+
+    Even though the IDL is shorter, two distinctively different concepts are conflated in the first
+    argument. This makes the second version remarkably less readable than the first.
+
+    Another consideration is that the prose for [=overloaded=] operations can be specified in
+    separate blocks, which can aid in both reading and writing specifications, but not [=optional
```
domenic: "reading and writing specifications. This is not the case for optional arguments."

index.bs
```diff
@@ -3341,6 +3341,219 @@ be the same.
...
+
+    In general, optionality is best expressed using the <emu-t>optional</emu-t> keyword, and not
+    using overloads.
+
+When the case fits none of the categories above, it is up to the specification author to choose the
+style, since it is most likely that either style would sufficiently and conveniently describe the
+intended behavior. However, the definition and <a href="#es-to-union">conversion algorithms</a> of
+[=union types=] are simpler to implement and reason about than [=overload resolution
+algorithm|those=] of [=overloaded|overloads=], and can discourage implementation mistakes. Thus,
+unless any other considerations apply, [=union types=] should be the default choice.
```
domenic: Maybe "discourage implementation mistakes" is not the real concern here, but instead "and usually result in more idiomatic APIs in the ECMAScript binding" or something.

domenic: "union types (and/or optional arguments)"?

TimothyGu:
@Domenic @annevk Changed to be more overloading-hostile :)

index.bs
```diff
@@ -3341,6 +3341,249 @@ be the same.
...
+language-level overloading. As such, overloads are rarely appropriate for new APIs, instead often
+appearing in legacy APIs or in specialized circumstances.
+
+That being said, we offer the following recommendations and examples in case of difficulties to
+determine what Web IDL language feature to use:
+
+*   In the unusual case where the operation must return values of different types for different
+    argument types, [=overloaded|overloading=] should be used. <span class="note">This is almost
+    never appropriate API design, and separate operations with distinct names should be used for
+    such cases.</span>
```
annevk: You cannot have should in a note. You probably do not want to use a note here.

TimothyGu: This entire section is non-normative. Does that mean I'm not allowed to use the RFC keywords anywhere in this section? What other words would you suggest?

annevk: "Could", "might", "can", "encouraged", "discouraged", "strongly encouraged", and "strongly discouraged" are the words I end up using at times. Some of this seems like it should be requirements, but I guess non-normative is fine for now.

TimothyGu:
I don't really see a problem with the keywords after all. [RFC 2119](https://tools.ietf.org/html/rfc2119) explicitly states:

> Note that the force of these words is modified by the requirement level of the document in which they are used.

----

@domenic 

> If you wrap this all in `<div class="non-normative">` these should be caught by the build process.

This entire section is already wrapped in it.

annevk:
@TimothyGu we're stricter than RFC 2119: https://infra.spec.whatwg.org/#conformance. The rationale is that not using these words avoids needless confusion.

TimothyGu:
Argh. Okay, I'll fix them.

TimothyGu:
@annevk Keywords are replaced.

index.bs
```diff
@@ -3474,32 +3474,32 @@ determine what Web IDL language feature to use:
...
-    Yet using <var ignore>value</var> as an [=optional argument=], the specification author must use
-    more boilerplate-style text to effectively replicate the [=overload resolution algorithm=].
+    Yet using <var ignore>value</var> as an [=optional argument=], the specification author has to
+    use more boilerplate-style text to effectively replicate the [=overload resolution algorithm=].
 
     <div algorithm="execute supports() with optional argument">
 
         The <code class="idl">supports(|propertyOrConditionText|, |value|)</code> method, when
-        called, must run these steps:
+        called, takes the following steps:
```
annevk: I think in this case it would actually be acceptable to use must, since it's an example.

PR #427, opened 2017-08-25
Link to ES for "array index"

### LINKED ISSUE #409: Explain why array index property names cannot be 2^32 - 1
https://github.com/heycam/webidl/blob/d76a9c24901a9734374e8fd9fe790307c3f3989f/index.bs#L12047

It seems pretty mysterious.  A note in the spec would be helpful.  I tried tracing it via blame, but got back to b7e3bf2b and gave up.

TimothyGu:
Kept an explicit algorithm for determining if a property name is an array index as someone (forgot who) found it to be helpful.

Fixes #409.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/array-index.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f869be8...TimothyGu:4f5c681.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -50,6 +50,8 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: Error; url: sec-error-objects
         text: SyntaxError; url: sec-native-error-types-used-in-this-standard-syntaxerror
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
+    type: abstract-op
+        text: CanonicalNumericIndexString; url: sec-canonicalnumericindexstring
     type: dfn
         text: NumericLiteral; url: sec-literals-numeric-literals
         text: ECMAScript error objects; url: sec-error-objects
@@ -96,7 +98,6 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %Promise%; url: sec-promise-constructor
         text: Property Descriptor; url: sec-property-descriptor-specification-type
         text: array index; url: array-index
-        text: CanonicalNumericIndexString; url: sec-canonicalnumericindexstring
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
@@ -12196,7 +12197,7 @@ internal method as follows.
     applied:
 
     1.  If [=Type=](|P|) is not String, then return <emu-val>false</emu-val>.
-    1.  Let |index| be [=!=] [=CanonicalNumericIndexString=](|P|).
+    1.  Let |index| be [=!=] <a abstract-op>CanonicalNumericIndexString</a>(|P|).
     1.  If |index| is <emu-val>undefined</emu-val>, then return <emu-val>false</emu-val>.
     1.  If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return
         <emu-val>false</emu-val>.
--- a/index.bs
+++ b/index.bs
@@ -95,7 +95,8 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %FunctionPrototype%; url: sec-properties-of-the-function-prototype-object
         text: %Promise%; url: sec-promise-constructor
         text: Property Descriptor; url: sec-property-descriptor-specification-type
-        text: array index; url: sec-array-exotic-objects
+        text: array index; url: array-index
+        text: CanonicalNumericIndexString; url: sec-canonicalnumericindexstring
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
@@ -11987,7 +11988,7 @@ The same applies for named properties.
 The [=indexed property getter=]
 that is defined on the derived-most interface that the
 legacy platform object implements is the one that defines the behavior
-when indexing the object with an array index.  Similarly for
+when indexing the object with an [=array index=].  Similarly for
 [=indexed property setters=].
 This way, the definitions of these special operations from
 ancestor interfaces can be overridden.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12188,18 +12189,19 @@ internal method as follows.
...
+    To determine if a property name |P| <dfn lt="is an array index|is not an array index"
+    oldids="dfn-array-index-property-name">is an [=array index=]</dfn>, the following algorithm is
+    applied:
+
+    1.  If [=Type=](|P|) is not String, then return <emu-val>false</emu-val>.
+    1.  Let |index| be [=!=] [=CanonicalNumericIndexString=](|P|).
+    1.  If |index| is <emu-val>undefined</emu-val>, then return <emu-val>false</emu-val>.
+    1.  If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return
+        <emu-val>false</emu-val>.
+    1.  Return <emu-val>true</emu-val>.
```
TimothyGu: NB: the algorithm is redone in ES terms (especially the use of [CanonicalNumericIndexString](https://tc39.github.io/ecma262/#sec-canonicalnumericindexstring), which brings symmetry with [ES's definition of array index](https://tc39.github.io/ecma262/#array-index)), but it is equivalent to the old algorithm. ToUint32 does some weird manipulation with NaN, -0, Infinity, and -Infinity, and also moduloes the input against 2<sup>32</sup>, but these irregularities were normalized by the ToString check below (which is also included in CanonicalNumericIndexString).

index.bs
```diff
@@ -12188,18 +12190,19 @@ internal method as follows.
...
+    To determine if a property name |P| <dfn lt="is an array index|is not an array index"
+    oldids="dfn-array-index-property-name">is an [=array index=]</dfn>, the following algorithm is
+    applied:
+
+    1.  If [=Type=](|P|) is not String, then return <emu-val>false</emu-val>.
+    1.  Let |index| be [=!=] <a abstract-op>CanonicalNumericIndexString</a>(|P|).
+    1.  If |index| is <emu-val>undefined</emu-val>, then return <emu-val>false</emu-val>.
+    1.  If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return
+        <emu-val>false</emu-val>.
+    1.  Return <emu-val>true</emu-val>.
```
TimothyGu: NB: the algorithm is redone in ES terms (especially the use of [CanonicalNumericIndexString](https://tc39.github.io/ecma262/#sec-canonicalnumericindexstring), which brings symmetry with [ES's definition of array index](https://tc39.github.io/ecma262/#array-index)), but it is equivalent to the old algorithm. ToUint32 does some weird manipulation with NaN, -0, Infinity, and -Infinity, and also moduloes the input against 2<sup>32</sup>, but these irregularities were normalized by the ToString check below (which is also included in CanonicalNumericIndexString).

index.bs
```diff
@@ -12188,18 +12190,19 @@ internal method as follows.
 
 <h4 id="legacy-platform-object-abstract-ops">Abstract operations</h4>
 
-<div algorithm>
+<div algorithm="to determine if a property name is an array index">
```
tobie: You don't need a value for the "algorithm" attribute if there's a DFN in it. (Ideally, all algorithm would have a dfn and wouldn't require anything but a boolean "algorithm" attribute.

index.bs
```diff
@@ -12188,18 +12190,19 @@ internal method as follows.
...
-    is an <dfn id="dfn-array-index-property-name" export>array index property name</dfn>,
-    which is a property name |P| for which the following algorithm returns true:
+    To determine if a property name |P| <dfn lt="is an array index|is not an array index"
+    oldids="dfn-array-index-property-name">is an [=array index=]</dfn>, the following algorithm is
+    applied:
+
+    1.  If [=Type=](|P|) is not String, then return <emu-val>false</emu-val>.
+    1.  Let |index| be [=!=] <a abstract-op>CanonicalNumericIndexString</a>(|P|).
+    1.  If |index| is <emu-val>undefined</emu-val>, then return <emu-val>false</emu-val>.
+    1.  If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return
```
tobie: To properly close #409, I think you want to put a small note right here about what this number represents (e.g. it's 1 less than the maximum array length as defined in ES).

index.bs
```diff
@@ -12188,18 +12190,19 @@ internal method as follows.
...
 <h4 id="legacy-platform-object-abstract-ops">Abstract operations</h4>
 
-<div algorithm>
+<div algorithm="to determine if a property name is an array index">
 
-    The name of each property that appears to exist due to an object supporting indexed properties
-    is an <dfn id="dfn-array-index-property-name" export>array index property name</dfn>,
-    which is a property name |P| for which the following algorithm returns true:
+    To determine if a property name |P| <dfn lt="is an array index|is not an array index"
+    oldids="dfn-array-index-property-name">is an [=array index=]</dfn>, the following algorithm is
```
tobie: I'm not too sure I like the name change. Aren't array indices numbers? And aren't we here verifying they're in fact a string? While more verbose, the old name made that a lot more obvious.

TimothyGu: Not really, array indices are strings according to [ECMAScript](https://tc39.github.io/ecma262/#array-index).

While the old name *is* more obvious, I'm explicitly moving the *definition* of that term out of IDL, as I do not want to provide a competing definition of what's essencially the same thing under a different name. Instead IDL should only offering an algorithm to verify that an ECMAScript value is indeed an array index, as ES does not provide that, not in algorithmic form at least.

index.bs
TimothyGu: If I leave it out, the generated `data-algorithm` attribute has `[=array index=]` instead of `array index`. We can regard that a yet to be fixed Bikeshed bug though. @tabatkins

index.bs
tobie: > Not really, array indices are strings according to ECMAScript.

Well that settles it, then. You can just ignore my comment. :)

index.bs
TimothyGu: Sure.

index.bs
tobie: Oh, that might be because it can't disambiguate the two names in the "lt" attribute. Just leave in the "algorithm" attribute, then.

tabatkins: Nah, that looks like a bug on my part. Curious!

TimothyGu:
@tobie Anything additional I need to work on to get this in?

tobie:
I was thinking you were going to add a note to this (as discussed above).

TimothyGu:
Oh right, sorry 😧

TimothyGu:
@tobie Comment fixed!

PR #435, opened 2017-09-08
Meta: Use OrdinarySetWithOwnDescriptor

### LINKED ISSUE #201: Use OrdinarySet instead of [[Set]]
Last step of the algorithm "to invoke the internal [[Set]] method of platform objects" in [Platform object [[Set]] method](https://heycam.github.io/webidl/#platformobjectset) instructs to perform steps 3-11 of the [default [[Set]] internal method](https://tc39.github.io/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-set-p-v-receiver), when it should instead rely on [OrdinarySet](https://tc39.github.io/ecma262/#sec-ordinaryset).

It might be possible to rework the algorithm to avoid plugging into the middle of the algorithm, which is always brittle.

Also, remove |ownDesc| from Ignored Vars.

tobie:
Closes #201.

***
[Preview](https://s3.amazonaws.com/pr-preview/heycam/webidl/tobie-OrdinarySetWithOwnDescriptor.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/931394e...2f447b5.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -101,7 +101,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
-        text: OrdinarySet; url: sec-ordinaryset
+        text: OrdinarySetWithOwnDescriptor; url: sec-ordinarysetwithowndescriptor
         text: equally close values; url: sec-ecmascript-language-types-number-type
         text: internal slot; url: sec-object-internal-methods-and-internal-slots
         text: JavaScript execution context stack; url: execution-context-stack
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12086,8 +12086,7 @@ and [[#legacy-platform-object-set]].
             1.  [=Invoke the named property setter=] with |P| and |V|.
             1.  Return <emu-val>true</emu-val>.
     1.  Let |ownDesc| be <a abstract-op>LegacyPlatformObjectGetOwnProperty</a>(|O|, |P|, <emu-val>true</emu-val>).
-    1.  Perform [=?=] [=OrdinarySet=](|O|, |P|, |V|, |Receiver|) from step 3 onwards
-        with its internal "ownDesc" variable set to |ownDesc|.
+    1.  Perform [=?=] [=OrdinarySetWithOwnDescriptor=](|O|, |P|, |V|, |Receiver|, |ownDesc|).
```
annevk: `|V|` should be `<emu-val>true</emu-val>` I suspect? And it also looks like we don't have a `|Receiver|`?

annevk:
Oh wait, I misread, this does look correct. Good 😊

PR #436, opened 2017-09-09
Various editorial updates related to ECMAScript

TimothyGu:
For details, see the specific commits. The last commit ("Use abstract-op appropriately") was done mostly automatedly.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/update-es.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/d28583e...TimothyGu:20fddb6.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6235,18 +6235,8 @@ and whose value is the specified string.
 
 <p id="ecmascript-abstractop">
     Algorithms in this section use the conventions described in [=ECMA-262 section 5.2=], such as the use of steps and substeps, the use of mathematical
-    operations, and so on.  The
-    [=ToBoolean=],
-    [=ToNumber=],
-    [=ToUint16=],
-    [=ToInt32=],
-    [=ToUint32=],
-    [=ToString=],
-    [=ToObject=],
-    [=IsAccessorDescriptor=] and
-    [=IsDataDescriptor=] abstract operations and the
-    [=Type|Type(x)=]
-    notation referenced in this section are defined in ECMA-262 sections 6 and 7.
+    operations, and so on.  This section may also reference abstract operations
+    and notations defined in other parts of ECMA-262.
 </p>
 
 When an algorithm says to
--- a/index.bs
+++ b/index.bs
@@ -5750,7 +5750,7 @@ The following extended attributes are <dfn for="extended attributes">applicable 
 
 <div algorithm>
     The <dfn for="IDL type" lt="extended attribute associated with|extended attributes associated with">extended attributes associated with</dfn>
-    a [=type=] |type| are determined as follows:
+    an IDL type |type| are determined as follows:
 
     1.  Let |extended attributes| be a new empty [=ordered set|set=].
     1.  If |type| appears as part of a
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6187,18 +6182,21 @@ five forms are allowed.
...
+\[[SetPrototypeOf]],
+\[[PreventExtensions]],
+\[[GetOwnProperty]],
 \[[DefineOwnProperty]],
-\[[GetOwnProperty]], and
-\[[Delete]].
+\[[Set]],
+\[[Delete]],
+\[[OwnPropertyKeys]], and
+\[[Call]].
```
TimothyGu: This order came from ECMA-262: https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots

index.bs
```diff
@@ -6268,11 +6256,11 @@ to its caller, and so on.
         Consider the following algorithm:
 
         1.  Let |x| be the ECMAScript value passed in to this algorithm.
-        1.  Let |y| be the result of calling [=ToString=](|x|).
+        1.  Let |y| be the result of calling <a abstract-op>ToString</a>(|x|).
```
TimothyGu: This is unfortunately necessary because of https://github.com/tabatkins/bikeshed/issues/809.

index.bs
```diff
@@ -5747,7 +5742,7 @@ The following extended attributes are <dfn for="extended attributes">applicable
 
 <div algorithm>
     The <dfn for="IDL type" lt="extended attribute associated with|extended attributes associated with">extended attributes associated with</dfn>
-    a [=type=] |type| are determined as follows:
```
TimothyGu: This used to link to https://tc39.github.io/ecma262/#sec-ecmascript-language-types.

tobie: Shouldn't we add a DFN for WebIDL types here, instead of removing the linking altogether?

index.bs
tobie: I'm wondering if we should get rid of the above list too. I'm not sure keeping it around adds any value.

index.bs
```diff
@@ -1223,7 +1219,7 @@ lie outside the valid range of values for its type, as given in
     1.  Let |S| be the sequence of characters matched by the
         <emu-t class="regex"><a href="#prod-float">float</a></emu-t> token.
     1.  Let |result| be the Mathematical Value that would be obtained if
-        |S| were parsed as an ECMAScript [=NumericLiteral=].
+        |S| were parsed as an ECMAScript <emu-nt>[=NumericLiteral=]</emu-nt>.
```
tobie: Have you checked that this doesn't create issues with the JS script that handles all of the grammar productions dynamically?

TimothyGu: Yep I've checked. That script doesn't link productions automatically.

index.bs
TimothyGu: There isn't a readily available definition for "type" in Web IDL that I can find, and if I were to add one it would be a substantial undertaking to link all corresponding "type"s in the document. As a result I'd rather leave that to a different PR.

index.bs
TimothyGu: I was thinking about that too. I'll remove it.

TimothyGu:
@tobie comments addressed. +1 on landing this unsquashed.

PR #447, opened 2017-09-22
Clarify nullable types for constants

### LINKED ISSUE #446: Why is a constant type nullable?
https://heycam.github.io/webidl/#prod-ConstType

```
ConstType ::
    PrimitiveType Null
    identifier Null
```

https://heycam.github.io/webidl/#idl-nullable-type

```webidl
interface MyConstants {
  const boolean? ARE_WE_THERE_YET = false;
};
```

I couldn't find a real world example that uses nullable constant types. Why is it nullable? How is `const boolean?` different from `const boolean`?

TimothyGu:
Fixes #446.

/cc @SaschaNaz

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/nullable-constant.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/f97da39...TimothyGu:3c4eb7b.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1141,9 +1141,9 @@ Constants can appear on [=interfaces=].
     Constants have in the past primarily been used to define
     named integer codes in the style of an enumeration.  The Web platform
     is moving away from this design pattern in favor of the use of strings.
-    Specification authors who wish to define constants are strongly advised to discuss
-    this on the <a href="mailto:public-script-coord@w3.org">public-script-coord@w3.org</a>
-    mailing list before proceeding.
+    Editors who wish to use this feature are strongly advised to discuss this
+    by <a href="https://github.com/heycam/webidl/issues/new?title=Intent%20to%20use%20Constants">filing an issue</a>
+    before proceeding.
 </p>
 
 <pre highlight="webidl" class="syntax">const type constant_identifier = 42;</pre>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5508,8 +5513,8 @@ the string “OrNull”.
     is written as <code class="idl">boolean?</code>:
 
     <pre highlight="webidl">
-        interface MyConstants {
-          const boolean? ARE_WE_THERE_YET = false;
+        interface NetworkFetcher {
+          void get(optional boolean? areWeThereYet = false);
```
tobie: Why is the constant gone from this example?

index.bs
```diff
@@ -1167,6 +1167,11 @@ If an [=identifier=] is used,
 it must reference a [=typedef=]
 whose type is a primitive type or a nullable primitive type.
 
+Note: For null to be able to become the value of a constant, at least one [=nullable type=] needs to
```
tobie: I'm unconvinced this really clarifies, tbh.

index.bs
TimothyGu: This example is used to illustrate how nullable types work, not how constants work. Plus, https://github.com/heycam/webidl/issues/446#issuecomment-331388615.

tobie: Oops. Sorry. I didn't know we had const scattered across examples in the spec. :-/

Issue #172, opened 2016-09-09
Various "terms defined by this specification" are miscapitalized


domenic:
This is partially a Bikeshed issue, and it's completely editorial, so not very important. Anyway...

https://heycam.github.io/webidl/#index-defined-elsewhere shows everything lowercased. This is because all of these are classified as "dfn"-type terms.

- Many of them should be "abstract-op"
- Probably the built-in constructors like `%Promise%` and should be "interface"?
- No idea what `@@toStringTag` and friends should be (not attributes, since they aren't "for" anything)
- No idea what `%ObjProto_values%` and friends should be (maybe a "method" for `%Object%`??)
- No idea what `%ArrayPrototype%` and friends should be (not an interface... it's an object)
- Several of the terms reference "ECMAScript" which should not be lowercased, but they are still probably "dfn"

@tabatkins any ideas here?

tabatkins:
All the IDL types are for, well, IDL; many of the ES-spec concepts should map there, but it may be awkward or incomplete.  Happy to expand the set of definition types to help handle this, as I do want ES specs to have an okay time with the reference database.  I'm not really sure what all these concepts *are*, as I'm not fluent in ES-speak.

> Several of the terms reference "ECMAScript" which should not be lowercased, but they are still probably "dfn"

Unfortunately I don't have a way to handle "parts of this term should have specific capitalization, but the rest shouldn't".  At least the casing only changes in the indexes, where it can be attributed to a stylistic effect?

domenic:
I wonder if we could come up with a term type like "JS object" or "programming language value" or similar that would encompass my "no idea" bullets. If we wanted to be more specific, I guess they would be: symbol (or property name?), function, and object.

For the ECMAScript issue, maybe hard-code in a list of proper names? Or just provide an override like data-dfn-dont-override-lt-case. Meh.

tabatkins:
Problem is that I *do* want to override the casing, just not for one specific part of the phrase. I dont' want it to uppercase the first word of "conventions of the ECMAScript specification" just because you started a sentence with it in your prose, for example.

[...]

PR #451, opened 2017-10-02
More ECMAScript biblio cleanup

TimothyGu:
***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/ecma.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/93f9ef0...TimothyGu:aa87708.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,16 +40,39 @@ urlPrefix: http://www.unicode.org/glossary/; spec: UNICODE
         text: Unicode scalar value; url: unicode_scalar_value
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
     type: interface; for: ECMAScript
+        text: Array; url: sec-array-objects
         text: ArrayBuffer; url: sec-arraybuffer-objects
         text: DataView; url: sec-dataview-objects
         text: Map; url: sec-map-objects
         text: Promise; url: sec-promise-objects
         text: Set; url: sec-set-objects
         text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
+        text: %ArrayPrototype%; url: sec-properties-of-the-array-prototype-object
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8618,9 +8611,9 @@ appears on an [=interface=]
 that is not defined to [=interface/inherit=]
 from another, it indicates that the internal \[[Prototype]]
 property of its [=interface prototype object=]
-will be the intrinsic object [=%ArrayPrototype%=] rather than
-[=%ObjectPrototype%=].  This allows
-Array methods to be used more easily
+will be the intrinsic object {{%ArrayPrototype%}} rather than
+{{%ObjectPrototype%}}.  This allows
+{{Array}} methods to be used more easily
```
tobie: I *thought* we had agreed on not linking ES standard lib objects. I think we should be consistent here. And if we change our mind, link them all up in a separate PR.

tobie: Well, you were actually more precise than that in https://github.com/heycam/webidl/pull/405#issue-249858976. But we're not linking Array anywhere else in the spec, are we?

TimothyGu: Oh heh, yeah. No, we are not currently linking to Array as an interface, but we did have a link to "Array methods" which I'm trying to eliminate in favor of something more generic, like this.

Would you want me to change this to `{{ECMAScript/Array}}` then?

tobie: I don't know. Should we strive for consistency and link Array as we do Map and Set? (Note I opened #452 to discuss this further if needed it and cement it in the spec.)

TimothyGu: Made it consistent with Map/Set. Also fixed #452.

tobie:
Does this address #172 completely or only partially?

PR #457, opened 2017-10-07
Standardize around typography of IDL true/false/null

### LINKED ISSUE #437: Typography: IDL true and false
In the spec, the two values of the IDL `boolean` type are currently written with `<code class=idl>` like "`true`". Yet the ECMAScript Boolean type values are written with `<emu-val>` like "**true**" in the same spec, and most other specs (e.g. the [URL Standard](https://url.spec.whatwg.org/#dom-urlsearchparams-has)) write boolean values without any particular styling like "true".

IMO we should remove the `<code class=idl>` styling. The ES vs. IDL distinction is okay (maybe even preferred).

TimothyGu:
I kept certain constants bolded in the spec for readability. The distinction is basically summarized in the updated Document Conventions:

> * IDL types: `unsigned long`
> * IDL values: true, false, null, 0.42, −0
> * ECMAScript classes: `Map`
> * ECMAScript language types: Object
> * ECMAScript values: generally the same notation as IDL values; bolded type (e.g. **true**) may be used when it is needed to distinguish between ECMAScript and IDL values, or if the ECMAScript value does not have a single corresponding IDL value of the same name (**undefined** and **NaN**).

The last bullet point is especially needed for things like:

> The IDL `boolean` value true is converted to the ECMAScript **true** value and the IDL boolean value false is converted to the ECMAScript **false** value.

which IMO looks odd with either both bolded or both regular-weighted. Also, ECMAScript true is usually not bolded for things like

> The property has the attributes { [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true }.

.

Happy to have a discussion about it.

Fixes #437.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/webidl/true.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/90efe5b...TimothyGu:b1dfa43.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -32,6 +32,9 @@ Default Biblio Status: current
 spec: infra; type: dfn; text: list
 spec: url; type: interface; text: URL
 spec: dom; type: interface; text: Document
+spec: ecma-262; type: dfn; for: /
+    text:internal method
+    text:internal slot
 </pre>
 
 <pre class="anchors">
--- a/index.bs
+++ b/index.bs
@@ -1536,7 +1536,7 @@ empty sequence <code>[]</code> cannot be used as the value of a
 
 The value of the boolean literal tokens <emu-t>true</emu-t> and
 <emu-t>false</emu-t> are the IDL {{boolean}} values
-<code class="idl">true</code> and <code class="idl">false</code>.
+true and false.
 
 <div algorithm="value of integer tokens">
 
@@ -1616,7 +1616,7 @@ Also, <emu-t>Infinity</emu-t>, <emu-t>-Infinity</emu-t> and <emu-t>NaN</emu-t> m
 be used as the value of a {{float}} or {{double}}.
 
 The value of the <emu-t>null</emu-t> token is the special
-<emu-val>null</emu-val> value that is a member of the
+null value that is a member of the
 [=nullable types=].  The type of
 the <emu-t>null</emu-t> token is the same as the
 type of the constant, dictionary member or optional argument it is being used as the value of.
@@ -2154,9 +2154,9 @@ corresponding argument omitted.
 
 <p class="advisement">
     It is strongly suggested not to use a [=optional argument/default value=]
-    of <emu-val>true</emu-val> for {{boolean}}-typed arguments,
+    of true for {{boolean}}-typed arguments,
     as this can be confusing for authors who might otherwise expect the default
-    conversion of <emu-val>undefined</emu-val> to be used (i.e., <emu-val>false</emu-val>).
+    conversion of <emu-val>undefined</emu-val> to be used (i.e., false).
 </p>
 
 If the type of an argument is a [=dictionary type=] or a [=union type=] that has a dictionary as one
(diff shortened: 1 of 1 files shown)


tobie:
Thanks @annevk.

Will review one last time and merge this evening (or tomorrow morning).

annevk:
FWIW, I think the notation in OP makes sense. Hopefully at some point we can merge some Infra and IDL values, but making them look the same is already a big plus.

index.bs
```diff
@@ -13519,9 +13522,14 @@ The following typographic conventions are used in this document:
 *   Grammar terminals: <emu-t>sometoken</emu-t>
 *   Grammar non-terminals: <emu-nt>ExampleGrammarNonTerminal</emu-t>
 *   Grammar symbols: <emu-t class="regex">identifier</emu-t>
-*   IDL types: {{unsigned long}}
-*   ECMAScript classes: {{ECMAScript/Map}}
+*   IDL types: <code class="idl">unsigned long</code> <!-- {{unsigned long}} -->
+*   IDL values: true, false, null, 0.42, −0
+*   ECMAScript classes: <code class="idl">Map</code> <!-- {{ECMAScript/Map}} -->
 *   ECMAScript language types: Object
+*   ECMAScript values: generally the same notation as IDL values; bolded type (e.g.
```
tobie: I'm not sure we have the same perspective, here. I feel like ES values should be clearly marked as such across the spec, as they are in the ES spec.

index.bs
```diff
@@ -11434,9 +11437,9 @@ then there must exist a property with the following characteristics:
     then the property exists on every object that implements the interface.
     Otherwise, the property exists on the [=interface prototype object=].
 *   The property has attributes
-    { \[[Writable]]: |B|, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: |B| },
-    where |B| is <emu-val>false</emu-val> if the stringifier is [=unforgeable=] on the interface,
-    and <emu-val>true</emu-val> otherwise.
+    { \[[Writable]]: |B|, \[[Enumerable]]: true, \[[Configurable]]: |B| },
```
tobie: I'd rather we stayed consistent with the ES spec, which marks-up ES values within property descriptors.

tobie:
Sorry I rebased this. I had planned to just merge it before I noticed this was broader than just fixing incorrectly marked-up IDL values. Hope it's not too much of a hindrance.

TimothyGu:
@tobie The main consideration in not marking up "ECMAScript booleans" as different from "IDL booleans" is https://github.com/whatwg/infra/issues/95. I expect Infra to eventually define booleans in a similar way to Infra lists (which is explicitly stated to be equivalent to ECMAScript Lists):

> Whenever JavaScript expects a List, a list as defined here can be used; they are the same type.

If that happens, IDL booleans will be the same things as ECMAScript booleans (which makes sense), both of which are going to be mapped directly to Infra booleans.

Also, the ECMAScript-based pieces in HTML don't capitalize "true", "false", "null", or even "undefined" at all. See e.g. [The `WindowProxy` exotic object](https://html.spec.whatwg.org/multipage/window-object.html#the-windowproxy-exotic-object).

tobie:
> The main consideration in not marking up "ECMAScript booleans" as different from "IDL booleans" is whatwg/infra#95. I expect Infra to eventually define booleans in a similar way to Infra lists (which is explicitly stated to be equivalent to ECMAScript Lists):

Can we wait for those until that's clarified in infra?

> Also, the ECMAScript-based pieces in HTML don't capitalize "true", "false", "null", or even "undefined" at all. See e.g. The WindowProxy exotic object.

s/capitalize/markup/ ? If so, I see what you mean. But it still seems bizarre to create our own flavor of ES markup, while style relying on the same tags the ES spec is using.

annevk:
@TimothyGu I think you're basically correct and I'd be happy to merge this. But at this point it needs rebasing. Is this still something you want to pursue?

PR #459, opened 2017-10-11
Add a note on the difference of SyntaxError

TakayoshiKochi:
Add a note on the difference of ECMAScript's SyntaxError and
"SyntaxError", which is one of the DOMException's names.

***
[Preview](https://s3.amazonaws.com/pr-preview/TakayoshiKochi/webidl/syntax_error.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/76c5192...TakayoshiKochi:358c37e.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4401,6 +4401,10 @@ for {{DOMException}}, a description, and legacy code values.
 
 Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
 
+Note: Do not confuse {{SyntaxError!!exception}} defined here with ECMAScript's
+{{ECMAScript/SyntaxError}}. The ECMAScript one should not be used outside
+ECMAScript and it is deliberately omitted from [=simple exception=].
+
 <table id="error-names" class="vert data">
     <thead>
         <tr><th>Name</th><th>Description</th><th>Legacy code name and value</th></tr>
--- a/index.bs
+++ b/index.bs
@@ -4402,8 +4402,9 @@ for {{DOMException}}, a description, and legacy code values.
 Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
 
 Note: Do not confuse {{SyntaxError!!exception}} defined here with ECMAScript's
-{{ECMAScript/SyntaxError}}. The ECMAScript one should not be used outside
-ECMAScript and it is deliberately omitted from [=simple exception=].
+{{ECMAScript/SyntaxError}}. The ECMAScript one is deliberately omitted from
+[=simple exception=], and cannot be used for reporting non ECMAScript errors
+such as parsing errors of CSS selector.
 
 <table id="error-names" class="vert data">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4401,6 +4401,10 @@ for {{DOMException}}, a description, and legacy code values.
 
 Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
 
+Note: Do not confuse {{SyntaxError!!exception}} defined here with ECMAScript's
+{{ECMAScript/SyntaxError}}. The ECMAScript one should not be used outside
+ECMAScript and it is deliberately omitted from [=simple exception=].
```
annevk: Cannot use "should not" in a note. If you make it a statement of fact by using "cannot" it's fine.

TakayoshiKochi: I see.  Does the revised note look okay?  If adding an example "such as~" is not acceptable, it would be okay to omit that part and simply say "The ECMAScript one is delibarately ommitted from simple exception".

index.bs
```diff
@@ -4401,6 +4401,11 @@ for {{DOMException}}, a description, and legacy code values.
 
 Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
 
+Note: Do not confuse {{SyntaxError!!exception}} defined here with ECMAScript's
```
tobie: Think it would make sense to use the (correct)  `"{{SyntaxError!!exception}}" {{DOMException}}` notation here.

index.bs
```diff
@@ -4401,6 +4401,11 @@ for {{DOMException}}, a description, and legacy code values.
...
+{{ECMAScript/SyntaxError}}. The ECMAScript one is deliberately omitted from
```
tobie: I'd write something like this again here:

> The ECMAScript {{ECMAScript/SyntaxError}} is deliberately omitted from [=simple exceptions=] as it is reserved for the ECMAScript parser.

And then add something about using "{{SyntaxError!!exception}}" {{DOMException}} for other kids of parsing erros (e.g. CSS parser) after that.

tobie:
Apologies for my terse comments above. I was on my phone and had not seen some of @annevk's comments.

Thanks a lot for adding this note. I think it indeeds helps clarify this pretty well. I'll just bake in the suggested changes for you, if that's OK.

TakayoshiKochi:
@tobie thanks for the comment!  I was on other stuff and haven't been back to this until now, but feel free to add/modify the note.

Just FYI, the context for this is
https://bugs.chromium.org/p/chromium/issues/detail?id=609415#c20
that while investigating the filed bug (throw "SyntaxError" for querySelector), we noticed that there are two SyntaxErrors...

tobie:
IMHO, the DOM standard should really say:

> Throw a "SyntaxError" DOMException.

and not:

> Throw a SyntaxError.

tobie:
Just created a PR for this in DOM: https://github.com/whatwg/dom/pull/518.

PR #462, opened 2017-10-16
Fix #461 - clarify applicability of overload resolution

### LINKED ISSUE #461: Missing argument handling
Browsers throw exceptions if DOM APIs are invoked with missing arguments, and it's been this way for a while.

This behaviour doesn't seem clearly specified here, however. The algorithm for resolving overloaded function calls will indeed cause `TypeError` to be thrown if the arguments don't match any overload. However, from the definition of overloaded operations above:

> If a regular operation or static operation defined on an interface has an identifier that is the same as the identifier of another operation on that interface of the same kind (regular or static), then the operation is said to be overloaded. 

this algorithm doesn't seem to apply to non-overloaded calls.

It seems sensible to have this mentioned somewhere - I'm happy to PR it in.

timruffles:
This algorithm applies to all function invocations, not
just those described in the spec as overloaded - see #461 for discussion.

***
[Preview](https://s3.amazonaws.com/pr-preview/timruffles/webidl/tr/specific.html) | [Diff](https://s3.amazonaws.com/pr-preview/heycam/webidl/6673a82...timruffles:c3e0be1.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10296,7 +10296,7 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
 <div algorithm>
 
-    In order to define how overloaded function invocations are resolved, the
+    In order to define how function invocations are resolved, the
     <dfn id="dfn-overload-resolution-algorithm" export>overload resolution algorithm</dfn>
     is defined.  Its input is an [=effective overload set=],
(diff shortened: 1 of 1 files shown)


bzbarsky:
Thank you for the fix!

PR #483, opened 2017-12-03
Add space to class string of iterator objects

### LINKED ISSUE #419: Class string of default iterator object seems to have been accidentally changed?

TimothyGu:
The first commit changes the notation of strings within Web IDL to Infra-mandated "`string`". This helps prevent errors like #419 to exist in the first place, as it can be difficult to tell the difference between " Iterator" and "Iterator" without monospace styling.

The second commit fixes a Bikeshed linking error.

The third commit is a normative change that actually fixes #419, by incorporating the consensus reached in https://github.com/heycam/webidl/issues/419#issuecomment-323790991.

***
<a href="https://pr-preview.s3.amazonaws.com/TimothyGu/webidl/pull/483.html" title="Last updated on Dec 7, 2017, 11:40 AM GMT (c596cef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/483/2394cc9...TimothyGu:c596cef.html" title="Last updated on Dec 7, 2017, 11:40 AM GMT (c596cef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4038,7 +4038,7 @@ that have [=members=] with these names.
 
         typeof SessionManager.prototype.values;            // Evaluates to "function"
         var it = sm.values();                              // values() returns an iterator object
-<!--    String(it);                                        // Evaluates to "[object SessionManagerIterator]"
+<!--    String(it);                                        // Evaluates to "[object SessionManager Iterator]"
                                                            // TODO: https://github.com/heycam/webidl/issues/419 -->
         typeof it.next;                                    // Evaluates to "function"
 
@@ -11854,7 +11854,7 @@ its index is set to 0.
 
 The [=class string=] of a [=default iterator object=] for a given [=interface=]
 is the result of concatenating the [=identifier=] of the [=interface=]
-and the string "<code>Iterator</code>".
+and the string "<code> Iterator</code>".
 
 
 <h5 id="es-iterator-prototype-object">Iterator prototype object</h5>
@@ -11918,7 +11918,7 @@ must be {{%IteratorPrototype%}}.
 
 The [=class string=] of an [=iterator prototype object=] for a given [=interface=]
 is the result of concatenating the [=identifier=] of the [=interface=]
-and the string "<code>Iterator</code>".
+and the string "<code> Iterator</code>".
 
 
 <h4 id="es-maplike">Maplike declarations</h4>
--- a/index.bs
+++ b/index.bs
@@ -32,6 +32,8 @@ Default Biblio Status: current
 spec: infra; type: dfn; text: list
 spec: url; type: interface; text: URL
 spec: dom; type: interface; text: Document
+spec: ecma-262; type: dfn; for: /; text: internal method
+spec: ecma-262; type: dfn; for: /; text: internal slot
 </pre>
 
 <pre class="anchors">
@@ -7828,10 +7830,9 @@ objects.
             return <emu-val>undefined</emu-val>.
         1.  Otherwise, return the result of performing any steps that were required to be run if the promise was rejected,
             with |reason| as the rejection reason.
-    1.  Let |then| be the result of calling the internal \[[Get]] method of |promise| with property name "<code>then</code>".
-    1.  If <a abstract-op>IsCallable</a>(|then|) is <emu-val>false</emu-val>, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  Return the result of calling |then| with |promise| as the <emu-val>this</emu-val> value and |onFulfilled| and |onRejected|
-        as its two arguments.
+    1.  Let |then| be [=?=] <a abstract-op>GetMethod</a>(|promise|, "<code>then</code>").
+    1.  If |then| is <emu-val>undefined</emu-val>, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+    1.  Return <a abstract-op>Call</a>(|then|, |promise|, «|onFulfilled|, |onRejected|»).
 </div>
 
 <p class="issue">
@@ -8269,7 +8270,7 @@ If the [{{Constructor}}]
 [=extended attribute=]
 appears on an [=interface=], it indicates that
 the [=interface object=] for this interface
-will have an \[[Construct]] internal method,
+will have an \[[Construct]] [=internal method=],
 allowing objects implementing the interface to be constructed.
 
 Multiple [{{Constructor}}] extended
@@ -8892,8 +8893,8 @@ If the [{{LegacyArrayClass}}]
 [=extended attribute=]
 appears on an [=interface=]
 that is not defined to [=interface/inherit=]
-from another, it indicates that the internal \[[Prototype]]
-property of its [=interface prototype object=]
+from another, it indicates that the \[[Prototype]]
+[=internal slot=] of its [=interface prototype object=]
 will be the intrinsic object {{%ArrayPrototype%}} rather than
 {{%ObjectPrototype%}}.  This allows
 {{ECMAScript/Array}} methods to be used more easily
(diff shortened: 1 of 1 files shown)


tobie:
I'd like to hear @annevk and @domenic's thoughts on whether all of the strings modified here fall under the infra string definition. Why it's unquestionably the case for some of the event. I'm unsure about others. For example, when you talk about the "foo interface property", should "foo" be marked-up as a string (`"<code>foo</code>"`)?

annevk:
If you're specifically talking about the name exposed somewhere, I'd think so. That's what I did in https://github.com/whatwg/html/pull/3242 anyway. If it's an informal reference of sorts I wouldn't do it.

tobie:
So would you say: 

```
Note: In the ECMAScript language binding, an interface that is iterable
will have "<code>entries</code>", "<code>forEach</code>", "<code>keys</code>", "<code>values</code>" and {{@@iterator}} properties on its [=interface prototype object=].
```

…or would you use another notation there? And if so, which one?

annevk:
I'm not sure. I guess I'd look what ECMAScript does.

I would add an Oxford comma.

For platform objects we always talk about the `foo()` method or `bar` attribute so it doesn't come up as much.

tobie:
So ECMAScript generally uses:

```
the <code>foo</code> property
```

… though it sometimes (albeit rarely) adds quotes:

```
the <code>"foo"</code> property
```

domenic:
In general I like referring to properties, methods, and class names without quotes, using `<code>` instead. That's also what's encouraged by Bikeshed's default conventions, e.g. when you do `{{EventTarget}}`, there are no quotes generated in the output. That speaks to examples like

```diff
Note: A leading <span class="char">"_"</span> is used to escape an identifier from looking
-like a reserved word so that, for example, an interface named “interface” can be
-like a reserved word so that, for example, an interface named "<code>interface</code>" can be
defined.  The leading <span class="char">"_"</span> is dropped to unescape the
identifier.
```

which should probably be just `<code>interface</code>`, or

```diff
{{ECMAScript/Set}} objects.  If the <emu-t>readonly</emu-t>
-keyword is used, this includes “entries”, “forEach”, “has”,
-“keys”, “values”, {{@@iterator}} methods and a “size” getter.
-For read–write setlikes, it also includes “add”, “clear”, and “delete” methods.
+keyword is used, this includes "<code>entries</code>", "<code>forEach</code>", "<code>has</code>",
+"<code>keys</code>", "<code>values</code>", {{@@iterator}} methods and a "<code>size</code>" getter.
+For read–write setlikes, it also includes "<code>add</code>", "<code>clear</code>", and "<code>delete</code>" methods.
```

where again the quotes are probably not ideal.

It's less clear what to do when talking about an identifier, or a "name" e.g. as in

```diff
-Although the “toJSON” [=identifier=] is not a [=reserved identifier=],
+Although the "<code>toJSON</code>" [=identifier=] is not a [=reserved identifier=],
```

or

```diff
Interfaces that [=support indexed properties=] must define
-an [=integer types|integer-typed=] [=attribute=] named “length”.
+an [=integer types|integer-typed=] [=attribute=] named "<code>length</code>".
```

Probably identifiers and names are strings? I imagine the line gets blurry sometimes. E.g. if we modified the methods listing above to say "methods named ..." instead of "... methods", we'd need to add quotes everywhere?

TimothyGu:
> E.g. if we modified the methods listing above to say "methods named ..." instead of "... methods", we'd need to add quotes everywhere?

I'm comfortable with drawing the line here.

This is also what ECMAScript spec does (7 instances of `property named "` versus 1 instance of `property named`; 7 instances of `"</code> property` versus more than 150 instances of `[^"]</code> property` or `[^"]</code> data property`). This also works well with the variable/value equivalence: "if *A* has a method named *P*" works better than "if *A* has a *P* method".

index.bs
```diff
@@ -3978,12 +3981,14 @@ is.
...
 For interfaces that [=support indexed properties=],
-the iterator objects returned by “entries”, “keys”, “values” and {{@@iterator}} are
+the iterator objects returned by <code class="idl">entries</code>,
+<code class="idl">keys</code>, <code class="idl">values</code>, and {{@@iterator}} are
 actual [=array iterator objects=].
 
 Interfaces with iterable declarations must not
 have any [=interface members=]
-named “entries”, “forEach”, “keys” or “values”,
+named "<code class="idl">entries</code>", "<code class="idl">forEach</code>",
```
tobie: Is this `"<code class="idl">entries</code>"` on purpose or should it be: `"<code>entries</code>"`?

TimothyGu: It’s that way on purpose. I decided to use idl class for identifiers that are not plain strings.

tobie: Mmm. You're consciously making the decision to mark those up differently than idl props (which aren't quoted), but you don't want to mark those as "infra" strings. Why not just `"entries"`, then?

TimothyGu:
Bugs:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1424160
- https://bugs.chromium.org/p/chromium/issues/detail?id=793204

bzbarsky:
@TimothyGu Are there web platform tests for this change, by any chance?

TimothyGu:
@bzbarsky I don't think WPT has any tests for toStringTag yet (for anything, iterator objects and regular objects), because of the fact that different browser vendors haven't implemented it. So, no.

bzbarsky:
OK, but we seem to sort of have browser consensus on the behavior specifically for iterator objects, right?  And the spec reflects that consensus?  If so, seems like we can add tests; tests don't need to wait on implementation...

TimothyGu:
@bzbarsky I've opened #501 to make the spec fully consistent with browser consensus. Tests for both this PR and #501 are in https://github.com/w3c/web-platform-tests/pull/8796.

index.bs
```diff
@@ -4033,7 +4038,7 @@ that have [=members=] with these names.
 
         typeof SessionManager.prototype.values;            // Evaluates to "function"
         var it = sm.values();                              // values() returns an iterator object
-<!--    String(it);                                        // Evaluates to "[object SessionManagerIterator]"
+<!--    String(it);                                        // Evaluates to "[object SessionManager Iterator]"
                                                            // TODO: https://github.com/heycam/webidl/issues/419 -->
```
ExE-Boss: This should’ve been uncommented in this PR:
```suggestion
        String(it);                                        // Evaluates to "[object SessionManager Iterator]"
````

I’m fixing this in <https://github.com/heycam/webidl/pull/858>.

Issue #516, opened 2018-01-31
Definition of "is an array index" makes "-0" an array index


foolip:
https://heycam.github.io/webidl/#dfn-array-index-property-name

This links to https://tc39.github.io/ecma262/#sec-canonicalnumericindexstring which special-cases "-0".

This problem was introduced with https://github.com/heycam/webidl/pull/427, the previous definition was OK:
https://web.archive.org/web/20170903233050/heycam.github.io/webidl/#dfn-array-index-property-name

If the current definition were used, it would mean that `document.all["-0"]` should behave like `document.all["0"]`, which it doesn't as implemented.

foolip:
@TimothyGu can you take a look?

TimothyGu:
I'm inclined to just say "If *P* is an array index" as ES does it, but since that's not preferred here just adding a `"-0"` as an early return is fine.

annevk:
Well, this affects more than just `document.all`, right? I think we very much want the Ecma array index definition for other indexed getters...

foolip:
Yes, it affects more things, and I don't think @TimothyGu was suggested a `document.all`-specific workaround.

PR #517, opened 2018-01-31
Disallow -0 in algorithm for checking array index

TimothyGu:
Fixes: #516 

CC @foolip

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -181,6 +181,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: error objects; for: ECMAScript; url: sec-error-objects
         text: function object; url: function-object
         text: immutable prototype exotic object; url: sec-immutable-prototype-exotic-objects
+        text: integer index; url: integer-index
         url: sec-object-internal-methods-and-internal-slots
             text: internal method
             text: internal slot
@@ -12625,6 +12626,10 @@ internal method as follows.
     applied:
 
     1.  If <a abstract-op>Type</a>(|P|) is not String, then return <emu-val>false</emu-val>.
+    1.  If |P| is "<code>-0</code>", then return <emu-val>false</emu-val>.
+
+        Note: <a abstract-op>CanonicalNumericIndexString</a> allows "<code>-0</code>" as a special
+        case, even though it is not an [=integer index=] or by extension an [=array index=].
     1.  Let |index| be [=!=] <a abstract-op>CanonicalNumericIndexString</a>(|P|).
     1.  If |index| is <emu-val>undefined</emu-val>, then return <emu-val>false</emu-val>.
     1.  If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return


index.bs
```diff
@@ -12625,6 +12626,10 @@ internal method as follows.
     applied:
 
     1.  If <a abstract-op>Type</a>(|P|) is not String, then return <emu-val>false</emu-val>.
+    1.  If |P| is "<code>-0</code>", then return <emu-val>false</emu-val>.
+
+        Note: <a abstract-op>CanonicalNumericIndexString</a> allows "<code>-0</code>" as a special
+        case, even though it is not an [=integer index=] or by extension an [=array index=].
```
annevk: If this indeed matches implementations, wouldn't it be cleaner to return false if index is -0 after the following step?

foolip:
I have tests for the implications fo `document.all` in https://chromium-review.googlesource.com/c/chromium/src/+/880621. They already pass in Safari, and will pass in Chrome with those changes.

annevk:
FWIW, `w(document.querySelectorAll("body")["-0"])` (unlike `w(document.querySelectorAll("body")[-0])`) indeed returns undefined so this is indeed a bug.

I wonder how many other algorithms have subtle -0 issues...

index.bs
foolip: The implementation for this in Chromium:
https://chromium.googlesource.com/v8/v8.git/+/842d424d99060375bf4a8b6fdfd2a0177e4c0477/src/api.cc#4128
https://chromium.googlesource.com/v8/v8.git/+/842d424d99060375bf4a8b6fdfd2a0177e4c0477/src/utils-inl.h#35

So no special handling of "-0" there because the code uses unsigned integers. ES uses ToNumber where negative zero is possible, so whatever this says won't be a close match.

tobie:
Wouldn't just changing step 4 from:

```
If |index| is less than 0 or is greater than or equal to 2<sup>32</sup> − 1, then return false .

```
to:

```
If |index| is not in the range +0 ≤ |index| < 2<sup>32</sup> − 1, then return false.
```

Maybe we could also mention -0 in the note right below, or just say that the range corresponds to the range of supported array indices in ECMAscript.

foolip:
With a note saying that it's supposed to be equivalent to https://tc39.github.io/ecma262/#array-index anything would do, but using just the ≤ operator in the normative definition seems wrong. Whether operating on IEEE doubles or real numbers, positive zero equals negative zero. So it's not possible to write that expression out as-is in code or in mathematics and have -0  not be in the range.

annevk:
Hopefully this will all be a bit clearer once https://github.com/whatwg/infra/issues/87 is solved.

tobie:
> but using just the ≤ operator in the normative definition seems wrong.

Yeah, I tend to agree, but as I lifted that off of the ES spec I get to blame someone else. :D

foolip:
Who can make a decision on this? Anything WFM.

tobie:
A bunch of suggestions were made to improve the initial PR. I'll make a last one: replace step 5 with the following:

```
1.  If |index| is not an [=array index=], then return false.
```

and fix or remove the related note.

tobie:
Realized how tautological my previous comment was. Pushed some changes to @TimothyGu's branch. Fixed the issue that we weren't filtering out non integers before.

@foolip can you give the latest a look?

index.bs
```diff
@@ -12627,11 +12628,12 @@ internal method as follows.
...
-        <emu-val>false</emu-val>.
-
-          Note: 2<sup>32</sup> − 1 is the maximum array length allowed by ECMAScript.
-
+    1.  If <a abstract-op>IsInteger</a>(|index|) is <emu-val>false</emu-val>,
+        then return <emu-val>false</emu-val>.
+    1.  If |index| is -0, then return <emu-val>false</emu-val>.
+    1.  If |index| is smaller than 0, then return <emu-val>false</emu-val>.
+    1.  If |index| is greater than or equal to 2<sup>32</sup> − 1,
+        then return <emu-val>false</emu-val>.
```
TimothyGu: I would use < and ≥, just to make things a bit more obvious. If you prefer English over mathematical symbols, s/smaller than/less than/.

Also I'd like to keep the note:

> Note: 2<sup>32</sup> − 1 is the maximum array length allowed by ECMAScript.

tobie: I was hesitant about removing the note, but if it's not just me, then I'll keep it around.

tobie:
Yay, thanks. Writing the test and merging this.

PR #525, opened 2018-02-22
Add bigint type

littledan:
This patch adds a bigint type to WebIDL with the following properties:
- bigint corresponds directly to BigInt
- The conversion from JS is based on ToBigInt, namely it throws on Number.
- bigint is not grouped as a "numeric" type, as this category is often
  used to refer to types which have a JavaScript representation of Number.
- bigint is included in the overloading resolution algorithm, and it's
  possible to overload a function for BigInt and numeric arguments.

Although it may be on the early side for such a change, the lack of
BigInt WebIDL integration is already coming up in some designs for integrating
BigInt with the Web Platform, such as https://github.com/w3c/IndexedDB/pull/231

***
<a href="https://pr-preview.s3.amazonaws.com/littledan/webidl/pull/525.html" title="Last updated on Nov 5, 2020, 4:57 PM UTC (4f8357a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/525/d87d389...littledan:4f8357a.html" title="Last updated on Nov 5, 2020, 4:57 PM UTC (4f8357a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -165,6 +165,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: ToBigInt; url: #sec-tobigint
         text: ToInt32; url: sec-toint32
         text: ToNumber; url: sec-tonumber
+        text: ToNumeric; url: sec-tonumeric
         text: ToObject; url: sec-toobject
         text: ToString; url: sec-tostring
         text: ToUint16; url: sec-touint16
@@ -7678,6 +7679,18 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
         value.
 </div>
 
+<div id="es-to-bigint" algorithm="convert an ECMAScript value to a numeric type or a bigint">
+
+    An ECMAScript value |V| is <dfn lt="converted to a numeric type or bigint">converted</dfn>
+    to an IDL [=numeric type=] |T| or {{bigint}} value by running the following algorithm:
+
+    1.  Let |x| be [=?=] [$ToNumeric$](|V|).
+    1.  If [$Type$](|x|) is BigInt, then
+        1.  Return the IDL {{bigint}} value that represents the same numeric value as |x|.
+    1.  Assert: [$Type$](|x|) is Number.
+    1.  Return the result of [=converted to an ECMAScript value|converting=] |x| to |T|.
+</div>
+
 
 <h4 id="es-DOMString">DOMString</h4>
 
@@ -8722,6 +8735,9 @@ that correspond to the union’s [=member types=].
         then return the result of
         [=converted to an IDL value|converting=]
         |V| to that type.
+    1.  If |types| includes a [=numeric type=] and {{bigint}},
+        then return the result of [=converted to a numeric type or bigint|converting=]
+        |V| to either that [=numeric type=] or {{bigint}}.
     1.  If |types| includes a [=numeric type=],
         then return the result of [=converted to an IDL value|converting=]
         |V| to that [=numeric type=].
--- a/index.bs
+++ b/index.bs
@@ -5924,14 +5924,12 @@ The [=type name=] of the
 
 <h4 id="idl-bigint" interface>bigint</h4>
 
-The {{bigint}} type is an arbitrary integer, unrestricted in range.
+The {{bigint}} type is an arbitrary integer type, unrestricted in range.
 
-{{bigint}} constant values in IDL are
-represented with <emu-t class="regex"><a href="#prod-bigint">bigint</a></emu-t>
-tokens.
+{{bigint}} constant values in IDL are represented with
+<emu-t class="regex"><a href="#prod-bigint">bigint</a></emu-t> tokens.
 
-The [=type name=] of the
-{{bigint}} type is “BigInt”.
+The [=type name=] of the {{bigint}} type is “<code>BigInt</code>”.
 
 
 <h4 oldids="dom-short" id="idl-short" interface>short</h4>
@@ -7228,9 +7226,9 @@ ECMAScript value type.
     1.  If <a abstract-op>Type</a>(|V|) is Number, then
         return the result of <a href="#es-to-unrestricted-double">converting</a> |V|
         to an {{unrestricted double}}.
-    1.  If <a abstract-op>Type</a>(|V|) is [=BigInt=], then
+    1.  If <a abstract-op>Type</a>(|V|) is BigInt, then
         return the result of <a href="#es-to-bigint">converting</a> |V|
-        to an {{bigint}}.
+        to a {{bigint}}.
     1.  If <a abstract-op>Type</a>(|V|) is String, then
         return the result of <a href="#es-DOMString">converting</a> |V|
         to a {{DOMString}}.
@@ -7667,18 +7665,17 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
     An ECMAScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{bigint}} value by running the following algorithm:
 
-    1.  Let |x| be [=?=] <a abstract-op>ToBigInt</a>(|V|).
-    1.  Return the IDL {{bigint}} value that represents the same numeric
-        value as |x|.
+    1.  Let |x| be [=?=] [$ToBigInt$](|V|).
+    1.  Return the IDL {{bigint}} value that represents the same numeric value as |x|.
 </div>
 
 <div id="bigint-to-es" algorithm="convert a bigint to an ECMAScript value">
 
     The result of [=converted to an ECMAScript value|converting=]
-    an IDL {{bigint}} value to an ECMAScript value is a [=BigInt=]:
+    an IDL {{bigint}} value to an ECMAScript value is a BigInt:
 
-    1.  Return the [=BigInt=] value that represents the same numeric value
-        as the IDL {{bigint}} value.
+    1.  Return the [=BigInt=] value that represents the same numeric value as the IDL {{bigint}}
+        value.
 </div>
 
 
@@ -8717,7 +8714,7 @@ that correspond to the union’s [=member types=].
         1.  If |types| includes a [=numeric type=],
             then return the result of [=converted to an IDL value|converting=]
             |V| to that [=numeric type=].
-    1.  If <a abstract-op>Type</a>(|V|) is [=BigInt=], then:
+    1.  If <a abstract-op>Type</a>(|V|) is BigInt, then:
         1.  If |types| includes {{bigint}},
             then return the result of [=converted to an IDL value|converting=]
             |V| to {{bigint}}
@@ -8725,15 +8722,13 @@ that correspond to the union’s [=member types=].
         then return the result of
         [=converted to an IDL value|converting=]
         |V| to that type.
-    1.  If |types| includes a [=numeric type=], then:
-        1.  If |types| includes {{bigint}},
-            [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-        1.  Return the result of [=converted to an IDL value|converting=]
-            |V| to that [=numeric type=].
+    1.  If |types| includes a [=numeric type=],
+        then return the result of [=converted to an IDL value|converting=]
+        |V| to that [=numeric type=].
     1.  If |types| includes {{boolean}},
         then return the result of [=converted to an IDL value|converting=]
         |V| to {{boolean}}.
-    1.  If |types| includes a {{bigint}},
+    1.  If |types| includes {{bigint}},
         then return the result of [=converted to an IDL value|converting=]
         |V| to {{bigint}}.
     1.  [=ECMAScript/Throw=] a {{ECMAScript/TypeError}}.
@@ -11065,7 +11060,7 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
             then remove from |S| all other entries.
 
-        1.  Otherwise: if <a abstract-op>Type</a>(|V|) is [=BigInt=]
+        1.  Otherwise: if <a abstract-op>Type</a>(|V|) is BigInt
             and there is an entry in |S| that has one of the following types at position |i| of its type list,
             *   {{bigint}}
             *   a [=nullable type|nullable=] {{bigint}}
--- a/index.bs
+++ b/index.bs
@@ -5932,16 +5932,6 @@ The [=type name=] of the
 {{octet}} type is "<code>Octet</code>".
 
 
-<h4 id="idl-bigint" interface>bigint</h4>
-
-The {{bigint}} type is an arbitrary integer type, unrestricted in range.
-
-{{bigint}} constant values in IDL are represented with
-<emu-t class="regex"><a href="#prod-bigint">bigint</a></emu-t> tokens.
-
-The [=type name=] of the {{bigint}} type is “<code>BigInt</code>”.
-
-
 <h4 oldids="dom-short" id="idl-short" interface>short</h4>
 
 The {{short}} type is a signed integer
@@ -6082,6 +6072,16 @@ tokens.
 
 The [=type name=] of the
 {{unrestricted double}} type is "<code>UnrestrictedDouble</code>".
+
+
+<h4 id="idl-bigint" interface>bigint</h4>
+
+The {{bigint}} type is an arbitrary integer type, unrestricted in range.
+
+{{bigint}} constant values in IDL are represented with
+<emu-t class="regex"><a href="#prod-bigint">bigint</a></emu-t> tokens.
+
+The [=type name=] of the {{bigint}} type is “<code>BigInt</code>”.
 
 
 <h4 oldids="dom-DOMString" id="idl-DOMString" interface>DOMString</h4>
--- a/index.bs
+++ b/index.bs
@@ -7701,7 +7701,7 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
         value.
 </div>
 
-<div id="es-to-bigint" algorithm="convert an ECMAScript value to a numeric type or a bigint">
+<div id="es-to-bigint-or-numeric" algorithm="convert an ECMAScript value to a numeric type or a bigint">
 
     An ECMAScript value |V| is <dfn lt="converted to a numeric type or bigint">converted</dfn>
     to an IDL [=numeric type=] |T| or {{bigint}} value by running the following algorithm:
--- a/index.bs
+++ b/index.bs
@@ -3655,7 +3655,7 @@ the following algorithm returns <i>true</i>.
             <th>bigint</th>
             <td class="belowdiagonal"></td>
             <td class="belowdiagonal"></td>
-            <td></td>
+            <td>(b)</td>
             <td>●</td>
             <td>●</td>
             <td>●</td>
@@ -3761,6 +3761,10 @@ the following algorithm returns <i>true</i>.
         1.  The two identified interface-like types are
             not the same, and no single [=platform object=] implements both
             interface-like types.
+        1.  The types are distinguishable, but there is
+            <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
+            overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
+            advice about using unions of these types</a>.
     </ol>
 
     <div class="example" id="example-distinguishability-diff-types">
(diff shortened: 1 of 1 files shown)


bzbarsky:
It's a little weird that this setup would allow passing `"1"` as a bigint but not allow passing `1`.

I assume bigint should not in fact be a "JSON type" because it can't be represented in JSON?  Looks to me like ToJSON throws on BigInts in the BigInt proposal.

domenic:
In particular this situation is interesting:

```webidl
interface X {
  void foo(bigint x);
  void bar(unsigned long long y);
}
```

What does calling `someX.foo("1")` do? What *should* it do?

bzbarsky:
> What does calling someX.foo("1") do?

The way the PR is written right now, it would call the `unsigned long long` overload, if you meant to name both your operations the same.  If you really did mean different names, then it would convert the string to bigint and call the one thing named "foo".

> What should it do?

Good question.  ;)

Here's a worse case:

    interface X {
      void foo(bigint x);
      void bar(bigint x);
      void bar(Node x);
    }

Here `foo("1")` would convert `"1"` to bigint, but `bar("1")` would throw, given the bits in the PR at the moment, because there is no case in https://heycam.github.io/webidl/#dfn-overload-resolution-algorithm step 12 that supports a declared type of bigint with an arg that does not have `Type(arg) == BigInt`.  So the exact changes we want to the overload resolution algorithm here require some thought.  Adding a new overload should generally not cause previously-valid calls to throw.

Also, we need corresponding changes to <https://heycam.github.io/webidl/#es-to-union>; those aren't in the PR afaict.

littledan:
> Also, we need corresponding changes to https://heycam.github.io/webidl/#es-to-union; those aren't in the PR afaict.

Added union support. I also added union support for Symbol (since I didn't see it added yet, and there are some funny cases where it could intersect).

bzbarsky:
Please don't do anything with Symbol here.  It's not in the algorithm there on purpose, because you can never have a union accepting a Symbol anyway.  The only way to pass Symbols in is via `any`.  See discussion in https://github.com/heycam/webidl/pull/377

littledan:
@bzbarsky Reverted `symbol` changes and added support in the overload algorithm for falling back to a `bigint` overload as the second-to-last resort (before an `any` overload).

index.bs
```diff
@@ -6864,6 +6913,9 @@ ECMAScript value type.
     1.  If <a abstract-op>Type</a>(|V|) is Number, then
         return the result of <a href="#es-to-unrestricted-double">converting</a> |V|
         to an {{unrestricted double}}.
+    1.  If <a abstract-op>Type</a>(|V|) is [=BigInt=], then
+        return the result of <a href="#es-to-bigint">converting</a> |V|
+        to an {{bigint}}.
```
tobie: s/an/a/ or s/an/an IDL/

littledan:
> What does calling someX.foo("1") do? What should it do?

I'd say, call the numeric overload over the bigint one. Doesn't this situation already exist in cases like the following, which would also call the numeric overload over the boolean one given a string argument?

```
interface X {
  void foo(boolean x);
  void foo(unsigned long long y);
}
```

bzbarsky:
> I'd say, call the numeric overload over the bigint one

Why?  The bigint conversion is the less lossy one in this case, I would think.

> which would also call the numeric overload over the boolean one given a string argument?

It would, yes.  Arguably conversion to numeric is less lossy than conversion to boolean.

More generally, it seems we're going out of our way to provide seamless conversions from strings (but not other primitives!) to bigints.  This is in contrast to numeric and boolean types, which can be converted from all primitives.  So in some sense, there is some sort of privileged relationship between bigint and strings that is not shared by other primitive values in JS or other primitive-like idl types...

bzbarsky:
Oh, and unlike a lot of overload scenarios, the "overload numeric and bigint" is not completely insane, because if you want to have an API that takes "numbers of arbitrary precision" your options are:

1. Have it just take `bigint`.  Callers can pass `BigInt(0)` or `"0"` but not `0`.
2. Have it take `(bigint or some-numeric-type)`.  Callers can pass `0` sanely.

Option 1 is slightly harder to use as an API, and the reasoning behind `"0"` being OK but `0` not being OK is subtle and likely to make anyone trying to just pass a small numeric constant be annoyed.

Option 2 involves an overload of bigint and numeric, and dataloss if you pass a long-enough string and it gets converted to the numeric.

littledan:
@bzbarsky I see; maybe we should then prohibit that particular overload and force an explicit string overload as well? There could still be some more obscure data loss cases, though.

bzbarsky:
There really isn't a mechanism so far for prohibiting an overload but allowing it of _more_ types are added.  We could make bigint and numeric not distinguishable, forcing option 1.  Or we could add some sort of mechanism along these lines.  Or we could make es-to-idl conversion for bigint more like https://tc39.github.io/proposal-bigint/#sec-bigint-constructor-number-value than like https://tc39.github.io/proposal-bigint/#sec-to-bigint I suppose....

littledan:
> Oh, and unlike a lot of overload scenarios, the "overload numeric and bigint" is not completely insane, because if you want to have an API that takes "numbers of arbitrary size" 

I'm not sure if that's in line with the goals of the BigInt proposal, to have new APIs which take both out of the box by design. I think most APIs fall into the category of "logically making sense for arbitrary-size integers" or "logically making sense for floating point values"; there's definitely stuff in the intersection ("logically making sense for only small integers"), and for that, Number works well. I was imagining overloading would mostly make sense when trying to upgrade an API from a Number version to a BigInt version, when we initially thought that it made sense only on small integers, but it turned out it was also important on large integers.

> Option 1 is slightly harder to use as an API, and the reasoning behind "0" being OK but 0 not being OK is subtle and likely to make anyone trying to just pass a small numeric constant be annoyed.

The hypothesis of the BigInt proposal design is that programmers will keep track of whether something is a BigInt operation or a Number operation, and the types of scalar values, and not do things generically. All the JS programmers I've heard from seem basically on board with this design. Part of this here is, even for small numeric constants, you put an `n` at the end of the literal. I don't want to encourage a proliferation of Web APIs which 

> Or we could make es-to-idl conversion for bigint more like https://tc39.github.io/proposal-bigint/#sec-bigint-constructor-number-value than like https://tc39.github.io/proposal-bigint/#sec-to-bigint I suppose....

If we did that, and privileged the bigint path, we'd get surprising exceptions for things like `f("3.2")`, wouldn't we?

I don't see a great answer here which makes everything beautiful, intuitive and not throw errors. Maybe we should start by making `bigint` and `number` indistinguishable, and make sure that unions throw in the ambiguous case. I uploaded a patch which does that.

bzbarsky:
> Part of this here is, even for small numeric constants, you put an n at the end of the literal.

Ah, if there are simple bigint literals that helps a lot.

> Maybe we should start by making bigint and number indistinguishable

Yes, please.

littledan:
> Yes, please.

How does [this patch](https://github.com/heycam/webidl/pull/525/commits/fa92cf3a23598a654df0f563921cebe0578d94fc) look?

inexorabletash:
Preface: this is all excellent discussion and @littledan's proactive approach to getting bigint integrated into the platform is awesome.

> Although it may be on the early side for such a change...

On that note: are there concrete examples of API that plan to consume/produce bigints, either via WebIDL or ES? (Indexed DB is unfortunately not a good example, since it requires custom integration)

I'm slightly nervous that since it's too early to have good examples of bigint usage we may be missing something. @littledan describes "the hypothesis of the BigInt proposal design" which makes sense to me wanting to keep mentally separate from the rest of the platform.... but what will eventual idiomatic use of bigints look like?

littledan:
WebIDL has a number of cases where it switches among all of the JS primitive types. If BigInt is a type in JavaScript, maybe we should have some description or other of what the behavior is when a BigInt ends up floating into certain operations, such as [converting an ES value to any](https://heycam.github.io/webidl/#es-any), even if we want to wait until we have more information to decide how to define what it means to overload with a `bigint` type.

One place BigInt will come up is the WebAssembly-JS API. If the [proposed BigInt/i64 integration](https://github.com/WebAssembly/spec/pull/707) is combined with the [proposed WebAssembly.Global](https://github.com/WebAssembly/mutable-global) interface, then I think we'd want to say that the type of the WebAssembly.Global.prototype.valueOf method is `bigint or unrestricted double`, for example. But maybe we could define something special in prose in that particular document, without updating WebIDL so generally.

Besides that, I don't have a particular other web platform feature that where I want to propose BigInt usage. All the examples I can think of (constant time 64-bit arithmetic operations, a date format representing time in nanoseconds since forever ago, etc) might make more sense in the JavaScript standard itself than as WebIDL-based APIs. Like the IndexedDB integration, I'm fine waiting until we have more motivation for this change.

As much as it's appropriate to wait for more applications to get input for the design of the BigInt WebIDL binding, I wanted to document a suggestion for best practices here, at least from the perspective of the way BigInt was designed in JavaScript, in particular, avoiding implicit conversions that lose information, and continuing to use Number exclusively everywhere it is currently used in WebIDL.

bzbarsky:
> The only way to pass Symbols in is via any

This may not be true.  I am checking with Domenic.

`any` definitely needs to handle all JS types, so we need something there even if we don't add a syntactic way to declare bigints.

littledan:
Another way to handle both the union and overloading semantics could be to do something similar to what the JS spec does: Call ToNumeric in the ambiguous cases, and make the judgment based on whether a Number or BigInt falls out. This is what we do for most overloaded operators such as `-`. ToNumeric converts strings to Numbers--this is required, for compatibility. Presumably a similar compatibility risk (but perhaps not as serious, depending how much the method is used) would exist if there's an existing method which takes a numeric type and wants to add a bigint overload, or a bigint member to a union.

littledan:
Any thoughts on whether we should support overloading between BigInt and Number or not? Note that one specification proposal where this arises is the integration of BigInt and Intl.NumberFormat (not implemented anywhere AFAIK): https://github.com/tc39/ecma402/pull/236 . That PR follows the ToNumeric algorithm as in https://github.com/heycam/webidl/pull/525#issuecomment-372964767 .

annevk:
I think it actually makes sense to not allow overloading/union BigInt with any other IDL numeric type as that allows for (maybe even encourages) the kind of APIs TC39 doesn't want. Where the API silently does BigInt(arg) if you pass it a non-BigInt after crossing the IDL boundary.

I'm also not sure why we'd support coercing string to BigInt, I think it makes sense to throw there as well.

annevk:
What does JavaScript do for its methods that take a BigInt?

littledan:
Well, the first built-in method proposed (not yet shipped anywhere AFAIK) is  tc39/ecma402#236 which works like operators like `*` and calls ToNumeric on the argument (which will convert a string to Number, or `{valueOf(){return 5n}}` to `5n`), then checks whether it's Number or BigInt. That's what I was suggesting in https://github.com/heycam/webidl/pull/525#issuecomment-444073662 , but if these semantics seem terrible, we could go with something more restrictive since AFAIK no overloaded methods have shipped, just overloaded operators (which have more legacy constraints).

annevk:
I think that's fine as long as we throw a TypeError if it turns out to be a Number and don't allow overloading/unioning a BigInt with a Number.

littledan:
OK, so your suggestion would be to say, the infix arithmetic operations are the only things overloaded, and otherwise we use different names?

annevk:
I thought the infix arithmetic operations were not overloaded? Beyond valueOf() that is?

I'd like to disallow
```webidl
a(double x);
a(BigInt x);
b((double or BigInt) x);
```
as per my understanding we don't want implicit `BigInt(x)`. However, if we have
```webidl
c(BigInt x);
```
and that does `ToNumber()` first for `valueOf()` similar to the infix arithmetic operations I think that'd be fine and nicely consistent.

littledan:
Infix arithmetic operations work on BigInt and Number, but both operands need to be either BigInt or Number, not mixed.

Overloading like 
```
a(double x);
a(bigint x);
```
would not cause an implicit conversion from Number to BigInt; it would just call `ToNumeric` on the argument, which would call `valueOf` on objects, and convert strings to Number, and leave Numbers and BigInts in place. That's just like what arithmetic operations like unary `-` do--still no implicit conversion from Number to BigInt, but also keeping in place all the existing conversions.

If you define `c(bigint x);`, this will call ToNumeric, and which calls valueOf but doesn't convert a Number to a BigInt; it will just throw on Number. (That's the case whether or not we allow the overloading.)

The patch currently out for review permits unions like `(double or bigint)` (I don't see any particular way to prohibit it), but if it's checking for membership of a numeric value in a union like that, it will throw an error.

annevk:
Right, so I'm saying we should try to prevent such overloading and unions (perhaps with a prose restriction) as otherwise they'll be used to do the kind of thing we don't want people to be doing.

littledan:
I see, sounds like we should change the Intl patch if we want to go this way.

littledan:
We discussed the overloading issue in TC39 and concluded that `Intl.NumberFormat.prototype.format` *should* overload between Number and BigInt (based on calling ToNumeric and then checking whether a Number or BigInt came out). So, we may want to permit that here, or we could still prohibit it. Any preferences? Based on a ping from @cynthia about TAG interest, I'd like to conclude on these issues land this PR soon.

domenic:
It seems like it would be easier to judge if there were web APIs wanting to use bigint. Are there any currently?

annevk:
Well, `Intl.NumberFormat.prototype.format`, per "jsidl".

We also have a case that shows that allowing a union/overload will likely lead to APIs with implicit coercion of Number to BigInt, which is something TC39 does not want, as far as I know: https://github.com/WebAssembly/JS-BigInt-integration/issues/12.

I don't think there's a way to enforce the difference through IDL. I guess you could try something along the lines of that when there's a union, the corresponding algorithm cannot contain a Number to BigInt operation, but try enforcing that realistically …

Ms2ger:
Looking into this some more, adding support for unions that include bigint and a numeric type (with the "call `ToNumeric` and see what falls out" semantics) is fairly straightforward. Adding support for overloading is significantly more hairy, especially when involving unions (such as in the example below). The only real use case I've seen so far is [`NumberFormat`](https://github.com/tc39/ecma402/pull/236), and unions are sufficient there. My inclination would be not to allow overloading at this point; I'm fine going either way for unions – we could discuss again if jsidl ever happens.

```idl
interface X {
  void foo((bigint or Node) x);
  void foo(long x);
};
```

Thoughts?

annevk:
I'm okay with allowing unions, with a warning to ask for advice (prolly filing an issue against this repository) and a high-level outline of the principles when it is and when it is not appropriate.

Ms2ger:
Updated some more; we can look into renaming "numeric types" separately.

index.bs
```diff
@@ -6529,6 +6579,19 @@ Each pair of [=flattened member types=]
...
+It is possible to create a union of {{bigint}} and a [=numeric type=].
+However, this is generally only supposed to be used for interfaces such as
+[[ECMA-402#numberformat-objects|NumberFormat]], which formats the values rather than using them in
+calculations.
+It would not be appropriate to accept such a union, only to then convert values of the
+[=numeric type=] to a {{bigint}} for further processing, as this runs the risk of introducing
+precision errors.
+Please
+<a href="https://github.com/heycam/webidl/issues/new?title=Intent%20to%20use%20bigint/numeric union">file an issue</a>
+to discuss before using this feature.
```
```suggestion
before using this feature.
```

index.bs
```diff
@@ -5768,6 +5807,7 @@ type.
         "boolean"
         "byte"
         "octet"
+        "bigint"
```
bathos: The addition of this new unnamed terminal means the `Other` production needs to be updated to include that terminal here https://github.com/heycam/webidl/blob/c387fed823d2d4ccf8940bbc065f358ccf563073/index.bs#L6998-L6999

(Unnamed/quoted terminals ‘delete the token’ from the set of values represented by a named terminal, so _ident_ is currently inclusive of the string "bigint," but will cease to be, hence the need to explicitly add it to Other, which is inclusive of every terminal except for a few specific punctuation marks.)

index.bs
```diff
@@ -14555,6 +14682,11 @@ expression syntax [[!PERLRE]]) as follows:
         <td><code>=</code></td>
         <td><code class="regex"><span class="mute">/</span>-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][+-]?[0-9]+)?|[0-9]+[Ee][+-]?[0-9]+)<span class="mute">/</span></code></td>
     </tr>
+    <tr>
+        <td id="prod-bigint"><emu-t class="regex">bigint</emu-t></td>
+        <td><code>=</code></td>
+        <td><code class="regex"><span class="mute">/</span>-?([1-9][0-9]*|0[Xx][0-9A-Fa-f]+|0[0-7]*)n<span class="mute">/</span></code></td>
```
bathos: The combined lexical goal can currently be realized case insensitively (assuming one handles unnamed terminals in a lookup step). That will cease to be true after this change because this "n" is case sensitive. This isn’t a problem, but I’m mentioning it because it could plausibly impact existing Web IDL lexers if they’ve made certain assumptions.

index.bs
```diff
@@ -5892,6 +5932,16 @@ The [=type name=] of the
 {{octet}} type is "<code>Octet</code>".
 
 
+<h4 id="idl-bigint" interface>bigint</h4>
```
bathos: This is placed between `octet` and `short` here in the Types list. Later in the ECMAScript type mapping section it’s placed between `unrestricted double` and `DOMString`. These lists currently use the same ordering and I think that should probably be preserved since it makes things a bit easier to follow. The latter positioning makes sense. Slotting it between two of the sized integer types seems strange:

- ± 8
- \+ 8
- ± ∞
- ± 16
- \+ 16
- ± 32
- \+ 32
- ± 64
- \+ 64

index.bs
Ms2ger: I don't think we actually need a bigint token; integer works fine. I did that instead.

Ms2ger:
I think I addressed the remaining comments. @bzbarsky, would you mind reviewing?

bzbarsky:
I will try to get to this, but there's a lot to page in and I am pretty swamped with other things at the moment.  I can't promise any timeline other than "not in the next two days"...

annevk:
So IDB doesn't depend on this as far as I can tell and it's not clear that's moving forward anyway. Are there other places that might use this through IDL? I'd rather not add it until we have a concrete consumer I think.

ExE-Boss:
Well, **WebAssembly** intends to use `bigint` for `i64` handling in **JavaScript** code: <https://github.com/WebAssembly/spec/pull/707>.

Currently, the `value` attribute type of `WebAssembly.Global` is `any`, because no other **WebIDL** type currently allows **bigint** values.

Ms2ger:
I guess the way WebAssembly uses `any` is problematic wrt the formal definition of `any` as a discriminated union, and fixing this would require a bigint type in IDL. Note future changes in the [reference-types](https://webassembly.github.io/reference-types/js-api/index.html#tojsvalue) proposal will extend `WebAssembly.Global` with a bunch of non-integer types.

annevk:
Ah, so that would be a problem for `postMessage()` and such too then? That does seem like something we ought to fix. (The weird thing with WebAssembly is that browsers don't seem to use IDL to implement it.)

littledan:
Right, if we don't want to permit bigint as an actual IDL type, we should make certain changes to WebIDL to let it "pass through". I've been hoping that this patch would land, so I didn't formulate another one for that subset. (I think WebIDL is just sort of incoherent right now, with BigInt existing as a type in JS.)

I've heard from a few people in the TAG that they are interested in this patch landing. Comments here about why would be appreciated. @cynthia @kenchris

If we *conclude* that this patch won't land, then let's formulate that more minimal patch and land that instead.

rtoy:
It would make the [API for a WebAudio noise generator](https://github.com/WebAudio/web-audio-api-v2/issues/8) a bit nicer if BigInt were supported.  We currently using an array of two 32-bit unsigned ints to specify the 64-bit seed we need.

domenic:
@rtoy the only thing this PR was waiting on was an actual consumer, so if you're providing one, then I think we should go ahead and merge this. @annevk?

annevk:
That seems reasonable. And it doesn't seem to want to use overloading so we don't have to address that for now.

rtoy:
Just to follow up, see https://github.com/WebAudio/web-audio-api-v2/issues/8#issuecomment-721439274 and https://github.com/WebAudio/web-audio-api-v2/issues/8#issuecomment-721803254.  Using BigInt makes the API a bit nicer, but an array of 2 32-bit unsigned integers works fine too.

domenic (approved):
I just did a re-review and this looks good to me. @TimothyGu, would you be able to give it a double-check and/or merge?

TimothyGu (approved):

TimothyGu:
CI seems to be failing due to infrastructural reasons.

PR #526, opened 2018-02-22
Add [Exposed=*] to expose on all globals

### LINKED ISSUE #468: Exposing an interface "on all globals"
For WebAssembly, I'm using WebIDL to specify the interface between WebAssembly and JavaScript [in this PR](https://github.com/WebAssembly/spec/pull/591). We're trying to avoid tying this into the Web Platform. One of the few ways that the current spec does tie into the Web is, to expose the interface to global objects, it uses `[Exposed=(Window,Worker)]`.

What we really want for Wasm is for it to be exposed on all global objects. One aspect was accidentally omitting it on Worklets, but the bigger aspect is omitting it for everything that's not the web platform. For example, Node.js is likely to eventually enable Wasm support, and so the interface should be created on its global object as well.

I imagine that a version of Exposed which is everywhere might be useful for the Web Platform as well. APIs like TextEncoder which don't do any particular I/O wouldn't need a change each time another type of global object is added--they could be like new libraries added in the JavaScript standard.

What would you think about an additional extended attribute, or mode of using `[Exposed]`, to support this case?

littledan:
This patch adds support in WebIDL for declaring an interface available
in all contexts.

Closes #468 

This PR follows the syntax and outline suggested by @tobie . Thanks for your clear instructions here.

***
<a href="https://pr-preview.s3.amazonaws.com/littledan/webidl/pull/526.html" title="Last updated on Feb 23, 2018, 2:43 PM GMT (2e23cfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/526/3750d7a...littledan:2e23cfb.html" title="Last updated on Feb 23, 2018, 2:43 PM GMT (2e23cfb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7002,6 +7002,17 @@ which form (or forms) it is in:
             <code>[Exposed=(Window,Worker)]</code>
         </td>
     </tr>
+    <tr>
+        <td>
+            <emu-nt><a href="#prod-ExtendedAttributeWildcard">ExtendedAttributeWildcard</a></emu-nt>
+        </td>
+        <td>
+            <dfn id="dfn-xattr-wildcard" for="extended attribute" export>takes a wildcard</dfn>
+        </td>
+        <td>
+            <code>[Exposed=*]</code>
+        </td>
+    </tr>
 
 </table>
 
@@ -7063,6 +7074,7 @@ five forms are allowed.
         "="
         "&gt;"
         "?"
+        "*"
         "ByteString"
         "DOMString"
         "FrozenArray"
@@ -7125,6 +7137,11 @@ five forms are allowed.
 <pre class="grammar" id="prod-ExtendedAttributeIdent">
     ExtendedAttributeIdent :
         identifier "=" identifier
+</pre>
+
+<pre class="grammar" id="prod-ExtendedAttributeWildcard">
+    ExtendedAttributeWildcard :
+        identifier "=" "*"
 </pre>
 
 <pre class="grammar" id="prod-ExtendedAttributeIdentList">
@@ -9588,14 +9605,30 @@ it indicates that the construct is exposed
 on that particular set of global interfaces.
 
 The [{{Exposed}}] [=extended attribute=] must either
-[=takes an identifier|take an identifier=] or
-[=takes an identifier list|take an identifier list=].
+[=takes an identifier|take an identifier=],
+[=takes an identifier list|take an identifier list=] or
+[=takes a wildcard|take a wildcard=].
 Each of the identifiers mentioned must be a [=global name=] and be unique.
-This list of identifiers is known as the construct's
-<dfn>own exposure set</dfn>.
+
+The <dfn>own exposure set</dfn> is the set of [=global name=]s which
+[{{Exposed}}] is applied to.
+
+If the [{{Exposed}}] [=extended attribute=]
+[=takes an identifier|take an identifier=] or
+[=takes an identifier list|take an identifier list=],
+the [=own exposure set=] is the list of identifiers.
+If the [{{Exposed}}] [=extended attribute=]
+[=takes a wildcard|take a wildcard=], the [=own exposure set=] is the
+set of interfaces which have the [{{Global}}] extended attribute.
+
+<p class="advisement">
+    <code>[Exposed=*]</code> should be used with care. It is only appropriate
+    when an interface does not expose significant new capabilities. If the
+    interface may be restricted or disabled in some environments, it is
+    recommended to list the globals explicitly.
+</p>
 
 <div algorithm>
-
     To get the <dfn id="dfn-exposure-set" export>exposure set</dfn> of a construct |C|,
     run the following steps:
 
--- a/index.bs
+++ b/index.bs
@@ -9608,24 +9608,25 @@ The [{{Exposed}}] [=extended attribute=] must either
 [=takes an identifier|take an identifier=],
 [=takes an identifier list|take an identifier list=] or
 [=takes a wildcard|take a wildcard=].
-Each of the identifiers mentioned must be a [=global name=] and be unique.
+Each of the identifiers mentioned must be a [=global name=] of some [=interface=] and be unique.
 
-The <dfn>own exposure set</dfn> is the set of [=global name=]s which
-[{{Exposed}}] is applied to.
+The <dfn>own exposure set</dfn> is the [=/set=] of identifiers defined as follows:
 
-If the [{{Exposed}}] [=extended attribute=]
-[=takes an identifier|take an identifier=] or
-[=takes an identifier list|take an identifier list=],
-the [=own exposure set=] is the list of identifiers.
-If the [{{Exposed}}] [=extended attribute=]
-[=takes a wildcard|take a wildcard=], the [=own exposure set=] is the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5963,6 +5963,17 @@ which form (or forms) it is in:
             <code>[Exposed=(Window,Worker)]</code>
         </td>
     </tr>
+    <tr>
+        <td>
+            <emu-nt><a href="#prod-ExtendedAttributeIdentList">ExtendedAttributeWildcard</a></emu-nt>
```
tobie: I think you want to fix the href here.

index.bs
```diff
@@ -7997,9 +8015,18 @@ defining which global environments the construct can be used in.
 The [=exposure set=]
 for a given construct is defined as follows:
 
-*   If the [{{Exposed}}]
-    [=extended attribute=]
-    is specified on the construct, then the [=exposure set=]
+*   If the [{{Exposed}}] [=extended attribute=] [=takes a wildcard=], then the
```
tobie: So unfortunately this bit has seen a lot of changes since you opened the initial bug report. I _think_ you should just be able to describe the **own exposure set** of constructs declared with [Exposed=*] as containing all of the interfaces which have a [{{Global}}] [=extended attribute=] and be done with it.

index.bs
littledan: Fixed

index.bs
littledan: Not sure how I forgot to rebase before putting this patch up. Fixed. What do you think of the new wording?

annevk:
I think before landing this we should try to address the question at https://github.com/heycam/webidl/issues/468#issuecomment-338774080.

index.bs
```diff
@@ -8560,13 +8577,33 @@ it indicates that the construct is exposed
...
+The <dfn>own exposure set</dfn> is the set of [=global name=]s which
+[{{Exposed}}] is applied to.
+
+If the [{{Exposed}}] [=extended attribute=]
+[=takes an identifier|take an identifier=] or
+[=takes an identifier list|take an identifier list=],
+the [=own exposure set=] is the list of identifiers.
+If the [{{Exposed}}] [=extended attribute=]
+[=takes a wildcard|take a wildcard=], the [=own exposure set=] is the
+the set of interfaces which have the [{{Global}}] extended attribute.
```
tobie: There's an extra "the" here.

index.bs
```diff
@@ -8560,13 +8577,33 @@ it indicates that the construct is exposed
...
+    <code>[Exposed=*]</code> should be used with care. It is only appropriate
+    when an interface does not expose significant new capabilities. If the
+    interface may be restricted or disabled in some environments, it is
+    recommended to list the globals explicitly.
+</p>
 
 <div algorithm>
+*   If the [{{Exposed}}] [=extended attribute=] [=takes a wildcard=], then the
+    [=exposure set=] is the set of all interfaces which have a [{{Global}}]
+    or [{{PrimaryGlobal}}] [=extended attribute=].
```
tobie: Is this bit necessary? Isn't that all handled by the below algorithm already?

index.bs
littledan: Fixed.

index.bs
littledan: Sorry, I just had this copied in there for my own use when writing the patch and forgot to remove it. Removed now.

littledan:
@annevk Thanks for making this link. I think we should work out an answer in conjunction with TC39 to this question, given that ECMAScript includes things that are pretty equivalent in power/appropriateness to put everywhere to TextEncoder and URL. I'm fine with letting this PR be blocked on this question.

littledan:
We should include the jsidl label on this issue, since all ES APIs will need to be exposed to all globals

leobalter:
@littledan I believe we should follow up on this PR for the proper ShadowRealms integration, right?

littledan:
We could add this for ShadowRealms, or we could just add `[Exposed=ShadowRealm]`. The choice depends on whether we want to go around adding lots of things to Worklets which are not currently exposed, and which we might decide to expose on ShadowRealm. In offline conversations with different HTML editors, I heard mixed opinions on this question, cc @domenic @annevk . Given that, I thought that it might be better to start with the more conservative approach of just adding things specifically to ShadowRealm.

leobalter:
it sounds like starting with just `[Exposed=ShadowRealm]` seems like a good ground work for further analysis to be used on follow up work.

domenic:
I think it'd be better to add `[Exposed=*]`? Saying that in worklets you have to access these APIs indirectly via shadow realms seems strange.

annevk:
In an offline discussion https://github.com/tc39/ecma262/issues/1120 came up, but since we did not change JavaScript to support minimal realms (i.e., a subset of what is defined in JavaScript), we should not do so here by excluding worklets. (As that would make worklets the new minimal realms due to organizational boundaries.)

As I understand it the primary motivation for having minimal realms in worklets was to avoid steering web developers down the wrong path, but in practice this hasn't been much of a problem.

This is a long way of saying that I agree with @domenic, but I thought it might be useful to add some context.

cc @padenot @codehag @Ms2ger

littledan:
OK, thanks for the clarifications everyone, sounds like I was the confused/misunderstanding one. Does that mean we should move ahead with [Exposed=*]?

Ms2ger:
Rebased and rephrased a bit - what's next here?

index.bs
```diff
@@ -9588,14 +9605,31 @@ it indicates that the construct is exposed
...
+The <dfn>own exposure set</dfn> is the [=/set=] of identifiers defined as follows:
+
+<dl class="switch">
+     :  If the [{{Exposed}}] [=extended attribute=] [=takes an identifier=] |I|
+     :: The [=own exposure set=] is the [=/set=] « |I| ».
+     :  If the [{{Exposed}}] [=extended attribute=] [=takes an identifier list=] |I|
+     :: The [=own exposure set=] is the [=/set=] |I|.
+     :  If the [{{Exposed}}] [=extended attribute=] [=takes a wildcard=]
+     :: The [=own exposure set=] is the [=/set=] of [=global names=] of [=interfaces=] which have the
+        [{{Global}}] [=extended attribute=].
```
ExE-Boss, line 9622: This should probably include the word “all”:
```suggestion
     :: The [=own exposure set=] is the [=/set=] of all [=global names=] of all [=interfaces=]
        which have the [{{Global}}] [=extended attribute=].
```

EdgarChen:
@Ms2ger it would be helpful if you could make a list of interfaces which `[Exposed=*]` is proposed somewhere publicly. Thanks!

TimothyGu (approved):
Looks good!

Ms2ger:
> @Ms2ger it would be helpful if you could make a list of interfaces which `[Exposed=*]` is proposed somewhere publicly. Thanks!

Besides the ones already listed in https://github.com/tc39/proposal-shadowrealm/issues/331, we might propose some parts of Crypto and Performance - please do let me know if you'd like me to present the list in some other way

littledan:
This change looks great, but I don't think it handles the ShadowRealm case yet, since that is not an interface.

annevk:
@littledan that's a good point. I suppose we should put that requirement in HTML along with the overall ShadowRealm integration?

PR #562, opened 2018-05-30
Disallow multiple keywords for SpecialOperation

### LINKED ISSUE #470: Can SpecialOperation be more simplified?
Currently [SpecialOperation](https://heycam.github.io/webidl/#prod-SpecialOperation) allows multiple keywords before RegularOperation:

```
SpecialOperation ::
    Special Specials RegularOperation
Specials ::
    Special Specials
    ε
Special ::
    getter
    setter
    deleter
```

This was required before #44 to allow `setter creator void set(DOMString name, any value);` but now it seems no IDL uses `getter setter`, `setter deleter`, etc.

So I think SpecialOperation can just be `Special RegularOperation`, what do you think?

saschanaz:
Closes #470 

This was to allow `setter creator` but is not useful anymore.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2371,13 +2371,7 @@ The following extended attributes are applicable to operations:
 
 <pre class="grammar" id="prod-SpecialOperation">
     SpecialOperation :
-        Special Specials RegularOperation
-</pre>
-
-<pre class="grammar" id="prod-Specials">
-    Specials :
-        Special Specials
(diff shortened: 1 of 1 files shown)


saschanaz:
#560 also has the same CI error, not sure what's wrong.

tobie:
Yeah, that's very weird. Need to investigate.

Issue #555, opened 2018-05-17
Is a dictionary which has an inherited dictionary with a non-JSON-type member a JSON type?


bzbarsky:
Consider this IDL:

    dictionary A {
      Node node;
    };

    dictionary B : A {
      DOMString str;
    };

Is `B` a JSON type?  The spec says "dictionaries where all of their members are JSON types" are JSON types, and all the members of B itself are JSON types...

I suspect this should be clarified to include the members of all the inherited dictionaries.

tabatkins:
Definitely, this seems like a straightforward accidental omission.

PR #571, opened 2018-07-12
A dictionary inheriting from a non-JSON type can't be a JSON type.

bzbarsky:
Fixes https://github.com/heycam/webidl/issues/555

***
<a href="https://pr-preview.s3.amazonaws.com/bzbarsky/webidl/pull/571.html" title="Last updated on Jul 16, 2018, 11:31 PM GMT (ec1f2c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/571/ebdaa2b...bzbarsky:ec1f2c9.html" title="Last updated on Jul 16, 2018, 11:31 PM GMT (ec1f2c9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2460,7 +2460,7 @@ The <dfn id="dfn-json-types" export lt="JSON type">JSON types</dfn> are:
 *   [=typedefs=] whose [=type being given a new name=] is a [=JSON type=],
 *   [=sequence types=] whose parameterized type is a [=JSON type=],
 *   [=frozen array types=] whose parameterized type is a [=JSON type=],
-*   [=dictionaries=] where all of their [=dictionary members|members=] are [=JSON types=],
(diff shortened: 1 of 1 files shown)


bzbarsky:
Hmm.  So it seems that part of the problem is that some places say "dictionary member" to refer to the syntactic construct (sometimes using "dictionary member declared on _D_" where _D_ is some dictionary) and some places say "dictionary member" to talk about the map entries for the map represented by a dictionary.  The former usage definitely excludes inherited members; otherwise https://heycam.github.io/webidl/#es-to-dictionary would not work right. 

https://heycam.github.io/webidl/#dfn-json-types is playing fast and  loose with this all by  using https://heycam.github.io/webidl/#dfn-dictionary when it should really be using <https://heycam.github.io/webidl/#idl-dictionary>, because it's being defined as an operation on types, not actual objects.

Given that, I think the right fix is to explicitly work with dictionary types, like https://heycam.github.io/webidl/#es-to-dictionary does.

bzbarsky:
@domenic @tobie How do you feel about the updated wording?

Issue #540, opened 2018-04-05
Is a subclass still a platform object?


annevk:
Related a bit to #533.

I'm wondering if a subclass of, .e.g., `EventTarget`, would still be considered a platform object. Otherwise we need to expand the definition of "relevant global object" and potentially other such concepts.

And even then we might need to clarify things here, e.g.,
```html
<iframe></iframe>
...<script>
class X extends frames[0].EventTarget {};
</script>
```

annevk:
FWIW, it seems that at least Chrome shifts the relevant global object in that case, see:
```html
.<iframe></iframe>
<script>
class X extends frames[0].EventTarget {};
obj = new X();
obj.addEventListener("hi", () => w(window.event));
obj.dispatchEvent(new Event("hi"));
</script>
```

annevk:
Yeah, Henri said that's what happened with his patch in https://bugzilla.mozilla.org/show_bug.cgi?id=218415. For Chrome it's window though. 😟

bzbarsky:
Hrm.  In the end you're executing `frames[0].EventTarget` and that's defined to create objects in its own global...  The WebIDL spec used to clearly say that when you create things via a constructor they are associated with that constructor's global, but I don't see that text in there now.  @tobie, @domenic, did it get lost in one of the refactorings or something?

Anyway, here's a less-magical testcase for this:

    <iframe></iframe>
    <script>
    class X extends frames[0].Image {};
    obj = new X();
    w(obj.ownerDocument == document);
    w(obj.ownerDocument == frames[0].document);
    </script>

(live at <http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=5856>).  And yes, Firefox and Chrome disagree here.  Firefox logs `false` then `true`, while Chrome logs `true` then `false`.

Safari logs `false` then `true` (but also doesn't create the object with X.prototype as the proto, of course).  Edge behaves like Safari.

Given that, sounds like Chrome is the odd one out in terms of what it does here....

bzbarsky:
Oh, also, even in today's spec https://heycam.github.io/webidl/#interface-object creation steps step 11 asserts that the newly-created thing's Realm is the realm of the constructor, not of the NewTarget.  Which is an assert Chrome would fail.

Looks like that assert is not clearly supported by any normative requirements, though.  Those normative requirements used to exist, but were removed in <https://github.com/heycam/webidl/commit/4a8f57e2> for some reason.  Before that, there were clear statements like:

[...]

Issue #258, opened 2016-12-18
Adding slots [initial braindump]


tobie:
I'd like to tackle adding slots to WebIDL soonish.

After some preliminary discussions in #whatwg on irc, it seems we want to start with a lightweight system and see how that goes first.

### What values can slots contain?

It seems slots should be able to indifferently hold WebIDL types, Infra [primitive data types](https://infra.spec.whatwg.org/#primitive-data-types) and [data structures](https://infra.spec.whatwg.org/#data-structures), and flags (boolean thingies that are either set or unset).

### Type coercion

We want to avoid type coercion. Editors need to be explicit about what values slots hold and make sure not to rely on type coercion for those.

Should we disallow type unions in slots?

### Syntax for defining slots

Should we define slots in WebIDL fragments? e.g.:

```webidl
interface Animal {
  slot children;
  readonly attribute DOMString name;
  attribute unsigned short age;
};
```

If so, what syntax should we use?

Should we include slot type in there? If so, how do we handle infra "types"?

How do we deal with inheritance?

Should we instead consider using tables for slots like the [Streams spec does](https://streams.spec.whatwg.org/#default-reader-internal-slots)?

### Syntax for using slots

We'd want to be able to use slots in specs like we use ES6 ones:

```
1.  If object.[[flag]] is unset, then:
    1.  Set object.[[state]] to "errored",
    1.  Etc.
```

### Bindings

Slots are private so no bindings. Is that worth being explicit about?

### Other

This is the first time I tackle adding a large feature to WebIDL, so I'm pretty sure I haven't considered all the things. Comments, thoughts, questions, etc. are most welcome.

annevk:
Prior discussion: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27354.

Given that we don't have a unified type system we should probably not block slots on that. So an initial version would not have explicit types (all done through prose).

It does seem good to define slots for an interface in a formalized way and have syntax for getting at them. We'll need to allow for mixins/partial too. Having a way to address "this" should probably come first. Being able to do "this.[[slot]]" would be enormously helpful. I suspect defining slots in IDL would be the most natural as you need to know about them when putting together an object.

[...]

Issue #135, opened 2016-07-19
State that by default all objects are created in the relevant realm of `this`


domenic:
This has come up recently in https://github.com/w3c/webcrypto/issues/85 and also in some internal work Blink is doing on refactoring their bindings layer.

It appears that in all browsers, document.createElement, innerHTML, and an Event created through button.click() are created in the relevant realm of `this`. All of these cases are currently unspecified and use spec text equivalent to "create a new X object".

We should specify a few things:

- If an IDL method or attribute is currently executing, "a new X object" means creating it in the relevant realm of `this`
- If an IDL constructor is currently executing, "a new X object" means creating it in the current realm
- If neither of these conditions hold, e.g. in steps that are being executed in parallel or in tasks that are posted to the event loop, "a new X object" *must* be accompanied by a specification of what realm to create the object in.

(At first I thought that we should follow the ES spec and use the current realm for all cases. That would better match e.g. `Array.prototype.map` or `Promise.prototype.then`. But it seems like that's not what browsers do, so we'll just have to live with the inconsistency.)

annevk:
This looks like a duplicate of https://www.w3.org/Bugs/Public/show_bug.cgi?id=24652.

annevk:
See also https://github.com/w3c/web-platform-tests/pull/1381 for tests which I'm about to land.

jyasskin:
Can we provide a default for "in parallel" blocks created from IDL methods and constructors too? We'd want to capture the relevant or current realm, respectively, at the point where the "in parallel" block starts, and use it as the "object-creation realm" within the block. Tasks posted from a context with an "object-creation realm" could inherit the same realm.

We still have some cases without an "object-creation realm", like https://webbluetoothcg.github.io/web-bluetooth/#notification-events. These tasks actually have an obvious realm to use, but it's lost when we pull the event loop out of the realm's settings object. We could extend "queue a task" to take its object-creation realm in order to have a default here too.

domenic:
That seems not great to me. When you're in a background thread, you should be very explicit about what realm you're creating objects in. Just inheriting a default is not going to work out well in general, I think.

IMO it's better to just state the realm when creating the object, and rely on ambient defaults as little as possible---only for the most obvious of cases. We should not be trying to create more situations that have ambient defaults available in non-obvious ways.

jyasskin:
Given:
1. Let |x| be a new object.
2. In parallel:
   1. Let |y| be a new object.

[...]

PR #595, opened 2018-12-09
Define "implements" checks using internal slots

### LINKED ISSUE #97: Add [[PlatformBrand]] internal slot to all platform objects

littledan:
- Define an imperative algorithm for instantiating a platform object,
  which everything theoretically calls into, and which sets the primary
  interface into a new internal slot [[PrimaryInterface]].
- Define the "implements" algorithm as checking the [[PrimaryInterface]]
  internal slot.

The algorithm here is roughly in correspondence with the V8 implementation,
where operation functions are allocated with a reference to a
FunctionTemplate object (which is in corresondence with WebIDL interfaces).
When checking the receiver of a method that came from WebIDL, the original
prototype chain is traversed by looking at the FunctionTemplate's parent.
The same FunctionTemplate is used in multiple JavaScript realms.

This PR procrastinates on defining an imperative algorithm for [Global]
interfaces, which would be a bit more text. It also doesn't define which
global object is associated with platform objects, or how this association
is represented.

Fixes #97

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7427,7 +7427,7 @@ values are represented by ECMAScript Object values (including [=function objects
     to an IDL [=interface type=] value by running the following algorithm (where |I| is the [=interface=]):
 
     1.  If <a abstract-op>Type</a>(|V|) is not Object, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  If |V| is a [=platform object=] that implements |I|, then return the IDL [=interface type=] value that represents a reference to that platform object.
+    1.  If |V| [=implements=] |I|, then return the IDL [=interface type=] value that represents a reference to that platform object.
     1.  If |V| is a [=user object=]
         that is considered to implement |I| according to the rules in [[#es-user-objects]], then return the IDL [=interface type=] value that represents a reference to that
         user object, with the [=incumbent settings object=] as the [=callback context=].
@@ -7441,7 +7441,6 @@ values are represented by ECMAScript Object values (including [=function objects
     value that represents a reference to the same object that the IDL
     [=interface type=] value represents.
 </p>
-
 
 <h4 id="es-dictionary">Dictionary types</h4>
 
@@ -7926,9 +7925,9 @@ that correspond to the union’s [=member types=].
     1.  If |V| is <emu-val>null</emu-val> or <emu-val>undefined</emu-val>, then:
         1.  If |types| includes a [=dictionary type=], then return the
             result of [=converted to an IDL value|converting=] |V| to that dictionary type.
-    1.  If |V| is a [=platform object=], then:
+    1.  If |V| [=is a platform object=], then:
         1.  If |types| includes an [=interface type=] that |V|
-            implements, then return the IDL value that is a reference to the object |V|.
+            [=implements=], then return the IDL value that is a reference to the object |V|.
         1.  If |types| includes {{object}}, then return the IDL value
             that is a reference to the object |V|.
     1.  If |V| is a {{DOMException}} platform object, then:
@@ -8352,7 +8351,7 @@ extended attribute must either
 The bare form, <code>[Constructor]</code>, has the same meaning as
 using an empty argument list, <code>[Constructor()]</code>.  For each
 [{{Constructor}}] extended attribute
-on the interface, there will be a way to construct an object that implements
+on the interface, there will be a way to construct an object that [=implements=]
 the interface by passing the specified arguments.
 
 The prose definition of a constructor must
@@ -8397,7 +8396,7 @@ for an interface is to be implemented.
     An ECMAScript implementation supporting these interfaces would
     have a \[[Construct]] property on the
     <code class="idl">Circle</code> interface object which would
-    return a new object that implements the interface.  It would take
+    return a new object that [=implements=] the interface.  It would take
     either zero or one argument.  The
     <code class="idl">NodeList</code> interface object would not
     have a \[[Construct]] property.
@@ -9212,7 +9211,7 @@ If the [{{LenientThis}}]
 appears on a [=regular attribute=],
 it indicates that invocations of the attribute’s getter or setter
 with a <emu-val>this</emu-val> value that is not an
-object that implements the [=interface=]
+object that [=implements=] the [=interface=]
 on which the attribute appears will be ignored.
 
 The [{{LenientThis}}] extended attribute
@@ -9307,7 +9306,7 @@ has the same meaning as using an empty argument list,
 <code>[NamedConstructor=<emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t>()]</code>.
 For each [{{NamedConstructor}}] extended attribute on the interface,
 there will be a way to construct an object that
-implements the interface by passing the specified arguments to the [=constructor=]
+[=implements=] the interface by passing the specified arguments to the [=constructor=]
 that is the value of the aforementioned property.
 
 The [=NamedConstructor identifier|identifier=] used for the named constructor must not
@@ -10244,7 +10243,7 @@ If the [{{Unscopable}}]
 [=extended attribute=]
 appears on a [=regular attribute=]
 or [=regular operation=], it
-indicates that an object that implements an interface with the given
+indicates that an object that [=implements=] an interface with the given
 interface member will not include its property name in any object
 environment record with it as its base object.  The result of this is
 that bare identifiers matching the property name will not resolve to
@@ -10372,9 +10371,9 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
             then remove from |S| all other entries.
 
-        1.  Otherwise: if |V| is a [=platform object=], and
+        1.  Otherwise: if |V| [=is a platform object=], and
             there is an entry in |S| that has one of the following types at position |i| of its type list,
-            *   an [=interface type=] that |V| implements
+            *   an [=interface type=] that |V| [=implements=]
             *   {{object}}
             *   a [=nullable type|nullable=] version of any of the above types
             *   an [=annotated type=] whose [=annotated types/inner type=] is one of the above types
@@ -10660,7 +10659,7 @@ as described in sections [[#es-constants]] and [[#es-operations]].
 
 If the [=interface=] is declared with a [{{Constructor}}] [=extended attribute=],
 then the [=interface object=] can be called as a [=constructor=]
-to create an object that implements that interface.
+to create an object that [=implements=] that interface.
 Calling that interface as a function will throw an exception.
 
 [=Interface objects=] whose [=interfaces=] are not declared
@@ -10702,7 +10701,7 @@ the <code>typeof</code> operator will return "function" when applied to an inter
             Rethrow any exceptions.
         1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |R|
             to an ECMAScript [=interface type=] value |I|.
-        1.  Assert: |O| is an object that implements |I|.
+        1.  Assert: |O| is an object that [=implements=] |I|.
         1.  Assert: |O|.\[[Realm]] is equal to |realm|.
         1.  Return |O|.
     1.  Let |constructorProto| be |realm|.\[[Intrinsics]].[[{{%FunctionPrototype%}}]].
@@ -10738,7 +10737,7 @@ implement the interface on which the
 [{{NamedConstructor}}] extended attributes appear.
 
 If the actions listed in the description of the constructor return normally,
-then those steps must return an object that implements interface |I|.
+then those steps must return an object that [=implements=] interface |I|.
 This object's relevant [=Realm=] must be the same as that of the [=named constructor=].
 
 <div algorithm>
@@ -10955,7 +10954,7 @@ is the concatenation of the [=interface=]’s
     is called with property key |P|, the following steps are taken:
 
     1.  Let |A| be the [=interface=] for the [=named properties object=] |O|.
-    1.  Let |object| be the sole object from |O|’s ECMAScript global environment that implements |A|.
+    1.  Let |object| be the sole object from |O|’s ECMAScript global environment that [=implements=] |A|.
 
         Note: For example, if the [=interface=] is the {{Window}} interface,
         then the sole object will be this global environment’s window object.
@@ -10972,7 +10971,7 @@ is the concatenation of the [=interface=]’s
         1.  Let |desc| be a newly created [=Property Descriptor=] with no fields.
         1.  Set |desc|.\[[Value]] to the result of [=converted to an ECMAScript value|converting=]
             |value| to an ECMAScript value.
-        1.  If |A| implements an interface with the
+        1.  If |A| [=implements=] an interface with the
             [{{LegacyUnenumerableNamedProperties}}]
             [=extended attribute=],
             then set |desc|.\[[Enumerable]] to <emu-val>false</emu-val>,
@@ -11036,7 +11035,7 @@ is the concatenation of the [=interface=]’s
 [=Constants=] are exposed on [=interface objects=],
 [=legacy callback interface objects=],
 [=interface prototype objects=], and
-on the single object that implements the interface,
+on the single object that [=implements=] the interface,
 when an interface is declared with the [{{Global}}] [=extended attribute=].
 
 <div algorithm>
@@ -11061,7 +11060,7 @@ when an interface is declared with the [{{Global}}] [=extended attribute=].
 [=Regular attributes=] are exposed on the [=interface prototype object=],
 unless the attribute is [=unforgeable=] or
 if the interface was declared with the [{{Global}}] [=extended attribute=],
-in which case they are exposed on every object that implements the interface.
+in which case they are exposed on every object that [=implements=] the interface.
 
 <div algorithm>
     To <dfn>define the regular attributes</dfn> of [=interface=] or [=namespace=] |definition| on |target|,
@@ -11126,9 +11125,9 @@ in which case they are exposed on every object that implements the interface.
                     specified.)
                     <!-- https://www.w3.org/Bugs/Public/show_bug.cgi?id=18547#c9 -->
                 1.  Otherwise, set |O| to the <emu-val>this</emu-val> value.
-                1.  If |O| is a [=platform object=], then [=perform a security check=], passing |O|,
+                1.  If |O| [=is a platform object=], then [=perform a security check=], passing |O|,
                     |attribute|'s [=identifier=], and "getter".
-                1.  If |O| is not a [=platform object=] that implements the interface |target|,
+                1.  If |O| is not a [=platform object=] that [=implements=] the interface |target|,
                     then:
                     1.  If |attribute| was specified with the [{{LenientThis}}]
                         [=extended attribute=], then return <emu-val>undefined</emu-val>.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12337,18 +12336,73 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    An ECMAScript value |value| <dfn id="es-is-platform-object">is a platform object</dfn>
+    if [$Type$](|value|) is Object and if |value| has a \[[PrimaryInterface]] internal slot.
+</div>
+
+<div algorithm>
+    An ECMAScript value |value| <dfn export>implements</dfn> an [=interface=] |interface|
+    if |value| [=is a platform object=] and one of the following is true:
+    <ul>
+        <li>|value|.\[[PrimaryInterface]] is |interface|.</li>
+        <li>The set of [=inherited interfaces=] of |value|.\[[PrimaryInterface]] contains |interface|.</li>
```
annevk: Theoretically it'd be better if these inherited interfaces were stored directly on the object, no?

littledan: Why's that? I thought it'd be better if the data structures were "normalized"--this makes it more clear to implementers that they don't need to denormalize either.

annevk: Because "inherited interfaces" as currently defined is global. To me it seems better to not rely on a global lookup for something like this. Note that inlining this would make this [[PlatformBrand]].

littledan: Right, this PR is trying to get at the same thing as you were proposing with `[[PlatformBrand]]`.

I don't see the downside to the implementation or specification strategy to use shared data structures here during the actual check itself. We will have to look up the brand set from these data structures at some point one way or another, just a question of whether we shift the lookup a bit earlier.

littledan: One thing we could do is explain what an interface is in terms of internal slots, so it doesn't sound like we're searching through all the interfaces to find the set of inherited interfaces, or "cache" the set of inherited interfaces. But this seems like excessive formalism for no particular purpose--it's pretty clear how to represent these with efficient data structures.

annevk: I guess as long as the "API" for other standards is clear it doesn't matter much how it's implemented here.

Implementation-wise, I know Gecko stores the set of interfaces on the object itself, do Chromium and WebKit do something else? If they all do that, what is the reason to not do it here and move the location where we invoke "inherited interfaces"? All things being equal I'd much rather implicitly recommend an optimal implementation strategy.

littledan: To clarify, what I like about this form of specification is that it makes it clear that, even when an object is in a deoptimized hashtable mode, no more than constant space is needed to represent the set of platform brands needed for a particular instance.

annevk: How does that avoid the brand being per Realm? It seems the "function template" would have to be scoped to an Agent, no?

littledan: Yep, I believe that's how it works.

littledan: Yeah, I guess it's cached in [GetWrapperImpl](https://cs.chromium.org/chromium/src/gin/wrappable.cc?type=cs&q=SetObjectTemplate&g=0&l=42) isolate-wide (i.e., per agent).

annevk:
If @bzbarsky has the time I'd love for him to have another look, but I think I'm okay with the general design here per the above discussion.

littledan:
Added a note to clarify https://github.com/heycam/webidl/issues/97#issuecomment-445765462

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    </ul>
+</div>
+
+<div class="note">
+    Specifications use various wordings to reference the "[=implements=]" algorithm,
```
bzbarsky: Maybe:

Specifications may reference the concept "|object| implements |interface|" in various ways, including "|object| is an |interface| object".

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
 object is associated with.
 
+<div algorithm>
+    To <dfn>create an object implementing the interface</dfn> |interface|, with optional
+    intenal slots |slots|, for an interface which is not declared with the
+    [{{Global}}] [=extended attribute=], perform the following steps:
+
+    1. If |slots| is provided, append \[[PrimaryInterface]] to |slots|.
+    1. Otherwise, let |slots| be &laquo; \[[PrimaryInterface]] &raquo;.
+    1. Let |proto| be the [=interface prototype object=] of |interface|.
```
bzbarsky: So this algorithm doesn't do subclassing.  I assume that's because it's meant to be done from specs that really want the default prototype?

Which global's interface prototype object is used here?  Seems like the right global should be input to this algorithm.

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    1. Let |instance| be [$ObjectCreate$](|proto|, |slots|).
+    1. Set |instance|.\[[PrimaryInterface]] to |interface|.
+    1. Let |realm| be the current [=Realm=].
```
bzbarsky: What if there isn't one?  Lots of objects get created when there is no current Realm around.  As a simple example, documents are (and this matters, because they have unforgeable things on them).  Same thing for events, actually.

This really needs to use the Realm of the passed-in global, not the current Realm.

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
 
+<div algorithm>
+    To <dfn>create an object implementing the interface</dfn> |interface|, with optional
+    intenal slots |slots|, for an interface which is not declared with the
+    [{{Global}}] [=extended attribute=], perform the following steps:
+
+    1. If |slots| is provided, append \[[PrimaryInterface]] to |slots|.
+    1. Otherwise, let |slots| be &laquo; \[[PrimaryInterface]] &raquo;.
+    1. Let |proto| be the [=interface prototype object=] of |interface|.
+    1. Let |instance| be [$ObjectCreate$](|proto|, |slots|).
```
bzbarsky: This probably needs to enter the Realm of the passed-in global or something.  At least to the extent that ObjectCreate cares about its Realm.

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    1. Set |instance|.\[[PrimaryInterface]] to |interface|.
+    1. Let |realm| be the current [=Realm=].
+    1. [=Define the unforgeable regular operations=] of |interface| on |instance|, given |realm|.
+    1. [=Define the unforgeable regular attributes=] of |interface| on |instance|, given |realm|.
+    1. Return |instance|.
+</div>
+
+<div class="note">
+    Specifications use various wordings to reference the "[=create an object implementing
```
bzbarsky: Specifications may refer to the "[=create an object implementing the interface=]" algorithm in various ways, including "a new |interface| object".

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    the interface=]" algorithm, such as "a new |interface| object".
+</div>
+
+<div class="note">
+
+    The set of interfaces that a [=platform object=] [=implements=] does not change over
+    the lifetime of the object.
+
+    Multiple [=interface objects=] associated with different [=Realm/global objects=]
+    will allocate [=platform objects=] which have \[[PrimaryInterface]] internal
```
bzbarsky: Interface objects don't allocate anything, unless there's a [Constructor]....  I think this is really trying to say that platform objects associated with different globals can have the same [[PrimaryInterface]]; this doesn't involve interface objects.

index.bs
```diff
@@ -12337,18 +12336,77 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    the particular [=interface=] which [=includes=] the [=interface mixin=].
+
+</div>
+
 The <dfn id="dfn-primary-interface" export>primary interface</dfn> of a platform object
-that implements one or more interfaces is the most-derived [=interface=]
-that it implements.  The value of the \[[Prototype]] [=internal slot=]
+that [=implements=] one or more interfaces is the value of the object's
+\[[PrimaryInterface]] internal slot, which is is the most-derived [=interface=]
+that it [=implements=].  The value of the \[[Prototype]] [=internal slot=]
```
bzbarsky: Except when subclassing is involved.... I'm not sure there's much of a point to this sentence about [[Prototype]] now.  It's neither normative nor very informative.

index.bs
annevk: FWIW, I've been hoping that for cases where we create something before we create a realm, we can move those around, at least specification-wise. It seems better to first create a realm and global, and then a document.

bzbarsky: Oh, we may have a Realm fine.  But it's not going to be current unless either script is running or someone takes steps to make it current.

index.bs
littledan: What do you mean by "do subclassing"? Are there any specs that don't want the default prototype?

My guess is that the current global object would make sense, but there's a lot to test to see if this is solidly web reality...

bzbarsky: > What do you mean by "do subclassing"?

I mean `class Foo extends Event` in JS.

> Are there any specs that don't want the default prototype?

Well, `new Foo` in my example does not want the default prototype.  What prototype would it get and why, as things stand now?

index.bs
littledan: I can see how this is an important issue, but I am not sure how to address it in this patch. If I add a realm argument which is required for cases where there is no current realm, that won't make everyone pass it.

index.bs
littledan: I guess it would be the operation of getting the interface prototype object that requires the realm. Not sure how to get people to pass in the realm, so i am inclined to stay vague for now (like the old spec was).

littledan: I guess it would be the operation of getting the interface prototype object that requires the realm. Not sure how to get people to pass in the realm, so i am inclined to stay vague for now (like the old spec was).

index.bs
domenic: In previous discussions we wanted to default to the relevant realm, not the current one. And indeed it should be required in some scenarios; it would be a spec bug not to pass it. https://github.com/heycam/webidl/issues/135

bzbarsky: > that won't make everyone pass it.

True, but that way we can work on fixing those places to pass it, or have some default convention for what Realm is used when calling this algorithm in various situations.

annevk:
#540 also has some discussion on subclassing and globals btw.

index.bs
littledan: Oh, of course, I was thinking about interfaces extending other interfaces. I didn't realize that browsers actually used new.target from platform constructors this way; changed the draft spec to reflect that.

index.bs
littledan: It seems like we have two modes where this algorithm will be called:
- Passing a new.target value (from all constructors)
- Somehow indicating a realm. I made this PR vaguely refer to the relevant realm of the relevant *this* value, but I'm not sure how we should formalize this or let folks pass in their own *this*/realm.

Could we say that, if you don't have a relevant *this* value, then you should explicitly pass in the new.target value? Specs could use wording like this:

> Create a new object implementing the XYZ interface, with new.target of the interface prototype object of the relevant realm of abc.

Or is that too much cryptic boilerplate?

index.bs
littledan: Good point, deleted.

index.bs
littledan: Right, fixed.

index.bs
littledan: Thanks, adopted

index.bs
littledan: Adopted

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+<div algorithm>
+    To <dfn>create an object implementing the interface</dfn> |interface|, with optional
+    intenal slots |slots| and |new.target| value |newTarget|, for an interface which is not
+    declared with the [{{Global}}] [=extended attribute=], perform the following steps:
+
+    1. If |slots| is provided, append \[[PrimaryInterface]] to |slots|.
+    1. Otherwise, let |slots| be &laquo; \[[PrimaryInterface]] &raquo;.
+    1. If |NewTarget| is provided, let |instance| be ? [$OrdinaryCreateFromConstructor$](|newTarget|, "%ObjectPrototype%")
+    1. Otherwise,
+        1. Let |realm| be the [=relevant Realm=] of the |this| value that led to the execution of this algorithm.
```
bzbarsky: This may not exist either.  Simple example: the HTML parser.  It's creating objects without any |this| value involved, or any script anywhere in sight.

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
 object is associated with.
 
+<div algorithm>
+    To <dfn>create an object implementing the interface</dfn> |interface|, with optional
+    intenal slots |slots| and |new.target| value |newTarget|, for an interface which is not
+    declared with the [{{Global}}] [=extended attribute=], perform the following steps:
+
+    1. If |slots| is provided, append \[[PrimaryInterface]] to |slots|.
+    1. Otherwise, let |slots| be &laquo; \[[PrimaryInterface]] &raquo;.
+    1. If |NewTarget| is provided, let |instance| be ? [$OrdinaryCreateFromConstructor$](|newTarget|, "%ObjectPrototype%")
```
bzbarsky: I don't think this does the right thing, because not all web platform objects are ordinary objects (due to things like legacygetters and whatnot), but OrdinaryCreateFromConstructor explicitly creates an ordinary object.  Not sure how best to deal with that.

Also, this does not match existing browser behavior, which I think is the right behavior, for the cases when a prototype cannot be derived from `newTarget`.  This would create the object with proto set to `%ObjectPrototype%`, but we want to create it with proto set to the default prototype for the given interface in the correct Realm (the current Realm, which I believe would in fact always exist when `newTarget` exists.

As a minor nit, is it `newTarget` or `NewTarget`?  Should be consistent.

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    1. Otherwise,
+        1. Let |realm| be the [=relevant Realm=] of the |this| value that led to the execution of this algorithm.
+        1. Let |proto| be the [=interface prototype object=] of |interface| in |realm|.
+        1. Let |instance| be [$ObjectCreate$](|proto|, |slots|).
```
bzbarsky: Again, this always creates ordinary objects...

The ES spec really doesn't have great hooks for creating exotic objects, does it?  :(

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+</div>
+
 Every [=platform object=] is associated with a global environment, just
 as the [=initial objects=] are.
 It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+    To <dfn>create an object implementing the interface</dfn> |interface|, with optional
```
```suggestion
     To <dfn export lt="new|create|create a new|create an object implementing|create an object implementing the interface">create an object implementing the interface</dfn> |interface|, with optional
```

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+    intenal slots |slots| and |new.target| value |newTarget|, for an interface which is not
```
```suggestion
     internal slots |slots| and `new.target` value |newTarget|, for an interface which is not
```

index.bs
domenic: Relatedly, I can never think of a case when a web platform specification would provide a newTarget value to this algorithm. I guess from the definition of a Web IDL constructor, but that's rare. Can we create a wrapper that does not allow passing new.target?

index.bs
domenic: I think |realm| should be a parameter, which callers are allowed to omit if they are in a context that contains such a this value.

domenic: We seem to use `<emu-val>this</emu-val> value` through the rest of the spec.

index.bs
domenic: Too much cryptic boilerplate. As I mentioned above, I think the usage should be one of

> Create a new X object

or, if there is no this value

> Create a new X object in the relevant realm of abc

index.bs
```diff
@@ -12337,18 +12337,81 @@ The value of the [=function object=]’s <code class="idl">name</code> property
...
+
+    1. If |slots| is provided, append \[[PrimaryInterface]] to |slots|.
+    1. Otherwise, let |slots| be &laquo; \[[PrimaryInterface]] &raquo;.
+    1. If |NewTarget| is provided, let |instance| be ? [$OrdinaryCreateFromConstructor$](|newTarget|, "%ObjectPrototype%")
+    1. Otherwise,
+        1. Let |realm| be the [=relevant Realm=] of the |this| value that led to the execution of this algorithm.
+        1. Let |proto| be the [=interface prototype object=] of |interface| in |realm|.
+        1. Let |instance| be [$ObjectCreate$](|proto|, |slots|).
+    1. Set |instance|.\[[PrimaryInterface]] to |interface|.
+    1. Let |realm| be the current [=Realm=].
```
domenic: The mismatch where you use relevant realm of this above, and current realm here, seems bad.

index.bs
domenic: I would omit allowing passing internal slots until or unless we figure out a story for #258. It'd be bad if we had a few specifications start using "create a new X with slots [[Y]], [[Z]], [[W]]" and then later told them all to transition to putting their slot declarations in their IDL blocks, or similar.

index.bs
littledan: Thanks, will fix.

index.bs
littledan: That makes sense. What I saw was some text in HTML to explain constructors which say that they create a new instance of the interface. I like your idea that we could just implicitly take in new.target in this particular case; I will think about what wording will make sense.

littledan: That makes sense. What I saw was some text in HTML to explain constructors which say that they create a new instance of the interface. I like your idea that we could just implicitly take in new.target in this particular case; I will think about what wording will make sense.

domenic: Hmm yeah it's true that currently most web platform spec constructors are done "return override" style, i.e. they have a step that says "create a new X object" which Web IDL calls as part of https://heycam.github.io/webidl/#create-an-interface-object. Maybe we need to rethink that. I can see either "create a new instance of the interface being constructed" (which will do the right new.target thing for you) or making it implicit, by e.g. Web IDL allocating "this" for you before calling the constructor steps.

I think we should figure out which direction we want before landing this, but switching over specs to use it can be done as part of the larger project to get specs to use these new dfns, so it's OK if we create something novel that doesn't mesh with existing practice. As long as we're aligned on the right direction.

littledan: Those directions both seem plausible. The second (WebIDL creates the instance) feels simpler and more JS-like at first glance. The only downside I can think of is that it doesn't map to current spec text very clearly.

domenic: Agreed. Since we'd need to change things anyway, I'd be happy to go with the second.

domenic:
@littledan or @Ms2ger, do you think you'll be able to push this over the finish line?

Ms2ger:
Yeah, it's on my todo list.

Issue #371, opened 2017-06-21
"converting to an ECMAScript value" needs to take a realm, at least sometimes


domenic:
E.g. if you convert a `sequence<DOMString>` to an array, you need to specify the realm/global in which the conversion happens.

bzbarsky:
So for the uses in the IDL spec itself, there is always a current Realm, I believe.

But I agree, for uses from other specs something needs to happen.

TimothyGu:
Is there a way to find out which specs call the "convert to an ECMAScript value" algorithm, other than Web IDL itself?

bzbarsky:
Not trivially.

You could look for specs that have `object` or `any` (or sequences thereof) return values.  And specs with `Promise<sequence<something>>` return values.  And so on....

It's not even trivial to tell from implementations, I suspect.  In Gecko promises will automatically pick a global based on various criteria when you resolve them with something...

That said, a non-exhaustive list (just `Promise<sequence<*>>` bits, not all of which actually invoke the WebIDL algorithm) is:

* https://w3c.github.io/ServiceWorker/#clients-getall (rather broken; creates array objects in parallel steps, _manipulates_ array objects in parallel steps; filed <https://github.com/w3c/ServiceWorker/issues/1240>).
* https://w3c.github.io/ServiceWorker/#dom-cache-keys step 5.4.3
* https://w3c.github.io/ServiceWorker/#dom-cachestorage-keys
* https://w3c.github.io/ServiceWorker/#cache-matchall -- specifies the realm, nice!
* getFilesAndDirectories in http://w3c.github.io/filesystem-api/ (not directly linkable?)
* https://drafts.csswg.org/css-font-loading/#dom-fontfaceset-load
* https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-enumeratedevices
* WebVR's `activeVRDisplays` (Can't find a spec draft for this)
* https://w3c.github.io/ServiceWorker/#navigator-service-worker-getRegistrations (filed <https://github.com/w3c/ServiceWorker/issues/1241>)
* https://notifications.spec.whatwg.org/#dom-serviceworkerregistration-getnotifications sort of (filed <https://github.com/whatwg/notifications/issues/113>)

PR #598, opened 2018-12-12
Specify that returned Promises are created in the relevant realm

littledan:
Note that these semantics differ from JavaScript async functions, which
create a Promise in the current Realm.

Verified in Chromium and Firefox that HTMLImageElement.decode and
CustomElementsRegistry.whenDefined return Promise instances from
the relevant realm.

(I'm not sure what's the preferred way to typeset "this", or if it would
be preferred to thread it explicitly through the algorithm. Also, it's
not quite necessary to use the relevant Promise.resolve function, since
resolve effectively uses the relevant realm of its receiver(!))

Resolves part of #135

***
<a href="https://pr-preview.s3.amazonaws.com/littledan/webidl/pull/598.html" title="Last updated on Dec 12, 2018, 4:50 PM GMT (688df8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/598/5a20e34...littledan:688df8c.html" title="Last updated on Dec 12, 2018, 4:50 PM GMT (688df8c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7855,10 +7855,14 @@ objects.
     An ECMAScript value |V| is [=converted to an IDL value|converted=]
     to an IDL <a interface lt="Promise">Promise<var ignore>T</var></a> value as follows:
 
-    1.  Let |promise| be [=!=] <a abstract-op>Call</a>({{%Promise_resolve%}}, {{%Promise%}}, «|V|»).
+    1.  Let |realm| be the [=relevant Realm=] of the |this| value that led to the execution of this algorithm.
+    1.  Let |promise| be [=!=] <a abstract-op>Call</a>(|realm|.\[[Intrinsics]].[[{{%Promise_resolve%}}]], |realm|.\[[Intrinsics]].[[{{%Promise%}}]], «|V|»).
     1.  Return the IDL [=promise type=] value that is a reference to the
         same object as |promise|.
 </div>
+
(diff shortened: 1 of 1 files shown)


annevk:
We don't have a convention yet for "this", but it would be really great if we made it available on the IDL side of things as well (i.e., for specification algorithms). https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301 requested that a long time ago. https://github.com/heycam/webidl/pull/495 adds it, but couples it with other things that were harder to reach agreement on.

Ms2ger:
It seems kinda strange to refer to a `this` value in `#dfn-convert-ecmascript-to-idl-value`. Should we pass a Realm into all variants of that algorithm?

littledan:
OK, seems like this patch (as well as #595) will need to block on the infrastructure for `this`.

Ms2ger:
Related: #371

annevk:
We now have "this". Do you still wish to pursue this?

Ms2ger:
Note that we don't have **this** in the "converted to an IDL value" algorithm. I don't think this PR makes a great deal of sense without fixing #371 first.

domenic:
Let's close this and continue trying for the more ambitious plan in https://github.com/heycam/webidl/issues/135#issuecomment-772716243.

PR #601, opened 2018-12-18
Editorial: Use some more abstract operations.

### LINKED ISSUE #42: explicitly invoke [[HasOwnProperty]], [[GetProperyOf]], etc. more often
There are many instances in the spec where it says something like "If _O_ has an own property with name _P_, then ...".  Given proxies, we really should be invoking [[HasOwnProperty]] here explicitly, because we'll want to be clear that the proxy handler is invoked.  (And it's not really clear whether a proxy has "real" own properties anyway.)

Ms2ger:
Notably, this removes the vague 'has an own property' language and fixes #42.

Only in the case of the prototype in the named property visibility algorithm
is this change potentially observable.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/601.html" title="Last updated on Jan 16, 2019, 9:42 AM UTC (e843626)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/601/6ec828c...Ms2ger:e843626.html" title="Last updated on Jan 16, 2019, 9:42 AM UTC (e843626)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -138,6 +138,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: NormalCompletion; url: sec-normalcompletion
         text: ObjectCreate; url: sec-objectcreate
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
+        text: OrdinaryDelete; url: sec-ordinarydelete
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
         text: OrdinarySetWithOwnDescriptor; url: sec-ordinarysetwithowndescriptor
@@ -12553,9 +12554,9 @@ Issue: Define those internal methods imperatively instead.
             1.  If |operation| was declared with a [=return type=] of {{boolean}}
                 and the steps returned <emu-val>false</emu-val>, then return <emu-val>false</emu-val>.
         1.  Return <emu-val>true</emu-val>.
-    1.  If |O| has an own property with name |P|, then:
+    1.  If <a abstract-op>OrdinaryGetOwnProperty</a>(|O|, |P|) is not <emu-val>undefined</emu-val>, then:
         1.  If the property is not configurable, then return <emu-val>false</emu-val>.
-        1.  Otherwise, remove the property from |O|.
+        1.  Remove the own property with name |P| from |O|.
     1.  Return <emu-val>true</emu-val>.
 </div>
 
@@ -12634,7 +12635,7 @@ internal method as follows.
     The algorithm operates as follows, with property name |P| and object |O|:
 
     1.  If |P| is not a [=supported property name=] of |O|, then return false.
-    1.  If |O| has an own property named |P|, then return false.
+    1.  If <a abstract-op>OrdinaryGetOwnProperty</a>(|O|, |P|) is not <emu-val>undefined</emu-val>, then return false.
 
           Note: This will include cases in which |O| has unforgeable properties,
           because in practice those are always set up before objects have any supported property names,


index.bs
```diff
@@ -12636,7 +12635,7 @@ internal method as follows.
     The algorithm operates as follows, with property name |P| and object |O|:
 
     1.  If |P| is not a [=supported property name=] of |O|, then return false.
-    1.  If |O| has an own property named |P|, then return false.
+    1.  If <a abstract-op>HasOwnProperty</a>(|O|, |P|) is <emu-val>true</emu-val>, then return false.
```
bzbarsky: Unfortunately, I think this is an infinite recursion.

Consider a call to [[GetOwnProperty]] as defined in <https://heycam.github.io/webidl/#legacy-platform-object-getownproperty>.  This calls https://heycam.github.io/webidl/#LegacyPlatformObjectGetOwnProperty which calls into https://heycam.github.io/webidl/#dfn-named-property-visibility for _O_.  The proposed change would then call https://tc39.github.io/ecma262/#sec-hasownproperty on _O_, which will call right back into _O_'s [[GetOwnProperty]].

What you probably need to do instead is invoke https://tc39.github.io/ecma262/#sec-ordinarygetownproperty directly or something.  But note that this uses exactly the same vague "has an own property" (or rather "does not have an own property") language that webidl already has...

index.bs
```diff
@@ -12647,7 +12646,8 @@ internal method as follows.
     1.  Let |prototype| be |O|.\[[GetPrototypeOf]]().
     1.  While |prototype| is not null:
         1.  If |prototype| is not a [=named properties object=],
-            and |prototype| has an own property named |P|, then return false.
+            and <a abstract-op>HasOwnProperty</a>(|prototype|, |P|) is <emu-val>true</emu-val>,
```
bzbarsky: Is the potentially-observable change here what happens when there's a proxy on the proto chain?  What do browsers currently do here?

TimothyGu:
The crux here is that in ECMAScript, every object implicitly has a list of properties that’s different from what’s exposed through internal methods, and these references are talking about the implicit properties. See https://github.com/tc39/ecma262/issues/1067 for some more background.

index.bs
Ms2ger: They're inconsistent, of course. I'll retake this part after #607 is fixed.

index.bs
```diff
@@ -12553,10 +12554,7 @@ Issue: Define those internal methods imperatively instead.
             1.  If |operation| was declared with a [=return type=] of {{boolean}}
                 and the steps returned <emu-val>false</emu-val>, then return <emu-val>false</emu-val>.
         1.  Return <emu-val>true</emu-val>.
-    1.  If |O| has an own property with name |P|, then:
-        1.  If the property is not configurable, then return <emu-val>false</emu-val>.
-        1.  Otherwise, remove the property from |O|.
-    1.  Return <emu-val>true</emu-val>.
+    1.  Return <a abstract-op>OrdinaryDelete</a>(|O|, |P|).
```
bzbarsky: So... UAs are kinda all over the map here (e.g. neither Firefox nor Chrome seems to implement the named property visibility algorithm the way the spec specs it!) but this is observably different from the old spec.  That's because OrdinaryDelete calls [[GetOwnProperty]], not OrdinaryGetOwnProperty.  This means we'll do the proto walk for the visibility algorithm twice, if UAs were actually implementing the spec.  Do we really want that?

Ms2ger: No, I forgot about the distinction between [[GetOwnProperty]] and OrdinaryGetOwnProperty here; will fix.

domenic:
At this point it's not clear what this change accomplishes. It replaces text that extracts the one relevant part of OrdinaryGetOwnProperty (step 2, "If O does not have an own property with key P, return undefined.") with a call to the whole of OrdinaryGetOwnProperty, throwing away the result of steps 3-9.

This doesn't accomplish the goal implied by #42, of triggering proxy traps. It seems like the above discussion concluded that wasn't desirable? In which case maybe the better resolution is to just close #42 as not desirable, with no changes to the spec text.

Ms2ger:
I feel we probably should have some clearer explanation that "own properties" only refers to the "real" ones, and doesn't imply a call to the overridden `[[GetOwnProperty]]` hook. I guess you're right that calling the Ordinary* abstract ops may not be the best approach.

annevk:
I guess you could potentially escalate this to TC39 as https://tc39.github.io/ecma262/#sec-object-type only talks about get/set access to these properties, not "has". So perhaps they should define a clearer primitive that can be linked.

annevk:
Closing this as we use language that's consistent with JavaScript. If JavaScript formalizes own properties more (perhaps in https://github.com/tc39/ecma262/pull/1438) we would follow.

Issue #151, opened 2016-08-24
"get a copy of the bytes held by the buffer source" should not fail for detached buffers


annevk:
Currently it seems that IDL throws for detached buffers. However, as far as I can tell implementations do not follow this and simply end up with the empty byte sequence whenever they encounter a detached buffer.

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/4414 is my demo for this (note the lack of exceptions/rejections).

(Note that there's also still an ask to clean up this aspect of IDL and make it much easier to work with: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28798 .)

bzbarsky:
Note that I think this may be a more general issue, in that the ES spec requires throwing in many cases when operating on detached stuff, but in practice UAs don't do that...  Worth checking with @jswalden at least; I recall him doing some work on that stuff.

annevk:
Yeah, I raised that two years ago when I last did research into this: https://esdiscuss.org/topic/arraybuffer-neutering. @bterlson is TC39 still hoping implementers will throw more for ArrayBuffer and friends? When are we going to throw in the towel?

annevk:
In any event, IDL doesn't have to follow suit and can just return the empty byte sequence if IsDetached returns true solving at least part of the problem.

bterlson:
I think we hope everyone will align and throw more often, but I admit that I haven't looked into this area for at least a year so things may have changed. There should be test262 collateral for all this (right, @tcare / @leobalter / @jugglinmike?) so I think it's just a matter of priorities. Hopefully the web doesn't depend on non-throwing behavior? (Seems likely given that detached buffers aren't very common, I suppose)

littledan:
test262 does include tests for this (thanks @leobalter) but I'm still concerned about web compatibility. I haven't figured out how to instrument TypedArrays to measure whether detached TypedArrays are ever accessed without decreasing TypedArray performance (we set the length to 0 and bounds checks handle accesses, so we don't need a separate detached check), so I'm not sure how to get the information from telemetry. Does anyone else have an idea for how to get solid data on this question?

PR #604, opened 2019-01-09
[[ViewedArrayBuffer]] cannot be undefined during usage

annevk:
As far as I can tell there's no way for a %TypedArray%'s [[ViewedArrayBuffer]] internal slot to be undefined. I'm pretty sure this has to be the case as implementations assume this algorithm to never throw (the detached issue is tracked separately by #151).

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/604.html" title="Last updated on Jan 9, 2019, 2:44 PM UTC (3391298)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/604/8f48c5a...3391298.html" title="Last updated on Jan 9, 2019, 2:44 PM UTC (3391298)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2047,7 +2047,7 @@ There are three kinds of operation:
 
 If an operation has an identifier but no <emu-t>static</emu-t>
 keyword, then it declares a <dfn id="dfn-regular-operation" export>regular operation</dfn>.
-If the operation has a 
+If the operation has a
 [=special keyword=]
 used in its declaration (that is, any keyword matching
 <emu-nt><a href="#prod-Special">Special</a></emu-nt>, or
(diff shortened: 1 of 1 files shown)


annevk:
(The only place where I've found ECMAScript says it's undefined is during allocation, when view is created, but the buffer isn't, but at that point no JavaScript can run and do something with the object, afaik.)

bmeck:
Per the usage of the assert in [IsDetachedBuffer](https://tc39.github.io/ecma262/#sec-isdetachedbuffer) and lack of checks prior to usage in various locations such as in [ValidateTypedArray](https://tc39.github.io/ecma262/#sec-validatetypedarray) it would be a bug if it ever could be `undefined`. This PR looks fine to me.

PR #605, opened 2019-01-09
Detached ArrayBuffer objects should not throw

### LINKED ISSUE #151: "get a copy of the bytes held by the buffer source" should not fail for detached buffers

annevk:
Instead treat them the same as the empty byte sequence as implementations have been doing since forever.

This also performs the IsDetachedBuffer abstract operation on the correct object.

Fixes #151.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/605.html" title="Last updated on Jan 10, 2019, 8:57 AM UTC (19fa3e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/605/8f48c5a...19fa3e2.html" title="Last updated on Jan 10, 2019, 8:57 AM UTC (19fa3e2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8171,8 +8171,8 @@ a reference to the same object that the IDL value represents.
         1.  Set |offset| to the value of |O|’s \[[ByteOffset]] [=internal slot=].
         1.  Set |length| to the value of |O|’s \[[ByteLength]] [=internal slot=].
     1.  Otherwise, set |length| to the value of |O|’s \[[ArrayBufferByteLength]] [=internal slot=].
-    1.  If <a abstract-op>IsDetachedBuffer</a>(|arrayBuffer|), then return a reference to or a copy
-        of (as required) the empty byte sequence.
+    1.  If <a abstract-op>IsDetachedBuffer</a>(|arrayBuffer|) is <emu-val>true</emu-val>, then
+        return a reference to or a copy of (as required) the empty byte sequence.
     1.  Let |data| be the value of |O|’s \[[ArrayBufferData]] [=internal slot=].
     1.  Return a reference to or copy of (as required) the |length| bytes in |data|
         starting at byte offset |offset|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8171,8 +8171,8 @@ a reference to the same object that the IDL value represents.
         1.  Set |offset| to the value of |O|’s \[[ByteOffset]] [=internal slot=].
         1.  Set |length| to the value of |O|’s \[[ByteLength]] [=internal slot=].
     1.  Otherwise, set |length| to the value of |O|’s \[[ArrayBufferByteLength]] [=internal slot=].
-    1.  If <a abstract-op>IsDetachedBuffer</a>(|O|), then
-        [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+    1.  If <a abstract-op>IsDetachedBuffer</a>(|arrayBuffer|) is <emu-val>true</emu-val>, then
+        return a reference to or a copy of (as required) the empty byte sequence.
```
domenic: I think you should just return the empty byte sequence. A copy or reference doesn't really make sense here.

littledan: I don't know what the effective difference between them would be. What does "as required" mean?

annevk: There's two ways to invoke the algorithm and the algorithm is expected depending on how it's invoked to either return a reference to bytes or a copy thereof. So this matches the existing expectations of callers and the existing way things are phrased for return values.

I agree that this could be done in a somewhat cleaner way, but that seems like a follow-up of sorts as it affects the whole algorithm (and potentially callers).

domenic: How do you have a reference or copy of something empty, though? (Presumably newly-created? Or is everyone sharing a reference to a single empty byte sequence?) You only reference or copy existing things. This failure path is special and shouldn't be doing reference or copy.

annevk: Yeah, I'm not sure what distinction I was trying to make anymore.

PR #606, opened 2019-01-10
Give Location an immutable prototype object

annevk:
Remaining outstanding issue of https://github.com/tc39/ecma262/issues/683.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10815,9 +10815,15 @@ with the [{{NoInterfaceObject}}] [=extended attribute=].
             \[[Writable]]: <emu-val>false</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>,
             \[[Configurable]]: <emu-val>true</emu-val>}.
         1.  Perform [=!=] <a abstract-op>DefinePropertyOrThrow</a>(|interfaceProtoObj|, {{@@unscopables}}, |desc|).
-    1.  If |interface| is declared with the [{{Global}}] [=extended attribute=], or
-        |interface| is in the set of [=inherited interfaces=] of an interface
-        that is declared with the [{{Global}}] [=extended attribute=], then:
+    1.  If one of the following is true
+
+        * |interface| is declared with the [{{Global}}] [=extended attribute=]
+        * |interface| is in the set of [=inherited interfaces=] of an interface that is declared
+          with the [{{Global}}] [=extended attribute=]
+        * |interface| is {{Location}} and declared with the [{{Exposed}}] [=extended attribute=]
+          whose value is {{Window}}
+
+        then:
         1.  Set the internal methods of |interfaceProtoObj|
             which are specific to [=immutable prototype exotic objects=]
             to the definitions specified in


annevk:
It seems the build script is using outdated packages and today is the day Travis CI is no longer having it?

littledan:
Rereading that thread makes me sad, but making things immutable makes me happy.

Doesn't HTML already cover this in https://html.spec.whatwg.org/multipage/history.html#location-setprototypeof ?

If we still need to do something  at the IDL level, the wording of this patch feels like a layering violation, to check for one interface here. I would prefer an ImmutablePrototype extended attribute. That's how I ended up implementing this in Blink anyway, for what it's worth (e.g., see [EventTarget's IDL](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/events/event_target.idl?q=Immutableprototype&sq=package:chromium&g=0&l=28); this avoids the need for the spec's nonlocal logic). Anyway even if you keep with the rationale from last time about not using the extended attribute for things in the global prototype chain (to make things easier for spec authors), I think it would be cleaner to do that here.

annevk:
HTML covers it for the `Location` object, not for the `Location` prototype object, which is created by IDL. (I don't think we have any prose definitions for prototype objects in the web platform.)

At least, if I understand it correctly. I was initially also somewhat confused by that, but I think I got it right.

As for the layering violation, I thought this was a one-off for `Location` which kinda made me think this would be okay as introducing extended attributes for one-offs also seems wrong (someone might use it for the wrong reasons). Global objects are already handled and I don't think `EventTarget` should have this. In particular, Chrome does not throw for `new EventTarget().__proto__ = {};` and I don't think it should, so I'm not sure how that extended attribute works.

littledan:
The `[ImmutablePrototype]` Blink-specific extended attribute does what you're trying to do here--it makes the `.prototype` property an immutable prototype exotic object. (It doesn't do what that HTML text does, of freezing the prototype of the instance, sorry for my confusion.)

```
> EventTarget.prototype.__proto__ = {}
< VM82:1 Uncaught TypeError: Immutable prototype object '#<EventTarget>' cannot have their prototype set
    at EventTarget.set __proto__ [as __proto__] (<anonymous>)
    at <anonymous>:1:33
```

I still wouldn't mind if we switched the logic in WebIDL from implicitly freezing all prototypes of objects which any global inherits from, to marking each of them with an extended attribute, but we already settled on this design, so I won't try to argue that again. I'm OK with the layering violation here.

If we end up using WebIDL for more kinds of non-web things, it might be helpful to do these kinds of refactorings, to be strict about what's HTML and what's just WebIDL, but we're not there yet. More important to fully specify everything to start. So I'm fine with the spec in its current state.

----

LGTM modulo the following:

Are there wpt tests that check that Location.prototype.__proto__ is not settable?

Is there implementation support for this change? It seems like Firefox, WebKit and Chromium currently all permit `Location.prototype.__proto__` to be set.

annevk:
Indeed, it's only the following that throws (though not in Firefox):
```js
Location.prototype.__proto__ = { __proto__: location}
```
I guess this wasn't the correct fix after all then and I'm not really sure how to resolve the original issue.

littledan:
Hah, I wasn't able to navigate away after giving Location a cyclic prototype chain in Firefox! But WebKit and Chromium do seem to do this check.

Given 2/3 web reality (and not awesome behavior from the third), I wonder if we should just have prose to explain that there is a cyclic prototype chain here, just like for ordinary objects (as that other thread started with, even if the ES spec refuses to put in clean hooks).

bzbarsky:
> Hah, I wasn't able to navigate away after giving Location a cyclic prototype chain in Firefox! 

You did that via the console, right?  That's a devtools bug!  Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1519154 to track that.

PR #609, opened 2019-01-14
Remove [NoInterfaceObject].

### LINKED ISSUE #430: No mixin [NoInterfaceObject] usage
Per https://github.com/heycam/webidl/pull/423#issuecomment-326332460 this exists. I'd like to study such usage and see if we can get rid of it.

Ms2ger:
Fixes #430.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/609.html" title="Last updated on Jan 14, 2019, 4:01 PM UTC (98de7f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/609/6ec828c...Ms2ger:98de7f7.html" title="Last updated on Jan 14, 2019, 4:01 PM UTC (98de7f7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1042,8 +1042,7 @@ limitations.  The following extended attributes must not
 be specified on partial interface definitions:
 [{{Constructor}}],
 [{{LegacyWindowAlias}}],
-[{{NamedConstructor}}],
-[{{NoInterfaceObject}}].
+[{{NamedConstructor}}].
 
 Note: The above list of [=extended attributes=]
 is all of those defined in this document that are applicable to
@@ -1067,12 +1066,10 @@ The following extended attributes are applicable to interfaces:
 [{{Global}}],
 [{{LegacyWindowAlias}}],
 [{{NamedConstructor}}],
-[{{NoInterfaceObject}}],
 [{{OverrideBuiltins}}], and
 [{{SecureContext}}].
 
-Non-callback [=interfaces=] which are not annotated
-with a [{{NoInterfaceObject}}] [=extended attribute=],
+Non-callback [=interfaces=]
 and [=callback interfaces=] which declare [=constants=]
 must be annotated with an [{{Exposed}}] [=extended attribute=].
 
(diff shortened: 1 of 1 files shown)


domenic:
I guess the real work here is PRing every spec that uses it, and then trying to ship and see if the result is web-compatible...

Ms2ger:
Indeed. I'm not sure if all the specs that use it are maintained, though. Perhaps one of the browser vendors should try shipping first and see what happens.

foolip:
Greping https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl reveals surprisingly few remaining uses of this beyond what's already linked in this PR:
https://drafts.csswg.org/css-font-loading/#font-face-source (@tabatkins)
https://drafts.csswg.org/css-regions/#the-region-interface (@atanassov @astearns)
https://w3c.github.io/mediacapture-main/#interface-definition (@aboba @jan-ivar)
https://w3c.github.io/webappsec-feature-policy/#the-policy-object (@clelland)

jan-ivar:
> https://w3c.github.io/mediacapture-main/#interface-definition (@aboba @jan-ivar)

*"Although this specification formally defines `ConstrainablePattern` as a WebIDL interface, it is actually a template or pattern for other interfaces and cannot be inherited directly since the return values of the methods need to be extended,"*, so 👍

foolip:
Ah, sounds like that interface should be marked up as an example in the spec. This would avoid it showing up in reffy-reports, and would avoid the need for `[NoInterfaceObject]`, which presumably isn't something the other interfaces should copy.

jan-ivar:
@foolip How do I do that?

foolip:
@jan-ivar wrapping in `<div class="example">` might to the trick. Or if you want to just avoid it being extracted by reffy, the `exclude` class was added in https://github.com/tidoust/reffy/pull/137 for that purpose.

saschanaz:
Are there still known blockers other than https://github.com/w3c/webappsec-feature-policy/issues/341 or is it the last one?

bzbarsky:
Well, there's the fact that as far as I know browsers haven't shipped the removals done so far, so we don't know whether they're web-compatible, right?

saschanaz:
Yes, although at least some of them are shipped in stable Chrome e.g. `External` from HTML and `DeviceMotionEventRotationRate` etc. from device orientation.

domenic:
We renamed it to [LegacyNoInterfaceObject] and all remaining uses are in Web GL. If Web GL shows signs of wanting to expose their interfaces then we can revisit, but for now let's close...

PR #610, opened 2019-01-14
Use map/exists to define dictionary/present.

### LINKED ISSUE #381: Dictionary members' "present" definition seems circular
[2.4 Dictionaries](https://heycam.github.io/webidl/#dfn-inherited-dictionaries) sets up a *dictionary* internally as an [ordered map](https://infra.spec.whatwg.org/#ordered-map) of [entries](https://infra.spec.whatwg.org/#map-entry).

So shouldn't the definition of [present](https://heycam.github.io/webidl/#dfn-present) be based on [exists](https://infra.spec.whatwg.org/#map-exists)?

The current definition seems circular: *"A dictionary value of type D can have key–value pairs corresponding to the dictionary members defined on D and on any of D’s inherited dictionaries. On a given dictionary value, the presence of each dictionary member is optional, unless that member is specified as required. When specified in the dictionary value, a dictionary member is said to be* ***present***, *otherwise it is* ***not present***."

It uses *"presence"* ahead of defining **present**, and the word *"specified"* sounds like we're trying to avoid saying "present" again, but I think it's the wrong word, as presence doesn't seem to rely on specification. Chicken and egg.

The **present** definition seems used both on input and output:

1. "In the ECMAScript binding, a value of undefined is treated as not present,"
2. "If the dictionary member named key is present in V, then:"

Ms2ger:
Fixes #381.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/610.html" title="Last updated on Jan 14, 2019, 4:31 PM UTC (2ca8a5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/610/6ec828c...Ms2ger:2ca8a5f.html" title="Last updated on Jan 14, 2019, 4:31 PM UTC (2ca8a5f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4483,8 +4483,10 @@ to the dictionary members defined on |D| and on any of |D|’s
 [=inherited dictionaries=].
 On a given dictionary value, the presence of each dictionary member
 is optional, unless that member is specified as required.
-When specified in the dictionary value, a dictionary member is said to be
-<dfn id="dfn-present" export lt="present|not present" for="dictionary member">present</dfn>, otherwise it is [=not present=].
+A dictionary member is said to be
+<dfn id="dfn-present" export lt="present|not present" for="dictionary member">present</dfn>
+in a dictionary value if the value [=map/exists|contains an entry with the key=]
+given by the member's [=identifier=], otherwise it is [=not present=].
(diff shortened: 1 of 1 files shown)


bzbarsky:
Thank you!

PR #612, opened 2019-01-15
Allow identifiers that start with a hyphen.

### LINKED ISSUE #564: identifier needs to allow a leading hyphen
[CSSOM spec](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-_dashed_attribute) describes about identifiers with leading dashs, but they conflict with current WebIDL spec, because now we allow `/_?[A-Za-z][0-9A-Z_a-z-]*/` in identifiers.

So I'd suggest to update the pattern to allow a hypen as a leading character to make both specs consistent with each other. Specifically, I suggest following two candidates.
- `[_-]?[A-Za-z][0-9A-Z_a-z-]*`
- `_?[A-Za-z-][0-9A-Z_a-z-]*`

Ms2ger:
Fixes #564.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/612.html" title="Last updated on Jan 15, 2019, 10:59 AM UTC (ed1765c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/612/d05c8cc...Ms2ger:ed1765c.html" title="Last updated on Jan 15, 2019, 10:59 AM UTC (ed1765c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13556,7 +13556,7 @@ expression syntax [[!PERLRE]]) as follows:
     <tr>
         <td id="prod-identifier"><emu-t class="regex">identifier</emu-t></td>
         <td><code>=</code></td>
-        <td><code class="regex"><span class="mute">/</span>_?[A-Za-z][0-9A-Z_a-z-]*<span class="mute">/</span></code></td>
+        <td><code class="regex"><span class="mute">/</span>_?[A-Za-z-][0-9A-Z_a-z-]*<span class="mute">/</span></code></td>
     </tr>
     <tr>
         <td id="prod-string"><emu-t class="regex">string</emu-t></td>


marcoscaceres:
cc @saschanaz. Tracking for WebIDL2.js.

Ms2ger:
I'm guessing Gecko allowed it for `_content`; not sure if there's a particular reason to allow it here.

(We'd best clarify "For all of these constructs, the identifier is the value of the identifier token with any leading U+005F LOW LINE ("_") character (underscore) removed." if so, though.)

annevk:
I filed #632 as follow-up.

PR #615, opened 2019-01-15
Editorial: Clarify the wording of an Exposed requirement.

### LINKED ISSUE #548: Poorly worded normative requirement could be improved
Filing this only because we implemented this incorrectly, and had to re-read it 10 more times before it [sort of] made sense :)

> The [Exposed] extended attribute must not be specified both on an interface, mixin, or namespace member, and on the partial interface, partial interface mixin, or partial namespace definition the member is declared on.

Two things: 

Clarify that its the member being described--this is revealed only via inspection of the hyperlink: e.g.,:

> The [Exposed] extended attribute must not be specified both on an interface **member**, mixin **member**, or namespace member, and on the partial interface, partial interface mixin, or partial namespace definition the member is declared on.

This makes it sound like its possible to have the same member defined both in a partial interface and non-partial interface. This obviously can't happen. A better phasing might be:

> The [Exposed] extended attribute must not be specified both on the members and on the member's host interface, where the host interface is a partial interface, partial mixin interface, or partial namespace.

That makes much more sense to me.

Ms2ger:
Fixes #548.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/615.html" title="Last updated on Jan 15, 2019, 2:22 PM UTC (a748560)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/615/d05c8cc...Ms2ger:a748560.html" title="Last updated on Jan 15, 2019, 2:22 PM UTC (a748560)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8613,7 +8613,7 @@ If [{{Exposed}}] appears on an [=overloaded=] [=operation=],
 then it must appear identically on all overloads.
 
 The [{{Exposed}}] extended attribute must not be specified both on
-an [=interface member|interface=], [=interface mixin member|mixin=], or [=namespace member=], and on
+an [=interface member=], [=interface mixin member=], or [=namespace member=], and on
 the [=partial interface=], [=partial interface mixin=], or [=partial namespace=] definition
 the [=member=] is declared on.
 


annevk:
Ta!

PR #616, opened 2019-01-15
Rename the 'float' production to 'real'.

### LINKED ISSUE #558: "float" is the name of two distinct things in the grammar
In the grammar, "float" is both the name of the [regular expression](https://heycam.github.io/webidl/#prod-float) matching a literal floating point value and the actual keyword ["float"](https://heycam.github.io/webidl/#index-prod-FloatType). The only visual distinction between the two is that the regular expression float is displayed in a sans-serif font, and the keyword is displayed in a monospace font - and honestly, I can't tell which is which without looking at the computed css. While it's reasonably easy to figure out which one is intended where based on the context, it's still a pretty confusing thing to come across - it would be nice if one was renamed, or more realistically, if it was pointed out in the [description of the grammar](https://heycam.github.io/webidl/#idl-grammar).

Ms2ger:
This should prevent confusion between the production and the keyword. Fixes #558.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/616.html" title="Last updated on Jan 28, 2019, 1:31 PM UTC (c04bd6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/616/d05c8cc...Ms2ger:c04bd6a.html" title="Last updated on Jan 28, 2019, 1:31 PM UTC (c04bd6a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1606,7 +1606,7 @@ one of the two boolean literal tokens (<emu-t>true</emu-t>
 and <emu-t>false</emu-t>),
 the <emu-t>null</emu-t> token, an
 <emu-t class="regex"><a href="#prod-integer">integer</a></emu-t> token,
-a <emu-t class="regex"><a href="#prod-float">float</a></emu-t> token,
+a <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token,
 or one of the three special floating point constant values
 (<emu-t>-Infinity</emu-t>, <emu-t>Infinity</emu-t> and <emu-t>NaN</emu-t>).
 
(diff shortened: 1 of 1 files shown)


Ms2ger:
(If someone has a better name, that's great too.)

littledan:
Bikeshedding: decimal? Infinity and NaN aren't really reals.

Ms2ger:
Infinity and NaN aren't included in this production either.

littledan:
Oh sorry, i don't have a strong opinion then. (FWIW, not all reals are included in the reals production...)

annevk:
Wouldn't [rational](https://en.wikipedia.org/wiki/Rational_number) be somewhat closer to the truth?

Will anything in the eventual BigInt number refactoring help us here?

Ms2ger:
Fair point about "rational" being better. Happy to change to that if you'd be willing to merge that :)

annevk:
(I've been hoping for somebody else to weigh in.)

So taking a closer look, it seems this is IDL's https://tc39.github.io/ecma262/#prod-DecimalLiteral, right? I think in that case Daniel is correct and we should adopt ECMAScript precedent. I'd be happy to merge that.

Ms2ger:
Makes sense, updated to use "decimal".

PR #619, opened 2019-01-16
Editorial: Make the opName argument required.

### LINKED ISSUE #534: "call a user object’s operation" should not have an optional argument in the middle
That makes callers ambiguous unless they use rather elaborate wording to get around it.

This is about _opName_.

Now, having said that, do we actually need _opName_? We don't want folks to use callback interfaces. Those in DOM only have a single operation (although they both do list _opName_, maybe it wasn't always optional). Are there any that have more than one operation?

Ms2ger:
Fixes #534.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/619.html" title="Last updated on Jan 16, 2019, 10:43 AM UTC (32e54b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/619/ce7b02c...Ms2ger:32e54b3.html" title="Last updated on Jan 16, 2019, 10:43 AM UTC (32e54b3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12836,7 +12836,7 @@ the special value “missing”, which represents a missing optional argument.
 
     To <dfn id="call-a-user-objects-operation" export>call a user object's operation</dfn>,
     given a [=Interface types|callback interface type=] value |value|,
-    sometimes-optional operation name |opName|, [=Web IDL arguments list=] |args|, and optional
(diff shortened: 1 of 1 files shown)


domenic:
This doesn't really seem editorial, as it changes the "public interface" of Web IDL. I will amend the commit description while merging (and also remove the period from the headline).

Issue #350, opened 2017-04-25
Consider renaming all legacy extended attributes


domenic:
I'm not sure this is worth the churn, but if we wanted to, let's do it all at once.

- LenientSetter -> LegacyLenientSetter
  - Used by fullscreen
- LenientThis -> LegacyLenientThis
   - Used by HTML
- NamedConstructor -> LegacyFactoryFunction https://github.com/heycam/webidl/issues/277
  - Used by HTML
- OverrideBuiltins -> LegacyOverrideBuiltins??
  - Used by HTML
- Replaceable -> LegacyReplaceable?
- TreatNonObjectAsNull -> LegacyTreatNonObjectAsNull
  - Used by HTML
- TreatNullAs=EmptyString -> LegacyNullToEmptyString (different naming scheme than LegacyTreatNonObjectAsNull since it operates on types, not callback functions)
  - Used by CSSOM, DOM Parsing, HTML, DOM
- Unforgeable -> LegacyUnforgeable???
  - Used by HTML, DOM

I have attached varying degrees of ?s for the ones I am unsure on.

annevk:
In general I'm very much in favor of this and I think they all make sense. I'm a little worried where we have something we recommend, EventHandler, depend on something that we don't, [TreatNonObjectAsNull], as I wouldn't want folks to start ignoring EventHandler. I think that'll probably be okay in practice though.

bzbarsky:
`Replaceable` is not legacy.  It's basically required for any new attribute we add on globals.  And even then it may not be enough; see https://bugzilla.mozilla.org/show_bug.cgi?id=1359022 for example.

annevk:
If [Replaceable] is limited to globals it seems reasonable to keep it with the existing name.

bzbarsky:
Hmm.  In Gecko, the only non-global use of Replaceable is `.stack` on `DOMException`.  Which is exciting for all sorts of reasons, not least because there is no mention of `.stack` in https://heycam.github.io/webidl/#es-exceptions

Of course there's no mention of it in ES either, for ES exceptions, but in practice browsers implement it (with non-identical semantics) and websites rely on it....

annevk:
It sounds like we could limit it to globals in standards then and deal with stack once someone figures out what to do there (which likely will need to be defined in JavaScript-prose anyway given how we define DOMException today).

TimothyGu:
Is this describing the same issue as #100? If so we should close one of them out.

tobie:
[...]

PR #620, opened 2019-01-16
Restrict the extended attributes for partial interfaces.

### LINKED ISSUE #154: [Exposed] doesn't seem to work well with partial interfaces
Previously at https://github.com/heycam/webidl/pull/121#discussion_r76247465 and https://github.com/heycam/webidl/pull/121#discussion_r76250672

The spec currently says

> "Any extended attribute specified on a partial interface definition is considered to appear on the interface itself"

Which does not seem to work well with [Exposed] (and possibly not well with [SecureContext] either). Probably that statement needs to be modified, but we also need to think about why it's there in the first place and for what extended attributes it does make sense

We are also missing a potential restriction which seems important, namely "If [Exposed] appears on a partial interface, then the partial interface's exposure set MUST be a subset of the interface's exposure set."

In general it's not clear what putting different [Exposed] values on an interface vs. a partial interface might mean, or on members of an interface or members of a partial interface.

Ms2ger:
It seems like a bad idea to allow [Global] on a partial interface.

The behavior of [Exposed] and [SecureContext] on partial interfaces is already
defined in the relevant algorithms.

The behavior of [OverrideBuiltins] is defined here with the same vague wording
used before. It might be better to disentangle the concept of an interface and
its syntactical structure.

Fixes #154.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/620.html" title="Last updated on Jan 16, 2019, 1:12 PM UTC (c0d8b12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/620/ce7b02c...Ms2ger:c0d8b12.html" title="Last updated on Jan 16, 2019, 1:12 PM UTC (c0d8b12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1037,28 +1037,6 @@ Note: A partial interface definition cannot specify that the interface
 Inheritance must be specified on the original [=interface=]
 definition.
 
-[=Extended attributes=] can be specified on
-[=partial interface=] definitions, with some
-limitations.  The following extended attributes must not
-be specified on partial interface definitions:
-[{{Constructor}}],
-[{{LegacyWindowAlias}}],
-[{{NamedConstructor}}],
-[{{NoInterfaceObject}}].
-
-Note: The above list of [=extended attributes=]
-is all of those defined in this document that are applicable to
-[=interfaces=] except for
-[{{Exposed}}],
-[{{Global}}],
-[{{OverrideBuiltins}}], and
-[{{SecureContext}}].
-
-Any [=extended attribute=] specified
-on a [=partial interface=] definition
-is considered to appear on the [=interface=]
-itself.
-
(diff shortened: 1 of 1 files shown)


domenic:
Are there any existing usages of [OverrideBuiltins] on a partial interface? My suspicion is not, in which case, it'd be better to ban it.

Ms2ger:
```
[OverrideBuiltins]
partial interface Document {
```

domenic:
Fair :/. I wonder if moving the annotation to DOM would be an improvement or just weird to have things spread out. Probably not worth addressing at this time.

annevk:
Given it's on the legacy list per #350 it doesn't seem worth bothering about it too much unless it starts blocking something else.

PR #622, opened 2019-01-16
Disallow a default value for required dictionary members in the grammar.

### LINKED ISSUE #569: We should disallow a default value for required dictionary members in the grammar
When we added TypeWithExtendedAttributes, we split up the DictionaryMember production into two cases:

    required TypeWithExtendedAttributes identifier Default ;
    Type identifier Default ;

We should remove the "Default" from the "required" case, since there is already prose that says:

    A required dictionary member must not have a default value.

It's just that back when we had a single grammar production for dictionary members, with an optional "required" keyword at the front and "Default" at the end, this was not forbidden on the grammar level.  But now it would be simple to forbid it there.

Ms2ger:
Fixes #569.


annevk:
It seems we can also remove

> A required dictionary member must not have a default value.

then, no?

Issue #607, opened 2019-01-11
Named properties object / named property visibility algorithm with property on Window.prototype


Ms2ger:
When there's a property on `Window.prototype`, the named property visibility algorithm called on `window` will return false in step 5.1, when *prototype* is `Window.prototype`. This implies that [[GetOwnProperty]] on the named properties object will fall back to calling `OrdinaryGetOwnProperty(O, P)`, which returns undefined.

However, the browsers I tested consistently return the result of the named getter.

Consider [the following code](http://software.hixie.ch/utilities/js/live-dom-viewer/saved/6505):

```html
<!DOCTYPE html>
<span id=foo></span>
...<script>
var wp = window.__proto__;
var npo = window.__proto__.__proto__;
wp.foo = 7;
w(wp)
w(npo)
w(window.foo)
w(wp.foo)
>>>> w(npo.foo);

w("------")

var ep = npo.__proto__;
ep.bar = 7;
w(ep)
w(window.bar)
w(wp.bar)
w(npo.bar);
w(ep.bar);
</script>
```

This logs the same in Firefox, Chrome dev and Epiphany:

```
log: object "[object WindowPrototype]" (4 props: foo=7, addEventListener=function addEventListener() {\n    [native code]\n}, removeEventListener=function removeEventListener() {\n    [native code]\n}, dispatchEvent=function dispatchEvent() {\n    [native code]\n})
log: object "[object WindowProperties]" (3 props: addEventListener=function addEventListener() {\n    [native code]\n}, removeEventListener=function removeEventListener() {\n    [native code]\n}, dispatchEvent=function dispatchEvent() {\n    [native code]\n})
log: 7
log: 7
>>>> log: object "[object HTMLSpanElement]" (242 props: click=function click() {\n    [native code]\n}, focus=function focus() {\n    [native code]\n}, blur=function blur() {\n    [native code]\n}, title=""...)
log: ------
```

(Edit: ignore the results below; Chrome is doing something weirder and I didn't notice because I failed to actually include the span.)

<del>

```
log: object "[object EventTargetPrototype]" (4 props: addEventListener=function addEventListener() {\n    [native code]\n}, removeEventListener=function removeEventListener() {\n    [native code]\n}, dispatchEvent=function dispatchEvent() {\n    [native code]\n}, bar=7)
log: 7
log: 7
log: 7
log: 7
```

</del>

annevk:
[...]

PR #625, opened 2019-01-17
Check ignoreNamedProps before the named property visibility algorithm.

Ms2ger:
Running the named property visibility algorithm is (or should be) observable
by inserting a proxy in the prototype chain. This changes the spec to match
Gecko's behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/625.html" title="Last updated on Jan 31, 2019, 8:21 AM UTC (4f60f42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/625/45438a1...Ms2ger:4f60f42.html" title="Last updated on Jan 31, 2019, 8:21 AM UTC (4f60f42)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12741,30 +12741,31 @@ internal method as follows.
             1.  Set |desc|.\[[Enumerable]] and |desc|.\[[Configurable]] to <emu-val>true</emu-val>.
             1.  Return |desc|.
         1.  Set |ignoreNamedProps| to true.
-    1.  If |O| [=support named properties|supports named properties=],
-        the result of running the [=named property visibility algorithm=] with
-        property name |P| and object |O| is true, and |ignoreNamedProps| is false, then:
-        1.  Let |operation| be the operation used to declare the named property getter.
-        1.  Let |value| be an uninitialized variable.
-        1.  If |operation| was defined without an [=identifier=], then
-            set |value| to the result of performing the steps listed in the interface description to
-            [=determine the value of a named property=]
-            with |P| as the name.
-        1.  Otherwise, |operation| was defined with an identifier.  Set |value| to the result
-            of performing the steps listed in the description of |operation| with |P| as the only argument value.
-        1.  Let |desc| be a newly created [=Property Descriptor=] with no fields.
-        1.  Set |desc|.\[[Value]] to the result of [=converted to an ECMAScript value|converting=]
-            |value| to an ECMAScript value.
-        1.  If |O| implements an interface with a [=named property setter=], then set
-            |desc|.\[[Writable]] to <emu-val>true</emu-val>, otherwise set it to
-            <emu-val>false</emu-val>.
-        1.  If |O| implements an interface with the
-            [{{LegacyUnenumerableNamedProperties}}]
-            [=extended attribute=],
-            then set |desc|.\[[Enumerable]] to <emu-val>false</emu-val>,
-            otherwise set it to <emu-val>true</emu-val>.
-        1.  Set |desc|.\[[Configurable]] to <emu-val>true</emu-val>.
-        1.  Return |desc|.
+    1.  If |O| [=support named properties|supports named properties=] and
+        |ignoreNamedProps| is false, then:
+        1.  If the result of running the [=named property visibility algorithm=] with
+            property name |P| and object |O| is true, then:
+            1.  Let |operation| be the operation used to declare the named property getter.
+            1.  Let |value| be an uninitialized variable.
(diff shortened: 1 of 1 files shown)


Ms2ger:
This assumes that we all agree that "and" short-circuits; I'm not actually sure that's well-defined.

bzbarsky:
Funtimes.  I think it might be clearer to just nest:

2. If O supports named properties and ignoreNamedProps is false, then:
   1. If the result of running the named property visibility algorithm with property name P and object O is true, then:

Either way, we should probably have tests.

annevk:
I somewhat prefer @bzbarsky's phrasing. If there can be side effects when determining a particular thing, let's do it as a separate step.

cc @domenic

Ms2ger:
Fixed.

index.bs
```diff
@@ -12741,30 +12741,32 @@ internal method as follows.
...
-            [=extended attribute=],
-            then set |desc|.\[[Enumerable]] to <emu-val>false</emu-val>,
-            otherwise set it to <emu-val>true</emu-val>.
-        1.  Set |desc|.\[[Configurable]] to <emu-val>true</emu-val>.
-        1.  Return |desc|.
+    1.  If |O| [=support named properties|supports named properties=] and
+        |ignoreNamedProps| is false:
+        1.  Let |visible| be the result of running the [=named property visibility algorithm=] with
+            property name |P| and object |O|.
+        1.  If |visible| is true:
```
annevk: Combine these two steps?

index.bs
```diff
@@ -12741,30 +12741,32 @@ internal method as follows.
...
-            <emu-val>false</emu-val>.
-        1.  If |O| implements an interface with the
-            [{{LegacyUnenumerableNamedProperties}}]
-            [=extended attribute=],
-            then set |desc|.\[[Enumerable]] to <emu-val>false</emu-val>,
-            otherwise set it to <emu-val>true</emu-val>.
-        1.  Set |desc|.\[[Configurable]] to <emu-val>true</emu-val>.
-        1.  Return |desc|.
+    1.  If |O| [=support named properties|supports named properties=] and
+        |ignoreNamedProps| is false:
```
annevk: Usually if statements are followed by: ", then:". Or does IDL use a different style?

bzbarsky:
@littledan It's worth testing in Safari.  Testing in Chrome... given https://github.com/heycam/webidl/issues/607#issuecomment-453564230 Chrome is not even trying to follow the spec here, so I doubt testing in Chrome will tell us much, but we could try.  Interpreting the results of black-box tests is pretty complicated when you have no idea what the underlying model might look like.

Ms2ger:
I haven't yet been able to observe the named property visibility algorithm in either Chrome or WebKit, which makes this change unobservable there.

<details>
<summary>Test case</summary>

```html
<!DOCTYPE html>
<span id=x></span>
<script>
var c = document.getElementsByTagName("span");
c.__proto__ = new Proxy(Storage.prototype, {
  has(_, p) { w("has " + p); return true; },
  get(_, p) { w("get " + p); return "from proto" },
});
w("===========")
w(c.x)
w("===========")
c["x"] = "foo";
w("===========")
w(c.x)
w("===========")
</script>
```
Chrome:
```
log: ===========
log: get x
log: from proto
log: ===========
log: ===========
log: get x
log: from proto
log: ===========
```
WebKit:
```
log: ===========
log: object "[object HTMLSpanElement]" (208 props: title="", lang="", translate=true, dir=""...)
log: ===========
log: ===========
log: object "[object HTMLSpanElement]" (208 props: title="", lang="", translate=true, dir=""...)
log: ===========
```
Gecko:
```
log: ===========
log: has x
log: get x
log: from proto
log: ===========
log: has x
log: ===========
log: has x
log: get x
log: from proto
log: ===========
```
</details>

bzbarsky:
I'm a little confused by that testcase, actually.  Shouldn't it use a proxy for "localStorage" or "sessionStorage" or something, not "Storage.prototype"?  There should be no named property visibility stuff involved with "Storage.prototype".

Ms2ger:
I'm more than a little confused by it myself now; looks like I somehow merged two different test cases. Should be better now, and still shows that the change isn't observable in Chrome/WebKit.

index.bs
annevk: I'd like this to be addressed, given that we do use then on the subsequent line. Perhaps some kind of editorial issue needs to be filed to make sure we're consistent.

(As far as I can tell this PR is otherwise okay.)

PR #626, opened 2019-01-18
Improve the named property visibility algorithm for Window.

### LINKED ISSUE #607: Named properties object / named property visibility algorithm with property on Window.prototype

Ms2ger:
In particular, if a string is a supported property name as well as an own
property of the window or Window.prototype object, we don't need to block
that property from existing on the named properties object, since it will
be shadowed automatically.

Fixes #607.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/626.html" title="Last updated on Feb 20, 2019, 11:13 AM UTC (dcab4de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/626/7769c5f...Ms2ger:dcab4de.html" title="Last updated on Feb 20, 2019, 11:13 AM UTC (dcab4de)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10964,7 +10964,7 @@ is the concatenation of the [=interface=]’s
         Note: For example, if the [=interface=] is the {{Window}} interface,
         then the sole object will be this global environment’s window object.
     1.  If the result of running the [=named property visibility algorithm=] with
-        property name |P|, object |object| and object |O| is true, then:
+        |P|, |object|, and |O| is true, then:
         1.  Let |operation| be the operation used to declare the named property getter.
         1.  Let |value| be an uninitialized variable.
         1.  If |operation| was defined without an [=identifier=], then
@@ -12545,7 +12545,7 @@ Issue: Define those internal methods imperatively instead.
     1.  If |O| [=support named properties|supports named properties=],
         |O| does not implement an [=interface=] with the [{{Global}}] [=extended attribute=]
         and the result of calling the [=named property visibility algorithm=]
-        with property name |P| and object |O| is true, then:
+        with |P| and |O| is true, then:
         1.  If |O| does not implement an interface with a [=named property deleter=], then return <emu-val>false</emu-val>.
         1.  Let |operation| be the operation used to declare the named property deleter.
         1.  If |operation| was defined without an [=identifier=], then:
@@ -12643,7 +12643,7 @@ internal method as follows.
     1.  If |P| is not a [=supported property name=] of |O|, then return false.
     1.  If |propertyHolder| has an own property named |P|, then return false.
 
-          Note: This will include cases in which |O| has unforgeable properties,
+          Note: This will include cases in which |propertyHolder| has unforgeable properties,
           because in practice those are always set up before objects have any supported property names,
           and once set up will make the corresponding named properties invisible.
 
@@ -12742,7 +12742,7 @@ internal method as follows.
     1.  If |O| [=support named properties|supports named properties=] and
         |ignoreNamedProps| is false, then:
         1.  If the result of running the [=named property visibility algorithm=] with
-            property name |P| and object |O| is true, then:
+            |P| and |O| is true, then:
             1.  Let |operation| be the operation used to declare the named property getter.
             1.  Let |value| be an uninitialized variable.
             1.  If |operation| was defined without an [=identifier=], then
--- a/index.bs
+++ b/index.bs
@@ -10964,7 +10964,7 @@ is the concatenation of the [=interface=]’s
         Note: For example, if the [=interface=] is the {{Window}} interface,
         then the sole object will be this global environment’s window object.
     1.  If the result of running the [=named property visibility algorithm=] with
-        property name |P| and object |object| is true, then:
+        property name |P|, object |object| and object |O| is true, then:
         1.  Let |operation| be the operation used to declare the named property getter.
         1.  Let |value| be an uninitialized variable.
         1.  If |operation| was defined without an [=identifier=], then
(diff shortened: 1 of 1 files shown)


Ms2ger:
https://github.com/web-platform-tests/wpt/pull/14936

index.bs
```diff
@@ -10963,7 +10963,7 @@ is the concatenation of the [=interface=]’s
         Note: For example, if the [=interface=] is the {{Window}} interface,
         then the sole object will be this global environment’s window object.
     1.  If the result of running the [=named property visibility algorithm=] with
-        property name |P| and object |object| is true, then:
+        property name |P|, object |object| and object |O| is true, then:
```
domenic: Let's remove the types from the call site: "with P, object, and O is true".

index.bs
```diff
@@ -12639,21 +12639,22 @@ internal method as follows.
...
-    The algorithm operates as follows, with property name |P| and object |O|:
+    The algorithm operates as follows, with property name |P|, object |O|, and
+    optionally object |propertyHolder|:
 
+    1.  If |propertyHolder| was not specified, set |propertyHolder| to |O|.
     1.  If |P| is not a [=supported property name=] of |O|, then return false.
-    1.  If |O| has an own property named |P|, then return false.
+    1.  If |propertyHolder| has an own property named |P|, then return false.
 
           Note: This will include cases in which |O| has unforgeable properties,
```
domenic: This note should now refer to |propertyHolder|, right?

index.bs
```diff
@@ -12639,21 +12639,22 @@ internal method as follows.
...
           because in practice those are always set up before objects have any supported property names,
           and once set up will make the corresponding named properties invisible.
 
     1.  If |O| implements an interface that has the [{{OverrideBuiltins}}]
         [=extended attribute=], then return true.
-    1.  Let |prototype| be |O|.\[[GetPrototypeOf]]().
+    1.  Let |prototype| be |propertyHolder|.\[[GetPrototypeOf]]().
     1.  While |prototype| is not null:
-        1.  If |prototype| is not a [=named properties object=],
-            and |prototype| has an own property named |P|, then return false.
```
domenic: Are we sure that removing this clause is safe for the other 3 or so call sites of "named property visiblity algorithm"?

Ms2ger: It'll only affect the other call sites if someone takes an object that supports named properties and sticks the named properties object into its proto chain. I don't think we need to care particularly about that scenario.

annevk: Well, it's observable, right? If so, what do implementations do?

Ms2ger: IIRC, it's only observable in Gecko, and Gecko matches the new spec text.

Ms2ger:
Any more thoughts?

index.bs
```diff
@@ -12636,21 +12636,22 @@ internal method as follows.
     is used to determine if a given named property is exposed on an object.
     Some named properties are not exposed on an object depending on whether
     the [{{OverrideBuiltins}}] [=extended attribute=] was used.
-    The algorithm operates as follows, with property name |P| and object |O|:
+    The algorithm operates as follows, with property name |P|, object |O|, and
+    optionally object |propertyHolder|:
```
annevk, line 12640: Where is this invoked with _propertyHolder_?

domenic:
@shvaikalesh this looks like it's in the same area as other stuff you've been doing. Would you be interested in picking this up (if it's still applicable)?

annevk:
That's great! (I didn't mean to close this by the way, it seems that somehow the branch rename didn't affect this PR and then when the master branch was removed it got automatically closed.)

ExE-Boss:
> it seems that somehow the branch rename didn't affect this PR and then when the master branch was removed it got automatically closed.

That happens when the fork repository or branch was deleted, leaving behind a dangling PR.

Issue #632, opened 2019-01-29
Better regular expression for identifiers


annevk:
See https://github.com/heycam/webidl/pull/612#pullrequestreview-197260832. Perhaps we need something that disallows `--`, but still meets the other criteria.

(This seems fairly minor since I'd expect those writing standards to keep to relatively good practice.)

saschanaz:
What was reason not to choose `[_-]?[A-Za-z][0-9A-Z_a-z-]*`?

peria:
I agree with @saschanaz . We need to discuss which is better before merging the patch.
Obviously, `_?[A-Za-z-][0-9A-Z_a-z-]*` allows `_-` and `--`, while `[_-]?[A-Za-z][0-9A-Z_a-z-]*` doesn't.

annevk:
This was proposed in https://github.com/heycam/webidl/issues/564#issuecomment-402591724 and nobody suggested a different take, so I'm not sure. Is there a reason to prefer that one? So CSS doesn't need `_` all over I guess?

peria:
As I said in the comment, that regexp has a concern to accept strange strings as identifiers. From this point of view, "`[_-]?[A-Za-z][0-9A-Z_a-z-]*`" is better, and now I prefer this.

In addition, leading `_` has an important roll in WebIDL spec.

>  the identifier is the value of the identifier token with any leading U+005F LOW LINE ("_") character (underscore) removed.

> If one of these keywords is used, it need not be escaped with a leading underscore.

So if we accept "`__`", should it be treated as "`_`"?

saschanaz:
>So if we accept "__", should it be treated as "_"?

The answer is probably yes, but I think that would be meaningless because [currently identifiers cannot start with `_`](https://heycam.github.io/webidl/#idl-names):
   >The identifier of any of the abovementioned IDL constructs must not be "constructor", "toString", or begin with a U+005F LOW LINE ("_") character. These are known as reserved identifiers.

saschanaz:
Looks like this can be closed now.

annevk:
Fixed by 81dc0403a56e8b3a66eaab869704d635b39c0fe2.

PR #633, opened 2019-01-30
Disallow underscore-and-hyphen only identifiers

peria:
This works for #632.
The previous regular expression accepts strange tokens, i.e. `_-` or `-123`, as identifier tokens. The new expression does not accept them.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13557,7 +13557,7 @@ expression syntax [[!PERLRE]]) as follows:
     <tr>
         <td id="prod-identifier"><emu-t class="regex">identifier</emu-t></td>
         <td><code>=</code></td>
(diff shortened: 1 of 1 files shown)


annevk:
It does not allow `-123` afaict.

annevk:
And this would allow a leading `-` which apparently isn't usual in other IDL languages, but could be okay here given that we only use `-` for negative numbers thus far.

peria:
It does. ` _?` matches ""(empty), `[A-Za-z-]` matches "-", and `[0-9A-Z_a-z-]*` matches "123". It also matches "`---`" in the same way.

I agree it is the only use case, and is unusual in other IDLs. But we already have it in CSSOM spec.
Andf for numbers, they are not identifer tokens, and our previous (before #612) regular expression did not allow them.

bzbarsky:
Looks fine to me too.  Thank you!  Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1524041

PR #635, opened 2019-01-31
Add an algorithm to create a platform object.

Ms2ger:
***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/635.html" title="Last updated on Feb 22, 2019, 2:01 PM UTC (976b7ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/635/7769c5f...976b7ff.html" title="Last updated on Feb 22, 2019, 2:01 PM UTC (976b7ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -125,6 +125,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: DefinePropertyOrThrow; url: sec-definepropertyorthrow
         text: DetachArrayBuffer; url: sec-detacharraybuffer
         text: Get; url: sec-get-o-p
+        text: GetFunctionRealm; url: sec-getfunctionrealm
         text: GetIterator; url: sec-getiterator
         text: GetMethod; url: sec-getmethod
         text: IsAccessorDescriptor; url: sec-isaccessordescriptor
@@ -189,6 +190,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
             text: internal method
             text: internal slot
         text: Number type; url: sec-ecmascript-language-types-number-type
+        text: Object; for: ECMAScript; url: sec-object-type
         for: ordinary object; url: sec-ordinary-object-internal-methods-and-internal-slots
             text: internal method
             text: internal slot
@@ -869,6 +871,19 @@ An interface must not be declared such that
 its inheritance hierarchy has a cycle.  That is, an interface
 |A| cannot inherit from itself, nor can it inherit from another
 interface |B| that inherits from |A|, and so on.
+
+<div algorithm>
+  The [=list=] of <dfn for=interface>inclusive inherited interfaces</dfn> of an [=interface=] |I|
+  is defined as follows:
+
+    1.  Let |result| be « ».
+    1.  Let |interface| be |I|.
+    1.  While |interface| is not null:
+        1.  [=list/Append=] |interface| to |result|.
+        1.  Set |interface| to the [=interface=] that |I| [=interface/inherits=] from, if any, and
+            null otherwise.
+    1.  Return |result|.
+</div>
 
 Note that general multiple inheritance of interfaces is not supported, and
 objects also cannot implement arbitrary sets of interfaces.
@@ -10700,11 +10715,14 @@ the <code>typeof</code> operator will return "function" when applied to an inter
         1.  Let &lt;|constructor|, |values|&gt; be the result
             of passing |S| and |args|.
             to the [=overload resolution algorithm=].
-        1.  Let |R| be the result of performing
-            the actions listed in the description of |constructor|
-            with |values| as the argument values.
+        1.  Let |object| be the result of [=internally create a new object implementing the
+            interface|internally creating a new object implementing=] |I|, with |realm| and
+            {{NewTarget}}.
+        1.  Perform the actions listed in the description of |constructor|
+            with |values| as the argument values
+            and |object| as the <emu-val>this</emu-val> value.
             Rethrow any exceptions.
-        1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |R|
+        1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |object|
             to an ECMAScript [=interface type=] value |I|.
         1.  Assert: |O| is an object that implements |I|.
         1.  Assert: |O|.\[[Realm]] is equal to |realm|.
@@ -12347,12 +12365,65 @@ It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+  To <dfn export lt=new>create a new object implementing the interface</dfn> |interface|, with a
+  Realm |realm|, perform the following steps:
+
+    1.  Return the result of [=internally create a new object implementing the
+        interface|internally creating a new object implementing=] |interface|, with |realm| and
+        <emu-val>undefined</emu-val>.
+</div>
+
+<div algorithm>
+  To <dfn>internally create a new object implementing the interface</dfn> |interface|, with a
+  Realm |realm| and a JavaScript value |newTarget|, perform the following steps:
+
+    1.  Assert: |interface| is [=exposed=] in |realm|.
+    1.  If |newTarget| is <emu-val>undefined</emu-val>, then:
+        1.  Let |prototype| be the [=interface prototype object=] for |interface| in
+            |realm|.
+    1.  Otherwise:
+        1.  Assert: [$IsCallable$](|newTarget|) is true.
+        1.  Let |prototype| be [=?=] [$Get$](|newTarget|, "prototype").
+        1.  If [$Type$](|prototype|) is not Object, then:
+            1.  Let |targetRealm| be [$GetFunctionRealm$](|newTarget|).
+            1.  Set |prototype| to the [=interface prototype object=] for |interface| in
+                |targetRealm|.
+    1.  Let |instance| be a newly created [=ECMAScript/object=] in |realm|.
+    1.  Set |instance|.\[[Prototype]] to |prototype|.
+    1.  Set |instance|'s essential internal methods to the definitions specified in
+        [=ECMA-262 Ordinary object internal methods and internal slots=].
+    1.  Let |interfaces| be the [=inclusive inherited interfaces=] of |interface|.
+    1.  [=list/iterate|For every=] [=interface=] |ancestor interface| in |interfaces|:
+        1.  [=Define the unforgeable regular operations=] of |ancestor interface| on |instance|,
+            given |realm|.
+        1.  [=Define the unforgeable regular attributes=] of |ancestor interface| on |instance|,
+            given |realm|.
+    1.  If |interface| is declared with the [{{Global}}] [=extended attribute=], then:
+        1.  [=Define the regular operations=] of |interface| on |instance|, given |realm|.
+        1.  [=Define the regular attributes=] of |interface| on |instance|, given |realm|.
+        1.  Set |instance|.\[[SetPrototypeOf]] as defined in [[#platform-object-setprototypeof]].
+    1.  Otherwise, if |interfaces| contains an [=interface=] which
+        [=support indexed properties|supports indexed properties=],
+        [=support named properties|named properties=], or both:
+        1.  Set |instance|.\[[GetOwnProperty]] as defined in
+            [[#legacy-platform-object-getownproperty]].
+        1.  Set |instance|.\[[Set]] as defined in
+            [[#legacy-platform-object-set]].
+        1.  Set |instance|.\[[DefineOwnProperty]] as defined in
+            [[#legacy-platform-object-defineownproperty]].
+        1.  Set |instance|.\[[Delete]] as defined in
+            [[#legacy-platform-object-delete]].
+        1.  Set |instance|.\[[PreventExtensions]] as defined in
+            [[#legacy-platform-object-preventextensions]].
+        1.  Set |instance|.\[[OwnPropertyKeys]] as defined in
+            [[#legacy-platform-object-ownpropertykeys]].
+    1.  Return |instance|.
+</div>
+
 The <dfn id="dfn-primary-interface" export>primary interface</dfn> of a platform object
 that implements one or more interfaces is the most-derived [=interface=]
-that it implements.  The value of the \[[Prototype]] [=internal slot=]
-of the platform object is the [=interface prototype object=]
-of the [=primary interface=]
-from the [=platform object=]’s associated global environment.
+that it implements.
 
 The global environment that a given [=platform object=]
 is associated with can <dfn id="dfn-change-global-environment" for="global environment" export>change</dfn> after it has been created.  When
@@ -12367,24 +12438,6 @@ a platform object that implements one or more interfaces
 must be the [=qualified name=] of
 the [=primary interface=]
 of the platform object.
-
-[=Platform objects=] which implement an [=interface=] which has a [{{Global}}]
-[=extended attribute=] have an internal \[[SetPrototypeOf]] method
-as defined in [[#platform-object-setprototypeof]].
-
-Within a [=Realm=] |realm|,
-for each [=interface=] |interface| implemented by a [=platform object=] |obj|,
-the following steps must be run as part of |obj|'s creation:
-
-1.  [=Define the unforgeable regular operations=] of |interface| on |obj|, given |realm|.
-1.  [=Define the unforgeable regular attributes=] of |interface| on |obj|, given |realm|.
-
-If within a [=Realm=] |realm| a [=platform object=] |obj| implements
-an interface |interface| which is declared with the [{{Global}}] [=extended attribute=], then
-the following steps must be run as part of |obj|'s creation:
-
-1.  [=Define the regular operations=] of |interface| on |obj|, given |realm|.
-1.  [=Define the regular attributes=] of |interface| on |obj|, given |realm|.
 
 Additionally, [=platform objects=] which implement an [=interface=]
 which has a [{{Global}}] [=extended attribute=]
@@ -12447,13 +12500,11 @@ Support for [=getters=] is handled in [[#legacy-platform-object-getownproperty]]
 and for [=setters=] in [[#legacy-platform-object-defineownproperty]]
 and [[#legacy-platform-object-set]].
 
-Additionally, [=legacy platform objects=] must have internal methods as defined in:
+Additionally, [=legacy platform objects=] have internal methods as defined in:
 
 * [[#legacy-platform-object-delete]],
 * [[#legacy-platform-object-preventextensions]], and
 * [[#legacy-platform-object-ownpropertykeys]].
-
-Issue: Define those internal methods imperatively instead.
 
 
 <h4 id="legacy-platform-object-getownproperty" dfn oldids="getownproperty" lt=GetOwnProperty for="legacy platform object">\[[GetOwnProperty]]</h4>
@@ -13325,10 +13376,8 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
 The <dfn constructor for="DOMException"><code>DOMException(|message|, |name|)</code></dfn>
 constructor, when invoked, must run these steps:
 
-1. Let |e| be a new {{DOMException}} object.
-1. Set |e|'s [=DOMException/name=] to |name|.
-1. Set |e|'s [=DOMException/message=] to |message|.
-1. Return |e|.
+1. Set <emu-val>this</emu-val>'s [=DOMException/name=] to |name|.
+1. Set <emu-val>this</emu-val>'s [=DOMException/message=] to |message|.
 
 The <dfn attribute for="DOMException"><code>name</code></dfn> attribute's getter must return this
 {{DOMException}} object's [=DOMException/name=].


Ms2ger:
This is partially based on the algorithm in #595, and there's still a bunch of declarative requirements that could move into this, but I wanted to take small steps.

index.bs
```diff
@@ -12349,6 +12362,48 @@ It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+  To <dfn export lt="new|create|create a new|create an object implementing|create an object
+  implementing the interface">create an object implementing the interface</dfn> |interface| in the
```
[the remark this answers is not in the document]
Ms2ger: I'm happy to go whichever way here.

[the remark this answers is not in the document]
littledan: At the moment, I think @Ms2ger 's approach is best to allow for existing specs. Over time, we can try to migrate to a preferred wording.

index.bs
```diff
@@ -12349,12 +12363,60 @@ It is the responsibility of specifications using Web IDL to state
...
+  Realm |realm|, with optional `new.target` value |newTarget|, perform the following steps:
+
+    1.  Assert: |interface| is [=exposed=] in |realm|.
+    1.  Let prototype be the [=interface prototype object=] for |interface| in
+        |realm|.
+    1.  If |newTarget| is provided, then:
+        1.  Issue: If we just pass {{NewTarget}} through in a constructor, would this not
+            always be defined (sometimes to the [=interface object=] itself)?
```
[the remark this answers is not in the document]
littledan: I'm not sure if it makes sense to take _newTarget_ as a parameter like this, as opposed to somehow retrieving it from the ambient algorithm. If we do that, then I imagine we'd have _newTarget_ provided if this algorithm is being invoked from a constructor, and we wouldn't have it provided if it's being invoked from some other algorithm.

index.bs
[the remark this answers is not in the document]
annevk: I think it's easier if we allow a single style. Then effectively everything is old and can be updated one-by-one to point here instead (I'm assuming we want it linked in due course). Allowing and exporting lots of styles does not help existing specifications (they still need to be updated to link) and also does not give them  guidance on style.

index.bs
```diff
@@ -12366,14 +12366,21 @@ object is associated with.
...
-  Realm |realm|, perform the following steps:
+  Realm |realm|, with optional `new.target` value |newTarget|, perform the following steps:
 
     1.  Assert: |interface| is [=exposed=] in |realm|.
+    1.  Let prototype be the [=interface prototype object=] for |interface| in
+        |realm|.
+    1.  If |newTarget| is provided, then:
+        1.  Issue: If we just pass {{NewTarget}} through in a constructor, would this not
+            always be defined (sometimes to the [=interface object=] itself)?
+        1.  Let |newTargetPrototype| be [=?=] [$Get$](|newTarget|, "prototype").
```
littledan: The logic here is a bit different from the JS spec, which is based on [OrdinaryCreateFromConstructor](https://tc39.github.io/ecma262/#sec-ordinarycreatefromconstructor).

Actually, it seems like we've opened up here a whole world of divergence between JS and the Web, as well as between browsers. Consider this test case:

```js
function foo() { }
function bar() { } bar.prototype = 1;

console.log(`
  ${ Reflect.construct(URL, ["file:///foo"], foo).constructor.name }
  ${ Reflect.construct(URL, ["file:///foo"], bar).constructor.name }
  ${ Reflect.construct(Uint8Array, [1], foo).constructor.name }
  ${ Reflect.construct(Uint8Array, [1], bar).constructor.name }
`);
```

WebKit results:

```
  URL
  URL
  foo
  Uint8Array
```

Chromium results:

```
  foo
  Object
  foo
  Uint8Array
```

Gecko results:

```
  foo
  URL
  foo
  Uint8Array
```

The Gecko results have the closest analogy to JavaScript here--they are as if we're following the OrdinaryCreateFromConstructor algorithm, with all of the platform interfaces included in the fallback list. If Chromium and WebKit folks are in favor of it, I'd suggest adopting these semantics.

index.bs
```diff
@@ -12349,12 +12363,54 @@ It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+  To create a <dfn export>new</dfn> object implementing the interface |interface| in the
+  Realm |realm|, with optional `new.target` value |newTarget|, perform the following steps:
```
domenic: I think we should not allow providing `new.target`. I cannot think of any specs that would use it.

index.bs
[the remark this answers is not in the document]
domenic: If I am understanding @littledan correctly, +1. Inside a constructor "Create a new X" should implicitly grab the NewTarget that exists. Elsewhere, it should just be the X constructor. In neither case should the caller of "new" pass anything.

(In the future I think constructors would be nicer if Web IDL created the object for you, and constructor spec text elsewhere just operated on the created object using a known variable name like `this` or something. But, that's a bigger project, and maybe not worth the churn. For now we have to work with the world where all constructors say "create a new X" as their first step or so.)

index.bs
domenic: Given the three-way divergence, and the extreme edge-casey nature of this `Reflect.construct` madness, I think it'll be pretty safe to pick one, write tests, and file bugs on everyone else. So I encourage us to pick Gecko. /cc @yuki3 @cdumez in case they want to weigh in, but IMO we shouldn't block.

Note that for custom elements we tried to copy ES more exactly: https://html.spec.whatwg.org/multipage/dom.html#html-element-constructors steps 7 and 8.

index.bs
```diff
@@ -12349,12 +12363,54 @@ It is the responsibility of specifications using Web IDL to state
...
+    1.  Let |interfaces| be the [=inclusive inherited interfaces=] of |interface|.
+    1.  [=list/iterate|For every=] [=interface=] |ancestor interface| in |interfaces|:
+        1.  [=Define the unforgeable regular operations=] of |ancestor interface| on |instance|,
+            given |realm|.
+        1.  [=Define the unforgeable regular attributes=] of |ancestor interface| on |instance|,
+            given |realm|.
+    1.  If |interface| is declared with the [{{Global}}] [=extended attribute=], then:
+        1.  [=Define the regular operations=] of |interface| on |instance|, given |realm|.
+        1.  [=Define the regular attributes=] of |interface| on |instance|, given |realm|.
+        1.  Set |instance|.\[[SetPrototypeOf]] as defined in [[#platform-object-setprototypeof]].
```
domenic: Oh, really sweet to make this imperative; great stuff.

index.bs
```diff
@@ -12370,24 +12426,6 @@ must be the [=qualified name=] of
...
-If within a [=Realm=] |realm| a [=platform object=] |obj| implements
-an interface |interface| which is declared with the [{{Global}}] [=extended attribute=], then
-the following steps must be run as part of |obj|'s creation:
-
-1.  [=Define the regular operations=] of |interface| on |obj|, given |realm|.
-1.  [=Define the regular attributes=] of |interface| on |obj|, given |realm|.
-
 Additionally, [=platform objects=] which implement an [=interface=]
 which has a [{{Global}}] [=extended attribute=]
 get properties declaratively from:
```
domenic: This can be deleted or rephrased too, I think.

index.bs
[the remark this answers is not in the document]
domenic: I guess we should also make the Realm optional inside constructors; it should always be the current Realm, and asking people to figure that out is a recipe for badness.

index.bs
```diff
@@ -13327,7 +13363,7 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
 The <dfn constructor for="DOMException"><code>DOMException(|message|, |name|)</code></dfn>
 constructor, when invoked, must run these steps:
 
-1. Let |e| be a new {{DOMException}} object.
+1. Let |e| be a [=new=] {{DOMException}} object in {{NewTarget}}'s Realm, with {{NewTarget}}.
```
domenic: It should be the current Realm, not NewTarget's realm. (But, see my comment above about making it optional.)

index.bs
yuki3: No objection.  Please go ahead.  :)

index.bs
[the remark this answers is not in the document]
annevk: (I think even if we improved how to define a constructor, you still need new, e.g., for tear-off objects and such.)

index.bs
Ms2ger: I tried to write a test for this, but failed. Would you mind giving me a few pointers?

index.bs
Ms2ger: I'm planning to define those properties imperatively in a follow-up, and then this can go, but for now, I'd keep it.

index.bs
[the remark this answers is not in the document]
Ms2ger: That would have some extremely surprising effects if interpreted to the letter. Consider the [`Response` constructor](https://fetch.spec.whatwg.org/#dom-response):

3. Let _r_ be a new `Response` object associated with a new response.
4. Set _r_’s headers to a new `Headers` object, whose header list is _r_’s response’s header list, and guard is "response".

If you just pull the NewTarget from the air, it would be used for both the `Response` and `Headers` objects.

index.bs
bzbarsky: Yeah, to be clear, the Gecko behavior, in the absence of an explicit spec, was written to match the ES semantics, since that was the only example of constructor behavior around.  I don't see any reason for deviating from those semantics in Web IDL, but obviously I'm biased, since I wrote the Gecko code here. ;)

index.bs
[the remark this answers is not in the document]
domenic: Yes, I guess you'd need to guard it to only apply to the creation of objects of the interface whose constructor is being interpreted.

index.bs
domenic: Oh wow, so, this is more complicated than I thought, and you are right that "in the current Realm" is not right. But I think just passing in "NewTarget's realm" might not be right either?

Here is what ES does, unrolled and inlined:

1. Let newTarget be NewTarget, if it is not undefined, or the active function object (i.e. the currently-executing DOMException constructor), if NewTarget is undefined.
1. Let proto be Get(newTarget, "prototype").
1. If Type(proto) is not Object, then:
   1. Let realm be GetFunctionRealm(newTarget).
   1. Set proto to the DOMException interface prototype object in realm.
1. Let object be ObjectCreate(proto, internalSlotsList).
1. ... do stuff to object ...
1. Return object.

In particular accounting for an undefined NewTarget (possible via Reflect.construct) seems like it makes just passing in NewTarget's realm not ideal.

I'm not sure what the ideal setup is, but here are my pointers:

1. We should have "new", with no arguments, do the right thing inside constructors. (When applied to the relevant constructor, as you point out.) By "the right thing" we mean, congruent with the unrolled ES spec I have done above.
1. In other cases, "new" should take a single Realm argument.

Let me know if that's enough to go on. Tricky stuff...

Ms2ger: I think we can assume that NewTarget is not undefined in a constructor, because of the check in https://heycam.github.io/webidl/#create-an-interface-object ; will consider the rest of your comment next week.

domenic: Ah, good catch, I was looking at NativeError, which has special behavior for undefined NewTarget. OK, that may simplify things, and make your existing text work. Let me edit my above unrolled and inlined comment to be more correct.

domenic: Yeah, OK, after writing that, it seems like the current Realm ends up not involved at all. So the only part of my comment that applies is that we should have good defaults, to avoid anyone else having to figure this out in the future :).

index.bs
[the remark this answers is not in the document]
Ms2ger: I still think that pulling the NewTarget from the ambient algorithm makes this unnecessarily hard to follow. Alternative proposal: when passing a NewTarget, we derive the realm from that. That at least means that specs defining constructors don't need to think about realms there.

index.bs
```diff
@@ -12346,12 +12361,54 @@ It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+  To create a <dfn export>new</dfn> object implementing the interface |interface|, with either
+  a Realm |realm| or a `new.target` value |newTarget|, perform the following steps:
+
+    1.  If |newTarget| is provided, then:
+        1.  Set |realm| to [$GetFunctionRealm$](|newTarget|).
```
Ms2ger: @bzbarsky makes an interesting point that this might not be the right realm after all. Assuming no other interfering bugs in the test case below, I see Gecko and WebKit using the constructor's realm, and Chrome using the NewTarget's.

<details>
<summary>Test</summary>

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/6603

```html
<!DOCTYPE html>
<script>
function f(i) {
var dp = new DOMParser();
w(dp.parseFromString("x", "text/html").documentURI)

dp = new i.contentWindow.DOMParser();
w(dp.parseFromString("x", "text/html").documentURI)

class X extends DOMParser {}

dp = new X();
w(dp.parseFromString("x", "text/html").documentURI)

class Y extends i.contentWindow.DOMParser {}

dp = new Y();
w(dp.parseFromString("x", "text/html").documentURI)
}
</script>
<iframe src=document onload=f(this)></iframe>
```

With `http://software.hixie.ch/utilities/js/` stripped from all results:

| Test | Gecko | WebKit | Chrome |
|-|-|-|-|
| parent `DOMParser` | `live-dom-viewer/` | `live-dom-viewer/` | `live-dom-viewer/` |
| subframe `DOMParser` | `live-dom-viewer/document` | `live-dom-viewer/document` | `live-dom-viewer/document` |
| subclass of parent `DOMParser` | `live-dom-viewer` | `live-dom-viewer` | `live-dom-viewer` |
| subclass of subframe `DOMParser` | `live-dom-viewer/document` | `live-dom-viewer/document` | `live-dom-viewer` |

</details>

index.bs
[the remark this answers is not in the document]
domenic: I don't quite understand the alternate proposal. Can you explain what would be required of spec authors using the "new" construct in your version?

From my perspective, I'm not very concerned with how hard to follow this algorithm is. I'm concerned about users of this algorithm, and ensuring they don't have to pass any explicit information in constructors, and only have to pass one thing (the realm) in other contexts.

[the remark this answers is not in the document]
Ms2ger: In the interest of making life easy for the users of the algorithm, I've pushed a change to do the object creation in WebIDL. If we're going to edit all the specifications that create objects anyway, we might as well make take the opportunity to simplify it as much as possible.

index.bs
```diff
@@ -12346,12 +12363,60 @@ It is the responsibility of specifications using Web IDL to state
...
+  To create a <dfn export>new</dfn> object implementing the interface |interface|, with a Realm
+  |realm|, perform the following steps:
+
+    1.  Return the result of creating a [=new-internal|new=] object implementing |interface|, with
+        |realm| and no target.
+</div>
+
+<div algorithm>
+  To create a <dfn lt=new-internal>new</dfn> object implementing the interface |interface|, with
+  a Realm |realm| and optionally a target |newTarget|, perform the following steps:
```
```suggestion
  a Realm |realm| and optionally a JavaScript function value |newTarget|, perform the following steps:
```

index.bs
```diff
@@ -13324,10 +13369,8 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
 The <dfn constructor for="DOMException"><code>DOMException(|message|, |name|)</code></dfn>
 constructor, when invoked, must run these steps:
 
-1. Let |e| be a new {{DOMException}} object.
-1. Set |e|'s [=DOMException/name=] to |name|.
-1. Set |e|'s [=DOMException/message=] to |message|.
-1. Return |e|.
+1. Set <emu-val>this</emu-val>'s [=DOMException/name=] to |name|.
```
domenic: Technically "this" is not really defined that well; that's https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301. However I support landing this as-is and as a follow-up working on that additional rigorization project. The text you have that says "as the this value" is pretty good for now.

In particular I think we'd want to define "this value" or "this" as something linkable, for both methods and constructors, and encourage a typography that either leaves it un-marked-up or cross-links to the Web IDL definition.

index.bs
```diff
@@ -10699,11 +10714,13 @@ the <code>typeof</code> operator will return "function" when applied to an inter
         1.  Let &lt;|constructor|, |values|&gt; be the result
             of passing |S| and |args|.
             to the [=overload resolution algorithm=].
-        1.  Let |R| be the result of performing
-            the actions listed in the description of |constructor|
-            with |values| as the argument values.
+        1.  Let |object| be a [=new-internal|new=] object implementing |I|, with |realm| and
```
domenic: Per my unrolling at https://github.com/heycam/webidl/pull/635#discussion_r254879231, it looks like we should use GetFunctionRealm(NewTarget), not _realm_, to match ES.

bzbarsky: Yeah, it looks like that's what ES does.  Given that slightly-older Firefox releases and Safari do something different (use current Realm), it's not clear to me whether the spec changed there or what.  But I agree we should just match ES.

index.bs
annevk: I'd prefer "this" over "this value" as it's extremely common.

index.bs
```diff
@@ -12346,12 +12364,60 @@ It is the responsibility of specifications using Web IDL to state
...
+        1.  Set |instance|.\[[Set]] as defined in
+            [[#legacy-platform-object-set]].
+        1.  Set |instance|.\[[DefineOwnProperty]] as defined in
+            [[#legacy-platform-object-defineownproperty]].
+        1.  Set |instance|.\[[Delete]] as defined in
+            [[#legacy-platform-object-delete]].
+        1.  Set |instance|.\[[PreventExtensions]] as defined in
+            [[#legacy-platform-object-preventextensions]].
+        1.  Set |instance|.\[[OwnPropertyKeys]] as defined in
+            [[#legacy-platform-object-ownpropertykeys]].
```
TimothyGu: This is slightly problematic, as I think we shouldn't access any internal methods (which define algorithms do) before they are fully initialized – at least ES doesn't. However, it seems certainly correct that the define algorithms should not be visible to these overridden internal methods.

Could we do something like
- Initialize these internal methods right after "Set *instance*'s essential internal methods"
- Use ValidateAndApplyPropertyDescriptor directly (or otherwise directly access the property table) instead of DefinePropertyOrThrow

I feel like this also matches implementations better, as I don't think they would directly call the internal methods during bootstrapping.

This also applies to the [[SetPrototypeOf]] assignment for [[Global]].

index.bs
```diff
@@ -12346,12 +12364,60 @@ It is the responsibility of specifications using Web IDL to state
 which global environment (or, by proxy, which global object) each platform
 object is associated with.
 
+<div algorithm>
+  To create a <dfn export>new</dfn> object implementing the interface |interface|, with a Realm
```
TimothyGu: I tend to prefer having a longer name for the &lt;dfn&gt; part, like "create a new object implementing an interface", which seems to be what we do with "create an operation function" etc. We can then have a shorter alternative link-text like just "new" or something, with a clarification in a note.

index.bs
```diff
@@ -12346,12 +12364,60 @@ It is the responsibility of specifications using Web IDL to state
...
+  |realm|, perform the following steps:
+
+    1.  Return the result of creating a [=new-internal|new=] object implementing |interface|, with
+        |realm| and no target.
+</div>
+
+<div algorithm>
+  To create a <dfn lt=new-internal>new</dfn> object implementing the interface |interface|, with
```
TimothyGu: Just to make it clearer, can we call this "to create internally" or something to that sense? Otherwise the call "create a new `URL` object in *realm*" could technically go into either of these definitions, as *newTarget* is optional here.

index.bs
Ms2ger: Please note the previous discussion about *not* providing multiple names for this.

index.bs
Ms2ger: This is already going wrong in the current spec text, whichever way you read it. I'm happy to fix the "define the attributes/operations" algorithms in a followup, but that's scope creep for this PR.

index.bs
annevk: I think what @TimothyGu is suggesting is that we still provide a single convention for callers `Let <var>x</var> be a <a for=/>new</a> <code>Y</code>`, but that the `<dfn>` part here is a bit longer. I'd be okay with that, though some additional rationale would help.

index.bs
```diff
@@ -869,6 +871,19 @@ its inheritance hierarchy has a cycle.  That is, an interface
 |A| cannot inherit from itself, nor can it inherit from another
 interface |B| that inherits from |A|, and so on.
 
+<div algorithm>
```
bzbarsky: I assume this can't just say "_I_ and _I_'s **inherited interfaces**" because we want a list, not a set?

index.bs
```diff
@@ -10699,14 +10714,18 @@ the <code>typeof</code> operator will return "function" when applied to an inter
         1.  Let &lt;|constructor|, |values|&gt; be the result
             of passing |S| and |args|.
             to the [=overload resolution algorithm=].
-        1.  Let |R| be the result of performing
-            the actions listed in the description of |constructor|
-            with |values| as the argument values.
+        1.  Let |targetRealm| be [$GetFunctionRealm$]({{NewTarget}}).
+        1.  Let |object| be the result of [=internally create a new object implementing the
```
bzbarsky: This looks wrong to me.

We want to create the object in the current Realm of the constructor, period.  That's how ES builtins work, last I checked, and we should not be inventing a new behavior.  All we should be using NewTarget for is the prototype object, which may end up coming from a different Realm.

In other words, we should be passing our current Realm, and then inside the "internally create" algorithm, if newTarget is not undefined, we can get the prototype from it, if an object use that, else fall back on the default proto in GetFunctionRealm(newTarget).  And fall back to the passed-in Realm if there is no newTarget.

index.bs
```diff
@@ -10699,14 +10714,18 @@ the <code>typeof</code> operator will return "function" when applied to an inter
...
+        1.  Perform the actions listed in the description of |constructor|
+            with |values| as the argument values
+            and |object| as the <emu-val>this</emu-val> value.
             Rethrow any exceptions.
-        1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |R|
+        1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |object|
             to an ECMAScript [=interface type=] value |I|.
         1.  Assert: |O| is an object that implements |I|.
-        1.  Assert: |O|.\[[Realm]] is equal to |realm|.
+        1.  Assert: |O|.\[[Realm]] is equal to |targetRealm|.
```
bzbarsky: Again, this is NOT the behavior we want.

index.bs
```diff
@@ -12346,12 +12365,62 @@ It is the responsibility of specifications using Web IDL to state
...
+    1.  Return the result of [=internally create a new object implementing the
+        interface|internally creating a new object implementing=] |interface|, with |realm| and
+        <emu-val>undefined</emu-val>.
+</div>
+
+<div algorithm>
+  To <dfn>internally create a new object implementing the interface</dfn> |interface|, with a
+  Realm |realm| and a JavaScript value |newTarget|, perform the following steps:
+
+    1.  Assert: |interface| is [=exposed=] in |realm|.
```
bzbarsky: Just as a note: this assert does NOT hold with the current definition of constructor behavior.  Last I checked, you could have a constructor from a Realm where the interface is exposed and a NewTarget from a Realm where the interface is _not_ exposed (at least with [SecureContext], where things can be same-origin yet different secure context state, last I checked) and the NewTarget Realm is what's getting passed in.  Yet another reason to not do that.  ;)

index.bs
```diff
@@ -12346,12 +12365,62 @@ It is the responsibility of specifications using Web IDL to state
...
+    1.  Let prototype be the [=interface prototype object=] for |interface| in
+        |realm|.
+    1.  If |newTarget| is not <emu-val>undefined</emu-val>, then:
```
bzbarsky: See above for my suggestions on how this should be modified.

index.bs
domenic: From what I can tell ES builtins actually create in the realm of NewTarget. See https://github.com/heycam/webidl/pull/635#discussion_r254879231 . I too was surprised.

Ms2ger: I used *realm* until you told me otherwise!

> Yeah, it looks like that's what ES does. Given that slightly-older Firefox releases and Safari do something different (use current Realm), it's not clear to me whether the spec changed there or what. But I agree we should just match ES.

bzbarsky: ES builtins get the default prototype to use if NewTarget's `.prototype` is not an object in the realm of NewTarget, which is what I thought https://github.com/heycam/webidl/pull/635#discussion_r254879231 was talking about.  But they create the object in the current Realm.  Which is not observable in ES so far (without weakrefs), because you can't observe a Realm for non-Function objects.  But if you look at the unrolled steps from ES, the ObjectCreate is happening at a point when the current Realm is the Realm of the constructor.

Note that ES is actually kinda internally inconsistent here.  In particular, if you look at https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget (the [[Construct]] for normal scripted functions), it does the ObjectCreate before entering the Realm of the function itself.  So this creation happens in whatever Realm you're in when you call the constructor, which can be neither the Realm of the constructor nor the Realm of the NewTarget.  This came up recently in https://phabricator.services.mozilla.com/D19760#519322 and is likely a spec issue if object realms become observable.

On the other hand, for a built-in function (like Map, say, or any of the Web IDL constructors) the [[Construct]] is defined as https://tc39.github.io/ecma262/#sec-built-in-function-objects-construct-argumentslist-newtarget which just defers to https://tc39.github.io/ecma262/#sec-built-in-function-objects-call-thisargument-argumentslist with step 10 changed.  Importantly, there is no object-creation here outside of step 10 doing whatever it does, and we enter the Realm of the constructor in step 9.  So let's look at the Map constructor, to consider something specific.  It's defined at https://tc39.github.io/ecma262/#sec-map-iterable and calls https://tc39.github.io/ecma262/#sec-ordinarycreatefromconstructor which calls `ObjectCreate` while in the Realm of the Map constructor itself.  Again, the proto comes from whatever NewTarget is, but the actual object comes from the current Realm.

index.bs
Ms2ger: Yeah. I've got a note on my todo list to see if we can get back to just one concept.

index.bs
Ms2ger: Okay, thanks for clarifying. I wrote some tests, and it turns out no two browsers agree. I updated the PR to match my understanding of what ES does, based on your comment. Note that this doesn't match Gecko if NewTarget has a prototype property that isn't an object.

<table>
<thead>
<tr><th>Test <th>Gecko <th>WebKit <th>Chrome
<tbody>
<tr><th colspan=4>new DOMParser()
<tr><th>Prototype <td colspan=3>DOMParser.&shy;prototype
<tr><th>Object realm <td colspan=3>parent window
<tr><th colspan=4>new i.&shy;contentWindow.&shy;DOMParser()
<tr><th>Prototype <td colspan=3>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype
<tr><th>Object realm <td colspan=3>child window
<tr><th colspan=4>Reflect.&shy;construct(i.&shy;contentWindow.&shy;DOMParser, [], DOMParser)
<tr><th>Prototype <td>DOMParser.&shy;prototype <td>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype <td>DOMParser.&shy;prototype
<tr><th>Object realm <td>child window <td>child window <td>parent window
<tr><th colspan=4>Reflect.&shy;construct(DOMParser, [], i.&shy;contentWindow.&shy;DOMParser)
<tr><th>Prototype <td>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype <td>DOMParser.&shy;prototype <td>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype
<tr><th>Object realm <td>parent window <td>parent window <td>child window
<tr><th colspan=4>Reflect.&shy;construct(i.&shy;contentWindow.&shy;DOMParser, [], cons)
<tr><th>Prototype <td>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype <td>i.&shy;contentWindow.&shy;DOMParser.&shy;prototype <td>Object.&shy;prototype
<tr><th>Object realm <td>child window <td>child window <td>parent window
<tr><th colspan=4>Reflect.&shy;construct(DOMParser, [], i.&shy;contentWindow.&shy;cons)
<tr><th>Prototype <td>DOMParser.&shy;prototype <td>DOMParser.&shy;prototype <td>i.&shy;contentWindow.&shy;Object.&shy;prototype
<tr><th>Object realm <td>parent window <td>parent window <td>child window
<tr><th colspan=4>Conclusion
<tr><th>Prototype <td>From NewTarget; fallback to interface object from constructor's realm <td>Interface object from constructor's realm <td>From NewTarget; fallback to Object.prototype from NewTarget's realm
<tr><th>Object realm <td>From constructor <td>From constructor <td>From NewTarget
</table>

<details>
<summary>Test</summary>

```html
<!DOCTYPE html>
<script>
function identify(proto, i) {
  if (proto === Object.prototype)
    return "Object.prototype";

  if (proto === i.contentWindow.Object.prototype)
    return "i.contentWindow.Object.prototype";

  if (proto === DOMParser.prototype)
    return "DOMParser.prototype";

  if (proto === i.contentWindow.DOMParser.prototype)
    return "i.contentWindow.DOMParser.prototype";

  return "???"
}

function object_realm(dp) {
  var url = DOMParser.prototype.parseFromString.call(dp, "x", "text/html").documentURI;
  if (url === "http://software.hixie.ch/utilities/js/live-dom-viewer/")
    return "parent window";

  if (url === "http://software.hixie.ch/utilities/js/live-dom-viewer/document")
    return "child window";

  return "???";
}

function f(i) {
var dp = new DOMParser();
w("new DOMParser()");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

dp = new i.contentWindow.DOMParser();
w("new i.contentWindow.DOMParser()");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

dp = Reflect.construct(i.contentWindow.DOMParser, [], DOMParser);
w("Reflect.construct(i.contentWindow.DOMParser, [], DOMParser)");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

dp = Reflect.construct(DOMParser, [], i.contentWindow.DOMParser);
w("Reflect.construct(DOMParser, [], i.contentWindow.DOMParser)");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

var cons = function() {};
cons.prototype = 7;

dp = Reflect.construct(i.contentWindow.DOMParser, [], cons);
w("Reflect.construct(i.contentWindow.DOMParser, [], cons)");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

var script = i.contentWindow.document.createElement("script");
script.textContent = "window.cons = function() {}; cons.prototype = 8;";
i.contentWindow.document.documentElement.appendChild(script);

dp = Reflect.construct(DOMParser, [], i.contentWindow.cons);
w("Reflect.construct(DOMParser, [], i.contentWindow.cons)");
w(identify(dp.__proto__, i));
w(object_realm(dp));
w("");

}
</script>
<iframe src=document onload=f(this)></iframe>
```
</summary>

bzbarsky:
> Note that this doesn't match Gecko if NewTarget has a prototype property that isn't an object.

Yep, Gecko doesn't have access to a thing that implements the per-spec behavior of GetFunctionRealm, for example.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1529757 and https://bugzilla.mozilla.org/show_bug.cgi?id=1317658

PR #640, opened 2019-02-06
Specify what typeof should return for exotic objects

### LINKED ISSUE #512: typeof operator and platform objects
The [`typeof` operator](https://tc39.github.io/ecma262/#table-35) is defined for "Object (non-standard exotic and does not implement [[Call]])" to return "Implementation-defined. Must not be "undefined", "boolean",  "function", "number", "symbol", or "string"."

From

> The semantics of the essential internal methods for ordinary objects and standard exotic objects are specified in clause 9.

it seems that "non-standard" means "non-ECMA-262". Then named property objects, legacy platform objects, etc. are all non-standard. We should say in Web IDL that they should all return "object"

TimothyGu:
Fixes #512.

***
<a href="https://pr-preview.s3.amazonaws.com/TimothyGu/webidl/pull/640.html" title="Last updated on Feb 9, 2019, 2:25 AM UTC (a6b9638)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/640/fc5da8f...TimothyGu:a6b9638.html" title="Last updated on Feb 9, 2019, 2:25 AM UTC (a6b9638)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6713,9 +6713,9 @@ of objects defined in this section has the value <emu-val>true</emu-val>.
 Unless otherwise specified, the \[[Prototype]] internal slot
 of objects defined in this section is {{%ObjectPrototype%}}.
 
-Unless otherwise specified, when the {{ECMAScript/typeof}} operator is called
+Unless otherwise specified, {{ECMAScript/typeof}} operator, when called
 on an exotic object that is not a [=function object=] defined in this section
-and other specifications, the string "<code>object</code>" is returned.
+and other specifications, must return the string "<code>object</code>".
 
 Some objects described in this section are defined to have a <dfn id="dfn-class-string" export>class string</dfn>,
 which is the string to include in the string returned from Object.prototype.toString.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6712,6 +6713,10 @@ of objects defined in this section has the value <emu-val>true</emu-val>.
 Unless otherwise specified, the \[[Prototype]] internal slot
 of objects defined in this section is {{%ObjectPrototype%}}.
 
+Unless otherwise specified, when the {{ECMAScript/typeof}} operator is called
+on an exotic object that is not a [=function object=] defined in this section
+and other specifications, the string "<code>object</code>" is returned.
```
Ms2ger: Do we need a "must" somewhere?

annevk:
I had missed the `typeof` issue, but looking into it I wonder if this is something we should perhaps sort completely on the ECMAScript side: https://github.com/heycam/webidl/issues/512#issuecomment-461328702.

domenic:
I think we should add the "must", merge, and potentially later revert if ES tightens itself up. I've found waiting on ES spec patches to land to be a bad time, so I'd rather not block on that.

littledan:
FWIW I wrote the ES PR: https://github.com/tc39/ecma262/pull/1441

But, my experience matches @domenic 's unfortunately; I'd be fine with merging this patch without waiting on the other one. It's great to have the ambiguity clarified.

TimothyGu:
"Must" added.

Issue #645, opened 2019-02-13
Should LegacyWindowAlias and LegacyNamespace be allowed together?


Ms2ger:

domenic:
Let's say no. They're not currently, and ideally they should not be used in the future.

PR #648, opened 2019-02-22
Imperatively define the global property references

Ms2ger:
***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/648.html" title="Last updated on Feb 25, 2019, 4:38 PM UTC (752f29b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/648/d534bbe...752f29b.html" title="Last updated on Feb 25, 2019, 4:38 PM UTC (752f29b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -122,6 +122,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: CreateDataProperty; url: sec-createdataproperty
         text: CreateIterResultObject; url: sec-createiterresultobject
         text: CreateMapIterator; url: sec-createmapiterator
+        text: CreateMethodProperty; url: sec-createmethodproperty
         text: CreateSetIterator; url: sec-createsetiterator
         text: DefinePropertyOrThrow; url: sec-definepropertyorthrow
         text: DetachArrayBuffer; url: sec-detacharraybuffer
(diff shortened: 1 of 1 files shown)


Ms2ger:
Commits to be squashed later, but I thought this would be easier to review.

index.bs
```diff
@@ -12434,6 +12432,12 @@ object is associated with.
             1.  Let |interfaceObject| be the result of [=create an interface object|creating
                 an interface object=] for |interface| with |id| in |realm|.
             1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+        1.  If the |interface| is declared with a [{{LegacyWindowAlias}}] [=extended attribute=],
+            and |target| implements the {{Window}} [=interface=], then:
+            1.  Assert: |interfaceObject| was created in the previous steps.
```
domenic: We tend to treat ifs as "block scope", so the interfaceObject variable went out of scope and shouldn't bused here. So, nesting under the previous bullet would be better. (But then we run into the issues you noted in #645.) You could alternately say "Let interfaceObject be null" in an outer scope.

index.bs
```diff
@@ -12438,6 +12436,12 @@ object is associated with.
             1.  [=list/iterate|For every=] [=LegacyWindowAlias identifier|identifier=] |id| in
                 [{{LegacyWindowAlias}}]'s [=LegacyWindowAlias identifier|identifiers=]:
                 1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+        1.  If the |interface| is declared with a [{{NamedConstructor}}] [=extended attribute=],
+            then:
+            1.  [=list/iterate|For every=] [=NamedConstructor identifier=] |id| on |interface|:
```
domenic: I'm prety sure there can only be on NamedConstructor identifier, so this loop is not necessary.

index.bs
```diff
@@ -12438,6 +12436,12 @@ object is associated with.
...
+                1.  Let |interfaceObject| be the result of [=create a named constructor|creating
```
domenic: interfaceObject is a bad name here; namedConstructor or maybe namedConstructorFactory would be better.

index.bs
```diff
@@ -12442,6 +12441,12 @@ object is associated with.
                 1.  Let |interfaceObject| be the result of [=create a named constructor|creating
                     a named constructor=] with |id| for |interface| in |realm|.
                 1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+    1.  [=list/iterate|For every=] [=callback interface=] |interface| that is [=exposed=] in
```
domenic: "Define the interface objects" is no longer a good name for this algorithm, as it creates interface objects, legacy callback interface objects, and (in the next commit) namespace objects. Maybe "Define the top-level constructs" or "Define the global property references" or something.

index.bs
domenic: Similarly _id_ has gone out of scope and should be pulled out.

index.bs
```diff
@@ -12447,6 +12447,11 @@ object is associated with.
         1.  Let |interfaceObject| be the result of [=create a legacy callback interface
             object|creating a legacy callback interface object=] for |interface| with |id| in |realm|.
         1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+    1.  [=list/iterate|For every=] [=namespace=] |namespace| that is [=exposed=] in
+        |realm|:
+        1.  Let |namespaceObject| be the result of [=create a namespace object|creating a namespace object=] for |namespace| in |realm|.
+        1.  Let |id| be |namespace|'s [=identifier=].
```
domenic: Nit: in other cases you grab id first.

index.bs
bzbarsky: Per spec there can be many.  https://heycam.github.io/webidl/#NamedConstructor explicitly says "Multiple [NamedConstructor] extended attributes may appear on a given interface.".

Now in practice no one does that for the uses we have for NamedConstructor right now, so maybe we should simplify things...

Ms2ger: Filed https://github.com/heycam/webidl/issues/653.

index.bs
```diff
@@ -12421,6 +12417,42 @@ object is associated with.
...
+                an interface object=] for |interface| with |id| in |realm|.
+            1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+            1.  If the |interface| is declared with a [{{LegacyWindowAlias}}] [=extended attribute=],
+                and |target| implements the {{Window}} [=interface=], then:
+                1.  [=list/iterate|For every=] [=LegacyWindowAlias identifier|identifier=] |id| in
+                    [{{LegacyWindowAlias}}]'s [=LegacyWindowAlias identifier|identifiers=]:
+                    1.  Perform [=!=] [$CreateMethodProperty$](|target|, |id|, |interfaceObject|).
+        1.  If the |interface| is declared with a [{{NamedConstructor}}] [=extended attribute=],
+            then:
+            1.  [=list/iterate|For every=] [=NamedConstructor identifier=] |id| on |interface|:
```
domenic: Nit: above it's "For every id in [LegacyWindowAlias]'s identifiers"; here it's "For ever NamedConstructor identifier id". I like the first one more.

PR #654, opened 2019-02-25
Define "implements" checks using internal slots

### LINKED ISSUE #97: Add [[PlatformBrand]] internal slot to all platform objects

Ms2ger:
- Update the steps for instantiating a platform object, which everything
  theoretically calls into, and which sets the primary interface into a new
  internal slot [[PrimaryInterface]].
- Define the "implements" algorithm as checking the [[PrimaryInterface]]
  internal slot.

The algorithm here is roughly in correspondence with the V8 implementation,
where operation functions are allocated with a reference to a
FunctionTemplate object (which is in corresondence with WebIDL interfaces).
When checking the receiver of a method that came from WebIDL, the original
prototype chain is traversed by looking at the FunctionTemplate's parent.
The same FunctionTemplate is used in multiple JavaScript realms.

Fixes #97.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/webidl/pull/654.html" title="Last updated on Feb 25, 2019, 4:24 PM UTC (1f1c78f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/654/c9246d4...Ms2ger:1f1c78f.html" title="Last updated on Feb 25, 2019, 4:24 PM UTC (1f1c78f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7442,7 +7442,7 @@ values are represented by ECMAScript Object values (including [=function objects
     to an IDL [=interface type=] value by running the following algorithm (where |I| is the [=interface=]):
 
     1.  If <a abstract-op>Type</a>(|V|) is not Object, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  If |V| is a [=platform object=] that implements |I|, then return the IDL [=interface type=] value that represents a reference to that platform object.
+    1.  If |V| [=implements=] |I|, then return the IDL [=interface type=] value that represents a reference to that platform object.
     1.  If |V| is a [=user object=]
         that is considered to implement |I| according to the rules in [[#es-user-objects]], then return the IDL [=interface type=] value that represents a reference to that
         user object, with the [=incumbent settings object=] as the [=callback context=].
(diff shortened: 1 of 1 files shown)


Ms2ger:
This was #595.

index.bs
```diff
@@ -12364,6 +12359,20 @@ The value of the [=function object=]’s <code class="idl">name</code> property
 
 <h3 id="es-platform-objects">Platform objects implementing interfaces</h3>
 
+<div algorithm>
+    An ECMAScript value |value| <dfn id="es-is-platform-object">is a platform object</dfn>
+    if [$Type$](|value|) is Object and if |value| has a \[[PrimaryInterface]] internal slot.
```
bzbarsky: So here is an interesting question:  How should WindowProxy interact with all this?  Probably followup fodder...

index.bs
```diff
@@ -12364,6 +12359,20 @@ The value of the [=function object=]’s <code class="idl">name</code> property
 
 <h3 id="es-platform-objects">Platform objects implementing interfaces</h3>
 
+<div algorithm>
+    An ECMAScript value |value| <dfn id="es-is-platform-object">is a platform object</dfn>
```
domenic: Just curious, why add a custom ID for this?

PR #664, opened 2019-02-27
Editorial: improve language around conversions to JS

Ms2ger:
This prose is converting IDL values to JS values, so they should not try to
convert to IDL types.

Separately, fix some grammar nits.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=20455.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/664.html" title="Last updated on Mar 1, 2019, 4:28 PM UTC (e325ade)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/664/c705ce6...e325ade.html" title="Last updated on Mar 1, 2019, 4:28 PM UTC (e325ade)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10725,8 +10725,7 @@ the <code>typeof</code> operator will return "function" when applied to an inter
         1.  Perform the actions listed in the description of |constructor|
             with |values| as the argument values
             and |object| as the <emu-val>this</emu-val> value.
-        1.  Let |O| be the result of [=converted to an ECMAScript value|converting=] |object|
-            to an ECMAScript [=interface type=] value |I|.
+        1.  Let |O| be |object|, [=converted to an ECMAScript value=].
         1.  Assert: |O| is an object that [=implements=] |I|.
         1.  Assert: |O|.\[[Realm]] is |realm|.
         1.  Return |O|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11343,15 +11341,14 @@ in which case they are exposed on every object that [=implements=] the interface
...
+                1.  Set |R| to the result of performing the actions listed in
                     |operation|'s [=corresponding default operation=] on |O|,
                     with |values| as the argument values.
             1.  Otherwise:
-                1.  Set |R| be the result of performing the actions listed in the
+                1.  Set |R| to the result of performing the actions listed in the
                     description of |operation|, on |O| if |O| is not <emu-val>null</emu-val>, with |values|
                     as the argument values.
-                1.  Return the result of [=converted to an ECMAScript value|converting=] |R| to
-                    an ECMAScript value of the type |op| is declared to return.
```
annevk: I think this precision is sometimes useful and could lead to different results. At least long term I want to operate on a byte sequence and be able to return it, and let IDL handle the conversion to ByteString or Uint8Array as applicable.

Similar, if all you have is a list of things, you need the type to know what to turn that list into, no?

domenic: But operations are not declared to return ECMAScript values, so the existing sentence is nonsensical.

annevk: That is true, but the type is necessary somehow and this proposal removes it.

Ms2ger: How about something like this:

```diff
-             1.  Let |R| be <emu-val>null</emu-val>.
+             1.  Let |R| be an undefined value of the type |op| is declared to return.
```

annevk: Maybe instead we should add an issue marker pointing to #674?

PR #665, opened 2019-02-27
Formalize the 'this' and 'given value' ambient values

Ms2ger:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/665.html" title="Last updated on Mar 18, 2019, 10:09 AM UTC (95b778a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/665/6dcfbec...95b778a.html" title="Last updated on Mar 18, 2019, 10:09 AM UTC (95b778a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1492,6 +1492,13 @@ that are exposed as a convenience to users of objects in the system.
 
 </div>
 
+Every [=regular operation=], [=regular attribute=] getter, and [=regular attribute=] setter's
+algorithm steps have access to a <dfn export>this</dfn> value, which is an IDL value of the type
+the member is declared on.
+
+[=Attribute=] setter's algorithm steps also have access to <dfn export>the given value</dfn>,
+which is an IDL value of the type the [=attribute=] is declared as.
+
 [=Interfaces=], [=interface mixins=], and [=namespaces=] each support a different set of [=members=],
 which are specified in [[#idl-interfaces]], [[#idl-interface-mixins]], and [[#idl-namespaces]],
 and summarized in the following informative table:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11144,23 +11144,25 @@ in which case they are exposed on every object that [=implements=] the interface
 
     1.  Let |steps| be the following series of steps:
         1.  Try running the following steps:
-            1.  Let |O| be <emu-val>null</emu-val>.
+            1.  Let |object| be null.
```
annevk: Perhaps this change (and below a couple times) should be made separately? Changing the variable name seems somewhat significant.

Ms2ger: Now depends on #666.

index.bs
```diff
@@ -11244,8 +11244,9 @@ in which case they are exposed on every object that [=implements=] the interface
                 </dd>
             </dl>
 
-        1.  Perform the actions listed in the description of |attribute| that occur on setting, on
-            |object| if it is not null.
+        1.  Perform the actions listed in the description of |attribute| that occur on setting,
+            with |idlValue| as <b>given value</b> and
```
domenic: "the" given value sounds nicer to me.

annevk:
FWIW, I think bold is reasonable to indicate they're special. Linking them would result in a lot of links to IDL, that are probably not that beneficial since this is really something you only need to learn once.

domenic:
I'm fine with not encouraging other specs to link but I think within IDL we should cross-link, e.g. in "Perform the actions listed in the description of |attribute| that occur on setting....".

(But, not in the DOMException definition, since that's the template other specs would follow.)

PR #666, opened 2019-02-27
Editorial: run platform algorithms on IDL objects

Ms2ger:
Currently we technically pass around JS objects to specifications
that use IDL to define their APIs.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/666.html" title="Last updated on Mar 14, 2019, 2:59 PM UTC (3cfbccf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/666/ff2b414...3cfbccf.html" title="Last updated on Mar 14, 2019, 2:59 PM UTC (3cfbccf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11110,23 +11110,25 @@ in which case they are exposed on every object that [=implements=] the interface
 
     1.  Let |steps| be the following series of steps:
         1.  Try running the following steps:
-            1.  Let |O| be <emu-val>null</emu-val>.
+            1.  Let |idlObject| be null.
             1.  If |target| is an [=interface=], and |attribute| is a [=regular attribute=]:
-                1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
-                    <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
+                1.  Let |esValue| be the <emu-val>this</emu-val> value, if it is not
+                    <emu-val>null</emu-val> or <emu-val>undefined</emu-val>, or |realm|'s
+                    [=Realm/global object=] otherwise.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11144,23 +11144,25 @@ in which case they are exposed on every object that [=implements=] the interface
 
     1.  Let |steps| be the following series of steps:
         1.  Try running the following steps:
-            1.  Let |O| be <emu-val>null</emu-val>.
+            1.  Let |object| be null.
             1.  If |target| is an [=interface=], and |attribute| is a [=regular attribute=]:
                 1.  If the <emu-val>this</emu-val> value is <emu-val>null</emu-val> or
-                    <emu-val>undefined</emu-val>, set |O| to |realm|'s [=Realm/global object=].
+                    <emu-val>undefined</emu-val>, let |O| be |realm|'s [=Realm/global object=].
```
TimothyGu: We don’t use the ECMA-262-like variable scoping scheme here, but the [Infra](https://infra.spec.whatwg.org/#variables)-based block-scoping rules.

Ms2ger: Can you show what this would look like? Ideally we'd have Rust-style `let x = if .. {} else {}`.

annevk: Basically you'd declare the _O_ upfront and then override it here.

annevk:
Another thing that might be nice while cleaning this up is to name the variables esValue and idlValue or some such?

index.bs
domenic: Rust style is done via "Let x be y if z, or w otherwise."

Ms2ger:
Rebased and renamed variables.

PR #668, opened 2019-02-28
Deduplicate unforgeable member functions

Ms2ger:
This avoids creating a new function object for every instance.

This change is already implemented in Gecko and Chromium.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28739.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/668.html" title="Last updated on Mar 1, 2019, 1:06 PM UTC (582d788)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/668/c705ce6...582d788.html" title="Last updated on Mar 1, 2019, 1:06 PM UTC (582d788)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10698,7 +10698,17 @@ the <code>typeof</code> operator will return "function" when applied to an inter
     1.  Let |constructorProto| be |realm|.\[[Intrinsics]].[[{{%FunctionPrototype%}}]].
     1.  If |I| inherits from some other interface |P|,
         then set |constructorProto| to the [=interface object=] of |P| in |realm|.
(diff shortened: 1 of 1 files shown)


bzbarsky:
For what it's worth, Gecko uses an object as well, though there are some complications.

Gecko hangs it off the prototype, not the constructor, but that's because we don't have an interface object at all for a [NoInterfaceObject] interface.  I vaguely recall that there were some changes in that area at some point...

Ms2ger:
That would work as well, but the interface prototype object creation is a little less well-defined at this point, and it doesn't really matter after #609. (It also doesn't matter in practice now, since unforgeable and nointerfaceobject aren't combined anywhere.)

bzbarsky:
Right, to be clear the PR here looks fine to me.

Issue #473, opened 2017-11-03
Order of includes statements is not a well-defined concept


bzbarsky:
https://heycam.github.io/webidl/#idl-interface-mixins says:

> The order of appearance of includes statements affects the order in which interface mixin are included by their host interface.

But "order of appearance of includes statements" is not really defined, since they can appear in multiple specifications.

As a data point, what Gecko does in practice right now (pre-mixin, in the "implements" and consequential interface world), is to sort the consequential interfaces by name, then include their members in that order.

tobie:
Yeah, this is what I was trying to get at here: https://github.com/heycam/webidl/issues/432.

Issue #432, opened 2017-09-01
Enumeration order of interface members


tobie:
Spec both says:

> The order that members appear [on an interface] has significance for property enumeration in the ECMAScript binding.

and

> The order of appearance of an interface definition and any of its partial interface definitions does not matter.

This seems contradictory. Which one is it?

annevk:
It's important to some extent for compatibility. E.g., we cannot change the order of some members on the 2D API as that would break sites. How much order is important in general is unknown and proposals have come up for randomizing it and seeing what breaks...

tobie:
So enumeration order matters between some members defined within the same interface or partial interface declaration. I see.

tobie:
So if we want to define this precisely, here's a proposal so there's some basis for discussion:

1.  Let |members| be a new [=ordered set=].
1.  Let |stack| be the result of invoking [=create an inheritance stack=] for [=interface=] |I|.
1.  [=iteration/While=] |stack| is not [=stack/empty=]:
    1.  Let |interface| be the result of [=stack/popping=] from |stack|.
    1.  [=set/Append=] [=For each|each=] |member| of |interface|'s
        [=interface member|members=] to |members|.
    1.  [=For each=] |partialInterface| of |interface| declared in the same specification as |interface|,
         in the order in which it appears in the specification:
         1.  [=set/Append=] [=For each|each=] |member| of |partialInterface|'s
             [=interface member|members=] to |members|.
    1.  [=For each=] |specification| that declares a [=partial interface=] of |interface|,
        lexicographically ordered by |specification|'s short name:
        1.  [=For each=] |partialInterface| of |interface| declared in |specification|,
            in the order in which it appears in |specification|:
            1.  [=set/Append=] [=For each|each=] |member| of |partialInterface|'s
                [=interface member|members=] to |members|.
    1.  [=For each=] |mixin| [=included=] by |interface|, declared in the same specification as |interface|,
        in the order in which it appears in the specification:
        1.  [=set/Append=] [=For each|each=] |member| of |mixin|'s
            [=mixin member|members=] to |members|.
        1.  [=For each=] |partialMixin| of |mixin| declared in the same specification as |mixin|,
            in the order in which it appears in the specification:
            1.  [=set/Append=] [=For each|each=] |member| of |partialMixin|'s
                [=mixin member|members=] to |members|.
        1.  [=For each=] |specification| that declares a [=partial mixin=] of |mixin|,
            lexicographically ordered by |specification|'s short name:
[...]

PR #675, opened 2019-03-03
Define Web IDL Modules

### LINKED ISSUE #592: Declaring modules in WebIDL
There's talk in various places about adding built-in modules to JavaScript and the Web Platform. For this, it could be useful to add support to WebIDL for defining a module.

`namespace`/`[LegacyNamespace=...]` could be a good model here, but lots of details are a bit different, as modules don't have all the flexibility of objects. For example, attributes in modules (which will always be readonly) shouldn't trigger any kind of evaluation/execution when they are accessed.

We could start with modules exporting only functions and interfaces (classes). It could look something like this:

```
module "std::foo" {
  double sum(double x, double y);
}
[Module="std::foo"]
interface Bar { }
```

Thoughts?

(this issue should be tagged jsidl)

littledan:
This PR defines the ability in WebIDL to define JavaScript built-in modules.

Modules are defined with a new `module` declaration, which can contain
regular operations, readonly attributes, and interfaces. For example:

```
[Exposed=Window, SecureContext]
module temporal {
  interface Timezone {
    readonly attribute USVString name;
    long long offsetMs(long long unixTime);
  };
  temporal.Timezone getCurrentTimezone();
  readonly attribute temporal.Timezone initialTimezone;
};
```

This PR is based on the proposed infrastructure at
https://github.com/tc39/proposal-javascript-standard-library/pull/44
These semantics set the module map, which makes sense in the context of
the import maps draft specification
https://wicg.github.io/import-maps/

Closes #592

***
<a href="https://pr-preview.s3.amazonaws.com/littledan/webidl/pull/675.html" title="Last updated on Aug 29, 2019, 2:18 PM UTC (753001a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/675/4ef3509...littledan:753001a.html" title="Last updated on Aug 29, 2019, 2:18 PM UTC (753001a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4478,7 +4478,7 @@ The [=module/module members=] of a [=module=] |module| are given by the followin
     1.  Let |result| be « ».
     1.  Let |specifier| be |module|'s [=module/specifier=].
     1.  Let |declaration| be the [=module declaration=] whose [=module declaration/specifier=] is
-        |specifier|:
+        |specifier|.
     1.  For every member |member| in |declaration|'s [=module declaration/module declaration
         members|members=]:
         1.  [=list/Append=] |member| to |result|.
--- a/index.bs
+++ b/index.bs
@@ -35,6 +35,7 @@ spec: dom; type: interface; text: Document
 spec: ecma-262; type: dfn; for: /; text: internal method
 spec: ecma-262; type: dfn; for: /; text: internal slot
 spec: ecma-262; type: dfn; for: /; text:realm
+spec:html; type:dfn; for:environment settings object; text:module map
 </pre>
 
 <pre class="anchors">
@@ -223,12 +224,17 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
 
 <pre class=biblio>
 {
-    "GEOMETRY": {
-        "aliasOf": "GEOMETRY-1"
+    "IMPORT-MAPS": {
+        "href": "https://wicg.github.io/import-maps/",
+        "title": "Import Maps",
+        "authors": ["Domenic Denicola"]
     },
     "JSSTDLIB": {
         "href": "https://github.com/tc39/proposal-javascript-standard-library/",
         "title": "Standard Library Proposal"
+    },
+    "GEOMETRY": {
+        "aliasOf": "GEOMETRY-1"
     }
 }
 </pre>
@@ -476,8 +482,10 @@ The different kinds of <dfn id="dfn-definition">definitions</dfn> that can appea
 [=partial namespace|partial namespace definitions=],
 [=dictionary|dictionaries=],
 [=partial dictionary|partial dictionary definitions=],
-[=typedefs=] and
-[=includes statements=].
+[=typedefs=],
+[=includes statements=],
+[=module declarations=]
+and [=partial module declarations=].
 These are all defined in the following sections.
 
 Each [=definition=]
@@ -512,6 +520,7 @@ in [[#es-extended-attributes]].
         Enum
         Typedef
         IncludesStatement
+        Module
 </pre>
 
 <div class="example">
@@ -717,23 +726,15 @@ in later sections.
 
 Within the set of [=IDL fragments=]
 that a given implementation supports,
-the [=identifier=] of every
-[=interface=],
-[=namespace=],
-[=dictionary=],
-[=enumeration=],
-[=callback function=],
-[=callback interface=] and
-[=typedef=]
-must not
-be the same as the identifier of any other
-[=interface=],
-[=namespace=],
-[=dictionary=],
-[=enumeration=],
-[=callback function=],
-[=callback interface=] or
-[=typedef=].
+the [=list=] that is the concatenation of the following [=lists=] must not
+contain the same [=list/item=] more than once:
+* the [=interface/scoped identifier=] of every [=interface=],
+* the [=identifier=] of every [=namespace=],
+* the [=identifier=] of every [=dictionary=],
+* the [=identifier=] of every [=enumeration=],
+* the [=identifier=] of every [=callback function=],
+* the [=identifier=] of every [=callback interface=], and
+* the [=identifier=] of every [=typedef=].
 
 Within an [=IDL fragment=], a reference
 to a [=definition=] need not appear after
@@ -831,11 +832,32 @@ expected that an object that implements a particular IDL interface
 provides ways to inspect and modify the object's state and to
 invoke the behavior described by the interface.
 
+[=Interfaces=] and [=partial interfaces=] have an <dfn for="interface,partial interface">enclosing
+module</dfn>, which is a [=module=] or null.
+Unless otherwise specified, it is null.
+
+<div algorithm>
+  [=Interfaces=] and [=partial interfaces=] have a
+  <dfn for="interface,partial interface">scoped identifier</dfn>,
+  which uniquely identifies an [=interface=].
+
+  It is given by the result of the following steps:
+
+    1.  Let |definition| be the given [=interface=] or [=partial interface=].
+    1.  Let |identifier| be |definition|'s [=identifier=].
+    1.  Let |module| be |definition|'s [=enclosing module=].
+    1.  If |module| is null, then:
+        1.  Return |identifier|.
+    1.  Let |specifier| be |module|'s [=module/specifier=].
+    1.  Return the [=concatenation=] of « |specifier|, |identifier| » with
+        separator <span class="char">U+002E FULL STOP (".")</span>.
+</div>
+
 An interface can be defined to <dfn id="dfn-inherit" for="interface" export>inherit</dfn> from another interface.
 If the identifier of the interface is followed by a
 <span class="char">U+003A COLON (":")</span> character
-and an [=identifier=],
-then that identifier identifies the inherited interface.
+and a <emu-t class="regex"><a href="#prod-scopedidentifier">scopedidentifier</a></emu-t> token,
+then that token is the [=interface/scoped identifier=] of the inherited interface.
 An object that implements an interface that inherits from another
 also implements that inherited interface.  The object therefore will also
 have members that correspond to the interface members from the inherited interface.
@@ -949,9 +971,9 @@ The IDL for interfaces can be split into multiple parts by using
 <dfn id="dfn-partial-interface" export>partial interface</dfn> definitions
 (matching <emu-t>partial</emu-t> <emu-t>interface</emu-t>
 <emu-nt><a href="#prod-PartialInterfaceRest">PartialInterfaceRest</a></emu-nt>).
-The [=identifier=] of a partial
+The [=partial interface/scoped identifier=] of a partial
 interface definition must be the same
-as the identifier of an interface definition.  All of
+as the [=interface/scoped identifier=] of an interface definition.  All of
 the members that appear on each of the partial interfaces are considered to be
 members of the interface itself.
 
@@ -998,6 +1020,7 @@ The following extended attributes are applicable to [=partial interfaces=]:
 
 [=Interfaces=] which are not annotated
 with a [{{NoInterfaceObject}}] [=extended attribute=]
+and whose [=interface/enclosing module=] is null,
 must be annotated with an [{{Exposed}}] [=extended attribute=].
 
 <div algorithm>
@@ -1009,6 +1032,10 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
     1.  Let |namespace| be the identifier argument of the [{{LegacyNamespace}}]
         [=extended attribute=].
     1.  Return the [=concatenation=] of « |namespace|, |identifier| » with
+        separator <span class="char">U+002E FULL STOP (".")</span>.
+1. If |interface|'s [=interface/enclosing module=] is not null, then:
+    1.  Let |specifier| be |interface|'s [=interface/enclosing module=]'s [=module/specifier=].
+    1.  Return the [=concatenation=] of « |specifier|, |identifier| » with
         separator <span class="char">U+002E FULL STOP (".")</span>.
 1. Return |identifier|.
 
@@ -1041,6 +1068,7 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
         "interface" PartialInterfaceOrPartialMixin
         PartialDictionary
         Namespace
+        PartialModule
 </pre>
 
 <pre class="grammar" id="prod-PartialInterfaceOrPartialMixin">
@@ -1075,7 +1103,7 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
 
 <pre class="grammar" id="prod-Inheritance">
     Inheritance :
-        ":" identifier
+        ":" scopedidentifier
         ε
 </pre>
 
@@ -1455,15 +1483,16 @@ must be annotated with an [{{Exposed}}] [=extended attribute=].
 
 <h3 id="idl-members">Members</h3>
 
-[=Interfaces=], [=interface mixins=], and [=namespaces=] are specifications of a set of
+[=Interfaces=], [=interface mixins=], [=namespaces=], and [=modules=] are specifications of a set of
 <dfn id="dfn-member" export lt="members">members</dfn> (respectively matching
 <emu-nt><a href="#prod-InterfaceMembers">InterfaceMembers</a></emu-nt>,
-<emu-nt><a href="#prod-MixinMembers">MixinMembers</a></emu-nt>, and
-<emu-nt><a href="#prod-NamespaceMembers">NamespaceMembers</a></emu-nt>),
+<emu-nt><a href="#prod-MixinMembers">MixinMembers</a></emu-nt>,
+<emu-nt><a href="#prod-NamespaceMembers">NamespaceMembers</a></emu-nt>, and
+<emu-nt><a href="#prod-ModuleMembers">ModuleMembers</a></emu-nt>),
 which are the [=constants=], [=attributes=], [=operations=], and
 other declarations that appear between the braces of their declarations.
 [=Attributes=] describe the state that an object
-implementing the [=interface=], [=interface mixin=], or [=namespace=] will expose,
+implementing the [=interface=], [=interface mixin=], [=namespace=], or [=module=] will expose,
 and [=operations=] describe the behaviors that can be invoked on the object.
 [=Constants=] declare named constant values
 that are exposed as a convenience to users of objects in the system.
@@ -1485,10 +1514,9 @@ declared on or that [=includes=] the [=interface mixin=] the member is declared 
 [=Attribute=] setter's algorithm steps also have access to <dfn export>the given value</dfn>,
 which is an IDL value of the type the [=attribute=] is declared as.
 
-[=Interfaces=], [=interface mixins=], [=callback interfaces=] and [=namespaces=] each support a
-different set of [=members=],
-which are specified in [[#idl-interfaces]], [[#idl-interface-mixins]],
-[[#idl-callback-interfaces]], and [[#idl-namespaces]],
+[=Interfaces=], [=interface mixins=], [=callback interfaces=], [=namespaces=], and [=modules=] each
+support a different set of [=members=], which are specified in [[#idl-interfaces]],
+[[#idl-interface-mixins]], [[#idl-callback-interfaces]], [[#idl-namespaces]], and [[#idl-modules]]
 and summarized in the following informative table:
 
 <table class="data complex non-normative">
@@ -1499,6 +1527,7 @@ and summarized in the following informative table:
             <th>[=Callback interfaces=]</th>
             <th>[=Interface mixins=]</th>
             <th>[=Namespaces=]</th>
+            <th>[=Modules=]</th>
         </tr>
     </thead>
     <tbody>
@@ -1508,12 +1537,14 @@ and summarized in the following informative table:
             <td>●</td>
             <td>●</td>
             <td></td>
+            <td></td>
         </tr>
         <tr>
             <th>[=Regular attributes=]</th>
             <td>●</td>
             <td></td>
             <td>●</td>
+            <td>Only [=read only=] attributes</td>
             <td>Only [=read only=] attributes</td>
         </tr>
         <tr>
@@ -1522,9 +1553,11 @@ and summarized in the following informative table:
             <td></td>
             <td></td>
             <td></td>
+            <td></td>
         </tr>
         <tr>
             <th>[=Regular Operations=]</th>
+            <td>●</td>
             <td>●</td>
             <td>●</td>
             <td>●</td>
@@ -1536,10 +1569,12 @@ and summarized in the following informative table:
             <td></td>
             <td>Only [=stringifiers=]</td>
             <td></td>
+            <td></td>
         </tr>
         <tr>
             <th>[=Static Operations=]</th>
             <td>●</td>
+            <td></td>
             <td></td>
             <td></td>
             <td></td>
@@ -1550,10 +1585,12 @@ and summarized in the following informative table:
             <td></td>
             <td></td>
             <td></td>
+            <td></td>
         </tr>
         <tr>
             <th>[=Maplike declarations=]</th>
             <td>●</td>
+            <td></td>
             <td></td>
             <td></td>
             <td></td>
@@ -1564,6 +1601,23 @@ and summarized in the following informative table:
             <td></td>
             <td></td>
             <td></td>
+            <td></td>
+        </tr>
+        <tr>
+            <th>[=Interfaces=]</th>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td>●</td>
+        </tr>
+        <tr>
+            <th>[=Partial interfaces=]</th>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td>●</td>
         </tr>
     </tbody>
 </table>
@@ -1821,7 +1875,8 @@ be retrieved and (in some cases) changed.  There are two kinds of attributes:
 If an attribute has no <emu-t>static</emu-t> keyword, then it declares a
 <dfn id="dfn-regular-attribute" export>regular attribute</dfn>.  Otherwise,
 it declares a [=static attribute=]. Note that in addition to being [=interface members=],
-[=read only=] [=regular attributes=] can be [=namespace members=] as well.
+[=read only=] [=regular attributes=] can be [=namespace members=] or [=module
+members=] as well.
 
 
 <div algorithm>
@@ -2046,7 +2101,8 @@ then it declares a special operation.  A single operation can declare
 both a regular operation and a special operation;
 see [[#idl-special-operations]] for details on special operations.
 Note that in addition to being [=interface members=],
-regular operations can also be [=callback interface members=] and [=namespace members=].
+regular operations can also be [=callback interface members=], [=namespace members, or
+[=module members=].
 
 If an operation has no identifier,
 then it must be declared to be a special operation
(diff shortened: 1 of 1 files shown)


littledan:
This came out much simpler than I expected, due to @domenic 's great work setting up related specifications. I'm not sure if this handles overloaded operations properly, or if it does enough to prohibit things that don't make sense; please consider this an early draft. cc @Ms2ger

index.bs
```diff
@@ -4445,6 +4465,139 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+    };
+</pre>
+
+Note: As with partial interface definitions, partial module definitions are
+intended for use as a specification editorial aide, allowing the definition
+of a module to be separated over more than one section of the document, and
+sometimes multiple documents.
+
+The order that members appear in has significance for property enumeration in
+the <a href="#es-modules">ECMAScript binding</a>.
```
annevk: Shall we try to define a deterministic order that also works for partial to avoid introducing another place where order might differ between implementations?

littledan: I'd be a little uncomfortable with the order being "alphabetization", as we previously discussed--this would be a little unidiomatic to require of JavaScript implementations of built-in modules.

I thought we'd handle this non-determinism in a separate, orthogonal PR. Another option would be to delay partial modules until later. I don't have any use cases in mind for partial modules, but I'm not sure if implementations are eager to split up IDL for a single module into multiple files.

(When I commented that I'm not sure if this handles partials correctly, I'm especially worried about partial implementations as module members, and if it's OK to just pretend that the partials don't exist in the later algorithms as they're already merged in.)

annevk: I'm sorry, I don't remember that discussion, but I also don't care what method we use to determine order. What would the approach be for the separate PR?

Ms2ger: Perhaps https://github.com/heycam/webidl/issues/473?

littledan: I was thinking of @tobie 's suggestion of lexicographic ordering in https://github.com/heycam/webidl/issues/432#issuecomment-334233518 . Looks like I mis-remembered it and he was actually suggesting sorting specifications by their "short name". However, @bzbarsky pointed out [issues](https://github.com/heycam/webidl/issues/432#issuecomment-334234875) with what "short name" means exactly.

Anyway, sorting *specifications* in some way doesn't seem so unidiomatic; what would be surprising is if all module members were sorted lexicographically, and the order within the specification weren't respected. But I guess no one is suggesting that.

annevk: Ah yeah, we should preserve order within a block, that's also needed for things outside modules. So basically, we need some way to compute a sort key for each block.

annevk:
A thing I'm missing here is infrastructure around scoping.
```webidl
module X {
  interface Y { ... };
};
interface Y { ... };
```
Is that okay? If not, what forbids it? If it is, what forbids the second interface declaration from being `partial interface Y`? (Or maybe this why `[Exposed]` is not allowed?)

littledan:
I was hoping that this would be OK to have two different interfaces named `Y`, one inside a module and one not. I hope modules will let us scale up spec development exactly for this kind of reason. In drafting this text, I was having trouble finding what I would have to change/clarify to make sure that a later `partial interface Y` wouldn't merge with the interface in the module.

`[Exposed]` and `[SecureContext]` are not allowed just because it'd be a bunch of extra complexity:
- On the spec side: if you have `partial module "std:foo" { partial interface Bar { void method(); } }`, you'd have to check five things to understand how the exposure-related extended attributes apply to `method`: the main module, the partial module, the main interface, the partial interface, and the operation. We could specify this, but it'd be a bit more complicated, and not simply the current algorithm of (roughly) "traverse up until you find the attribute".
- On the implementation side: if you write a built-in module in JavaScript, you'd have to provide an entirely different version to omit certain named exports, as the list of exported names is static and indicated in the JavaScript code. And if you use a JavaScript class, you'd have to mutate the result to update visibility.
- On use cases: I think modules will be a bit more granular than some of the mega-interfaces that the Web has been evolving over time. So I think that these attributes make less sense for things inside of modules.

If you think full expressiveness here is important, I could update this patch to do that.

annevk:
That all sounds good to me, modulo what comes out of #676 but that can be addressed at that point and isn't needed for v1. (I'm not entirely sure what needs changing to clarify scope unfortunately.)

index.bs
```diff
@@ -4445,6 +4465,139 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+lack of clear use cases. Please <a
+href="https://github.com/heycam/webidl/issues/new?title=Fine%20grained%20exposure%20in%20modules">
+file an issue</a> if you see a need for this feature.
+
+Each module may have <dfn>module evaluation steps</dfn> which are executed
+when the module is first imported. These steps may throw an [=exception=], which
+prevents the module from being accessed.
+
+Rather than defining behavior when accessed, the attributes of a [=module=] act
+as a simple storage location. The steps to "set a module attribute" in a
```
Ms2ger: xref "set a module attribute"?

index.bs
```diff
@@ -13202,6 +13361,67 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+
+    1.  Let |map| be the [=module map=] associated with |realm|'s [=relevant
+        settings object=].
+    1.  Assert: |map|[|specifier|] exists.
+    1.  Let |moduleRecord| be |map|[|specifier|].
+    1.  Assert: |moduleRecord|.\[[ExportNames]] [=list/contains=] |attribute|.
+    1.  Perform ! [=SetSyntheticModuleExport=](|moduleRecord|, |attribute|,
+        |value|).
+</div>
+
```
Ms2ger: Nit: probably two empty lines

index.bs
littledan: I didn't reference this since that would reference the ECMAScript version; each language binding is responsible for defining its own algorithm.

index.bs
littledan: Done

index.bs
```diff
@@ -4445,6 +4465,139 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+          Timezone getCurrentTimezone();
+          readonly attribute Timezone initialTimezone;
+        };
+    </pre>
+
+    <blockquote>
+        The [=module evaluation steps=] of "std:timezone" are as follows:
+            1.  [=set a module attribute|Set the attribute=]
+                "initialTimezone" of "std:timezone" in the
+                [=current Realm=] to an instance of <code>Timezone</code>
```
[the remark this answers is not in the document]
littledan: I believe so, since JavaScript will set the current realm here to that of the module that was just constructed. I don't think "relevant realm" would be well-defined.

index.bs
```diff
@@ -13202,6 +13361,67 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+
+Modules are reified as entries in the [=module map=] as [=Synthetic Module
+Records=], allowing them to be imported as ECMAScript modules.
+
+<div algorithm>
+  To <dfn>create a synthetic module record</dfn> for a [=module=] |module| in a
+  [=Realm=] |realm|, perform the following steps:
+
+    1.  Let |exports| be « ».
+    1.  For each member |member| of |module|:
```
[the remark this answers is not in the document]
littledan: That's what's done in the definition of module members. Is this precise enough? I'll improve the references here.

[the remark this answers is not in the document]
littledan: One thing I'd like to do is actually make sure this list has just one entry for an overloaded operation, and just one entry for all partial interfaces of the same name put together. But I don't really see other algorithms doing anything explicit for overloaded operations.

index.bs
```diff
@@ -13202,6 +13361,67 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+                    |interfaceObject|).
+            1.  Otherwise, if |member| is an [=operation=]:
+                1.  Let |method| be the result of [=creating an operation function=]
+                    given |member|, |module|, and |realm|.
+                1.  Perform ! [=SetSyntheticModuleExport=](|moduleRecord|, |id|,
+                    |method|).
+            1.  Otherwise:
+                1.  Assert: |member| is an [=attribute=].
+                1.  Note: Attributes are expected to be initialized by the
+                          module evaluation steps.
```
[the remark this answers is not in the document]
littledan: That's a good idea; I will add this MUST. This makes it irrelevant whether bindings start out as undefined or in TDZ (https://github.com/tc39/proposal-javascript-standard-library/pull/44#issuecomment-469066281).

index.bs
```diff
@@ -13202,6 +13361,67 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+    1.  For every [=module=] |module| that is [=exposed=] in |interface|,
+        1.  Let |name| be the [=module specifier=] of |module|.
+        1.  Let |record| be the result of [=create a synthetic module record=]
+            for |module| in |realm|.
+        1.  [=map/Set=] |map|[|name|] to |record|.
+</div>
+
+<div algorithm>
+  To <dfn>set a module attribute</dfn> |attribute| of a module specifier |specifier| in a
+  [=Realm=] |realm| to an ECMAScript value |value|:
```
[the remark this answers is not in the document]
littledan: Done.

index.bs
```diff
@@ -4445,6 +4465,139 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+sometimes multiple documents.
+
+The order that members appear in has significance for property enumeration in
+the <a href="#es-modules">ECMAScript binding</a>.
+
+Note that like namespaces, modules do not create types.
+
+Of the extended attributes defined in this specification, only the [{{Exposed}}]
+and [{{SecureContext}}] extended attributes are applicable to modules. No
+extended attributes are applicable to [=partial modules=].
```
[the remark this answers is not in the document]
littledan: This "applicable" wording is widespread in the WebIDL text; maybe we should make a separate issue to clean it up?

index.bs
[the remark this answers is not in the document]
littledan: Done

index.bs
```diff
@@ -13202,6 +13361,67 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+        1.  Run the [=module evaluation steps=] for the [=module specifier=] of
+            |module|, if any are present.
+    1.  Return |moduleRecord|.
+</div>
+
+<div algorithm>
+  To <dfn>initialize the module map</dfn> of a [=Realm=] |realm| with [=interface=] |interface|:
+
+    1.  Let |map| be the [=module map=] associated with |realm|'s [=relevant settings object=].
+    1.  For every [=module=] |module| that is [=exposed=] in |interface|,
```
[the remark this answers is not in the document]
littledan: Done

index.bs
```diff
@@ -4445,6 +4465,139 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
 </div>
 
 
+<h3 id="idl-modules">Modules</h3>
+
+A <dfn id="dfn-module" export>module</dfn> is a definition (matching
```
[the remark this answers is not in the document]
littledan: I really like this text, so I've adopted it. I think you have a stronger vision for how this should be written than I do, so I'll defer to you to reword the members and the existing parts of the specification.

bzbarsky:
@littledan Is there a summary somewhere of what the actual syntax this PR implements is?

littledan:
@bzbarsky Added an example in the PR description;  is this the level of detail you're looking for?

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+The IDL for modules can be split into multiple parts by using <dfn
+id="dfn-partial-module" export>partial module</dfn> definitions (matching
+<emu-t>partial</emu-t> <emu-nt><a href="#prod-Module">Module</a></emu-nt>).
+The module specifier of a partial module definition must be the same as the
+identifier of a module definition. All of the members that appear on each of
+the partial module definitions are considered to be members of the module
+itself.
+
+<pre class="syntax">
+    module <mark>"std:example"</mark> {
```
domenic: As discussed previously, I am hopeful we can centralize the prefix in one place. (Whether it be HTML, or more likely it seems, Web IDL.) Having specs need to remember seems bad.

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+      /* module_members... */
+    };
+
+    partial module <mark>"std:example"</mark> {
+      /* module_members... */
+    };
+</pre>
+
+Note: As with partial interface definitions, partial module definitions are
```
domenic: Nit: xref partial interface definitions

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+intended for use as a specification editorial aide, allowing the definition
+of a module to be separated over more than one section of the document, and
+sometimes multiple documents.
+
+The order that members appear in has significance for property enumeration in
+the <a href="#es-modules">ECMAScript binding</a>.
+
+Note that like namespaces, modules do not create types.
```
domenic: xref namespaces

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+href="https://github.com/heycam/webidl/issues/new?title=Fine%20grained%20exposure%20in%20modules">
+file an issue</a> if you see a need for this feature.
+
+Each module may have <dfn>module evaluation steps</dfn> which are executed
+when the module is first imported. These steps may throw an [=exception=], which
+prevents the module from being accessed.
+
+Rather than defining behavior when accessed, the attributes of a [=module=] act
+as a simple storage location. The steps to "set a module attribute" to an IDL
+value of the type of the attribute. Each [=attribute=] which is a [=module
```
domenic: "The steps..." sentence seems to be incomplete or missing words.

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+member=] must be initialized in [=module evaluation steps=] to set the module
```
domenic: "the" module evaluation steps

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+          };
+          Timezone getCurrentTimezone();
+          readonly attribute Timezone initialTimezone;
+        };
+    </pre>
+
+    <blockquote>
+        The [=module evaluation steps=] of "std:timezone" are as follows:
+            1.  [=set a module attribute|Set the attribute=]
+                "initialTimezone" of "std:timezone" to a [=new=]
```
domenic: Typographically I think `initialTimezone` should be code and should cross-link to the definition. (Not sure cross-linking is possible/easy with example code though). And "of "std:timezone"" should not be necessary here, ideally, although I'm not sure how to rigorize that.

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+                <code>Timezone</code> representing the User Agent's preferred
+                time zone.
+    </blockquote>
+
+    An ECMAScript implementation would then expose a global <code class="idl">"std:timezone"</code>
```
domenic: Unsure what "global" means here

index.bs
```diff
@@ -4445,6 +4493,144 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+    module which includes a constructor for the interface and a function for the operation:
+
+    <pre highlight="js">
+        import * from "std:timezone" as timezone;
+        Object.keys(timezone);                      // Evaluates to ["Timezone", "getCurrentTimezone", "currentTimezone"]
+        typeof timezone.Timezone;                   // Evaluates to "function"
+        typeof timezone.getCurrentTimezone;         // Evaluates to "function"
+        typeof timezone.currentTimezone;            // Evaluates to "object"
```
domenic: currentTimezone -> initialTimezone

index.bs
```diff
@@ -13202,6 +13394,69 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+            1.  Otherwise, if |member| is an [=operation=]:
+                1.  Let |method| be the result of [=creating an operation function=]
+                    given |member|, |module|, and |realm|.
+                1.  Perform ! [=SetSyntheticModuleExport=](|moduleRecord|, |id|,
+                    |method|).
+            1.  Otherwise:
+                1.  Assert: |member| is an [=attribute=].
+                1.  Assert: The module evaluation steps will set each attribute.
+        1.  Run the [=module evaluation steps=] for the [=module/specifier=] of
+            |module|, if any are present.
```
domenic: I don't think they're "for the specifier of _module_", I think they're just "for _module_", right?

index.bs
littledan: There's a lot of open questions about how the module prefix will be determined, and at what level of specification, including:
https://github.com/tc39/proposal-javascript-standard-library/issues/14
https://github.com/tc39/proposal-javascript-standard-library/issues/12
https://github.com/tc39/proposal-javascript-standard-library/issues/20

I don't understand the "remember" point--presumably spec authors will have to use the module prefix in their sample code and tests. It doesn't seem like a lot to remember, either.

I'd like to separate the layering discussions from this PR, so I'll change these samples to avoid any particular prefix, and point to open issues, and suggest that we may do some kind of appending in the future, depending what's decided.

Is there any issue I can point to on the HTML/WebIDL side that includes your centralization idea?

index.bs
```diff
@@ -13202,6 +13394,69 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+
+<div algorithm>
+  To <dfn>initialize the module map</dfn> of a [=Realm=] |realm|:
+
+    1.  Let |map| be the [=module map=] associated with |realm|'s [=relevant settings object=].
+    1.  For every [=module=] |module| that is [=exposed=] in |realm|,
+        1.  Let |name| be the [=module/specifier=] of |module|.
+        1.  Let |record| be the result of [=create a synthetic module record=]
+            for |module| in |realm|.
+        1.  [=map/Set=] |map|[|name|] to |record|.
```
littledan: This line doesn't really make sense. The module map is keyed by URLs, whereas specifiers are strings (even if we concatenate a prefix on). Should we make built-in module specifiers be URLs? Or make the map somehow heterogeneous, with special handling in HTML's module specifier resolution algorithm? Or something else?

domenic: https://github.com/WICG/import-maps/tree/master/reference-implementation/lib has been working off the assumption that the module map keys would be `import:x` for built-in modules with specifier `x`. https://github.com/WICG/import-maps/blob/master/reference-implementation/lib/parser.js#L110

littledan:
Thinking about the comments on ambiguity from @domenic and @annevk a bit more, I'm thinking that it could be best to switch to a model where you declare a module with an *identifier* and then refer to interfaces declared within it as `moduleIdentifer.Interface` (whether you're inside or outside of a module). It'd be really complicated to do lookup based on location, and hurt expressiveness.

The only thing I can see potentially missing is that you can't use `/` within an identifier. This can be addressed in a few different possible ways (allowing identifiers to contain `/` in general, or letting module identifers have this special right) which we could address when someone proposes a built-in module which contains `/` in its module specifier.

It'll be a little while until I have a chance to get back to this; if @Ms2ger or someone else is interested in picking this up, that'd be welcome, or otherwise I will a bit later.

Ms2ger:
This is ready for review, btw.

littledan:
A couple additional suggestions which @domenic and I recently discussed offline. I think any of these could be follow-ons and don't need to block the PR.

- This is a new feature that's not quite ready to use, given surrounding ongoing specification efforts. There should probably be a note to this effect, to avoid confusing people.

- Maybe we should use `export` rather than `readonly attribute` as the syntax?

- We could consider moving the synthetic module record definition into WebIDL, being explicit that it could move back into the ES spec at some point later if there is interest (in a note?). Editorially, we could also consider defining synthetic module records' contents during "create a synthetic module record", rather than during the evaluate phase steps (either way this goes, we could include a note explaining that there are multiple implementation options, both literally lazy-loading JS code and also setting them up more statically).

index.bs
```diff
@@ -801,11 +807,32 @@ expected that an object that implements a particular IDL interface
...
 invoke the behavior described by the interface.
 
+[=Interfaces=] and [=partial interfaces=] have an <dfn for="interface,partial interface">enclosing
+module</dfn>, which is a [=module=] or null.
+Unless otherwise specified, it is null.
+
+<div algorithm>
+  [=Interfaces=] and [=partial interfaces=] have a
+  <dfn for="interface,partial interface">scoped identifier</dfn>,
+  which uniquely identifies an [=interface=].
```
domenic: Why is this a separate concept from qualfiied name? The difference seems to be that for modules we allow two same-named interfaces in different modules, but we don't allow two same-named interfaces in [LegacyNamespace]. Is that mismatch desirable?

index.bs
```diff
@@ -4422,6 +4483,203 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+<div algorithm>
+A module has a [=list=] of
+<dfn id="dfn-module-member" export lt="module member" for=module>module members</dfn> which are
+[=interfaces=], [=read only=] [=regular attributes=], and [=regular operations=].
+
+The [=module/module members=] of a [=module=] |module| are given by the following steps:
+
+    1.  Let |result| be « ».
+    1.  Let |specifier| be |module|'s [=module/specifier=].
+    1.  For each [=module declaration=] |declaration| whose [=module declaration/specifier=] is
```
domenic: How could there be more than one?

index.bs
```diff
@@ -4422,6 +4483,203 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+</div>
+
+For all [=modules=] |module|, all [=interfaces=] and [=partial interfaces=] that are in |module|'s
+[=module/module members=] have their [=interface/enclosing module=] set to |module|.
+
+<div algorithm>
+  The <dfn for=module>url</dfn> of a [=module=] |module| is given by the following steps:
+
+    1.  Let |specifier| be |module|'s [=module/specifier=].
+    1.  Return the [=concatenation=] of « "<code>import:std:</code>", |specifier| ».
```
domenic: Currently I am leaning toward it just being `std:` for built-in modules, but I am not certain until I write proper spec text for the import maps resolver. I would keep it as `std:` for now though.

index.bs
```diff
@@ -4422,6 +4483,203 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+an IDL value of the type of the [=attribute=].
+
+Note: See the ECMAScript binding's definition of [=set a module attribute=].
+
+Each module may have <dfn for=module>evaluation steps</dfn> which are executed when the module is
+first imported.
+These steps have access to a <dfn>this module</dfn> value, which is the [=module=] being
+evaluated, and a [=Realm=].
+They may throw an [=exception=], which prevents the module from being accessed.
+The [=module/evaluation steps=] for a module |m| must invoke the "set a module attribute"
```
domenic: I'm not a big fan of this "quotes surrounding a term" in place of cross-links, but I understand that the alternative is defining a language-agnostic term and a ES-specific term which is also annoying. I'm not sure what to do here.

index.bs
```diff
@@ -4422,6 +4483,203 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+    <pre highlight="js">
+        import * from "std:temporal" as temporal;
+        Object.keys(temporal);                      // Evaluates to ["Timezone", "getCurrentTimezone", "currentTimezone"]
+        typeof temporal.Timezone;                   // Evaluates to "function"
+        typeof temporal.getCurrentTimezone;         // Evaluates to "function"
+        typeof temporal.initialTimezone;            // Evaluates to "object"
+    </pre>
+</div>
+
+### Syntax ### {#idl-modules-syntax}
```
domenic: I note that nothing else has syntax subsections, although I do agree it's a nice thematic break here. Not sure what to do. (Although it doesn't matter much.)

index.bs
```diff
@@ -13183,6 +13446,71 @@ The characteristics of a namespace object are described in [[#namespace-object]]
...
+                1.  Let |interfaceObject| be the result of [=create an interface object|creating
+                    an interface object=] for |member| with |id| in |realm|.
+                1.  Perform [=!=] [=SetSyntheticModuleExport=](|moduleRecord|, |id|,
+                    |interfaceObject|).
+            1.  Otherwise, if |member| is an [=operation=]:
+                1.  Let |method| be the result of [=creating an operation function=] given
+                    |member|, |module|, and |realm|.
+                1.  Perform [=!=] [=SetSyntheticModuleExport=](|moduleRecord|, |id|, |method|).
+            1.  Otherwise:
+                1.  Assert: |member| is an [=attribute=] or a [=partial interface=].
```
domenic: Seeing this here makes me think that yes, we should exclude partial interfaces. (Which is an inline issue above.)

littledan: I don't think there's a real issue here; I think we should continue supporting partial interfaces in modules. For a use case, imagine if the Temporal proposal defined `CivilDate` in ECMA-262, in a built-in module, and then defined `CivilDate.prototype.toLocaleString` in ECMA-402 -- it would make sense to define the `toLocaleString` method in a partial interface inside a partial module. (We currently use a weirder system for Intl, where one spec has one toLocaleString definition, and the other overrides it, but I think this is even messier.)

I think we're concerned about this for the wrong reasons--this is really  more of an editorial mismatch vs the rest of the specification than an actual difficulty being faced. For example, when [defining the global property references](https://heycam.github.io/webidl/#define-the-global-property-references), there is a loop over "every non-callback interface that is exposed in realm.". That is another way of skipping partial interfaces, just less imperatively. I think the eventual fix to #432 will make it even more obvious that things work the same way between these two cases.

domenic: Oh, for sure, we should support partial interfaces inside modules. But should they really be in the member list? Or should only the primary interfaces be in the member list, with the partial interfaces just merging their interface-members into the primary interface?

Ms2ger: The main reason I left them in the member list is to formally define "enclosing module" for partial interfaces. I agree it's a little awkward, though.

littledan: Thanks for clarifying, you two. I don't have an opinion on this phrasing.

domenic:
Just want to check that this is not blocked on me at this point. I'm excited to see this land, but I also appreciate how every new commit from @Ms2ger corrects more subtle issues that I totally missed, so I'm happy to wait it out.

annevk:
Things that would be nice to see clarified:

* What's the plan for testing?
* What are implementers planning? Browsers as well as the various tools that will need updates.

index.bs
Ms2ger: Changing how LegacyNamespace works in this PR seems out of scope to me. Note that, if I connected everything correctly, using the qualified name would mean you'd need changes like this to specifications that use it:

```diff
 namespace WebAssembly {
-    Promise<Module> compile(BufferSource bytes);
+    Promise<WebAssembly.Module> compile(BufferSource bytes);
 };
```

Maybe we want that, but let's discuss that separately.

index.bs
Ms2ger: Let's leave it for now.

index.bs
Ms2ger: If I ever get time, I'd like to see if this split makes sense for the other sections as well, but that's still a while off.

Ms2ger:
> What's the plan for testing?

Wait until someone uses the feature, and then write tests. Possibly https://tc39.github.io/proposal-temporal/ or https://wicg.github.io/kv-storage/ might be the first.

> What are implementers planning? Browsers as well as the various tools that will need updates.

I don't have a clue. Maybe @saschanaz has thoughts on webidl.js.

Ms2ger:
I think I'm happy to squash and land this. (At least until the next time I read through the diff.) I'll do that some time next week, if I hear no objections.

saschanaz:
>Maybe @saschanaz has thoughts on webidl.js.

Will happily start to add support when this PR merges.

annevk:
I'm a little worried about landing something this big without a downstream user PR and browser implementation commitments. We kinda stopped adding "speculative" things a while ago. And while this isn't completely speculative, it still seems good to work it out end-to-end, no?

index.bs
```diff
@@ -4358,6 +4424,237 @@ Of the extended attributes defined in this specification, only the [{{Exposed}}]
...
+<div data-fill-with="grammar-PartialDefinition"></div>
+
+<pre class="grammar" id="prod-Module">
+    Module :
+        "module" identifier "{" ModuleMembers "}" ";"
+</pre>
+
+<pre class="grammar" id="prod-PartialModule">
+    PartialModule :
+        "module" identifier "{" ModuleMembers "}" ";"
```
saschanaz: Why do we have two exactly same grammars `Module` and `PartialModule`?

Ms2ger: To distinguish the two cases in the definition of "module declaration" and "partial module declaration". There may be a better way to do this.

saschanaz: I think the following two sentences are ambiguous become the grammars are same:

* A module declaration is a definition matching the Module nonterminal symbol.
* A partial module declaration is a definition matching the PartialModule nonterminal symbol.

littledan:
> I think I'm happy to squash and land this. 

I'd be happy to see this land, but could you make a big "work in progress within the draft, do not use yet" at the top? @domenic and I have been talking about whether interfaces should switch into some "new defaults" when inside of a module, which would bring sizable semantics changes.

littledan:
cc @mattijs

domenic:
@annevk, are you OK with landing this? Chrome is interested in implementing, and my understanding is that folks like @lukewagner and @tschneidereit were also interested. (@Ms2ger, there are some merge conflicts.)

annevk:
Is there a downstream PR? What about the sizeable semantic changes @littledan mentioned?

domenic:
I'm unsure what you mean by "downstream"?

We indeed might make sizable semantics changes before shipping; those are captured in e.g. #718 and #719. But it's good to make incremental progress, so that specs can use modules at all, before we contemplate exactly what might change by default inside modules.

annevk:
I meant a PR for a spec that wants to use this as-is.

domenic:
Ah, I'll work on that today.

domenic:
For posterity, @Ms2ger notes that he's already started in https://github.com/WICG/kv-storage/compare/master...Ms2ger:idl?expand=1, and in fact the modules part is done there. (It appears that parts that are unfinished are related to #720.)

domenic:
I think we should be able to land this after rebasing. https://github.com/WICG/kv-storage/pull/68 is what wants to use this.

domenic:
@annevk ping on whether you're OK to land this. I am going to land https://github.com/WICG/kv-storage/pull/68 today with reference to this pull request, but basing it on the actual IDL spec would be even better.

annevk:
For clarity, while Mozilla is supportive of KV Storage, it's still a little unclear what is happening with Builtin Modules (see https://github.com/mozilla/standards-positions/issues/147) which means I'm not comfortable moving forward with this.

littledan:
Given the current state of built-in modules in TC39, I think we're not ready for WebIDL support for built-in modules yet. I'm going to close this PR and defer to the efforts of @msaboff @mattijs and others, to build the foundation before we consider higher-level aspects.

PR #679, opened 2019-03-07
Correct the 'indexed getter' cross-reference

Ms2ger:
***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/679.html" title="Last updated on Mar 7, 2019, 2:23 PM UTC (b549ba8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/679/a94e4b4...b549ba8.html" title="Last updated on Mar 7, 2019, 2:23 PM UTC (b549ba8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4205,7 +4205,7 @@ A maplike interface and its [=inherited interfaces=]
 must not have an
 [=iterable declaration=],
 a [=setlike declaration=], or
-an [=indexed getter=].
+an [=indexed property getter=].
 
 <div data-fill-with="grammar-ReadOnlyMember"></div>
 
@@ -4298,7 +4298,7 @@ A setlike interface and its [=inherited interfaces=]
 must not have an
 [=iterable declaration=],
 a [=maplike declaration=], or
-an [=indexed getter=].
+an [=indexed property getter=].
 
 <div data-fill-with="grammar-ReadOnlyMember"></div>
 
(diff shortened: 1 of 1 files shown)


Ms2ger:
@tabatkins I accidentally pointed at https://drafts.css-houdini.org/css-typed-om-1/#cssnumericarray-indexed-getter here; I'm surprised that spec defines such a generic term globally. Is that intentional?

tabatkins:
> I'm surprised that spec defines such a generic term globally

It doesn't, the definition is scoped to CSSNumericArray: `<dfn data-dfn-for="CSSNumericArray" data-dfn-type="dfn" data-export="" id="cssnumericarray-indexed-getter">indexed getter<a class="self-link" href="#cssnumericarray-indexed-getter"></a></dfn>`

However, this might be the *only* defining instance of that term, so an unqualified autolink will match the dfn without complaining.  Was that term always "indexed property getter" in WebIDL? If so, I should update TypedOM to use the right word. ^_^

Ms2ger:
>  Was that term always "indexed property getter" in WebIDL? If so, I should update TypedOM to use the right word. ^_^

At least as far back as https://www.w3.org/TR/2011/WD-WebIDL-20110927/ :)

tabatkins:
Cool, I've just renamed the Typed OM term properly, and had it link to the WebIDL term. ^_^

PR #686, opened 2019-03-11
Disallow constants to be null

### LINKED ISSUE #448: Investigate removing nullables from allowed constant types
Looks like nullable types aren't actually used by constants on the platform, so we might as well disallow them altogether.

See #446.

saschanaz:
Closes #448.

No known specs are using constant null, and we discourage adding new constants (with some specs as exceptions), so I think we can safely remove this.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/webidl/pull/686.html" title="Last updated on Mar 11, 2019, 11:02 AM UTC (6094723)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/686/cc2fa40...saschanaz:6094723.html" title="Last updated on Mar 11, 2019, 11:02 AM UTC (6094723)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1748,7 +1748,7 @@ The following extended attributes are applicable to constants:
 
 <pre class="grammar" id="prod-Const">
     Const :
-        "const" PrimitiveType identifier "=" ConstValue ";"
+        "const" ConstType identifier "=" ConstValue ";"
 </pre>
 
 <pre class="grammar" id="prod-ConstValue">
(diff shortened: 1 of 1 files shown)


Ms2ger:
Travis is complaining about the grammar, it seems.

domenic:
Can someone explain "restore ConstType to not break WebGL"? It seems better not to have it.

Ms2ger:
WebGL uses typedefs for integer types, so we kinda need to support that.

domenic:
How is that related to the ConstType grammar production?

saschanaz:
WebGL has a lot of 'const GLint', so removing ConstType will break it.

bzbarsky:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1535647 on Gecko.

PR #688, opened 2019-03-15
Make the grammar check more robust

Ms2ger:
In particular, fail immediately if the server returns with an error, and yield
a readable error if no grammar snippets are found in the specification.


annevk:
Nit: commit title shouldn't end in a dot.

PR #690, opened 2019-03-19
Remove support for overloading callbacks

Ms2ger:
This was only supported in the definition of the effective overload set, but
this wasn't actually used anywhere, and there is no syntax to overload a
callback anyway.

The six types of IDL constructs are the five that were in the specification
before this change, and legacycallers. The count was not updated when these
were removed.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/690.html" title="Last updated on Mar 19, 2019, 10:21 AM UTC (404e94f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/690/b0621a8...404e94f.html" title="Last updated on Mar 19, 2019, 10:21 AM UTC (404e94f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3215,7 +3215,7 @@ constructor (specified with [{{Constructor}}]
 or [{{NamedConstructor}}]), or
 [=callback function=].
 The algorithm to compute an [=effective overload set=]
-operates on one of the following six types of IDL constructs, and listed with them below are
+operates on one of the following four types of IDL constructs, and listed with them below are
 the inputs to the algorithm needed to compute the set.
 
  :  For regular operations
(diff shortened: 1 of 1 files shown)


bzbarsky:
> and there is no syntax to overload a callback anyway.

It sort of depends on what "overload" means.  But for purposes of effective overload sets, those are computed in the following places:

1) Steps and length for constructors.
2) Steps and length for named constructors.
3) Steps and length for operation functions.

Callback functions don't use overload sets and don't call into the overload resolution algorithm.  They go through https://heycam.github.io/webidl/#es-invoking-callback-functions which presumes we already have a list of Web IDL things.

bzbarsky:
And in particular, nothing in the spec actually enforces that the list passed to https://heycam.github.io/webidl/#es-invoking-callback-functions has types that match the function declaration.  Maybe we should have something, but it would need to look a bit different from the overload resolution algorithm.

PR #696, opened 2019-03-25
Remove unused callback interface functionality

### LINKED ISSUE #661: Should callback interfaces with more than one operation be supported?
Looking at Gecko, there's `EventListener`, `NodeFilter` and `XPathNSResolver`, all with one operation; `ObserverCallback`, also with one operation, but shouldn't even exist (<https://bugzilla.mozilla.org/show_bug.cgi?id=1530702>); and `MozIdleObserver`, which isn't exposed to script. WebKit doesn't seem to have any others either.

There's already a note that suggests using dictionaries instead of callback interfaces, so I think we should bite the bullet and mark the entire thing as legacy, and reduce the spec to only support what's actually used.

Ms2ger:
Fixes #661.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/696.html" title="Last updated on Apr 2, 2019, 1:59 PM UTC (fdeb0a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/696/7ce9ef8...fdeb0a9.html" title="Last updated on Apr 2, 2019, 1:59 PM UTC (fdeb0a9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -931,21 +931,23 @@ Note: See also the similarly named [=callback function=] definition.
 
 [=Callback interfaces=]
 must not [=interface/inherit=]
-from any non-callback interfaces, and non-callback interfaces must not
+from any interfaces, and non-callback interfaces must not
 inherit from any callback interfaces.
 Callback interfaces must not [=include=] any [=interface mixins=].
 
-[=Static attributes=] and
+[=Regular attributes=],
+[=static attributes=] and
 [=static operations=] must not
 be defined on a [=callback interface=].
+
+[=Callback interfaces=] must define exactly one [=regular operation=].
 
 <div class="advisement">
 
     Specification authors should not define
     [=callback interfaces=]
-    that have only a single [=operation=],
     unless required to describe the requirements of existing APIs.
-    Instead, a [=callback function=] should be used.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -931,19 +931,21 @@ Note: See also the similarly named [=callback function=] definition.
...
 [=static operations=] must not
 be defined on a [=callback interface=].
 
+[=Callback interfaces=] must define exactly one [=regular operation=].
+
 <div class="advisement">
 
     Specification authors should not define
     [=callback interfaces=]
-    that have only a single [=operation=],
```
littledan: Are you sure about this deletion?

Ms2ger: We don't want people to use callback interfaces at all in new APIs, so yes. (And given the other changes, there's no point in having this line, because all callback interfaces only have a single operation.)

index.bs
```diff
@@ -931,19 +931,21 @@ Note: See also the similarly named [=callback function=] definition.
...
     unless required to describe the requirements of existing APIs.
     Instead, a [=callback function=] should be used.
```
domenic: Maybe "or dictionary" to replace the below deleted example?

Issue #698, opened 2019-03-26
Define that interface objects are constructors


Ms2ger:
ES claims the following (twice):

> Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.

We don't normatively identify the interface object as a constructor, even if it has a `[Constructor]`. I guess we should.

A less obvious question is what we should do if there's no `[Constructor]`. Right now, I think we're aiming to install a [[Construct]] internal method that always throws, but that makes the interface object pass [IsConstructor()](https://tc39.github.io/ecma262/#sec-isconstructor).

bzbarsky:
> unless otherwise specified in the description of a particular function

This is the key, no?  https://heycam.github.io/webidl/#Constructor specifies otherwise:

> If the [Constructor] extended attribute appears on an interface, it indicates that the interface object for this interface will have an [[Construct]] internal method

So I think we're good here: thing with a `[Constructor]` have `[[Construct]]` and things without do not, as desired.

bzbarsky:
Though I guess that leaves open the question of the no-`[Constructor]` case.  I just checked Gecko, and it _does_ have a `[[Constructor]]` in that case, albeit one that always throws.  I wonder whether the output of `IsConstructor()` is observable, other than by the exceptions that get thrown.

bzbarsky:
I guess you could do something where you use a DOM interface object as the `.constructor` of an array and see what happens in https://tc39.github.io/ecma262/#sec-arrayspeciescreate step 6.

bzbarsky:
And also whether a proxy for a DOM interface object ends up with a construct trap being called.  So simple testcase:
```
<script>
  var p = new Proxy(Node, { construct: () => console.log("constructing") });
  new p()
</script>
```

bzbarsky:
On that testcase, Chrome and Firefox log "constructing"; Safari does not (and the exception it throws is about `p` not being a constructor).  Yay interop.  ;)

But OK, if we want to spec Chrome/Firefox behavior here then we need to change something in the idl spec.

bzbarsky:
And admittedly it would be good if the "is a constructor" bit was near the bit that defines the steps, not way off somewhere else...

domenic:
I've always assumed they had constructors (and [[Construct]] methods), just throwing ones.

I think it'd be very strange to have callable interface objects exist at all, and be reachable via the `.constructor` property, if they are not also [[Construct]]able.

[...]

PR #700, opened 2019-03-29
Stop using extended attributes for constructors

### LINKED ISSUE #636: Stop using extended attributes for constructors
There has been some discussion about this in #485; I'm filing this to consider changing `[Constructor]` in particular, since nobody seemed to particularly object to it.

I'd like to propose the following syntax:

```webidl
interface Foo {
  constructor(...);
};
```

Without further changes, I believe the only WebIDL-defined extended attributes that would potentially start applying are `[Exposed]`, `[SecureContext]` and `[Unforgeable]`. It seems fine to allow `[Exposed]` and `[SecureContext]`, but I would disallow `[Unforgeable]`.

We should probably not allow constructors to be specified within mixins.

Please comment asap if you have thoughts on this.

CC @yuki3 @littledan @bzbarsky @tabatkins @domenic @annevk

Ms2ger:
Fixes #636.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/700.html" title="Last updated on Aug 27, 2019, 2:00 PM UTC (dba5577)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/700/68dbd23...dba5577.html" title="Last updated on Aug 27, 2019, 2:00 PM UTC (dba5577)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -543,8 +543,9 @@ in [[#es-extended-attributes]].
           attribute DOMString imageURL;
         };
 
-        [Exposed=Window, Constructor]
+        [Exposed=Window]
         interface GraphicalWindow {
+          constructor();
           readonly attribute unsigned long width;
           readonly attribute unsigned long height;
 
@@ -571,12 +572,13 @@ in [[#es-extended-attributes]].
     objects; each ECMAScript object that implements <code class="idl">GraphicalWindow</code>
     will have that prototype object in its prototype chain.
 
-    The [{{Constructor}}] that appears on <code class="idl">GraphicalWindow</code>
-    is an [=extended attribute=].
-    This extended attribute causes a [=constructor=] to exist in ECMAScript implementations,
+    The [=constructor operation=] that appears on <code class="idl">GraphicalWindow</code>
+    causes a [=constructor=] to exist in ECMAScript implementations,
     so that calling <code>new GraphicalWindow()</code> would return a new object
     that implemented the interface.
 
+    All [=interfaces=] have the [{{Exposed}}] [=extended attribute=], which ensures the interfaces
+    are only available in [=Realms=] whose [=Realm/global object=] is a {{Window}} object.
 </div>
 
 
@@ -992,7 +994,6 @@ The relevant language binding determines how interfaces correspond to constructs
 in the language.
 
 The following extended attributes are applicable to interfaces:
-[{{Constructor}}],
 [{{Exposed}}],
 [{{Global}}],
 [{{LegacyWindowAlias}}],
@@ -1061,7 +1062,7 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
 
 <pre class="grammar" id="prod-PartialInterfaceRest">
     PartialInterfaceRest :
-        identifier "{" InterfaceMembers "}" ";"
+        identifier "{" PartialInterfaceMembers "}" ";"
 </pre>
 
 <pre class="grammar" id="prod-InterfaceMembers">
@@ -1072,6 +1073,18 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
 
 <pre class="grammar" id="prod-InterfaceMember">
     InterfaceMember :
+        PartialInterfaceMember
+        Constructor
+</pre>
+
+<pre class="grammar" id="prod-PartialInterfaceMembers">
+    PartialInterfaceMembers :
+        ExtendedAttributeList PartialInterfaceMember PartialInterfaceMembers
+        ε
+</pre>
+
+<pre class="grammar" id="prod-PartialInterfaceMember">
+    PartialInterfaceMember :
         Const
         Operation
         Stringifier
@@ -2166,7 +2179,8 @@ language bindings.
     </pre>
 </div>
 
-An operation is considered to be <dfn id="dfn-variadic" export>variadic</dfn>
+An operation or [=constructor operation=] is considered to be
+<dfn id="dfn-variadic" export>variadic</dfn>
 if the final argument uses the <emu-t>...</emu-t> token just
 after the argument type.  Declaring an operation to be variadic indicates that
 the operation can be invoked with any number of arguments after that final argument.
@@ -2185,8 +2199,7 @@ is the final argument in the operation’s argument list.
 
 [=Extended attributes=]
 that [=takes an argument list|take an argument list=]
-([{{Constructor}}] and
-[{{NamedConstructor}}], of those
+([{{NamedConstructor}}], of those
 defined in this specification) and [=callback functions=]
 are also considered to be [=variadic=]
 when the <emu-t>...</emu-t> token is used in their argument lists.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2540,6 +2542,82 @@ in which case the [=default toJSON operation=] is exposed instead.
...
 
 
+<h4 id=idl-constructors oldids="Constructor" dfn>Constructor methods</h4>
+
+If an [=interface=] has a [=constructor method=] member (matching
+<emu-nt><a href="#prod-Constructor">Constructor</a></emu-nt>), it indicates that it is possible to
+create objects that [=implement=] the [=interface=] using a [=constructor=].
+
+Multiple [=constructor methods=] may appear on a given [=interface=].
+For each [=constructor method=] on the [=interface=], there will be a way to construct an instance
```
annevk: might be* (it could throw)

index.bs
```diff
@@ -2540,6 +2542,82 @@ in which case the [=default toJSON operation=] is exposed instead.
...
+          readonly attribute double circumference;
+        };
+    </pre>
+
+    An ECMAScript implementation supporting these interfaces would have a \[[Construct]] property
+    on the <code class="idl">Circle</code> interface object which would return a new object that
+    [=implements=] the interface.
+    It would take either zero or one argument.
+    The <code class="idl">NodeList</code> interface object would not have a \[[Construct]]
+    property.
```
annevk: I thought it would have a throwing one? And the way we'd explain how an implementation was able to create it was by passing some magical token.

index.bs
```diff
@@ -9432,15 +9428,12 @@ will not exist for the interface in the ECMAScript binding.
...
-A [{{NamedConstructor}}] extended attribute is fine,
-however.
-
 The [{{NoInterfaceObject}}] extended attribute
 must not be specified on an interface that has any
-[=static operations=] defined on it.
+[=constructors=] or [=static operations=] defined on it.
+
+Note: Combining the [{{NoInterfaceObject}}] and [{{NamedConstructor}}] extended attribute is not
+forbidden, however.
```
annevk: Hmm, maybe it should be? I guess that would be a separate change.

marcoscaceres:
@annevk is right. I can also help update legacy ReSpec specs using the old constructor. 

@saschanaz and I can take care of things on WebIDL2.js, which WPT relies on.

index.bs
Ms2ger: I wanna remove `[NoInterfaceObject]` anyway.

index.bs
Ms2ger: Unclear: https://github.com/heycam/webidl/issues/698.

domenic: We should either have this be an XXX unclear box, or block this on resolving #698.

marcoscaceres:
@foolip @domenic, quick list from grepping wpt's interfaces directory:
https://gist.github.com/marcoscaceres/98241f92092132cba703e9be89500061

Gives us some idea... it looks like we would need to update ~75 specs.

marcoscaceres:
Here is the list of matching files/specs: https://gist.github.com/marcoscaceres/c4d507839522d6bc96df23d460526c6a

foolip:
https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl is the source of the IDL files in WPT and what I've been treating as "all the web's IDL" for some time now.

I just finished fixing specs missed in the `[NoInterfaceObject] interface`+`implements` to `interface mixin`+`includes` transition, so making sure *all* changes have propagated to reffy-reports before using `[Constructor]` becomes invalid and is dropped from webidl2.js would be much appreciated.

saschanaz:
What is this waiting for now? Should issues be filed for browser implementations?

bzbarsky:
Well, for starters it's not clear whether implementors are in fact on board...

domenic:
Could you speak to that for Gecko? Chromium is on board.

bzbarsky:
I am still sorting out the Gecko position here.

In terms of the actual change, after this you won't be able to use "constructor" as an argument name, dictionary member name, attribute name, or method name, unless you write it as `constructor_`.  For attribute and method names this seems OK.  For argument names, I'd suspect we want to add it to the _ArgumentNameKeyword_ production, right?  For dictionary members, do we want to allow it?  It'd require a new production for dictionary member names, but that seems fine to me... and it would be potentially useful if people want to extract things like ES class info from objects via dictionaries.

I did just check and there are no existing IDLs in Gecko that use "constructor" in any of those situations, by the way.

bzbarsky:
@domenic @ms2ger it looks like the proposal disallows constructors on partial interfaces in prose, but it seems like we could do it on the syntax level now that there is an explicit syntactic construct.  Could we do that?  That check would be a lot more likely to actually get implemented in tools that way, I suspect, since just updating the grammar would do it.

saschanaz:
A quick search on reffy-reports says the only place that use the name `constructor` is `CustomElementRegistry`.

bzbarsky:
OK, apart from the grammar issues mentioned above, I think Gecko is on board.

bzbarsky:
Though I guess `constructor` is explicitly a reserved identifier (such that you can't use `_constructor` either, even without these changes).

In that case, the only grammar concern I have for now is the partial interface situation.

Ms2ger:
Made the change for partial interfaces. I think that resolves everything besides the comment about "constructor method". I'll change that to "operation" tomorrow, but the PR should be ready for re-review now.

Who works on this for WebKit now? @rniwa?

index.bs
```diff
@@ -1065,6 +1066,18 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
 
 <pre class="grammar" id="prod-InterfaceMember">
     InterfaceMember :
+        PartialInterfaceMember
+        Constructor
+</pre>
+
+<pre class="grammar" id="prod-PartialInterfaceMembers">
+    InterfaceMembers :
```
bzbarsky: This line should say `PartialInterfaceMembers :`

index.bs
```diff
@@ -3218,7 +3308,7 @@ definitions.
         };
     </pre>
 
-    Note that the [{{Constructor}}] and
+    Note that [=constructor methods=] and
```
bzbarsky: Do we still need the prose mention here now that the grammar enforces it?

index.bs
```diff
@@ -8286,87 +8375,6 @@ for the specific requirements that the use of
...
-
-The prose definition of a constructor must either initialize the value passed as <b>[=this=]</b>,
-or throw an exception.
-
-The [{{Constructor}}] and
-[{{NoInterfaceObject}}]
-extended attributes must not be specified on the same interface.
-
-The [{{Constructor}}] and [{{Global}}] [=extended attributes=] must not be specified on the same
-[=interface=].
```
bzbarsky: This requirement has disappeared.  It should be reinstated.

index.bs
Ms2ger: It's just a note; I think it's okay to keep.

Ms2ger:
> Looks reasonable to me. Would still like feedback from WebKit.

@cdumez @youennf

Ms2ger:
I propose we merge if Apple doesn't comment by next Monday, the 19th.

youennf:
I see some potential future benefits but this seems quite a big syntactic change.
Is there any concrete plan for specs to use this new ability?

In terms of WebKit implementation, our constructor-specific keywords (canthrow, callwith, custom) have counterparts for operations, this seems fine and I quite like it in fact.

I wonder what the plan is for NamedConstructor.
WebKit implementation would need to keep these constructor-specific keywords around for NamedConstructor if it is not migrated as well.

Ms2ger:
I'm expecting to file spec bugs based on a combination of scraping browser source code and adding errors in idlharness.js.

I think changing NamedConstructor as well would make sense, but don't have a timeline for that yet.

domenic:
I'm not so sure about NamedConstructor, since that adds a factory function to the global, and doesn't add a member of the class (or effect the class in any way). Keeping it as-is seems better to me... Or if anything, making it a separate top-level declaration.

tabatkins:
You write out an Infra list literal like `« "a", "b", "c", "a" »`, and do map iteration like `key → value`. These non-ASCII characters are nice for grabbing attention and being unambiguous in text, but they're annoying to type - I tend to just go copy-paste from the Infra spec.

We should come up with an ASCII way to write these literals in your source, that'll turn into the nice unicode characters in the output.

/cc @marcoscaceres to match up in syntax.

PR #713, opened 2019-04-15
Editorial: Shorten the conditions in the [[Set]] algorithm

Ms2ger:
An interface can only have a (named or indexed) setter if it also has the
corresponding getter, so checking for both is unnecessary.


Ms2ger:
[[DefineOwnProperty]] is different because it has to handle the case where the interface has a getter but no setter. [[Set]] doesn't.

PR #715, opened 2019-04-16
Limit the special case in the [[Set]] algorithm to [OverrideBuiltins] interfaces

### LINKED ISSUE #630: Consider removing call to named property setter in legacy platform object [[Set]]
Test:
```html
<!DOCTYPE html>
<script>
sessionStorage.clear()
var a = new Proxy(Object.getPrototypeOf(sessionStorage), {
  set(...args) { w(args); return false; },
});
Object.setPrototypeOf(sessionStorage, a);
sessionStorage.x = "foo";
w(sessionStorage.x)
</script>
```
Gecko and Chrome call proto.\[[Set]]() rather than taking the early return, as shown by the log:
```
log: object "[object Storage],x,foo,[object Storage]" (4 props: 0="[object Storage]", 1="x", 2="foo", 3="[object Storage]")
log: undefined
```
but WebKit may be following the spec:

```
log: foo
```

Ms2ger:
This is the case where it makes most sense, since the prototype chain can't
shadow any properties.

This change does not affect normal usage; it is only detectable by messing with
the prototype of the object.

The majority of implementations (Gecko and Chrome) already follow the proposed
change; WebKit follows the existing spec.

Fixes #630.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12534,6 +12534,7 @@ Additionally, [=legacy platform objects=] have internal methods as defined in:
             1.  [=Invoke the indexed property setter=] with |P| and |V|.
             1.  Return <emu-val>true</emu-val>.
         1.  If |O| [=implements=] an interface with a [=named property setter=]
+            and the [{{OverrideBuiltins}}] [=extended attribute=],
             and <a abstract-op>Type</a>(|P|) is String, then:
             1.  [=Invoke the named property setter=] with |P| and |V|.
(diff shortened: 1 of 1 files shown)


ExE-Boss:
If we do this, then <https://github.com/web-platform-tests/wpt/blob/master/webstorage/set.window.js> will start failing in the same way as **WebIDL2JS**, which currently has a buggy implementation of this (<https://github.com/jsdom/webidl2js/pull/218>).

PR #719, opened 2019-04-26
Introduce an UnenumerableOperations extended attribute

Ms2ger:
***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/719.html" title="Last updated on Jun 17, 2019, 2:50 PM UTC (38f2af8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/719/059491b...38f2af8.html" title="Last updated on Jun 17, 2019, 2:50 PM UTC (38f2af8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10098,6 +10098,8 @@ See [[#es-DOMString]] for the specific requirements that the use of [{{TreatNull
 
 <h4 id="UnenumerableOperations" extended-attribute lt="UnenumerableOperations">[UnenumerableOperations]</h4>
 
+The [{{UnenumerableOperations}}] [=extended attribute=] is only applicable to [=interfaces=].
+
 If the [{{UnenumerableOperations}}] [=extended attribute=] appears on a [=interface=], it indicates
 that all the [=interface=]'s [=operations=] are unenumerable.
 
--- a/index.bs
+++ b/index.bs
@@ -10096,6 +10096,18 @@ See [[#es-DOMString]] for the specific requirements that the use of [{{TreatNull
 </div>
 
 
+<h4 id="UnenumerableOperations" extended-attribute lt="UnenumerableOperations">[UnenumerableOperations]</h4>
+
+If the [{{UnenumerableOperations}}] [=extended attribute=] appears on a [=interface=], it indicates
+that all the [=interface=]'s [=operations=] are unenumerable.
+
+The [{{UnenumerableOperations}}] [=extended attribute=] must
+[=takes no arguments|take no arguments=].
+
+See [[#es-operations]] for the specific requirements that the use of [{{UnenumerableOperations}}]
+entails.
+
+
 <h4 id="Unforgeable" extended-attribute lt="Unforgeable">[Unforgeable]</h4>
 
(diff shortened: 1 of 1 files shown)


domenic:
Thanks for working on this! I have two thoughts:

- Should we provide some sort of guidance on when to use or not use this? It seems quite experimental. The nightmare scenario for me is various spec authors sprinkling this in their specs because they think enumerability is weird, or non-enumerability is more JavaScript like. ("Various spec authors", like myself 5 years ago.)

- Alternately, we could make this part of the "new defaults" inside modules, and not have a separate extended attribute at all.

index.bs
```diff
@@ -10096,6 +10096,18 @@ See [[#es-DOMString]] for the specific requirements that the use of [{{TreatNull
 </div>
 
 
+<h4 id="UnenumerableOperations" extended-attribute lt="UnenumerableOperations">[UnenumerableOperations]</h4>
+
+If the [{{UnenumerableOperations}}] [=extended attribute=] appears on a [=interface=], it indicates
```
littledan: Nit: Should there be a requirement that this extended attribute only be used on interfaces?

annevk:
I agree with @domenic.

If we want this for modules, but are unsure outside of modules, let's limit it to modules.

If we are sure outside of modules, let's make it the default and sprinkle `LegacyEnumerableOperations` around.

Ms2ger:
I am not sure this is desirable anywhere, but I updated the PR to make this the behaviour for modules.

index.bs
```diff
@@ -11215,14 +11232,17 @@ in which case they are exposed on every object that [=implements=] the interface
     To <dfn>define the operations</dfn> |operations| of [=interface=] or [=namespace=] |definition| on |target|,
     given [=Realm=] |realm|, run the following steps:
 
+    1.  Let |enumerable| be <emu-val>false</emu-val> if |definition| is an [=interface=] that has
+        the [{{UnenumerableOperations}}] [=extended attribute=] and <emu-val>true</emu-val>
+        otherwise.
```
ExE-Boss: This&nbsp;should also&nbsp;be&nbsp;defined for&nbsp;[**Stringifiers**](https://heycam.github.io/webidl/#es-stringifier), [**Common&nbsp;iterator&nbsp;behavior** §&nbsp;`forEach`](https://heycam.github.io/webidl/#es-forEach) and&nbsp;[**Iterable&nbsp;declarations**](https://heycam.github.io/webidl/#es-iterable), like&nbsp;I’m&nbsp;doing&nbsp;in&nbsp;<https://github.com/heycam/webidl/pull/825>.

littledan:
This PR was motivated by using WebIDL in a way that bridges the gap a bit with JS classes (which might relate to some built-in modules proposals, to facilitate accurate polyfilling) and TC39 conventions. It seems like we're not currently moving in that direction very much at the moment, so leaving this PR hanging (or closing it?) seems like right call.

domenic:
Let's close this until we have a concrete consumer.

Issue #490, opened 2017-12-06
Import (and fix) the promises guide


annevk:
Most of the algorithms in https://www.w3.org/2001/tag/doc/promises-guide for resolving and rejecting should be part of IDL (and formalized in the process). Resolving from "in parallel" should be something that HTML defines I think.

tobie:
Mmm. I was sure we already had an open issue for this but can't find it.

annevk:
It's not linked from https://github.com/w3ctag/promises-guide/issues/27 (and now this is).

domenic:
As background, one of the major things wrong with the promises guide is that it was written before I was bought into the Web IDL ecosystem, so it tries to be usable from JS specs as well. A lot of the definitions take as arguments JS values or JS functions, whereas you really want to be using them with IDL values or lists of steps.

https://www.w3.org/2001/tag/doc/promises-guide#shorthand-phrases is the section in question, mainly.

Here is what I would do for this issue, if I had the time to do so.

- Redefine "a promise resolved with" to use PromiseResolve(%Promise%, x converted from Web IDL to JS).
  - Potentially separate out two variants of it https://github.com/w3ctag/promises-guide/issues/56#issuecomment-438499875
  - Nobody uses "x resolved as a promise"; delete that
- Redefine "a promise rejected with" to use %Promise_reject%.
- Fix all call sites of "resolve a promise" etc. to explicitly queue a task, instead of the current implicit task queuing. This fixes https://github.com/w3ctag/promises-guide/issues/52.
- Make the Web IDL promise type a wrapper around a PromiseCapability record, not just a JS Promise object.
- Formalize "resolve" and "reject" on Web IDL promise types using the [[Resolve]] and [[Reject]] record fields. Insert type conversions as appropriate. Potentially separate out "resolved with" and "resolved with an arbitrary value".
- Re-do Web IDL's "perform some steps once a promise is settled" to use PerformPromiseThen (with no resultCapability).
    - Add examples of how to use "perform some steps once a promise is settled". Draw on existing usages of "upon fulfillment" and "upon rejection" in promise-heavy specs (e.g. Streams, Service Worker, maybe HTML) to ensure that this pattern is ergonomic.
    - If the pattern is not ergonomic, then update "upon fulfillment" and "upon rejection" to be specializations of "perform some steps once a promise is settled". If it is ergonomic, delete those terms and update all specs to use "perform some steps once a promise is settled".
[...]

PR #720, opened 2019-05-01
Add async_iterable support

### LINKED ISSUE #580: Async iterators
It'd be nice if we could create async iterators in a similar way to iterators.

```webidl
interface Whatever {
  // …
  async_iterable<Something>;
};
```

The prose would need to be a little more complicated than `iterable`, probably written like a generator.

For example:

```webidl
interface SlowCounter {
  // …
  async_iterable<unsigned long long>;
};
```

To asynchronously iterate over a `SlowCounter`, run the following steps:

1. Run these steps *in parallel*:
    1. Let *i* be 0.
    1. While true:
        1. Wait 1 second.
        1. Yield the value of *i* using the networking task source.
        1. Increment *i* by 1.

"Yield" would queue a task on the given task source to resolve the iteration's promise with the given value. The algorithm would not advance until the next iteration was requested.

I guess it's tricky as task sources are an HTML thing, but I can't see how it can be handled correctly otherwise.

Ms2ger:
Fixes #580.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/720.html" title="Last updated on Aug 7, 2019, 10:20 AM UTC (6ac2d3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/720/13c7e8f...6ac2d3d.html" title="Last updated on Aug 7, 2019, 10:20 AM UTC (6ac2d3d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11783,6 +11783,33 @@ for the interface.
 The \[[Prototype]] [=internal slot=] of an [=iterator prototype object=]
 must be {{%IteratorPrototype%}}.
 
+<div algorithm>
+  The <dfn>iterator result</dfn> for a pair of IDL values |pair| and a kind |kind| is given by the
+  following steps:
+
+    1.  Let |result| be a value determined by the value of |kind|:
+        <dl class="switch">
+             :  "<code>key</code>"
+             :: 1.  Let |idlKey| be |pair|’s key.
+                1.  Let |key| be the result of [=converted to an ECMAScript value|converting=] |idlKey| to an ECMAScript value.
+                1.  |result| is |key|.
+             :  "<code>value</code>"
+             :: 1.  Let |idlValue| be |pair|’s value.
+                1.  Let |value| be the result of [=converted to an ECMAScript value|converting=] |idlValue| to an ECMAScript value.
+                1.  |result| is |value|.
+             :  "<code>key+value</code>"
+             :: 1.  Let |idlKey| be |pair|’s key.
+                1.  Let |idlValue| be |pair|’s value.
+                1.  Let |key| be the result of [=converted to an ECMAScript value|converting=] |idlKey| to an ECMAScript value.
+                1.  Let |value| be the result of [=converted to an ECMAScript value|converting=] |idlValue| to an ECMAScript value.
+                1.  Let |array| be the result of performing [$ArrayCreate$](2).
+                1.  Call [$CreateDataProperty$](|array|, "<code>0</code>", |key|).
+                1.  Call [$CreateDataProperty$](|array|, "<code>1</code>", |value|).
+                1.  |result| is |array|.
+        </dl>
+    1.  Return [$CreateIterResultObject$](|result|, <emu-val>false</emu-val>).
+</div>
+
 <div algorithm="to invoke the next property of iterators">
 
     An [=iterator prototype object=] must have a <code class="idl">next</code> data property with attributes
@@ -11807,27 +11834,7 @@ must be {{%IteratorPrototype%}}.
         return <a abstract-op>CreateIterResultObject</a>(<emu-val>undefined</emu-val>, <emu-val>true</emu-val>).
     1.  Let |pair| be the entry in |values| at index |index|.
     1.  Set |object|’s index to |index| + 1.
-    1.  Let |result| be a value determined by the value of |kind|:
-        <dl class="switch">
-             :  "<code>key</code>"
-             :: 1.  Let |idlKey| be |pair|’s key.
-                1.  Let |key| be the result of [=converted to an ECMAScript value|converting=] |idlKey| to an ECMAScript value.
-                1.  |result| is |key|.
-             :  "<code>value</code>"
-             :: 1.  Let |idlValue| be |pair|’s value.
-                1.  Let |value| be the result of [=converted to an ECMAScript value|converting=] |idlValue| to an ECMAScript value.
-                1.  |result| is |value|.
-             :  "<code>key+value</code>"
-             :: 1.  Let |idlKey| be |pair|’s key.
-                1.  Let |idlValue| be |pair|’s value.
-                1.  Let |key| be the result of [=converted to an ECMAScript value|converting=] |idlKey| to an ECMAScript value.
-                1.  Let |value| be the result of [=converted to an ECMAScript value|converting=] |idlValue| to an ECMAScript value.
-                1.  Let |array| be the result of performing <a abstract-op>ArrayCreate</a>(2).
-                1.  Call <a abstract-op>CreateDataProperty</a>(|array|, "<code>0</code>", |key|).
-                1.  Call <a abstract-op>CreateDataProperty</a>(|array|, "<code>1</code>", |value|).
-                1.  |result| is |array|.
-        </dl>
-    1.  Return <a abstract-op>CreateIterResultObject</a>(|result|, <emu-val>false</emu-val>).
+    1.  Return the [=iterator result=] for |pair| and |kind|.
 </div>
 
 The [=class string=] of an [=iterator prototype object=] for a given [=interface=]
--- a/index.bs
+++ b/index.bs
@@ -83,6 +83,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: Promise; url: sec-promise-objects
         text: Set; url: sec-set-objects
         text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
+        text: %AsyncIteratorPrototype%; url: sec-asynciteratorprototype
         text: %ErrorPrototype%; url: sec-properties-of-the-error-prototype-object
         text: %FunctionPrototype%; url: sec-properties-of-the-function-prototype-object
         text: %IteratorPrototype%; url: sec-%iteratorprototype%-object
@@ -106,6 +107,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %PromiseProto_then%; url: sec-promise.prototype.then
     type: const; for: ECMAScript
         url: sec-well-known-symbols
+            text: @@asyncIterator
             text: @@iterator
             text: @@toStringTag
             text: @@unscopables
@@ -134,6 +136,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: GetFunctionRealm; url: sec-getfunctionrealm
         text: GetIterator; url: sec-getiterator
         text: GetMethod; url: sec-getmethod
+        text: IfAbruptRejectPromise; url: sec-ifabruptrejectpromise
         text: IsAccessorDescriptor; url: sec-isaccessordescriptor
         text: IsCallable; url: sec-iscallable
         text: IsConstructor; url: sec-isconstructor
@@ -144,6 +147,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: IteratorStep; url: sec-iteratorstep
         text: IteratorValue; url: sec-iteratorvalue
         text: NewModuleEnvironment; url: sec-newmoduleenvironment
+        text: NewPromiseCapability; url: sec-newpromisecapability
         text: NormalCompletion; url: sec-normalcompletion
         text: ObjectCreate; url: sec-objectcreate
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
@@ -1070,6 +1074,7 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
         Stringifier
         StaticMember
         Iterable
+        AsyncIterable
         ReadOnlyMember
         ReadWriteAttribute
         ReadWriteMaplike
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4091,6 +4104,138 @@ The following extended attributes are applicable to [=iterable declarations=]:
 </pre>
 
 
+<h4 id="idl-async-iterable">Asynchronously iterable declarations</h4>
```
domenic: Nit: I've never seen it spelled out as "asynchronously iterable". Maybe an "asynchronous iterable", as in the title of https://github.com/tc39/proposal-async-iteration. But "async iterable" is much more common.

index.bs
```diff
@@ -4091,6 +4104,138 @@ The following extended attributes are applicable to [=iterable declarations=]:
...
+Prose accompanying an [=interface=] with an [=asynchronously iterable declaration=] must define a
+<dfn id="dfn-get-the-next-iteration-result">get the next iteration result</dfn> algorithm.
+This algorithm receives a <b>[=this=]</b> value, which is an instance of the [=interface=] that it
+is defined for, and the <dfn export>current state</dfn>.
+It must return a {{Promise}} that either resolves with undefined – to signal the end of the
+iteration – or a tuple with three elements:
+
+1.  a value of the first type given in the declaration;
+1.  a value of the second type given in the declaration;
+1.  an opaque value that is passed back to the next invocation of the algorithm,
```
domenic: Hmm. This wasn't what was discussed in https://github.com/heycam/webidl/issues/580. I prefer the formulation over there, if you think it's possible to spec.

index.bs
```diff
@@ -4091,6 +4104,138 @@ The following extended attributes are applicable to [=iterable declarations=]:
...
+The following extended attributes are applicable to [=asynchronously iterable declarations=]:
+[{{Exposed}}],
+[{{SecureContext}}].
+
+Issue: these [=extended attributes=] are not currently taken into account.
+When they are, the effect will be as you would expect.
+
+<pre class="grammar" id="prod-AsyncIterable">
+    AsyncIterable :
+        "async_iterable" "&lt;" TypeWithExtendedAttributes "," TypeWithExtendedAttributes "&gt;" ";"
```
domenic: For https://wicg.github.io/native-file-system/#filesystemdirectoryhandle I believe we'll want a version that only does values.

For https://streams.spec.whatwg.org/#rs-asynciterator / https://streams.spec.whatwg.org/#rs-get-iterator we'll want a version that only does values and also can be aliased to a function that takes arguments.

For the values-only version I'd stay conservative and not add keys/values/entries for now as this is uncharted territory. (Even though sync iterables do that even for values-only cases like `Set`.)

All this can be done as a follow-up, but it may be helpful to keep it in mind while designing.

index.bs
Ms2ger: With "yield"? I added a commit to https://github.com/WICG/kv-storage/pull/68 to see what that looks like. It seems kinda weird to be throwing and yielding out of promise handlers and event listeners, but if you think it makes sense, I can try to draft prose for IDL to make it well-defined.

domenic: I'll need to think on this harder. It does seem like we have two rather different ways spec authors could do this, one of which makes sense when promises are involved, one of which makes sense when background threads are involved. Thanks for the added commit to give something concrete for us to look at.

My guess, before thinking hard, is that we'll either:

- Allow both styles, at the spec author's option; or
- Start with what kv-storage wants, then expand to what file system wants, etc., figuring it out as we go.

domenic: Let's stick with what you've got. Mixing yield/throw and promises is too weird, indeed.

domenic: Can we omit the opaque value, and have people just store any relevant state on their `this`?

index.bs
```diff
@@ -11577,6 +11725,45 @@ and the String "<code>values</code>"
...
+If the [=interface=] has an [=asynchronously iterable declaration=], then a property must exist
+whose name is the {{@@asyncIterator}} symbol, with attributes
+{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
+and whose value is a [=function object=].
+
+The location of the property is determined as follows:
+
+*   If the interface was declared with the [{{Global}}] [=extended attribute=],
+    then the property exists on the single object that [=implements=] the interface.
+*   Otherwise, the property exists solely on the interface’s [=interface prototype object=].
```
domenic: Kind of sad to add to the declaratively-added properties list, but oh well, this is probably easier for now.

index.bs
domenic: Streams will also need some side effects on acquiring the iterator (i.e. calling the function)

index.bs
```diff
@@ -11852,34 +12105,131 @@ must be {{%IteratorPrototype%}}.
...
+
+    An [=asynchronous iterator prototype object=] must have a <code class="idl">next</code> data
+    property with attributes
+    { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
+    and whose value is a [=built-in function object=] that behaves as follows:
+
+    1.  Let |interface| be the [=interface=] for which the
+        [=asynchronous iterator prototype object=] exists.
+
+    1.  Let |thisValidationPromiseCapability| be [=!=] [$NewPromiseCapability$]({{%Promise%}}).
```
domenic: If we had a PromiseReject abstract op, you could just create and return a new promise inline each time. I guess we'd wait for #490 for that.

index.bs
```diff
@@ -11852,34 +12105,131 @@ must be {{%IteratorPrototype%}}.
...
 
+<h5 id="es-default-asynchronous-iterator-object">Default asynchronous iterator objects</h5>
+
+A <dfn id="dfn-default-asynchronous-iterator-object" export>default asynchronous iterator object</dfn>
+for a given [=interface=], target and iteration kind is an object whose \[[Prototype]]
+[=internal slot=] is the [=asynchronous iterator prototype object=] for the [=interface=].
+
+A [=default asynchronous iterator object=] has internal values:
+
+*   its <em>target</em>, which is an object whose values are to be iterated,
```
domenic: Can we made these dfns, and cross-link to them?

index.bs
```diff
@@ -11852,34 +12105,131 @@ must be {{%IteratorPrototype%}}.
...
+
+    1.  Otherwise:
+        1.  Run |nextSteps| and set |object|'s ongoing promise to the result.
+
+    1.  Return |object|'s ongoing promise.
+</div>
+
+Issue: The [=class string=] of an [=asynchronous iterator prototype object=] for a given
+[=interface=] is the result of concatenating the [=identifier=] of the [=interface=] and the string
+"<code> Iterator</code>".
```
domenic: I guess `" AsyncIterator"`?

index.bs
domenic: Oh, I see we cannot, because it needs to be *per iterator instance*, not *per interface instance*. Can you add an explanation to that effect?

bzbarsky:
This could probably use review from other implementors before it merges....

index.bs
Ms2ger: Looking at Streams, it seems like we'll need to allow specs to stick data on the iterator object anyway, so maybe we should drop this as well.

domenic:
@bzbarsky would you have time to review this soon? I am planning to land https://github.com/WICG/kv-storage/pull/68 with a reference to this pull request, but it would be even better if I could reference the base spec.

bzbarsky:
This week is a Mozilla all-hands, so my time is very unpredictable through Friday.  I can make this a priority on Monday, though.  Is that good enough?

domenic:
That would be great, thank you!

ricea:
I have some concerns about how we will apply this to the Streams standard. The `iterable` definition has this text:

> If a single type parameter is given, then the interface has a value iterator and provides values of the specified type

but `async_iterable` lacks a similar provision. Since a ReadableStream has no keys, this makes application of the standard text difficult. There's nothing meaningful that a `keys()` method could iterate over, and I don't know how `values()` would be different from `entries()`.

Ms2ger:
I've got a WIP for value-only iterators in https://github.com/heycam/webidl/compare/async-iterator-values?expand=1.

index.bs
```diff
@@ -11766,34 +12032,146 @@ must be {{%IteratorPrototype%}}.
...
+A [=default asynchronous iterator object=] has internal values:
+
+*   its <dfn for="default asynchronous iterator object">target</dfn>, which is an object whose
+    values are to be iterated,
+*   its <dfn for="default asynchronous iterator object">kind</dfn>, which is the iteration kind,
+*   its <dfn for="default asynchronous iterator object">ongoing promise</dfn>, which is a
+    {{Promise}} or undefined,
+*   its <dfn for="default asynchronous iterator object">state</dfn>, which is an opaque value used
+    to store the position of the iterator by the algorithm to [=get the next iteration result=], or
+    null.
```
bzbarsky: Why null rather than undefined?

index.bs
```diff
@@ -11766,34 +12032,146 @@ must be {{%IteratorPrototype%}}.
...
+        1.  Issue: [=Realm=] check?
+        1.  Let |error| be a new {{ECMAScript/TypeError}}.
+        1.  Perform [=!=] [$Call$](|thisValidationPromiseCapability|.\[[Reject]],
+            <emu-val>undefined</emu-val>, « |error| »).
+        1.  Return |thisValidationPromiseCapability|.\[[Promise]].
+
+    1.  Let |nextSteps| be the following steps:
+        1.  Let |nextPromiseCapability| be [=!=] [$NewPromiseCapability$]({{%Promise%}}).
+        1.  Let |oldState| be |object|'s [=default asynchronous iterator object/state=].
+        1.  If |oldState| is null, then:
```
bzbarsky: So the state management here is a bit wonky.  Nothing before this point said anything about the state getting _set_ to null.  How would it get there?  And the initial state bit is not really set anywhere, and explicitly labeled as "not given" at the start.  I guess the intent is that "not given" is not the same thing as "null", but it's not very clear.  It would be good to be very clear about the value space of the state (IDL values, ES values, some other set of values) and when it's set to which of those.  It sounds like the intent is that it's in some other value space, which contains all possible opaque values, and two magical values "start of iteration" ("not given") and "stop iteration" ("null")?

index.bs
Ms2ger: Added the side effects, but I'd like to defer full Streams support to a followup.

Ms2ger:
Thanks for the review; I think I covered most of the issues.

index.bs
```diff
@@ -4116,6 +4129,142 @@ The following extended attributes are applicable to [=iterable declarations=]:
...
+The following extended attributes are applicable to [=asynchronously iterable declarations=]:
+[{{Exposed}}],
+[{{SecureContext}}].
+
+Issue: these [=extended attributes=] are not currently taken into account.
+When they are, the effect will be as you would expect.
+
+<pre class="grammar" id="prod-AsyncIterable">
+    AsyncIterable :
+        "async_iterable" "&lt;" TypeWithExtendedAttributes "," TypeWithExtendedAttributes "&gt;" ";"
```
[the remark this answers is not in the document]
Ms2ger: No strong feelings from my side on this.

index.bs
```diff
@@ -4116,6 +4129,142 @@ The following extended attributes are applicable to [=iterable declarations=]:
...
+This algorithm receives a <b>[=this=]</b> value, which is an instance of the [=interface=] that it
+is defined for, and the <dfn export>current state</dfn>.
+It must return a {{Promise}} that either resolves with undefined – to signal the end of the
+iteration – or a tuple with three elements:
+
+1.  a value of the first type given in the declaration;
+1.  a value of the second type given in the declaration;
+1.  an opaque value that is passed back to the next invocation of the algorithm,
+
+The prose may also define <dfn>asynchronous iterator initialization steps</dfn> for the
```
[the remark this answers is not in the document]
Ms2ger: This is supposed to be a step towards handling https://streams.spec.whatwg.org/#rs-get-iterator ; not sure if it's necessary to restrict anything here.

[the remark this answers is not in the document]
bzbarsky: Are we OK with the steps here calling into arbitrary script, say?  (We might be; I haven't audited very carefully.)

bzbarsky:
Generally looks good, thanks!  I think I would still prefer `async iterator` without the underscore.

ricea:
> Generally looks good, thanks! I think I would still prefer `async iterator` without the underscore.

@yuki3 Would `async iterator` as two tokens cause problems for Blink's IDL parser?

yuki3:
I think that it should be just fine.  We already have "partial dictionary", "partial interface", "callback interface", etc.  I don't think that a space between "async" and "iterator" will be a problem.

inexorabletash:
What's the support in bikeshed and WPT's respective IDL parsers?

marcoscaceres:
For WTP, I don't think we support this yet in WebIDL2.js, right @saschanaz? Would be trivial to add tho.

saschanaz:
@marcoscaceres Right, we don't. I can cook a PR soon.

saschanaz:
Just to make sure: `async_iterable` or `async iterable`?

marcoscaceres:
"async iterable" by looking at the example.

marcoscaceres:
Ok, @saschanaz added support (https://github.com/w3c/webidl2.js/pull/365). We will wait for this to merge. Probably want to bugs filed on browser engines too.

PR #726, opened 2019-05-10
Reduce the restrictions on members whose names match auto-generated properties

### LINKED ISSUE #49: Improve restrictions on attributes/operations when maplike/setlike are used
Two changes:

1)  Use of maplike/setlike should not affect static attributes/operations (but should affect constants, as it does now, since those go on the proto too).

2)  The mutator methods (clear, add, set, delete) should not be allowed to shadow a method on an inherited interface.  But having them on the interface itself or its consequential interfaces should be OK (as now) and just inhibit autogeneration of the method.

Ms2ger:
Fixes #49.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3981,12 +3981,19 @@ the iterator objects returned by <code class="idl">entries</code>,
 <code class="idl">keys</code>, <code class="idl">values</code>, and {{@@iterator}} are
 actual [=array iterator objects=].
 
-Interfaces with iterable declarations must not
-have any [=interface members=]
-named "<code>entries</code>", "<code>forEach</code>",
-"<code>keys</code>", or "<code>values</code>",
-or have any [=inherited interfaces=]
-that have [=members=] with these names.
+Interfaces with iterable declarations must not have any [=regular attributes=] or [=constants=]
+named
+"<code>entries</code>",
+"<code>forEach</code>",
+"<code>keys</code>", or
+"<code>values</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
+[=constants=] with these names.
+
+Note: Operations with these names are allowed on interfaces with iterable declarations and will
+prevent the default implementation of these methods being added to the
+[=interface prototype object=] in the ECMAScript language binding.
+This allows the default behavior of these operations to be overridden.
 
 <div class="example">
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3981,12 +3981,19 @@ the iterator objects returned by <code class="idl">entries</code>,
...
+Interfaces with iterable declarations must not have any [=regular attributes=] or [=constants=]
+named
+"<code>entries</code>",
+"<code>forEach</code>",
+"<code>keys</code>", or
+"<code>values</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
+[=constants=] with these names.
+
+Note: Operations with these names are allowed on interfaces with iterable declarations and will
```
bzbarsky: Why this change?  Regular operations with these names didn't use to be allowed, and they still shouldn't be, imo.

index.bs
```diff
@@ -4132,23 +4139,30 @@ keyword is used, this includes <code class="idl">entries</code>,
...
+[=constants=] with that name.
+
+Read–write maplike interfaces must not have any [=regular attributes=] or [=constants=] named
+"<code>clear</code>",
+"<code>delete</code>", or
+"<code>set</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
+[=constants=] with these names.
+
+Note: Operations with these names are allowed on maplike
```
bzbarsky: Again, looks wrong.  The existing "can be overridden" thing was for methods that might need to be hooked because they modify the map and might need to update other internal data structures.  But we don't really want hooking for random read operations, right?

index.bs
```diff
@@ -4226,22 +4240,28 @@ keyword is used, this includes <code class="idl">entries</code>,
...
+[=constants=] with that name.
+
+Read–write setlike interfaces must not have any [=regular attributes=] or [=constants=] named
+"<code>add</code>",
+"<code>clear</code>", or
+"<code>delete</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
+[=constants=] with these names.
+
+Note: Operations with these names are allowed on setlike
```
bzbarsky: Same here.

index.bs
```diff
@@ -4132,23 +4132,26 @@ keyword is used, this includes <code class="idl">entries</code>,
...
-Read–write maplike interfaces must not
-have any [=attributes=]
-or [=constants=] named
-"<code>clear</code>", "<code>delete</code>",
-or "<code>set</code>", or have any [=inherited interfaces=]
-that have [=attributes=] or [=constants=] with these names.
-
-Note: Operations named "<code>clear</code>", "<code>delete</code>",
-or "<code>set</code>" are allowed on read–write maplike
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
```
bzbarsky: This is repeating things twice about inherited interfaces, right?

index.bs
```diff
@@ -11579,7 +11583,8 @@ if the interface has a [=setlike declaration=].
 
 <h5 id="es-forEach">forEach</h5>
 
-If the [=interface=] has any of the following:
+If the [=interface=] does not declare a [=regular operation=] with [=identifier=]
+<code class="idl">forEach</code> and has any of the following:
```
bzbarsky: This change is not needed, since things named "forEach" are disallowed in that case.  Similar for the other methods...

index.bs
```diff
@@ -4226,22 +4227,23 @@ keyword is used, this includes <code class="idl">entries</code>,
...
-have any [=attributes=]
-or [=constants=] named
-"<code>add</code>", "<code>clear</code>",
-or "<code>delete</code>", or have any [=inherited interfaces=]
-that have [=attributes=] or [=constants=] with these names.
-
-Note: Operations named "<code>add</code>", "<code>clear</code>",
-or "<code>delete</code>" are allowed on read–write setlike
+Setlike interfaces and their [=inherited interfaces=] must not have any [=regular operations=],
+[=regular attributes=] or [=constants=] named
```
Ms2ger: This forbids
```webidl
interface A { void entries(); };
interface B : A { readonly setlike<long>; };
```
but not
```webidl
interface A { readonly setlike<long>; };
interface B : A { void entries(); };
```

Should it?

bzbarsky: That is a really good question....  Seems like disallowing it until there's a use case that might want it, and then deciding how it should work, is reasonable...

Ms2ger: Done.

Ms2ger:
@bzbarsky I think this is ready for another review.

bzbarsky:
Sorry for the lag; work week interfered... Will look sometime this week.

index.bs
```diff
@@ -4132,23 +4133,25 @@ keyword is used, this includes <code class="idl">entries</code>,
...
-Read–write maplike interfaces must not
-have any [=attributes=]
-or [=constants=] named
-"<code>clear</code>", "<code>delete</code>",
-or "<code>set</code>", or have any [=inherited interfaces=]
-that have [=attributes=] or [=constants=] with these names.
-
-Note: Operations named "<code>clear</code>", "<code>delete</code>",
-or "<code>set</code>" are allowed on read–write maplike
+Maplike interfaces, their [=inherited interfaces=], and any [=interfaces=] that
```
bzbarsky: Is it worth factoring out the "interface, its inherited interfaces, and all interfaces that inherit from it" concept so it can be reused, instead of copy/pasting it?

index.bs
```diff
@@ -4132,23 +4133,25 @@ keyword is used, this includes <code class="idl">entries</code>,
...
+"<code>keys</code>",
+"<code>size</code>", or
+"<code>values</code>".
+
+Read–write maplike interfaces must not have any [=regular attributes=] or [=constants=] named
+"<code>clear</code>",
+"<code>delete</code>", or
+"<code>set</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
+[=constants=] with these names.
```
bzbarsky: Can interfaces inheriting from the read-write maplike have these things?  Seems like they shouldn't...

index.bs
```diff
@@ -4226,22 +4229,24 @@ keyword is used, this includes <code class="idl">entries</code>,
...
+"<code>has</code>",
+"<code>keys</code>",
+"<code>size</code>", or
+"<code>values</code>".
+
+Read–write setlike interfaces must not have any [=regular attributes=] or [=constants=] named
+"<code>add</code>",
+"<code>clear</code>", or
+"<code>delete</code>",
+or have any [=inherited interfaces=] that have [=regular operations=], [=regular attributes=] or
```
bzbarsky: Again, what about interfaces that inherit from the read-write setlike?

PR #728, opened 2019-05-17
Remove the Error IDL type

Ms2ger:
***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/728.html" title="Last updated on May 20, 2019, 7:48 AM UTC (fc82701)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/728/41e0d1e...fc82701.html" title="Last updated on May 20, 2019, 7:48 AM UTC (fc82701)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4766,8 +4766,7 @@ listed in the [=error names table=] below.
 The table also indicates the {{DOMException}}'s integer code for that error name,
 if it has one.
 
-There is one type that can be used to refer to exception objects:
-{{DOMException}}.
+Note: As {{DOMException}} is an [=interface type=], it can be used as a type in IDL.
 This allows for example an [=operation=]
 to be declared to have a {{DOMException}}
 [=return type=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4767,13 +4766,11 @@ listed in the [=error names table=] below.
 The table also indicates the {{DOMException}}'s integer code for that error name,
 if it has one.
 
-There are two types that can be used to refer to exception objects:
-{{DOMException}} and {{Error!!interface}},
-where the latter encompasses [=simple exceptions=] and {{DOMException}}.
+There is one type that can be used to refer to exception objects:
+{{DOMException}}.
```
domenic: I don't think this is a type any more? Or maybe it's just an interface type. In which case I think this paragraph is confusing; everything defined as an interface is an interface type. A paragraph like this makes me think I should find special conversion rules for the type.

PR #731, opened 2019-05-29
The GetIterator operation in ECMA262 now takes three arguments

bathos:
The signature changed from (obj, method) to (obj, hint, method) when @@asyncIterator was introduced. Because the method is already known to be defined at this point, I don’t believe the hint argument will actually be used, but the argument still seems to be needed to put the method arg in the right position.

I was unsure what formatting "sync" should get here. It’s a constant symbolic value rather than a variable or a string. I didn’t see any pre-existing examples in the IDL spec, so it may need additional markup.

***
<a href="https://pr-preview.s3.amazonaws.com/bathos/webidl/pull/731.html" title="Last updated on May 29, 2019, 10:04 AM UTC (14536c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/731/4ef3509...bathos:14536c1.html" title="Last updated on May 29, 2019, 10:04 AM UTC (14536c1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7635,7 +7635,7 @@ ECMAScript Array values.
     given an iterable |iterable| and an iterator getter |method|,
     perform the following steps:
 
-    1.  Let |iter| be [=?=] <a abstract-op>GetIterator</a>(|iterable|, |method|).
+    1.  Let |iter| be [=?=] <a abstract-op>GetIterator</a>(|iterable|, sync, |method|).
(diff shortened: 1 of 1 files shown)


annevk:
Thanks! We already use various `emu-` elements. I guess it would be appropriate for this PR to add `emu-const` and the corresponding CSS.

PR #746, opened 2019-06-27
Disallow Promise types in unions at the syntax level.

### LINKED ISSUE #740: NonAnyType vs SingleType as home of PromiseType ε alternative
- The `NonAnyType` production is only used within [`UnionMemberType`][1] and [`SingleType`][2].
- `NonAnyType` [includes the alternative `PromiseType ε`][3].
- A Union type comprises [‘two or more other types’][4].
- Promise types are [not distinguishable from any other type][5].
- [Each pair of flattened member types in a union type, T and U, must be distinguishable.][6]

I may be missing or misunderstanding something (perhaps in the last item?), but it seems like, if all of the above statements are true, that a Promise type is never a valid union member. If that’s true, then the `PromiseType ε` alternative should probably be moved from `NonAnyType` to `SingleType` so that this constraint gets enforced syntactically.

[1]: https://heycam.github.io/webidl/#prod-UnionMemberType
[2]: https://heycam.github.io/webidl/#index-prod-SingleType
[3]: https://heycam.github.io/webidl/#prod-NonAnyType
[4]: https://heycam.github.io/webidl/#dfn-union-type
[5]: https://heycam.github.io/webidl/#example-distinguishability-promises
[6]: https://heycam.github.io/webidl/#ref-for-dfn-flattened-union-member-types%E2%91%A8

Ms2ger:
Fixes #740.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/746.html" title="Last updated on Jun 28, 2019, 9:34 AM UTC (b5551b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/746/c0d927a...b5551b7.html" title="Last updated on Jun 28, 2019, 9:34 AM UTC (b5551b7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5386,7 +5386,7 @@ type.
     SingleType :
         DistinguishableType
         "any"
-        PromiseType ε
+        PromiseType
 </pre>
 
 <pre class="grammar" id="prod-UnionType">
--- a/index.bs
+++ b/index.bs
@@ -5384,8 +5384,9 @@ type.
 
 <pre class="grammar" id="prod-SingleType">
     SingleType :
-        NonAnyType
+        DistinguishableType
         "any"
+        PromiseType ε
 </pre>
 
 <pre class="grammar" id="prod-UnionType">
@@ -5395,7 +5396,7 @@ type.
 
 <pre class="grammar" id="prod-UnionMemberType">
     UnionMemberType :
-        ExtendedAttributeList NonAnyType
+        ExtendedAttributeList DistinguishableType
         UnionType Null
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5384,8 +5384,9 @@ type.
 
 <pre class="grammar" id="prod-SingleType">
     SingleType :
-        NonAnyType
+        DistinguishableType
         "any"
+        PromiseType ε
```
bzbarsky: Do we need the "ε" bit?

Ms2ger: I suppose not (but could be proved wrong by someone whose grammar knowledge is more recent than mine)

bathos: I found the ε confusing when working with the grammar. Was the original intention just to emphasize that Null deliberately isn’t included in this alternative?

PR #747, opened 2019-06-27
Explicitly include annotated types in the various type categories.

### LINKED ISSUE #670: Is `long?` an integer type?
Per spec, looks like no.  That means that `[Clamp] long?` should be an error, right?  Is that the intent?

@domenic

Ms2ger:
Fixes #670.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5313,22 +5313,28 @@ The following types are known as <dfn id="dfn-integer-type" export>integer types
 {{unsigned short}},
 {{long}},
 {{unsigned long}},
-{{long long}} and
-{{unsigned long long}}.
+{{long long}},
+{{unsigned long long}},
+and any [=annotated types=] whose [=annotated types/inner type=] is an [=integer type=].
 
 The following types are known as <dfn id="dfn-numeric-type" export>numeric types</dfn>:
 the [=integer types=],
 {{float}},
 {{unrestricted float}},
-{{double}} and
-{{unrestricted double}}.
+{{double}},
+{{unrestricted double}},
+and any [=annotated types=] whose [=annotated types/inner type=] is a [=numeric type=].
 
 The <dfn id="dfn-primitive-type" export>primitive types</dfn> are
-{{boolean}} and the [=numeric types=].
+{{boolean}},
+the [=numeric types=],
+and any [=annotated types=] whose [=annotated types/inner type=] is a [=primitive type=].
 
 The <dfn id="dfn-string-type" export>string types</dfn> are
 {{DOMString}}, all [=enumeration types=],
-{{ByteString}} and {{USVString}}.
+{{ByteString}},
+{{USVString}},
+and any [=annotated types=] whose [=annotated types/inner type=] is a [=string type=].
 
 The <dfn id="dfn-typed-array-type" export>typed array types</dfn> are
 {{Int8Array}},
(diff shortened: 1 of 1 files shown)


bzbarsky:
@ms2ger I don't think this fixes #670.  This PR makes `[Clamp] long` be an integer type, but the question in #670 is around whether `long?` is an integer type, no?

Ms2ger:
You're correct, I got confused. I think this is better.

Ms2ger:
@bzbarsky any more thoughts?

index.bs
```diff
@@ -5313,22 +5313,32 @@ The following types are known as <dfn id="dfn-integer-type" export>integer types
...
+any [=nullable types=] whose [=nullable types/inner type=] is an [=integer type=],
+and any [=annotated types=] whose [=annotated types/inner type=] is an [=integer type=].
 
 The following types are known as <dfn id="dfn-numeric-type" export>numeric types</dfn>:
 the [=integer types=],
 {{float}},
 {{unrestricted float}},
-{{double}} and
-{{unrestricted double}}.
+{{double}},
```
bzbarsky: So... With this change there are a few things that are a bit odd:

1) https://heycam.github.io/webidl/#es-overloads step 12.12 no longer needs to worry about "a nullable numeric type" or "an annotated type whose inner type is one of the above types", since those will be included under "numeric type", no?  Same thing for step 12.14.  And 12.13 given the "string type" changes below.

2) There are various other places in the spec that mention "numeric types" but they all already handle unwrapping nullable and annotated types.  Maybe this is OK; it's just a little confusing to consider the table in the distinguishability section, for example, when "numeric types" could in theory include `long?` but in practice can't because we unwrapped that already...

index.bs
```diff
@@ -5313,22 +5313,32 @@ The following types are known as <dfn id="dfn-integer-type" export>integer types
...
+{{unrestricted double}},
+any [=nullable types=] whose [=nullable types/inner type=] is a [=numeric type=],
+and any [=annotated types=] whose [=annotated types/inner type=] is a [=numeric type=].
 
 The <dfn id="dfn-primitive-type" export>primitive types</dfn> are
-{{boolean}} and the [=numeric types=].
+{{boolean}},
```
bzbarsky: Again, this leads to the slightly odd situation that the `PrimitiveType` production no longer matches the set of "primitive types".  Further, the prose around ConstType says:

> The type of a constant (matching ConstType) must not be any type other than a primitive type.

with that last it linking to this definition but it would now actually means a stronger restriction than what that link provides.

index.bs
```diff
@@ -5338,18 +5348,24 @@ The <dfn id="dfn-typed-array-type" export>typed array types</dfn> are
 {{Uint16Array}},
 {{Uint32Array}},
 {{Uint8ClampedArray}},
-{{Float32Array}} and
-{{Float64Array}}.
+{{Float32Array}},
+{{Float64Array}},
+any [=nullable types=] whose [=nullable types/inner type=] is a [=typed array type=],
```
bzbarsky: But here we couldn't simplify overload resolution step 12.7 because for the annotated/nullable types the name would not match...

index.bs
```diff
@@ -5338,18 +5348,24 @@ The <dfn id="dfn-typed-array-type" export>typed array types</dfn> are
...
+and any [=annotated types=] whose [=annotated types/inner type=] is a [=typed array type=].
 
-The <dfn id="dfn-buffer-source-type" export>buffer source types</dfn>
-are {{ArrayBuffer}},
+The <dfn id="dfn-buffer-source-type" export>buffer source types</dfn> are
```
bzbarsky: With this part changed, the claim that "buffer source types are simply references to objects" stops being true, no?

And "Values of the IDL buffer source types are represented by objects of the corresponding ECMAScript class" also stops beign true.

Issue #602, opened 2018-12-18
Consider requiring optional dictionary defaulting explicitly


annevk:
Instead of
```webidl
foo(optional dictionaryType arg);
```
perhaps we should require writing
```webidl
foo(optional dictionaryType arg = {});
```
for consistency with other optional arguments and for clarity for the reader. At least anecdotally this seems to be one of the most confusing things about IDL that's not a legacy feature.

bzbarsky:
Or `optional dictionaryType arg = null`, though either one is probably ok... We just need to introduce new syntax to do `= {}`, last I checked, while `= null` should just work.

annevk:
I suspect that `= null` might have the same problem and could even lead to prose that branches on `null`.

bzbarsky:
OK.  As long as people are clear on `= {}` meaning "empty dictionary", not "dictionary initialized from a `{}` JS object literal", since those are not the same thing...

bzbarsky:
@annevk One question: would we still require the `= {}` on trailing (or all?) dictionary args to be present for the IDL to be valid?  That seems generally desirable to me...

annevk:
Yeah, I was thinking it would be a syntax error without it. I suspect we should require it always, to make it clear it's what happens when you pass undefined and you cannot branch on "_dict_ not given" or some such.

Issue #585, opened 2018-09-27
Is it true that dictionaries can't have a default value specified?


bzbarsky:
There's informative text in https://heycam.github.io/webidl/#idl-operations that says:

> Dictionary types cannot have a default value specified explicitly

however it seems to me that `null` would be a valid default value for a dictionary (meaning empty dictionary, once you go through the es-to-dictionary steps).

@tobie @domenic @marcoscaceres @annevk

annevk:
I think we decided not to have a default value (e.g., `{}`) since it always has a default value.

bzbarsky:
For example, `DOMQuadInit` at https://drafts.fxtf.org/geometry/#dictdef-domquadinit sure seems to assume that its `DOMPointInit` members get default values.  It has no provisions for those members being missing.

@zcorpan

zcorpan:
Good catch. Yeah, I think I assumed that they would act like `{}`. 

The other option is to default with prose in the algorithm, though that seems less nice.

PR #750, opened 2019-06-28
Define a way to specify a default value for dictionaries (the literal "{}") and require it to be specified for the dictionary arguments that are required to be optional.

bzbarsky:
Fixes https://github.com/heycam/webidl/issues/585
Fixes https://github.com/heycam/webidl/issues/602

@tabatkins @annevk @domenic @ms2ger thoughts?

***
<a href="https://pr-preview.s3.amazonaws.com/bzbarsky/webidl/pull/750.html" title="Last updated on Jun 28, 2019, 11:37 PM UTC (a740e17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/750/a029460...bzbarsky:a740e17.html" title="Last updated on Jun 28, 2019, 11:37 PM UTC (a740e17)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1616,8 +1616,8 @@ or one of the three special floating point constant values
 (<emu-t>-Infinity</emu-t>, <emu-t>Infinity</emu-t> and <emu-t>NaN</emu-t>).
 
 Note: These values – in addition to strings and the empty sequence – can also be used to specify the
-[=dictionary member/default value|default value of a dictionary member=] or [=optional argument/default value|of an optional argument=].  Note that strings and the
-empty sequence <code>[]</code> cannot be used as the value of a
+[=dictionary member/default value|default value of a dictionary member=] or [=optional argument/default value|of an optional argument=].  Note that strings, the
+empty sequence <emu-t>[]</emu-t>, and the default dictionary <emu-t>{}</emu-t> cannot be used as the value of a
 [=constant=].
 
 The value of the boolean literal tokens <emu-t>true</emu-t> and
@@ -2242,12 +2242,12 @@ corresponding argument omitted.
     conversion of <emu-val>undefined</emu-val> to be used (i.e., <emu-val>false</emu-val>).
 </p>
 
-If the type of an argument is a [=dictionary type=] or a [=union type=] that has a dictionary as one
+If the type of an argument is a [=dictionary type=] or a [=union type=] that has
+a [=dictionary type=] as one
 of its [=flattened member types=], and that dictionary type and its ancestors have no
 [=required dictionary member|required members=], and the argument is either the final argument or is
-followed only by [=optional arguments=], then the argument must be specified as optional. Such
-arguments are always considered to have a [=optional argument/default value=] of an empty
-dictionary, unless otherwise specified.
+followed only by [=optional arguments=], then the argument must be specified as
+optional and have a default value provided.
 
 <div class="note">
 
@@ -2255,11 +2255,10 @@ dictionary, unless otherwise specified.
     empty dictionary value when they wish only to use the dictionary’s
     default values.
 
-    Dictionary types cannot have a default value specified explicitly, so the
-    “unless otherwise specified” clause above can only be invoked for
-    a [=union type=] that has a
-    dictionary type as one of its [=flattened member types=].
-
+    Usually the default value provided will be <emu-t>{}</emu-t>, but in the
+    case of a [=union type=] that has a dictionary type as one of its
+    [=flattened member types=] a default value could be provided that
+    initializes some other member of the union.
 </div>
 
 When a boolean literal token (<emu-t>true</emu-t> or <emu-t>false</emu-t>),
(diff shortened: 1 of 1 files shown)


bzbarsky:
Note that this will require a ton of spec updates, by the way, to add all the now-required `= {}`.

index.bs
```diff
@@ -2242,24 +2242,23 @@ corresponding argument omitted.
     conversion of <emu-val>undefined</emu-val> to be used (i.e., <emu-val>false</emu-val>).
```
Ms2ger: Consider adding the new token to

> Note: These values – in addition to strings and the empty sequence – can also be used to specify the default value of a dictionary member or of an optional argument. Note that strings and the empty sequence [] cannot be used as the value of a constant.

annevk:
Created PRs for https://github.com/whatwg/dom/pull/771 and https://github.com/whatwg/html/pull/4745. This will require Bikeshed changes too, tracked at https://github.com/tabatkins/bikeshed/issues/1491 and https://github.com/plinss/widlparser/issues/39 for the IDL parser backend.

bzbarsky:
> My first thought was to use = null

Mine was too, and is in fact what Gecko has long implemented, but @annevk convinced me that this is more what people are likely to expect.

index.bs
```diff
@@ -2337,6 +2343,31 @@ that has a [=sequence type=] in its [=flattened member types=].
...
+
+    The following [=IDL fragment=]
+    defines an [=interface=]
+    with an operation that takes a dictionary argument:
+
+    <!-- Should be `highlight="webidl"`, but that gives bikeshed conniptions
+         because it does not know about the syntax we are introducing -->
+    <pre>
+        dictionary LookupOptions {
+          boolean caseSensitive = true;
```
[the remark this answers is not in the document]
bzbarsky: OK.  Used `false` as the default value there.

index.bs
bzbarsky: Done.

bzbarsky:
@annevk thank you for filing all those!

bzbarsky:
https://bugzilla.mozilla.org/show_bug.cgi?id=1562680 tracks Gecko implementing this.

bzbarsky:
We probably need to get a bunch of other spec issues filed...  Can we hope the bikeshed changes will drive that, or should we go through the list of changes I needed to make to IDLs in Gecko and try to file issues based on that?

annevk:
If you can dump a file list in a new issue for triage that'd be good I think. The other problem with updating IDL in the specifications is that we also need testharness support as it automatically gets the IDL from the specifications.

bzbarsky:
@annevk Do you want a list of files or a list of interface names?  They are generally 1-1 but not always (though a given file should always come from a single spec).

Ms2ger:
webidl2.js is fixed in https://github.com/w3c/webidl2.js/pull/338, but the copy in wpt isn't updated yet.

marcoscaceres:
Just recording that the relevant bug on the Blink side is: https://bugs.chromium.org/p/chromium/issues/detail?id=948139

Do we know if the WebKit folks are aware of this change?

bzbarsky:
> Do we know if the WebKit folks are aware of this change?

Argh.  I thought I had cced @rniwa above!

@rniwa, how do you feel about this change?  Please cc as needed if someone else should be looking at this on the WebKit side.

annevk:
@foolip @lukebjerring could either of you confirm whether web-platform-tests can deal with this change? I think that's the main thing holding me back from landing this in various specifications at this point.

@bzbarsky note that due to the IDL parser used for specifications not having universal knowledge (and no parameters for it either as of yet), it cannot require `= {}` so specifications will continue to have errors with this until that's fully sorted.

lukebjerring:
Because we go via reffy, there was initially an issue (https://github.com/tidoust/reffy-reports/issues/34) with the syntax. I believe it is now resolved.

bzbarsky:
So the issue is that the IDL parser used in specs doesn't actually know which type the identifier resolves to?

If wpt (as I understand it) pulls IDL from specs directly, then as long as wpt enforces this we will get test failures for non-updated specs, no?  And then we can fix those as needed....

lukebjerring:
wpt contains (in the /interfaces/ dir) scrapes from the specs directly, correct. So, yes, that sounds right - if the idlharness changes to enforce it, the PR making that change would produce failures for currently non-updated specs.

annevk:
@bzbarsky yeah, see https://github.com/plinss/widlparser/issues/39#issuecomment-515273312. This is true for Bikeshed-specifications for now, dunno about ReSpec. On the WPT side, supposedly it uses https://github.com/w3c/webidl2.js/pull/338/files which does not seem to enforce this. @lukebjerring was idlharness updated already to use the newer webidl2.js? That's the other important part.

Ms2ger:
WPT currently uses webidl2.js v22.1.0, which allows the new syntax but doesn't require it. I believe upstream webidl2.js is updated to require the new syntax, so we'd need to update the IDL in wpt before updating. @saschanaz probably knows for sure.

saschanaz:
There is a relevant discussion on WPT side: https://github.com/web-platform-tests/wpt/issues/16267

The current intent is to run validator as a subtest.

kainino0x:
In both webidl2.js and bikeshed right now, this code is accepted:

```webidl
dictionary A {
  required long x;
};

dictionary B {
  A a = {};
};
```

but it seems like this ought to be invalid since `{}` is not a valid `A`. `= null` seems like it should equally not work.

Semi-related, I'm not sure I know what `dictionary C { A a; }` means. Does this allow `a` to be "A or undefined"? It seems that way in Chrome's bindings generator right now. We want it to work this way if possible in WebGPU.

https://github.com/gpuweb/gpuweb/issues/420

bzbarsky:
> but it seems like this ought to be invalid since {} is not a valid A

That would be an option, yes.

> = null seems like it should equally not work.

Per spec that does not work, right.

> Does this allow a to be "A or undefined"?

It allows `a` to be not present (that is the `Get()` on the JS object returns `undefined`), but if it's present (any other value is returned) then it only allows values that can be converted to `A` without failing.

Chrome's bindings generator has all sorts of bugs around dictionaries, so I wouldn't base too much on what it does, but the desire for the `dictionary C { A a; }` case where `A` has required members and allowing `{}` to be passed for a `C` instance was present in a number of specs, so the IDL spec explicitly allows that.

kainino0x:
That's really helpful, thanks for the info!

> > = null seems like it should equally not work.
> 
> Per spec that does not work, right.

(fwiw, webidl2.js and bikeshed do allow this, too.)

saschanaz:
>(fwiw, webidl2.js and bikeshed do allow this, too.)

I neither wouldn't base too much on what they allow, because both have been just parsers without semantics checks, while only recently webidl2.js started providing some.

rniwa:
@cdumez

bzbarsky:
> (fwiw, webidl2.js and bikeshed do allow this, too.)

Right, it arguably used to be OK (and was supported in Gecko) until the changes in this PR.  At this point it's clearly not OK, though.  IDL `null` (not to be confused with JS `null`!) is not a valid value for a dictionary type.

PR #756, opened 2019-07-02
Editorial: Better explanation of the TreatNonObjectAsNull extended attribute.

Ms2ger:
When the name and behaviour of the extended attribute were changed back in
4043b051d492d4a2f21064f0edf91e642c5a6249, various prose around the
specification became somewhat unclear. This tries to make everything agree on
what happens.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6241,6 +6241,9 @@ An [=identifier=] that identifies
 a [=callback function=] is used to refer to
 a type whose values are references to objects that are functions with the given signature.
 
+Note: If the [{{TreatNonObjectAsNull}}] [=extended attribute=] is specified on the [=definition=] of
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5961,6 +5961,9 @@ An [=identifier=] that identifies
 a [=callback function=] is used to refer to
 a type whose values are references to objects that are functions with the given signature.
 
+Note: If the [{{TreatNonObjectAsNull}}] [=extended attribute=] specified on the [=definition=] of
```
bzbarsky: s/specified/ is specified/

bzbarsky:
```suggestion
Note: If the [{{TreatNonObjectAsNull}}] [=extended attribute=] is specified on the [=definition=] of
```

PR #772, opened 2019-08-20
Import the Promises Guide.

### LINKED ISSUE #490: Import (and fix) the promises guide

Ms2ger:
Fixes #490.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/772.html" title="Last updated on Aug 27, 2019, 9:45 AM UTC (c3fcded)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/772/34ef605...c3fcded.html" title="Last updated on Aug 27, 2019, 9:45 AM UTC (c3fcded)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -105,6 +105,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %Promise_reject%; url: sec-promise.reject
         text: %Promise_resolve%; url: sec-promise.resolve
         text: %PromiseProto_then%; url: sec-promise.prototype.then
+        text: all(); for: Promise; url: #sec-promise.all
     type: const; for: ECMAScript
         url: sec-well-known-symbols
             text: @@asyncIterator
@@ -218,6 +219,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
             text: internal method
             text: internal slot
         text: own property; url: sec-own-property
+        text: PromiseCapability; url: sec-promisecapability-records
         text: Property Descriptor; url: sec-property-descriptor-specification-type
         text: Source Text Module Record; url: sourctextmodule-record
         text: realm; url: realm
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
...
-    When a specification says to perform some steps once a promise is settled,
-    the following steps must be followed:
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
+    given a <a interface lt="Promise">Promise&lt;|T|&gt;</a> |promise| and one or two sets of steps to perform,
+    covering when the promise is fulfilled, rejected, or both, perform the following steps:
 
-    1.  Let |promise| be the promise object of type <a interface lt="Promise">Promise&lt;|T|&gt;</a>.
-    1.  Let |onFulfilled| be a new [=built-in function object=] whose
-        behavior when invoked is as follows:
+    1.  Let |onFulfilled| be [=!=] [$CreateBuiltinFunction$](the following steps, « »):
```
annevk: Unless there's precedent for this pattern I'd prefer assigning the following steps to a variable and then using that variable in the creation of this builtin function.

index.bs
```diff
@@ -8083,21 +8079,391 @@ objects.
...
+    <a interface lt="Promise">Promise&lt;<var ignore>T</var>&gt;</a> in a [=Realm=] |realm|,
+    perform the following steps:
+
+    1.  Let |constructor| be |realm|.\[[Intrinsics]].[[{{%Promise%}}]].
+    1.  Return [=?=] [$NewPromiseCapability$](|constructor|).
+</div>
+
+<div algorithm>
+    To create a <dfn export for=Promise>resolved promise</dfn> of type
+    <a interface lt="Promise">Promise&lt;|T|&gt;</a>, with |x| (a value of type |T|) in a [=Realm=]
```
annevk: Maybe make the parenthetical a type assertion step?

Ms2ger: I think it makes more sense to make it part of the "signature" of the algorithm, so to say.

index.bs
Ms2ger: I thought there was precedent, but now I can't find it. Changed.

index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
 
 <h4 id="es-promise">Promise types — Promise&lt;|T|&gt;</h4>
 
-IDL [=promise type=] values are
-represented by ECMAScript {{ECMAScript/Promise}}
-objects.
+IDL [=promise type=] values are represented by ECMAScript {{ECMAScript/PromiseCapability}} records.
```
domenic: Nit: This is not an interface, so shouldn't be using IDL syntax or be tagged as such in the Bikeshed metadata.

index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
...
-    When a specification says to perform some steps once a promise is settled,
-    the following steps must be followed:
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
+    given a <a interface lt="Promise">Promise&lt;|T|&gt;</a> |promise| and one or two sets of steps to perform,
+    covering when the promise is fulfilled, rejected, or both, perform the following steps:
 
-    1.  Let |promise| be the promise object of type <a interface lt="Promise">Promise&lt;|T|&gt;</a>.
-    1.  Let |onFulfilled| be a new [=built-in function object=] whose
-        behavior when invoked is as follows:
+    1.  Let |onFulfilledSteps| be the following steps:
```
domenic: I think "given an argument |V|" would be fine here and remove some of the future boilerplate. (Same for onRejected.)

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
                 return <emu-val>undefined</emu-val>.
             1.  Otherwise, return the result of performing any steps that were required to be run if the promise was fulfilled,
                 with |value| as the promise’s value.
```
domenic: maybe "given |value|" (same below)

index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
...
 
 <div algorithm>
 
-    One can <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>.
-    There can be one or two sets of steps to perform,
-    covering when the promise is fulfilled, rejected, or both.
-    When a specification says to perform some steps once a promise is settled,
-    the following steps must be followed:
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
+    given a <a interface lt="Promise">Promise&lt;|T|&gt;</a> |promise| and one or two sets of steps to perform,
```
domenic: Add `<code>` too :). Here and below.

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+
+    Note: If the provided steps return a value of type |T|, this algorithm returns a
+    <a interface lt="Promise">Promise&lt;|T|&gt;</a>.
 </div>
 
-<p class="issue">
-    Include an example of how to write spec text using this term.
-</p>
+<div algorithm>
+    To <dfn export for=Promise lt="a new promise">create</dfn> a new
```
domenic: Thank you for the "a new promise" linking text

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+    <a interface lt="Promise">Promise&lt;<var ignore>T</var>&gt;</a> in a [=Realm=] |realm|,
+    perform the following steps:
+
+    1.  Let |constructor| be |realm|.\[[Intrinsics]].[[{{%Promise%}}]].
+    1.  Return [=?=] [$NewPromiseCapability$](|constructor|).
+</div>
+
+<div algorithm>
+    To create a <dfn export for=Promise>resolved promise</dfn> of type
```
domenic: "create a promise resolved with" or "a promise resolved with" is the common usage; it would be good to have those either as the main dfn text or as alternate linking text.

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+This phrase is useful when you wish to aggregate the result of multiple promises, and react to them
+all together, in the same way that {{Promise/all()|Promise.all()}} functions for JavaScript code.
+
+<div algorithm>
+    To
+    <dfn id="waiting-for-all-promise" export lt="get a promise to wait for all|getting a promise to wait for all">get a promise for waiting for all</dfn>
+    with a [=list=] of <a interface lt="Promise">Promise&lt;|T|&gt;</a> values |promises| and a
+    [=Realm=] |realm|, perform the following steps:
+
+    1.  Let |promise| be [=Promise/a new promise=] of type Promise&lt;sequence&lt;|T|>> in |realm|.
```
domenic: `Promise&lt;sequence&lt;|T|>>` needs `<code>`, and probably ideally some links.

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+<div algorithm="delay">
+
+    The <code>delay(|ms|)</code> method steps are:
+
+    1.  Let |realm| be <b>this</b>'s [=relevant Realm=].
+    1.  If |ms| is NaN, let |ms| be +0; otherwise let |ms| be the maximum of |ms| and +0.
+    1.  Let |p| be [=Promise/a new promise=] in |realm|.
+    1.  Run the following steps [=in parallel=]:
+        1.  Wait |ms| milliseconds.
+        1.  [=Promise/Resolve=] |p| with the {{void}} value.
```
domenic: It'd be nice if we could make the argument optional for cases like this; WDYT?

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+    1.  Return |p|.
+</div>
+
+The equivalent function in JavaScript would be
+
+<pre highlight="js">
+function delay(ms) {
+    // Verify the this value is correct.
```
```suggestion
    // Omitted: verify the this value is correct.
```

and the same below.

Although, I am unsure if we really need to preserve these JavaScript versions, given the Web IDL spec audience. They are somewhat problematic, depending on how strongly you take "equivalent".

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+        };
+    </pre>
+
+<div algorithm="batchRequest">
+
+    The <code>batchRequest(|urls|)</code> method steps are:
+
+    1.  Let |responsePromises| be « ».
+    1.  [=list/For each=] |url| of |urls|:
+        1.  Let |p| be the result of calling {{WindowOrWorkerGlobalScope/fetch()}} with |url|.
```
domenic: Maybe better "running the steps specified by the fetch() method"

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
+        1.  [=list/Append=] |p| to |responsePromises|.
+    1.  Let |p| be [=getting a promise to wait for all=] with |responsePromises|.
```
```suggestion
    1.  Let |p| be the result of [=getting a promise to wait for all=] with |responsePromises|.
```

index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
...
 
 <p id="promise-to-es">
     The result of [=converted to an ECMAScript value|converting=]
     an IDL [=promise type=] value to an ECMAScript
-    value is the {{ECMAScript/Promise}} value that represents a reference to the same object that the
+    value is the value of the \[[Promise]] field of the record that
     IDL [=promise type=] represents.
 </p>
 
 <div algorithm>
```
domenic: Maybe introduce a subheading here, something like "Creating and manipulating promises"?

index.bs
```diff
@@ -8041,38 +8043,32 @@ ECMAScript Object values.
...
 
-    One can <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>.
-    There can be one or two sets of steps to perform,
-    covering when the promise is fulfilled, rejected, or both.
-    When a specification says to perform some steps once a promise is settled,
-    the following steps must be followed:
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
```
domenic: Maybe move this below all the creation/resolve/reject operations, i.e. before "wait for all"

domenic: I think we should also add "upon fulfillment" and "upon rejection" shorthands.

index.bs
```diff
@@ -8083,21 +8079,395 @@ objects.
...
             |R| to an IDL value of type {{any}}.
         1.  If there are no steps that are required to be run if the promise was rejected, then
             return <emu-val>undefined</emu-val>.
         1.  Otherwise, return the result of performing any steps that were required to be run if the promise was rejected,
             with |reason| as the rejection reason.
-    1.  Return [=!=] [$PerformPromiseThen$](|promise|, |onFulfilled|, |onRejected|).
+    1.  Let |onRejected| be [=!=] [$CreateBuiltinFunction$](|onRejectedSteps|, « »):
+    1.  Let |constructor| be |promise|.\[[Promise]].\[[Realm]].\[[Intrinsics]].[[{{%Promise%}}]].
+    1.  Let |newCapability| be ? [$NewPromiseCapability$](|constructor|).
+    1.  Return [=!=] [$PerformPromiseThen$](|promise|.\[[Promise]], |onFulfilled|, |onRejected|, |newCapability|).
```
domenic: You don't need to pass newCapability since it's unused. Just pass three arguments.

domenic: Well, current "upon fulfillment" etc. usages don't have a return value... that's where you'd use "transforming". But I see you seem to be allowing this to stand in for both, hmm.

index.bs
```diff
@@ -8062,53 +8062,72 @@ IDL [=promise type=] values are represented by ECMAScript {{ECMAScript/PromiseCa
...
     1.  Let |newCapability| be ? [$NewPromiseCapability$](|constructor|).
     1.  Return [=!=] [$PerformPromiseThen$](|promise|.\[[Promise]], |onFulfilled|, |onRejected|, |newCapability|).
 
     Note: If the provided steps return a value of type |T|, this algorithm returns a
-    <a interface lt="Promise">Promise&lt;|T|&gt;</a>.
+    <code>{{Promise}}<var ignore>T</var>&gt;</code>.
+</div>
+
+<div algorithm>
+    To perform some steps <dfn export for=Promise>upon fulfillment</dfn> of a
```
domenic: Realistically I think most of these don't need for=Promise; there's no competition to disambiguate with. At least, not the ones that explicitly have "promise" in their name, for sure...

index.bs
```diff
@@ -8041,63 +8043,367 @@ ECMAScript Object values.
...
-IDL [=promise type=] values are
-represented by ECMAScript {{ECMAScript/Promise}}
-objects.
+IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=] records.
 
 <div id="es-to-promise" algorithm="convert an ECMAScript value to promise">
 
     An ECMAScript value |V| is [=converted to an IDL value|converted=]
-    to an IDL <a interface lt="Promise">Promise<var ignore>T</var></a> value as follows:
+    to an IDL <code>{{Promise}}&lt;<var ignore>T</var>></code> value as follows:
```
domenic: This pattern results in nested `<code>`. I guess that's not technically invalid? But what would work better would be the verbose

```html
<code><a interface>Promise</a>&lt;<var ignore>T</var>></code>
```

index.bs
domenic: I guess we can just update all callers to say "Let _x_ be the result of performing the following steps upon fulfillment of _p_, given _v_" instead of the current "Let _x_ be the result of transforming _p_ by the following fulfillment steps, given _v_". That's not very different.

index.bs
```diff
@@ -8041,63 +8043,367 @@ ECMAScript Object values.
...
+        1.  Let |result| be the result of performing any steps that were required to be run if the
+            promise was rejected, given |reason|.
+        1.  Return |result|, [=converted to an ECMAScript value=].
+    1.  Let |onRejected| be [=!=] [$CreateBuiltinFunction$](|onRejectedSteps|, « »):
+    1.  Let |constructor| be |promise|.\[[Promise]].\[[Realm]].\[[Intrinsics]].[[{{%Promise%}}]].
+    1.  Let |newCapability| be ? [$NewPromiseCapability$](|constructor|).
+    1.  Return [=!=] [$PerformPromiseThen$](|promise|.\[[Promise]], |onFulfilled|, |onRejected|, |newCapability|).
+
+    Note: If the provided steps return a value of type |T|, this algorithm returns a
+    <code>{{Promise}}&lt;<var ignore>T</var>&gt;</code>.
```
domenic: Maybe we should use a different letter than T here, since it could be anything. And we should note that if they throw, it'll return a rejected promise? Should we also talk about what happens if they return promises?? Hmm, maybe just saying that it behaves like .then() would summarize better?

index.bs
```diff
@@ -8041,63 +8043,367 @@ ECMAScript Object values.
...
+            1.  Set |result|[|promiseIndex|] to |arg|.
+            1.  Set |fullfilledCount| to |fullfilledCount| + 1.
+            1.  If |fullfilledCount| equals |total|, then perform |successSteps| given |result|.
+        1.  Let |fulfillmentHandler| be [=!=] [$CreateBuiltinFunction$](|fulfillmentHandler|, « »):
+        1.  Perform [$PerformPromiseThen$](|promise|, |fulfillmentHandler|, |rejectionHandler|).
+        1.  Set |index| to |index| + 1.
+</div>
+
+This phrase is useful when you wish to aggregate the result of multiple promises, and react to them
+all together, in the same way that {{Promise/all()|Promise.all()}} functions for JavaScript code.
```
domenic: Probably the clause "in the same way..." should move to the "get a promise for waiting for all". I think this was just missed when I separated the two operations in the promises guide.

index.bs
```diff
@@ -8041,63 +8043,367 @@ ECMAScript Object values.
...
+
+This phrase is useful when you wish to aggregate the results of multiple promises, and then produce
+another promise from them.
+
+
+<h6 id="es-promise-examples">Examples</h6>
+
+<div class="example" id="promise-example-delay">
+
+    <code>delay</code> is a function that returns a promise that will be fulfilled in a number of
```
```suggestion
    <code>delay</code> is an operation that returns a promise that will be fulfilled in a number of
```

Or method? Anyway, same for the others...

index.bs
```diff
@@ -8041,63 +8043,367 @@ ECMAScript Object values.
...
+        1.  [=Promise/Reject=] |promise| with |reason|.
+    1.  [=Promise/Wait for all=] with |promises|, given |successSteps| and |failureSteps|.
+    1.  Return |promise|.
+</div>
+
+This phrase is useful when you wish to aggregate the results of multiple promises, and then produce
+another promise from them.
+
+
+<h6 id="es-promise-examples">Examples</h6>
```
domenic: My intention with the "add a heading" comment was for this to stay a h5, i.e. have two siblings under the promise section. Arguably this section has examples of the general promise type definitions too.

index.bs
domenic: Well, the intent is to make it clear what happens when author code overrides `fetch()`.

index.bs
Ms2ger: Right, I meant to call that out explicitly, but it seems I forgot. I could add a note that it's not necessary to create the PromiseCapability

PR #791, opened 2019-09-03
Handle conversion failures in 'upon settling'.

### LINKED ISSUE #782: Promise handling algorithms are not very clear about what happens if type conversion fails
https://heycam.github.io/webidl/#dfn-perform-steps-once-promise-is-settled

What if the type is `Promise<Node>` and someone passes `Promise.resolve(3)`?

I think the best we can do is go down the rejection path? Although that makes using "upon fulfillment" in isolation pretty dangerous for non-`any` _T_s.

Ms2ger:
Fixes #782.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8236,8 +8236,15 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=] 
     steps:
 
     1.  Let |onFulfilledSteps| be the following steps given argument |V|:
-        1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
-            value of type |T|.
+        1.  Try running the following step:
+            1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
+                value of type |T|.
+
+            And then, if <a lt="an exception was thrown">an exception |exception| was thrown</a>:
+
+            1.  If there are steps that are required to be run if the promise was rejected, let
+                |result| be the result of performing those steps, given |exception|.
+            1.  [=ECMAScript/throw=] |exception|.
         1.  If there are no steps that are required to be run if the promise was fulfilled, then
             return <emu-val>undefined</emu-val>.
         1.  Let |result| be the result of performing any steps that were required to be run if


index.bs
```diff
@@ -8236,8 +8236,15 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=]
...
-        1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
-            value of type |T|.
+        1.  Try running the following step:
+            1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
+                value of type |T|.
+
+            And then, if <a lt="an exception was thrown">an exception |exception| was thrown</a>:
+
+            1.  If there are steps that are required to be run if the promise was rejected, let
+                |result| be the result of performing those steps, given |exception|.
```
domenic: |result| is unused. Should it be returned, instead of throwing exception?

Ms2ger: Hm, good question. This matters if:

1. we want to do a type conversion
1. the returned promise is used (former "transforming")
1. we want to do something else than passing the rejection along

Are you aware of an existing case to inform the decision? From https://github.com/heycam/webidl/issues/782#issuecomment-526519971, `waitUntil`/`complete` fail (1), `respondWith`/`updateWith` fail (2), WebAssembly fails (3).

Writing this, I'm realizing that we never actually pass the rejection along, the way WebAssembly needs. Should we pass **undefined** instead of `onRejectedSteps` if there's no steps?

domenic: > Are you aware of an existing case to inform the decision?

I'm not aware of any. However I generally think aligning with promise.then() will be more intuitive for spec authors, if we have that degree of freedom. So I'd prefer that if possible.

> Writing this, I'm realizing that we never actually pass the rejection along, the way WebAssembly needs. Should we pass undefined instead of onRejectedSteps if there's no steps?

Definitely.

annevk:
@Ms2ger do you plan to complete this?

Ms2ger:
Probably not.

PR #792, opened 2019-09-03
Use "react" for the spec promise.then() analog

### LINKED ISSUE #783: "Perform some steps once a promise is settled" awkward to use in return value form
What prose do we expect spec authors to write to replace the "transforming" from the promises guide? The best I've got is

> Return the result of [=upon settling|performing the following steps=] once _p_ is settled, with the fullfilment handler which takes an argument |x|:

This seems kind of terrible. What could work better? Should we reintroduce "transforming"?

Related problem: the algorithm name is "perform some steps once a promise is settled", but you cannot link to it by that name, because it has `lt="upon settling"`. All usage sites use "upon settling".

domenic:
Closes #783.

/cc @jakearchibald

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/792.html" title="Last updated on Sep 3, 2019, 7:36 PM UTC (1f89230)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/792/182b487...1f89230.html" title="Last updated on Sep 3, 2019, 7:36 PM UTC (1f89230)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8230,8 +8230,8 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=] 
 
 <div algorithm>
 
-    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
-    given a <code><a interface>Promise</a>&lt;|T|&gt;</code> |promise| and one or two sets of steps
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="react|reacting" for="promise">react</dfn>
+    to a <code><a interface>Promise</a>&lt;|T|&gt;</code> |promise|, given one or two sets of steps
     to perform, covering when the promise is fulfilled, rejected, or both, perform the following
     steps:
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
I didn't realise it was ok to do for="promise" here (since none of the other algorithms do).

domenic:
I did it since "react" is pretty generic, whereas e.g. "upon fulfillment" is only ever going to be used in a promise context.

index.bs
```diff
@@ -8230,8 +8230,8 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=]
 
 <div algorithm>
 
-    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="upon settling">perform some steps once a promise is settled</dfn>,
-    given a <code><a interface>Promise</a>&lt;|T|&gt;</code> |promise| and one or two sets of steps
+    To <dfn id="dfn-perform-steps-once-promise-is-settled" export lt="react|reacting" for="promise">react</dfn>
```
Ms2ger: "promise" or "Promise"?

domenic: "promise" is the concept, `Promise<T>` is the type, so I think lowercase is good here.

PR #797, opened 2019-09-17
Define "usual constructor steps"

domenic:
This is for use in [HTMLConstructor] in https://github.com/whatwg/html/pull/4915.

There are alternative approaches and wordings that might be better, e.g.

- Say "perform the following steps or any others specified by specs" inline instead of allowing the usual steps to be replaced using a paragraph at the bottom
- Have an explicit "overridden constructor steps" value for each interface object, which [HTMLConstructor] sets and this algorithm consults

But I thought I'd start here to get the ball rolling, as it's the first thing that occurred to me. Happy to tweak as desired.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/797.html" title="Last updated on Sep 18, 2019, 8:36 AM UTC (4554894)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/797/db74ef4...4554894.html" title="Last updated on Sep 18, 2019, 8:36 AM UTC (4554894)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11130,6 +11130,13 @@ An interface may have <dfn export>overridden constructor steps</dfn>, which can
 change the behavior of the [=interface object=] when called or constructed. By
 default interfaces do not have such steps.
 
+<p class="advisement">
+    In general, constructors are described by defining a [=constructor operation=] and its behavior.
+    The [=overridden constructor steps=] are used only for more complicated situations.
+    Editors who wish to use this feature are strongly advised to discuss this by
+    <a href="https://github.com/heycam/webidl/issues/new?title=Intent%20to%20use%20constructor%20steps">filing an issue</a> before proceeding.
+</p>
+
 <div algorithm>
 
     The [=interface object=] for a given [=interface=] |I|
--- a/index.bs
+++ b/index.bs
@@ -11130,7 +11130,7 @@ An interface may have <dfn export>overridden constructor steps</dfn>, which can
 change the behavior of the [=interface object=] when called or constructed. By
 default interfaces do not have such steps.
 
-<div>
+<div algorithm>
 
     The [=interface object=] for a given [=interface=] |I|
     with [=identifier=] |id| and in [=Realm=] |realm|
--- a/index.bs
+++ b/index.bs
@@ -11130,7 +11130,7 @@ An interface may have <dfn export>overridden constructor steps</dfn>, which can
 change the behavior of the [=interface object=] when called or constructed. By
 default interfaces do not have such steps.
 
-<div algorithm="create an interface object">
+<div>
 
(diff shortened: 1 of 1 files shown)


Ms2ger:
I'm not a big fan of this approach, personally; I think it'd be better to have something like

1. Let _steps_ be [other steps (for interface _I_)], if they exist, or the [usual constructor steps (for interface _I_)] otherwise.

Any reason you're referring to the interface object, in particular?

domenic:
Sounds like my second option, right? Can do.

> Any reason you're referring to the interface object, in particular?

Nope, interface makes sense too.

domenic:
@Ms2ger I pushed a new approach, but I feel like there should be some warning text talking about how these steps are not general-purpose and are only used in special scenarios. Would you be able to help draft such text? I'm having some trouble doing so myself.

index.bs
```diff
@@ -11126,13 +11126,18 @@ and is described in more detail in [[#interface-prototype-object]].
 Note: Since an [=interface object=] is a [=function object=]
 the <code>typeof</code> operator will return "function" when applied to an interface object.
 
+An interface may have <dfn export>overridden constructor steps</dfn>, which can
+change the behavior of the [=interface object=] when called or constructed. By
+default interfaces do not have such steps.
+
```
Ms2ger: Maybe add a warning like we have a few already for legacy features, saying something like

> In general, constructors are described by defining a [=constructor operation=] and its behavior.
> The [=overridden constructor steps=] are used only for more complicated situations. 
> Editors who wish to use this feature are strongly advised to discuss this by filing an issue before proceeding.

index.bs
```diff
@@ -11126,13 +11126,18 @@ and is described in more detail in [[#interface-prototype-object]].
...
     The [=interface object=] for a given [=interface=] |I|
     with [=identifier=] |id| and in [=Realm=] |realm|
     is <dfn lt="create an interface object">created</dfn> as follows:
 
-    1.  Let |steps| be the following steps:
+    1.  Let |steps| be |I|'s [=overriden constructor steps=] if they exist, or
+        the following steps otherwise:
         1.  If |I| was not declared with a [=constructor operation=],
             then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
         1.  If {{NewTarget}} is <emu-val>undefined</emu-val>, then
```
Ms2ger: Should we keep the first two steps, even when overriding the rest of the algorithm?

Maybe also the last four steps?

domenic: I guess that would work for [HTMLConstructor], although it feels cleaner to just have a general primitive that replaces the whole thing.

domenic: I'll keep it like this for now, but we can refactor later if you think it's much better.

PR #798, opened 2019-09-23
Remove use of [Constructor] extended attribute

clelland:
This removes [Constructor] from DOMException and from examples, replacing
it as necessary with constructor operations.

***
<a href="https://pr-preview.s3.amazonaws.com/clelland/webidl/pull/798.html" title="Last updated on Sep 23, 2019, 2:42 PM UTC (2a37a9b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/798/db74ef4...clelland:2a37a9b.html" title="Last updated on Sep 23, 2019, 2:42 PM UTC (2a37a9b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2910,8 +2910,9 @@ a [=static attribute=].
     <code class="idl">name</code> attribute:
 
     <pre highlight="webidl">
-        [Exposed=Window, Constructor]
+        [Exposed=Window]
         interface Student {
+          constructor();
           attribute unsigned long id;
           stringifier attribute DOMString name;
         };
@@ -2935,8 +2936,9 @@ a [=static attribute=].
     not specified in the IDL itself.
 
     <pre highlight="webidl">
-        [Exposed=Window, Constructor]
+        [Exposed=Window]
         interface Student {
+          constructor();
           attribute unsigned long id;
           attribute DOMString? familyName;
           attribute DOMString givenName;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6799,7 +6802,7 @@ which form (or forms) it is in:
             <dfn id="dfn-xattr-argument-list" for="extended attribute" export>takes an argument list</dfn>
         </td>
         <td>
-            <code>[Constructor(double x, double y)]</code>
+            Not currently used; previously used by <code>[Constructor(double x, double y)]</code>
```
Ms2ger: Not even by NamedConstructor?

clelland: Nope -- it uses named argument list, rather than just argument list:

> [The [NamedConstructor] extended attribute must either take an identifier or take a named argument list.](https://heycam.github.io/webidl/#NamedConstructor)

It's possible that implementer-specific attributes still use this (I haven't checked Blink's list); that's the only reason I didn't try to remove it entirely.

Ms2ger: D'oh!

domenic: Interesting. We should probably create a thread to discuss removing this entirely then, and let implementers chime in there.

clelland: The only other usage I could find is in the Blink-specific 'CustomConstructor' attribute, but I suspect we could probably replace usages of

```
[ CustomConstructor(float x, int y) ] interface Z { ... };
```

with

```
interface Z {
  [Custom] constructor(float x, int y);
  ...
};
```

and be rid of that attribute.

PR #802, opened 2019-09-24
Tweak async iterator algorithms

domenic:
Make both the "async iterator initialization steps" and the "get the next iteration result" take two arguments: the instance, and the async iterator. This ends up being more ergonomic than using the "current state" concept.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/802.html" title="Last updated on Oct 16, 2019, 10:54 AM UTC (2d6954b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/802/3f89daf...2d6954b.html" title="Last updated on Oct 16, 2019, 10:54 AM UTC (2d6954b)">Diff</a>

PR #805, opened 2019-09-25
Allow async iterators to specify return algorithms

### LINKED ISSUE #804: Async iterators should be allowed to have an optional return() algorithm
So that we can implement https://streams.spec.whatwg.org/#rs-asynciterator-prototype-return

We can do the boilerplate of returning { value: argumentPassed, done: true } for you.

domenic:
Closes #804. Notably this allows consumers to not worry about resolving their promise with the correct iterator result.

This is on top of the stage-setting commits in #802; only the last commit is worth reviewing.

One choice I made here where the right answer wasn't immediately obvious is that calling return() while a call to next() is outstanding will reject with no side effects. This mirrors what we've done in the Streams design, and seems simpler than e.g. waiting for 1+ outstanding next()s to settle. This will only occur when manually manipulating the async iterator (as opposed to using for-await-of) so it's not a very mainline scenario. /cc @ricea in case he has any thoughts.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/805.html" title="Last updated on Oct 16, 2019, 11:04 AM UTC (01691c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/805/9d11c76...01691c3.html" title="Last updated on Oct 16, 2019, 11:04 AM UTC (01691c3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4310,6 +4310,22 @@ iteration, or resolves with a tuple containing two elements:
 1.  a value of the first type given in the declaration;
 1.  a value of the second type given in the declaration.
 
+The prose may also define an <dfn export>asynchronous iterator return</dfn> algorithm. This
+algorithm receives the instance of the [=interface=] that is being iterated, the async iterator
+itself, and a single argument value. This algorithm is invoked in the case of premature termination
+of the async iterator. It must return a {{Promise}}; if that promise fulfills, its fulfillment value
+will be ignored, but if it rejects, that failure will be passed on to users of the async iterator
+API.
+
+<p class="note">In the ECMAScript binding, this algorithm allows customizing the behavior when the
+async iterator's <code>return()</code> method is invoked. This most commonly occurs when a
+<code>break</code> or <code>return</code> statement causes an exit from a
+<code>for</code>-<code>await</code>-<code>of</code> loop.
+
+<p class="note">We could add a similar hook for <code>throw()</code>. So far there has been no need,
+but if you are creating an API that needs such capabilities, please
+<a href="https://github.com/heycam/webidl/issues/new?title=Enhancement%20request%20for%20Async%20Iterables">file an issue</a>.
+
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Reported @fallenAngelSpeaks2Me

index.bs
```diff
@@ -4307,6 +4307,21 @@ iteration, or resolves with a tuple containing two elements:
 1.  a value of the first type given in the declaration;
 1.  a value of the second type given in the declaration.
 
+The prose may also define an <dfn export>asynchronous iterator return</dfn> algorithm. This
+algorithm receives the instance of the [=interface=] that is being iterated, the async iterator
+itself, and a single argument value. This algorithm is invoked in the case of premature termination
```
Ms2ger: And the single argument value is an {{any}} value?

index.bs
```diff
@@ -12595,7 +12610,70 @@ The \[[Prototype]] [=internal slot=] of an [=asynchronous iterator prototype obj
...
+            <emu-val>undefined</emu-val>, « |e| »).
+        1.  Return |returnPromiseCapability|.\[[Promise]].
+
+    1.  If |object| is not a [=default asynchronous iterator object=] for |interface|, then:
+        1.  Let |error| be a new {{ECMAScript/TypeError}}.
+        1.  Perform [=!=] [$Call$](|returnPromiseCapability|.\[[Reject]],
+            <emu-val>undefined</emu-val>, « |error| »).
+        1.  Return |returnPromiseCapability|.\[[Promise]].
+
+    1.  If |object|'s [=default asynchronous iterator object/ongoing promise=] is not undefined, then:
```
Ms2ger: Check if this undefined needs an emu-val? I don't remember which way this one goes.

domenic: Mmm nice catch. Since we made it an IDL Promise, we should probably union it with null (another IDL value), not undefined (which is always `<emu-val>undefined</emu-val>`, a JS value). I switched all instances to do so.

Issue #813, opened 2019-10-07
Seemingly missing alternatives in Other and ArgumentNameKeyword


bathos:
It appears as if the Other non-terminal is meant to be inclusive of every terminal except for the three brace pairs and comma `()[]{},`. This makes sense because it’s an open-ended point for extension; users are free to define custom grammars for specific local extended attributes, so long as the braces are balanced and comma doesn’t appear outside of braces.

There is an informative note that makes this explicit:

> Note: The Other non-terminal matches any single terminal symbol except for (, ), [, ], {, } and ,.

However it does not actually permit the following terminals:

- `Promise`
- `mixin`
- `readonly`
- `record`
- `symbol`

I’m guessing these might be accidental oversights due to these terminals having been defined later than `Other`. Any time a new identlike terminal is defined, it appears that there is a need to update either `Other` or `ArgumentNameKeyword`, but this is not super obvious.

Following the existing pattern, it appears that `Other` itself should add the following alternatives, because they are terminals used in `Type`:

- `Promise`
- `record`
- `symbol`

And that `ArgumentNameKeyword`, which appears as though it’s intended to permit all the identlike terminals except for those used in `Type` and is an alternative for `Other`, should add:

- `mixin`
- `readonly`

PR #814, opened 2019-10-07
Add missing terminals to Other and ArgumentNameKeyword (#813)

bathos:
See #813.

***
<a href="https://pr-preview.s3.amazonaws.com/bathos/webidl/pull/814.html" title="Last updated on Oct 7, 2019, 7:42 PM UTC (619a953)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/814/8c71e7c...bathos:619a953.html" title="Last updated on Oct 7, 2019, 7:42 PM UTC (619a953)">Diff</a>


Ms2ger:
@bzbarsky ?

I wonder if we could lint for this.

bathos:
You mentioned on #815:

> At the very least, we should import that code into this repo. I wouldn't mind seeing it rewritten in python either, but that's not urgent.

I’d be happy to have a go at both integrating the postprocessor code and adding a test to check-grammar\* that confirms that Other describes the set of all terminals save the seven intended exceptions. Even if they don’t share code directly, they would benefit from a bit more uniformity since they are doing very similar things. However I’d like to do so as part of a more general effort to clean the build and test scripts and, in particular, to make them platform agnostic.

Aside from the deploy script, the platform assumptions currently made seem to be avoidable. I’m on windows, so I already had to write a js equiv of the makefile to jump the initial barrier to participation, and I could polish/share that.

I’m afraid though that if I do this, stuff will either remain JS or, in the base of bash stuff, _become_ JS 😱. My Python is pretty limited.

\* I now see this script repeats the "float" issue from #815, which might hint that there are additional worthwhile grammar checks that are missing.

bathos:
@bzbarsky 

The check-grammar.js script uses [syntax-cli’s](https://github.com/DmitrySoshnikov/syntax#readme) `Grammar` class —

https://github.com/heycam/webidl/blob/4203404f78d72d448d7d32c563733a35fa99b0ac/check-grammar.js#L40

— which exposes methods like [`getTerminals()`](https://github.com/DmitrySoshnikov/syntax/blob/212a6720d20e7d7b87b78106f7f6de8d37ca8d20/src/grammar/grammar.js#L332-L352)

I haven’t dug in enough to say for sure, but it appears that the syntax-cli module provides the grammar introspection tools which would be needed to perform the test.

PR #818, opened 2019-10-24
Use Travis-CI deployment support.

Ms2ger:


domenic:
This isn't as secure as using a per-repo deploy key; in particular GitHub access tokens give access to all repositories on your account.

The setup I've been using recently is still based on encoded deploy keys. See e.g. https://github.com/WICG/origin-policy/commit/c3be6b3b84c92a8e49fce1a5eca91a7eb70c4158#diff-354f30a63fb0907d4ad57269548329e3

Ms2ger:
Thanks for the pointer, I'll take a look.

marcoscaceres:
(remember you can also use GitHub actions... they do the same thing and can be easier to mange)

Ms2ger:
@marcoscaceres do you have a pointer to a repo that does that?

marcoscaceres:
An example of how we publish Gamepad spec: 
https://github.com/w3c/gamepad/blob/gh-pages/.github/workflows/pr-push.yml

But obviously need to look up how to get the equivalent of `deploy` from Github actions. The tokens can be stored in the repo settings... there is new "Secrets" option in the settings side bar.

.travis.yml
```diff
@@ -4,10 +4,18 @@ node_js:
...
+deploy:
+  local_dir: out/
+  target_branch: gh-pages
+  keep_history: true
+  provider: pages
+  skip_cleanup: true
+  github_token: $GH_ACCESS_TOKEN
+  committer_from_gh: true
+  on:
+    branch: main
```
ExE-Boss:
```suggestion
    branch: master
```

.travis.yml
```diff
@@ -4,10 +4,18 @@ node_js:
 
 script:
   - bash ./deploy.sh
```
ExE-Boss: This is why the build is failing with exit code `127`:

```suggestion
```

annevk:
Closing as this doesn't seem like the desired way forward.

Issue #484, opened 2017-12-05
Consider semantic changes to WebIDL to make suitable for ECMAScript


littledan:
If we were going to use WebIDL for the ECMAScript specification, we'd need a few more features, such as:
- Non-enumerable methods
- (possibly) declarations of own properties
- ... (there is probably more; I will come back to edit this issue to fill more in)

bzbarsky:
> Non-enumerable methods

This came up in https://github.com/heycam/webidl/issues/226 (and in general there is other discussion in there that may be relevant here).

> (possibly) declarations of own instance data properties

https://www.w3.org/Bugs/Public/show_bug.cgi?id=27346 discusses that; looks like it was not migrated to a github issue.

domenic:
- Methods and getters that don't brand check

TimothyGu:
- [Class strings of prototype objects](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28244)

domenic:
The latter is fixed by #357, FWIW.

littledan:
This is an easy one, but we'll want more types, such as to express ToIndex.

PR #825, opened 2019-12-11
feat: Add `[NonEnumerableMethods]` extended attribute

ExE-Boss:
This&nbsp;adds a&nbsp;`[NonEnumerableMethods]` extended&nbsp;attribute to&nbsp;let&nbsp;specs match&nbsp;what **ECMAScript**&nbsp;does with&nbsp;its&nbsp;built‑ins and&nbsp;the&nbsp;`class`&nbsp;construct.

It&nbsp;also&nbsp;provides `[LegacyEnumerableMethod]` for&nbsp;the&nbsp;legacy&nbsp;cases that&nbsp;would&nbsp;break if&nbsp;the&nbsp;methods on&nbsp;which it’s&nbsp;specified would&nbsp;be&nbsp;made&nbsp;non‑enumerable.

---

This&nbsp;is&nbsp;need to&nbsp;address&nbsp;<https://github.com/heycam/webidl/issues/226>, <https://github.com/heycam/webidl/issues/484>, <https://github.com/WebAssembly/spec/issues/1110> and&nbsp;<https://github.com/whatwg/streams/issues/963>,

Supersedes and&nbsp;closes #719

***
<a href="https://pr-preview.s3.amazonaws.com/EB-Forks/webidl/pull/825.html" title="Last updated on Dec 12, 2019, 2:04 PM UTC (23ebaa5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/825/513c5ea...EB-Forks:23ebaa5.html" title="Last updated on Dec 12, 2019, 2:04 PM UTC (23ebaa5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12321,8 +12321,9 @@ If the [=interface=] has any of the following:
 *   a [=setlike declaration=]
 
 then a <code class="idl">forEach</code> data property must exist with attributes
-{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
-and whose value is a [=function object=].
+{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: |E|, \[[Configurable]]: <emu-val>true</emu-val> }
+and whose value is a [=function object=], where |E| is <emu-val>false</emu-val> if the interface
+is declared with [{{NonEnumerableMethods}}] and <emu-val>true</emu-val> otherwise.
 
 The location of the property is determined as follows:
 
@@ -12418,8 +12419,9 @@ property is the String value "<code>forEach</code>".
 
 If the [=interface=] has an [=iterable declaration=] or an [=asynchronously iterable declaration=],
 then an <code class="idl">entries</code> data property must exist with attributes
-{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
-and whose value is a [=function object=].
+{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: |E|, \[[Configurable]]: <emu-val>true</emu-val> }
+and whose value is a [=function object=], where |E| is <emu-val>false</emu-val> if the interface
+is declared with [{{NonEnumerableMethods}}] and <emu-val>true</emu-val> otherwise.
 
 The location of the property is determined as follows:
 
@@ -12441,8 +12443,9 @@ the value of the {{@@asyncIterator}} property.
 
 If the [=interface=] has an [=iterable declaration=] or an [=asynchronously iterable declaration=],
 then a <code class="idl">keys</code> data property must exist with attributes
-{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
-and whose value is a [=function object=].
+{ \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: |E|, \[[Configurable]]: <emu-val>true</emu-val> }
+and whose value is a [=function object=], where |E| is <emu-val>false</emu-val> if the interface
+is declared with [{{NonEnumerableMethods}}] and <emu-val>true</emu-val> otherwise.
 
 The location of the property is determined as follows:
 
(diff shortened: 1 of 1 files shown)


bzbarsky:
A few questions:

1) This only applies to operations, not attributes?  We want attributes to keep being enumerable?
2) It looks like the actual processing model changes needed to make these extended attributes do something are missing.  In particular https://heycam.github.io/webidl/#define-the-operations presumably needs to be modified to not always do `[[Enumerable]]: true` and instead compute that value on a per-operation basis based on these extended attributes.

ExE-Boss:
> 1. This only applies to operations, not attributes? We want attributes to keep being enumerable? What about constants?

This only affects operations, everything else is unaffected.

> 2. It looks like the actual processing model changes needed to make these extended attributes do something are missing.

You’re right, I missed that.

bzbarsky:
More questions:

A few questions:

3) What happens if a mixin is mixed into interfaces with different enumerability behavior?  Presumably the processing model changes will pin this down.
3) If an operation is inside a mixin that is mixed into both interfaces that are `[NonEnumerableMethods]` and ones that are not, can that operation be `[LegacyEnumerableMethod]`?  This is entirely unclear, given the language about how that attr most not be used outside of `[NonEnumerableMethods]` interfaces.
3) What should happen with stringifiers and their enumerability?  That's not covered by <https://heycam.github.io/webidl/#define-the-operations > but specified separately in <https://heycam.github.io/webidl/#es-stringifier>.
3) Similar for the iterable forEach stuff defined by <https://heycam.github.io/webidl/#es-forEach>.  And other iterable-induced methods.
3) Similar for iterator prototype `.next` methods.  This is complicated by the fact that there is no explicit syntactic interface declaration for the iterator stuff.

ExE-Boss:
> 3. What happens if a mixin is mixed into interfaces with different enumerability behavior?  Presumably the processing model changes will pin this down.
> 4. If an operation is inside a mixin that is mixed into both interfaces that are `[NonEnumerableMethods]` and ones that are not, can that operation be `[LegacyEnumerableMethod]`?  This is entirely unclear, given the language about how that attr most not be used outside of `[NonEnumerableMethods]` interfaces.

I would expect that if `[NonEnumerableMethods]` is defined on the `interface mixin`, then it would only affect methods defined on that mixin, similarly methods transcluded from mixins `[NonEnumerableMethods]` would be kept enumerable, but I don't really know how all the moving parts of WebIDL interact with each other, so I don’t know which should be the right behaviour.

> 5. What should happen with stringifiers and their enumerability?

This should probably default to whether or not `[NonEnumerableMethods]` is specified on the interface.

> 6. Similar for the iterable forEach stuff defined by https://heycam.github.io/webidl/#es-forEach. And other iterable-induced methods.
> 7. Similar for iterator prototype `.next` methods. This is complicated by the fact that there is no explicit syntactic interface declaration for the iterator stuff.

¯\\\_(ツ)\_/¯

domenic:
What is the motivation for this feature? In particular, what specs want to take on a different behavior from the rest of the web platform by using this?

ExE-Boss:
> What is the motivation for this feature?

To&nbsp;make it&nbsp;possible for&nbsp;specifications using&nbsp;**WebIDL** to&nbsp;align&nbsp;more&nbsp;closely with&nbsp;how&nbsp;**ECMAScript** platform&nbsp;objects&nbsp;behave.

> In particular, what specs want to take on a different behavior from the rest of the web platform by using this?

**ECMAScript** and&nbsp;**Streams** come&nbsp;to&nbsp;mind (you&nbsp;even&nbsp;opened&nbsp;<https://github.com/whatwg/streams/issues/963>)

It’s&nbsp;also&nbsp;probable that&nbsp;**WebAssembly** might&nbsp;want to&nbsp;use&nbsp;this.

domenic:
> ECMAScript

I'd like to see some evidence of TC39 wanting to use Web IDL for ECMAScript before merging this

> Streams

We will make all methods enumerable

> WebAssembly

I don't see why they would change away from their current approach of aligning with all other web APIs.

ExE-Boss:
> > WebAssembly
> 
> I don't see why they would change away from their current approach of aligning with all other web APIs.

**WebAssembly**&nbsp;intends to&nbsp;be&nbsp;more of&nbsp;an&nbsp;**ECMAScript** platform&nbsp;feature, so&nbsp;I&nbsp;would&nbsp;expect its&nbsp;built‑ins&nbsp;to&nbsp;match **ECMAScript**&nbsp;conventions.

I’ve&nbsp;opened <https://github.com/WebAssembly/spec/issues/1110> to&nbsp;ask about&nbsp;this.

Ms2ger:
Please explain the difference between this and #719.

ExE-Boss:
> Please explain the difference between this and #719.

This&nbsp;also&nbsp;provides the&nbsp;`[LegacyEnumerableMethod]` extended&nbsp;attribute and&nbsp;defines how&nbsp;`stringifier`s and&nbsp;other dynamically&nbsp;created&nbsp;methods are&nbsp;affected.

ExE-Boss:
@domenic
> > Streams
> 
> We will make all methods enumerable

Where was&nbsp;that decided? Because&nbsp;I&nbsp;think enumerable&nbsp;methods are&nbsp;bad.

domenic:
I think they're good. (I don't anticipate this being a productive discussion, so don't plan to engage further.)

ExE-Boss:
@domenic That&nbsp;doesn’t&nbsp;answer my&nbsp;question&nbsp;about **_where_**&nbsp;it&nbsp;was&nbsp;decided to&nbsp;change&nbsp;methods in&nbsp;the&nbsp;**Streams&nbsp;API** to&nbsp;be&nbsp;enumerable.

ExE-Boss:
This&nbsp;should&nbsp;maybe have&nbsp;[the&nbsp;<kbd>jsidl</kbd>&nbsp;label](https://github.com/heycam/webidl/labels/jsidl).

bzbarsky:
@petervanderbeken @EdgarChen

domenic:
Let's close this until we have a concrete consumer.

Issue #827, opened 2019-12-12
Type association restrictions for extended attributes break with unions


bzbarsky:
Consider IDL that says:
```
  void foo([AllowShared] ArrayBufferView);
```
Per spec as written right now, this is invalid, because https://heycam.github.io/webidl/#AllowShared says:

> A type that is not a buffer source type must not be associated with the [AllowShared] extended attribute.

and `ArrayBufferView` is in fact not a buffer source type.   The rules in https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with propate the extended attribute to all the things in the union (ignoring for the moment the typedef complications), but don't remove the currently-invalid association with the outer union type...

We should really fix both this and https://github.com/heycam/webidl/issues/670, presumably, but it's not clear to me what the goals of these restrictions really were and hence how they should be loosened while preserving those goals.

@domenic @annevk

annevk:
Would changing it to

> A type that is not a union or a buffer source type must not be associated with the [AllowShared] extended attribute.

work in that we don't really care about unions as type other than using them as a grouping mechanism?

That would forbid `[AllowShared] (ArrayBuffer or DOMString)`, but that seems okay?

bzbarsky:
Hmm.  That would be one option.  The other would be to:

1) Define a concept of "type A that may be type B" which would then be defined to do sane things for unions and nullables.
2) Allow `[AllowShared]` on any type that may be a buffer source type.

That's equivalent to the proposal in https://github.com/heycam/webidl/issues/827#issuecomment-565102739 as far as observable behavior for `[AllowShared] ArrayBufferView` and also allows `[AllowShared] ArrayBufferView?`, though we would need some changes to https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with to propagate through nullables.  It seems like we should do that, since fundamentally `?` is pretty much like a union type but with different syntax...

If we wanted to also allow `[AllowShared] (ArrayBuffer or DOMString)` we could fix union propagation to only propagate if the extended attr applies to the inner type, right?  I don't really see a reason to forbid it...

bzbarsky:
In particular, do we want to have `[AllowShared]` on things like the arg to XHR `send()`?

domenic:
> In particular, do we want to have [AllowShared] on things like the arg to XHR send()?

I don't think we do? Instead you should annotate the buffer-source types specifically, i.e. it should change to

```webidl
[...]

PR #841, opened 2020-02-10
Added a definition of {{StringContext}} extended attribute.

koto:
WIP: This extended attribute can be defined on `DOMString` and `USVString`.

This is to hook up the Trusted Types validation during the ES->IDL type
conversion to avoid funky issues with its default policy.

See https://github.com/w3c/webappsec-trusted-types/issues/248,
https://github.com/w3c/webappsec-trusted-types/issues/176

The actual logic of the string validation calls into HTML, as will be something similar to https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6562,7 +6562,8 @@ annotate are called <dfn export for="annotated types" lt="inner type">inner type
 The following extended attributes are <dfn for="extended attributes">applicable to types</dfn>:
 [{{AllowShared}}],
 [{{Clamp}}],
-[{{EnforceRange}}], and
+[{{EnforceRange}}],
+[{{StringContext}}] and
 [{{TreatNullAs}}].
 
 <div algorithm>
@@ -7622,6 +7623,13 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
 
     An ECMAScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{DOMString}} value by running the following algorithm:
+
+    1.  If the conversion is to an IDL type [=extended attribute associated with|associated with=] the
+        [{{StringContext}}] extended attribute, then set |V| to the result of [=validate the string in context=], passing
+        [=this=], |V|, the {{StringContext}} extended attribute [=identifier=], and the [=identifier=]
+        of the [{{StringContext}}] extended attribute [=related construct=].
+
+        Note: That algorithm may [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
 
     1.  If |V| is <emu-val>null</emu-val> and the conversion is to an IDL type
         [=extended attribute associated with|associated with=] the [{{TreatNullAs}}] extended
@@ -10496,6 +10504,40 @@ that does specify [{{SecureContext}}].
     </pre>
 </div>
 
+<h4 id="StringContext" extended-attribute lt="StringContext">[StringContext]</h4>
+
+If the [{{StringContext}}] [=extended attribute=] appears on {{DOMString}} or {{USVString}}, it
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>html</code>", "<code>script-url</code>" and "<code>script</code>".
+
+A type annotated with the [{{StringContext}}] extended attribute must not appear in a construct
+that is not a [=regular attribute=] or a [=regular operation=]. A type annotated with the [{{StringContext}}]
+extended attribute must not appear in a [=read only=] attribute. The construct that the type annotated with
+the [{{StringContext}}] extended attribute appears in is its <dfn>related construct</dfn>.
+
+A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
+the [{{StringContext}}] extended attribute.
+
+See the rules for converting ECMAScript values to the IDL types in [[#es-DOMString]] and [[#es-USVString]]
+for the specific requirements that the use of [{{StringContext}}] entails.
+
+<div class="example">
+
+    In the following [=IDL fragment=],
+    a [=variadic=] [=operation=] is declared
+    that uses the [{{StringContext}}] [=extended attribute=]
+    on all its arguments:
+
+    <pre highlight="webidl">
+        interface Document {
+          void write([StringContext=html] DOMString... text);
+        };
+    </pre>
+</div>
+
 
 <h4 id="TreatNonObjectAsNull" extended-attribute lt="TreatNonObjectAsNull">[TreatNonObjectAsNull]</h4>
 
@@ -10825,6 +10867,21 @@ allowed.  The security check takes the following three inputs:
     getter or setter function of an IDL attribute).
 
 Note: The HTML Standard defines how a security check is performed. [[!HTML]]
+
+Certain algorithms in [[#es-type-mapping]] are defined to
+<dfn id="dfn-validate-the-string-in-context" export>validate the string in context</dfn> on a given
+value. This check is used to determine whether a given value
+is appropriate for its {{StringContext}}. This validation takes the following four inputs:
+
+1.  the [=platform object=] on
+    which the operation invocation or attribute access is being done,
+1.  the value to validate,
+1.  the {{StringContext}} [=identifier=], and
+1.  the [=identifier=] of the operation or attribute.
+
+The algorithm returns an ECMAScript String value, or [=ECMAScript/throws=] a {{ECMAScript/TypeError}}.
+
+Note: The HTML Standard defines how the validation is performed. [[!HTML]]
 
 
 <h3 id="es-overloads">Overload resolution algorithm</h3>


domenic:
This overall makes sense to me. I'd like to get @bzbarsky's take on this architecture. In particular, defining the extended attribute here, and having the "validate the string in context" be in HTML, is a clever split; it echoes the "perform a security check" split we already have. But there are other choices., such as defining [StringContext] entirely in HTML or in the Trusted Types spec, and I'd love further opinions on which split is best.

bzbarsky:
I think I'd like to see more multi-vendor buy-in on the trusted types spec before we add hooks to IDL for it, honestly...

bzbarsky:
That said, the basic principle of defining the extended attribute and where it fits in the processing model in the IDL spec makes sense, as does having the actual checks live in HTML or Trusted Types.  The other options sort of involve either monkeypatching IDL or doing the check on the post-IDL level.

I have not carefully looked at whether the actual arguments being passed to the checking algorithm make sense in general.

bzbarsky:
One thing I would like to understand, though: why is "check on the post-IDL level" not considered desirable here?

mikewest:
After some conversations with @annevk, we realized that the existing model of running JavaScript in the middle of a DOM operation was a bad idea; https://github.com/w3c/webappsec-trusted-types/issues/248 walks through the thought process that landed us at this solution as an alternative that should help us ensure that we only run script _before_, and not during, DOM tree mutation. https://github.com/w3c/webappsec-trusted-types/issues/248#issuecomment-576373688 is a summary of where we ended up.

bzbarsky:
> we realized that the existing model of running JavaScript in the middle of a DOM operation was a bad idea

Sure.  The question is why it can't run after IDL argument work but before anything else (just like CEReactions run after everything else on methods that have them).

koto:
IDL level is a single place, where the check may be applied effectively via extended attribute (alternatively, multiple places in DOM and HTML would have to be modified separately). Additionally, it solves the [issue with the default policy](https://github.com/w3c/webappsec-trusted-types/issues/248#issuecomment-563365620) - in short, it's advisable to run TT operations before DOM gets to mutate the tree.

index.bs
```diff
@@ -10496,6 +10504,40 @@ that does specify [{{SecureContext}}].
...
+
+If the [{{StringContext}}] [=extended attribute=] appears on {{DOMString}} or {{USVString}}, it
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>html</code>", "<code>script-url</code>" and "<code>script</code>".
+
+A type annotated with the [{{StringContext}}] extended attribute must not appear in a construct
+that is not a [=regular attribute=] or a [=regular operation=]. A type annotated with the [{{StringContext}}]
```
bzbarsky: OK, reading this more carefully, I don't understand what this means.  Is `void foo(sequence<[StringContext=html] DOMString>);` ok?  Why or why not?  What does "construct" even mean here?

What about a record?  A dictionary argument?  A union?  Is a dictionary a "separate construct" while a record is not, just because a dictionary has a separate IDL definition while a record or union is inferred?  Or are they both "separate constructs"?  But my impression was that you wanted to allow this new thing inside unions, which are just as "separate construct" as records.  Or is that out of date and the intent is to limit this thing to types of attributes and operation arguments?  If so, that seems like the right thing to say explicitly.

koto:
> Sure. The question is why it can't run after IDL argument work but before anything else (just like CEReactions run after everything else on methods that have them).

TT perform data validation, running them after the data was already processed (e.g. inserted into the DOM) in a separate microtask like CEReactions is too late to surface the violation to the developers.

bzbarsky:
> where the check may be applied effectively via extended attribute.  alternatively, multiple places in DOM and HTML would have to be modified separately

This was not the case for [CEReactions], right?  Why is this case fundamentally different?

> in short, it's advisable to run TT operations before DOM gets to mutate the tree.

As I said above, that does not require doing it as part of IDL argument conversions.

> running them after the data was already processed (e.g. inserted into the DOM) in a separate microtask like CEReactions

I didn't say to use the same semantics as CEReactions, which is obviously a daft thing to suggest, so I'm not sure why we're discussing that strawman. I asked why we can't have the semantics of "run before any of the operation steps".  Sort of like how all CanvasRenderingContext2D methods have a first step of "validate all the doubles".  Those are written out by hand in the spec right now, but they could be done via extended attribute (and are implemented that way in Gecko, fwiw).

index.bs
koto: It's the latter. I'll fix the language.

bzbarsky: OK.  So no more unions involved; this mechanism is replacing that?

If so, then the right place to put this might be not inside generic DOMString/USVString conversion but specifically in operation argument and attribute setter type conversions.  That would avoid the ambient "related construct" bit and make it much clearer when these things actually have an effect.

bzbarsky: Though again, at that point it could in theory also just happen at the top of the operation steps instead of intertwined with the other argument conversions, in which case it would not need any IDL hooks at all.  It's not clear to me which is easier or harder to reason about and implement in practice, or if there's a practical difference between them. 

All this assuming I now understand the proposal correctly...

koto: > OK. So no more unions involved; this mechanism is replacing that?

Yes, the extended attribute is to avoid using type unions in most of the places.

> If so, then the right place to put this might be [...] in operation argument and attribute setter type conversions. 
So something like https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute ? That's the previous attempt.

koto:
> I asked why we can't have the semantics of "run before any of the operation steps".

That is a possible alternative - in fact this is what we used [before](https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute). We could have that language in the other specs instead, possibly added via an extended attribute. The move to IDL type mapping is to avoid having type unions and branching on the types in each operation. Additional small advantage to changing this in IDL type conversion is that it aligns with what we implement (and what we think is the best way to implement TT).

index.bs
koto: I updated the PR such that "the intent is to limit this thing to types of attributes and operation arguments" is explicit.

bzbarsky:
> We could have that language in the other specs instead, possibly added via an extended attribute.

OK.  I'd like to understand what the problems are with that.

> The move to IDL type mapping is to avoid having type unions and branching on the types in each operation. 

I'm not quite following this, sorry.  Can you explain further what you mean here?

> and what we think is the best way to implement TT

That _really_ depends on how type conversion is implemented and how generic or special-cased it is.  For example, the type conversion code may well not have the attribute/operation involved available to it (as trusted types requires), because they are not needed in that code right now.

That said, I would really like to avoid diving into implementation specifics, which differ greatly across engines, I suspect, and sort out the conceptual right place to have this live.  Now that I think I understand the setup (though I thought that before too!), it seems like the possible places to put this are:

1) Inside DOMString/USVString conversions, and thread through the information those need, like which operation/attribute the conversion is for.  This could happen _before_ or _after_ conversion to an IDL value.  Looks like this PR does it "before".
2) At the places where args to operations or setters are converted.  That would correspond to https://heycam.github.io/webidl/#es-overloads step 11.5 and step 15.5, and https://heycam.github.io/webidl/#dfn-attribute-setter step 6.
3) After conversion but before invoking the spec-described steps.  That would correspond to https://heycam.github.io/webidl/#dfn-create-operation-function step 2.1.8 (well, after 2.1.5 and before 2.1.8), and the https://heycam.github.io/webidl/#dfn-attribute-setter step 7 (or rather beween steps 6 and 7).
4) At the front of the spec-described steps, whether via extended attribute or manually.  Manually seems undesirable.

Now I have my own opinions of how these various options compare, but I'd love a summary of what you view as the tradeoffs and what led to choosing option 1 with the "before" behavior, especially if this is already written down somewhere.  In terms of author-observable behavior, there are three different behaviors here: "option 1 with 'before'" is one, "option 1 with 'after'" and "option 2" are another, I think (in that they behave identically), and "options 3"/"option 4", which also behave identically.

Note that I did read through the two issues listed in the original PR, but there is a _lot_ of back-and-forth in there and it wasn't clear to me exactly why we landed on this outcome.

index.bs
bzbarsky: > So something like

In terms of my options in https://github.com/heycam/webidl/pull/841#issuecomment-586354298 this is option 3 or 4, I think, right?  I was talking about option 2 here.

bzbarsky:
Actually, there _is_ an important behavior difference between options 3 and 4 that I just thought of.  When other specs invoke the setter or operation directly, not via the JS reflection, option 4 will lead to the values they pass being sanitized while option 3 (or 2 or 1) will not, right?  I don't know how common such direct-invoking is for the cases we care about here, but if it happens it would definitely be observable to script via the callbacks, not to mention possibly affecting the sanitization behavior.

bzbarsky:
Just to clarify, the intent is that this is valid:
```
  void foo(([StringContext] DOMString or Node) arg);
```
right?  But probably not:
```
  void foo([StringContext] (DOMString or Node) arg);
```
per the discussion in https://github.com/heycam/webidl/issues/827 ?

And in particular, in this case the sanitizer should be invoked if the passed-in thing is not a Node, but _not_ invoked if it is a Node?

koto:
> > The move to IDL type mapping is to avoid having type unions and branching on the types in each operation.

> I'm not quite following this, sorry. Can you explain further what you mean here?

To my understading, previous version required us to use: 

```
partial interface mixin HTMLIFrameElement : HTMLElement {
  [TrustedTypes=TrustedHTML] attribute (DOMString or TrustedHTML) srcdoc;
};
```
and then the setter for `srcdoc` would have to branch on the value type. Perhaps I was misunderstanding to which extent the extended attributes are allowed to interfere with the value type, and the type union there is actually unneccessary, and the extended attribute can, in the end, cause the setter to abort early or accept a DOMString.

> Now that I think I understand the setup (though I thought that before too!), it seems like the possible places to put this are:
> 
> 1. Inside DOMString/USVString conversions, and thread through the information those need, like which operation/attribute the conversion is for.  This could happen _before_ or _after_ conversion to an IDL value.  Looks like this PR does it "before".

Correct. IIUC doing this "after" https://heycam.github.io/webidl/#es-DOMString step 2 is undesirable, as we'd lose the JS type information. The logic for the sanitization of values is roughly:

```
if TTNotEnforced:
  return v.toString()

if isCorrectTrustedType(v, context):
  return v.toString();
elseif realm.defaultPolicy:
  v = realm.defaultPolicy.sanitize(v, context) // this might throw.
return v.toString() 
```

so the JS type information needs to be available at the time of the checks (in HTML or TT or CSP).

> 2. At the places where args to operations or setters are converted.  That would correspond to https://heycam.github.io/webidl/#es-overloads step 11.5 and step 15.5, and https://heycam.github.io/webidl/#dfn-attribute-setter step 6.
> 3. After conversion but before invoking the spec-described steps.  That would correspond to https://heycam.github.io/webidl/#dfn-create-operation-function step 2.1.8 (well, after 2.1.5 and before 2.1.8), and the https://heycam.github.io/webidl/#dfn-attribute-setter step 7 (or rather beween steps 6 and 7).
> 4. At the front of the spec-described steps, whether via extended attribute or manually.  Manually seems undesirable.

This seems roughly what the previous iteration did. The downside, apart from the type branching required (we need to preserve the JS type for the sanitization check), was exactly what you point out: 

> When other specs invoke the setter or operation directly, not via the JS reflection, option 4 will lead to the values they pass being sanitized


The fact that the invoking the setter from outside the JS bypasses the sanitizing is actually desirable. For example, that allows the HTML parser algorithm to create the DOM elements without needing to carve out the exception for it (we are guarding e.g. innerHTML setter, but when the string actually gets allowed to be parsed - or is parsed after navigation - arbitrary nodes with arbitrary attributes are allowed). In other words, Trusted Type try to prevent the JS code from interacting with risky APIs, not the browser algorithms.

> Now I have my own opinions of how these various options compare, but I'd love a summary of what you view as the tradeoffs and what led to choosing option 1 with the "before" behavior, especially if this is already written down somewhere. In terms of author-observable behavior, there are three different behaviors here: "option 1 with 'before'" is one, "option 1 with 'after'" and "option 2" are another, I think (in that they behave identically), and "options 3"/"option 4", which also behave identically.

The intended behavior is that the sanitization logic gets access to the JS type used, for the value to skip the sanitization ste). The simplest way I see it to hook that as early as possible, and convert to the IDL type later. 

 - Option 1 with "before" allows us to do that by reusing the DOMString / USVString type. 
 - Option 1 with "after" loses the type, IIUC.
 - Option 2 actually looks OK too. What is exactly the author-observable difference to option 1 + before? 
- Option 3 loses the type, similarly option 4. We can have the type information if we use a type union in the IDL definition of the attribute / operation. 

I don't think there's something fundamentally wrong with using type unions in IDL definitions for our purpose - the annotated type route was taken as a preferred option for simplicity (one place to change things, and upstream specs remain to only see DOMStrings). Perhaps @domenic can weigh in here.

> Note that I did read through the two issues listed in the original PR, but there is a _lot_ of back-and-forth in there and it wasn't clear to me exactly why we landed on this outcome.

Thanks a lot, and sorry for looping you in that late into the discussion. Some of the them were triggered internally, so it's indeed not clear from GitHub alone why certain approaches were taken. I also am definitely not an expert in IDL, and I learn as I go. Thank you for taking your time to review this.

> Just to clarify [...] in this case [`void foo([StringContext] (DOMString or Node) arg)`] the sanitizer should be invoked if the passed-in thing is not a Node, but not invoked if it is a Node?

Good point. I didn't consider this case, but indeed it would be fine for the sanitizer to be skipped *if* the extra type was included in the IDL. In the case of `[StringContext] DOMString arg` a Node should undergo JS->IDL `DOMString` conversion, with sanitization invoked.

bzbarsky:
Thank you, that is very helpful.

So the key is that there is still a concept of "already validated" on the caller side (which was not clear to me from the above) and that we want to only do validation if the thing we have is not already validated.  OK, that makes things much clearer.

> The logic for the sanitization of values is roughly

OK, I see.  One thing that's not entirely clear to me is whether the default policy case should be passed the initial value, or a stringification of the original value, or a stringification with the USVString fixups applied, or whether it doesn't matter much.  But ok, now I understand why you want to at least invoke this algorithm before stringification of the original value.

> In other words, Trusted Type try to prevent the JS code from interacting with risky APIs, not the browser algorithms

OK.  That's a reasonable tradeoff, as long as you make sure that all the relevant boundary bits are annotated.

> Option 2 actually looks OK too. What is exactly the author-observable difference to option 1 + before?

I had been thinking of option 2 as a post-conversion step, so coming at a later point than "option 1 + before".  It looks like you're asking about doing it as a pre-conversion step.

The main observable difference, I guess, is what happens with a union type.  "Option 2" would invoke the sanitization stuff no matter what the value is, and while "option 1 + before" will only invoke it if we determine that we're taking the DOMString branch of the union, right?  Presumably we want the latter behavior, in general?

bzbarsky:
Er, clicked submit too early.

> the annotated type route was taken as a preferred option for simplicity 

Yes, this makes sense.

> and sorry for looping you in that late into the discussion

No need to apologize.  I don't have the bandwidth to be in all discussions all the time, and certainly didn't for this one...

I think I finally understand what the proposed setup here is.  From an IDL and implementation point of view (now with my Mozilla hat on for the latter), the main complexity is threading through the information about what operation/setter is involved to the conversion site.  If we didn't need to worry about unions we could do the check as in "option 2" _before_ doing the arg conversion, and that's _fairly_ simple to spec and implement, depending on what information about the operation is needed.  Specifically, it doesn't allow distinguishing overloads, but neither does any variant of option 1; the only way to do that is options 3 or 4.

But if we want to have the semantics that `void foo(([StringContext=html] DOMString or Node) arg);` only invokes sanitization for non-Nodes, then it seems like the approach in this PR is the only really viable option, and we need to deal with threading the information about what operation/setter is involved through to the specific string type conversions...

koto:
> OK, that makes things much clearer.

Yes, sorry about not being explicit about that early on. I don't think we have a precedent for such behavior in the web platform. It's familiar and intuitive to our team, as we're working on the thing, but understandably, it's a bit of a surprising twist.

> 
> So the key is that there is still a concept of "already validated" on the caller side (which was not clear to me from the above) and that we want to only do validation if the thing we have is not already validated. OK, that makes things much clearer.
> 
> > The logic for the sanitization of values is roughly
> 
> OK, I see. One thing that's not entirely clear to me is whether the default policy case should be passed the initial value, or a stringification of the original value, or a stringification with the USVString fixups applied, or whether it doesn't matter much. 

It would be preferable if we could get the initial value in the default policy. The current spec draft is worded differently, but we're thinking of changing that, as that allows the authors to allow a selection of "their" types to pass through. It's not a strong requirement though, just a recent idea we had. 

> > In other words, Trusted Type try to prevent the JS code from interacting with risky APIs, not the browser algorithms
> 
> OK. That's a reasonable tradeoff, as long as you make sure that all the relevant boundary bits are annotated.

👍
 
> > Option 2 actually looks OK too. What is exactly the author-observable difference to option 1 + before?
> 
> I had been thinking of option 2 as a post-conversion step, so coming at a later point than "option 1 + before". It looks like you're asking about doing it as a pre-conversion step.
> 
> The main observable difference, I guess, is what happens with a union type. "Option 2" would invoke the sanitization stuff no matter what the value is, and while "option 1 + before" will only invoke it if we determine that we're taking the DOMString branch of the union, right? Presumably we want the latter behavior, in general?

Yes, the latter would be preferable.

> the main complexity is threading through the information about what operation/setter is involved to the conversion site

What we'd be using in the sanitizing callout is:
 - the [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) realm of `this` (to check the CSP list),
 - the IDL identifier name of the attribute / operation (string of it would be enough) - this is to put details into a CSP violation object,
 - some identifier of the type to convert to (a string is fine),


> But if we want to have the semantics that `void foo(([StringContext=html] DOMString or Node)` arg); only invokes sanitization for non-Nodes

If that helps, I only see a single case for us where that would be relevant: https://w3c.github.io/webappsec-trusted-types/dist/spec/#enforcement-in-timer-functions, but we can special case it in HTML I think.

bzbarsky:
> the relevant global object realm of this (to check the CSP list),

OK.  Back with my implementor hat on, this is currently not readily available in the guts of type conversions in Gecko, fwiw...  With my spec hat on, it's sort of ambient state, kinda, if you know your type conversion is happening in the right places....

The CSP check doesn't need to know what method is involved?

> the IDL identifier name of the attribute / operation (string of it would be enough) - this is to put details into a CSP violation object,

Again, not really something that's available in the guts of type conversions right now, either in the spec or in (some) impls.  This one would be much more clearly available in the "option 2, before" scenario, of course, since that has access to the operation (in the spec).

Again, threading it through is possible, but needs to be done; definitely on the spec side.

> some identifier of the type to convert to (a string is fine),

This is the value of the extended attribute, so is conceptually available wherever we need it, assuming we can tell that we have the extended attribute at all at that point.

bzbarsky:
> The CSP check doesn't need to know what method is involved?

Ah, I guess it just needs to know "the type to convert to".

koto:
> > The CSP check doesn't need to know what method is involved?
> 
> Ah, I guess it just needs to know "the type to convert to".

the method as well (a name of it). Technically, we call the default policy function, the second argument to which is, say, "Element.innerHTML", or "Document.write" (see [Example 14](https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute)).

koto:
So, summarizing the options the integration as outline is OK given the constraints and the intention of the API, but there's a slight preference to do option 2 + before, at least for the simplicity for the implementors (the operation data is already available there):

> At the places where args to operations or setters are converted. That would correspond to https://heycam.github.io/webidl/#es-overloads step 11.5 and step 15.5, and https://heycam.github.io/webidl/#dfn-attribute-setter step 6.

Correct?

I have a folowup question - would it be preferable to merge with IDL now-ish, or should we monkey-patch IDL in TT? I don't know what the usual convention is for new APIs.

yuki3:
I have a question.  `[StringContext=XXX] DOMString value` seems expected to behave like the following.

https://github.com/heycam/webidl/pull/841#issuecomment-586413864
```
if TTNotEnforced:
  return v.toString()

if isCorrectTrustedType(v, context):
  return v.toString();
elseif realm.defaultPolicy:
  v = realm.defaultPolicy.sanitize(v, context) // this might throw.
return v.toString()
```

where `isCorrectTrustedType(v, context)` requires knowledge about `TrustedHTML`, `TrustedScript`, and `TrustedScriptURL` IDL interfaces.  Is this expected in Web IDL?  Or, is this okay if it's monkey-patched in other specs?

I'm now just wondering if this is cleaner or clearer than `(TrustedHTML or [StringContext=TrustedHTML] DOMString)` or not.  It seems to me like we're doing another type resolution during ES-to-IDL value conversion although overload resolution (= type resolution) is already done at this point.

If this doesn't matter much, please move on.

annevk:
@yuki3 that does seem quite a bit better to me. That would also more clearly allow for new endpoints that only take a type.

koto:
@annevk That is quite similar to how it is defined currently: We typedef [HTMLString](https://w3c.github.io/webappsec-trusted-types/dist/spec/#typedefdef-htmlstring) union and [set the attribute on the union](https://w3c.github.io/webappsec-trusted-types/dist/spec/#extensions-to-the-document-interface). We could set it for the `DOMString` only, like @yuki3 proposes, that's a minor detail I think. 

The real issue is that, to my understanding, the original design as outlined above caused issues with the default policy for DOM (https://github.com/w3c/webappsec-trusted-types/issues/248) - which triggered the redesign and this PR. It seems to me like now we're going back to where we were before. Does that not cause default policy issues in the end? The intention is for most spec algos to continue dealing with strings only, and for the sanitization happening before they run. If we can achieve that by rephrasing https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute, and without modifying WebIDL, I'm sold.

domenic:
I don't think we can achieve that without introducing a modification to the string type. A union type will always cause specs to branch on both forks of the union.

I think the better resolution to @yuki3's concern is to move [StringContext] to another spec, if he thinks it is bad for Web IDL to have knowledge of it. But I stand by my earlier comment that the division of putting [StringContext] here and putting the "validate the string in context" in another spec is a reasonable division.

koto:
> I don't think we can achieve that without introducing a modification to the string type. A union type will always cause specs to branch on both forks of the union.

The actual operation for the TrustedXYZ though is just stringification. DOMString is already a validated value, if stars are aligned correctly. So the modification to the algoritm could just be to stringify an argument (perhaps that logic could also be provided in an extended attribute definition).

yuki3:
> The intention is for most spec algos to continue dealing with strings only,

I don't know how much important this is, but given it's important, I'm fine with the current plan.  (I understand that it's itchy to add a line "convert the argument XXX to a DOMString with using a certain process YYY" into each spec, but I don't know how much important it is.  Anyway, we need to update each spec to change the argument type, I think.)

annevk:
I feel like I'm missing something, but would something like `[TrustedTypeCallbacks] TrustedHTML or [TrustedTypeCallbacks] DOMString` not preserve doing callbacks at IDL time while also not requiring changes to strings?

domenic:
> I feel like I'm missing something, but would something like [TrustedTypeCallbacks] TrustedHTML or [TrustedTypeCallbacks] DOMString not preserve doing callbacks at IDL time while also not requiring changes to strings?

The main problem with this solution is that it's a union type. This means that, like all union types, specification writers need to branch on which type they recieved. So specification writers need to write something like:

1. If _arg_'s type is `[TrustedTypeCallbacks] TrustedHTML`, then extract the string (somehow) and use it.
1. If _arg_'s type is `[TrustedTypeCallbacks] DOMString`, then use the string directly.

The desired experience, by using `[StringContext=html] DOMString`, is to remove this branching, and let spec authors say

1. Use _arg_ as a string.

koto:
I agree with @domenic here, we're avoiding type unions to avoid branching in other spec algos, and centralize the behavior in WebIDL (to define the extended attribute and callouts) and HTML (to implement the TT-specific check). 

The current Blink implementation follows the PR above ("option 1 + before" from https://github.com/heycam/webidl/pull/841#issuecomment-586354298), as we need the original type information when checking. In our impl piping through the context ended up being quite simple, and we get the benefits of removing many custom changes at the actual setters (the latter would be possible with other options too, as long as we don't choose the "type unions" approach). I think the decision process for choosing option 1 is neatly summarized in https://github.com/heycam/webidl/pull/841#issuecomment-586421175.

bzbarsky:
@petervanderbeken @EdgarChen

lukewarlow:
What's the latest status of this change?

index.bs
```diff
@@ -10496,6 +10504,41 @@ that does specify [{{SecureContext}}].
...
 </div>
 
+<h4 id="StringContext" extended-attribute lt="StringContext">[StringContext]</h4>
+
+If the [{{StringContext}}] [=extended attribute=] appears on {{DOMString}} or {{USVString}}, it
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>html</code>", "<code>script-url</code>" and "<code>script</code>".
```
lukewarlow, line 10514: The TrustedTypes spec as it currently is uses the TrustedXXX as the identifiers. It would be good to get clarity on whether that is correct or if this is correct?

index.bs
koto: Resolving, since we chose the option 1 here in the end.

index.bs
```diff
@@ -6353,8 +6353,10 @@ The following extended attributes are <dfn for="extended attributes">applicable
 [{{AllowResizable}}],
 [{{AllowShared}}],
 [{{Clamp}}],
-[{{EnforceRange}}], and
-[{{LegacyNullToEmptyString}}].
+[{{EnforceRange}}],
+[{{LegacyNullToEmptyString}}] and
```
annevk, line 6357: Please bring back the Oxford comma.

index.bs
```diff
@@ -7581,6 +7583,13 @@ value when its bit pattern is interpreted as an unsigned 64-bit integer.
     A JavaScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{DOMString}} value by running the following algorithm:
 
+    1.  If the conversion is to an IDL type [=extended attribute associated with|associated with=] the
+        [{{StringContext}}] extended attribute, then set |V| to the result of [=validate the string in context=], passing
```
annevk, line 7587: validating*

index.bs
```diff
@@ -10259,6 +10268,41 @@ that does specify [{{SecureContext}}].
...
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>html</code>", "<code>script-url</code>" and "<code>script</code>".
+
+[{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
+a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
+extended attribute must not appear in a [=read only=] attribute. The [=regular attribute=] or
+a [=regular operation=] argument that the type annotated with the [{{StringContext}}] extended
```
annevk, line 10283:
```suggestion
[=regular operation=] argument that the type annotated with the [{{StringContext}}] extended
```

index.bs
```diff
@@ -7581,6 +7583,13 @@ value when its bit pattern is interpreted as an unsigned 64-bit integer.
     A JavaScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{DOMString}} value by running the following algorithm:
 
+    1.  If the conversion is to an IDL type [=extended attribute associated with|associated with=] the
+        [{{StringContext}}] extended attribute, then set |V| to the result of [=validate the string in context=], passing
+        [=this=], |V|, the {{StringContext}} extended attribute [=identifier=], and the [=identifier=]
+        of the [{{StringContext}}] extended attribute [=related construct=].
```
annevk, line 7589: This has some extreme magic that I don't really understand.

As far as I know when we do type conversion there is no "this" and there is no access to something like a "related construct".

I'd like to hear what @domenic and @Ms2ger think, but it seems to me you have to patch algorithms such as https://webidl.spec.whatwg.org/#dfn-attribute-setter instead.

Ms2ger: I agree that this doesn't work with the layering as it currently exists.

index.bs
```diff
@@ -11056,6 +11100,21 @@ allowed.  The security check takes the following three inputs:
...
+value. This check is used to determine whether a given value
+is appropriate for its {{StringContext}}. This validation takes the following four inputs:
+
+1.  the [=platform object=] on
+    which the operation invocation or attribute access is being done,
+1.  the value to validate,
+1.  the {{StringContext}} [=identifier=], and
+1.  the [=identifier=] of the operation or attribute.
+
+The algorithm returns an ECMAScript String value, or [=ECMAScript/throws=] a {{ECMAScript/TypeError}}.
```
Ms2ger, line 11114: Note that https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context doesn't necessarily return a string, and if it did, the conversion algorithm above does some unnecessary work.

index.bs
domenic: Yes, you cannot access "this" here. Some call sites don't have a "this".

Putting it in the appropriate call sites instead (e.g., the overload resolution algorithm and attribute setters) seems like the best approach.

mbrodesser-Igalia:
Can the PR preview please be fixed?

CC @koto

index.bs
```diff
@@ -10259,6 +10268,41 @@ that does specify [{{SecureContext}}].
...
+[{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
+a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
+extended attribute must not appear in a [=read only=] attribute. The [=regular attribute=] or
+a [=regular operation=] argument that the type annotated with the [{{StringContext}}] extended
+attribute appears in is its <dfn>related construct</dfn>.
+
+A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
+the [{{StringContext}}] extended attribute.
+
+See the rules for converting ECMAScript values to the IDL types in [[#es-DOMString]] and [[#es-USVString]]
```
lukewarlow, line 10289: These two es-X references are fails in bikeshed which is breaking the preview

index.bs
```diff
@@ -11056,6 +11100,21 @@ allowed.  The security check takes the following three inputs:
 
 Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
+Certain algorithms in [[#es-type-mapping]] are defined to
```
lukewarlow, line 11103: This es-X is likewise a bikeshed fail

PR #848, opened 2020-03-09
Align a bit with Infra

annevk:
This will make IDL adopt https://infra.spec.whatwg.org/#javascript-string-convert rather than have its own algorithm.

Depends on https://github.com/whatwg/infra/pull/292.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/848.html" title="Last updated on Mar 9, 2020, 12:23 PM UTC (b85f284)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/848/225a115...b85f284.html" title="Last updated on Mar 9, 2020, 12:23 PM UTC (b85f284)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6060,43 +6060,6 @@ of specifications using Web IDL might want to obtain a sequence of
 [=Unicode scalar values=] given a particular sequence of
 [=code units=].
 
-<div algorithm>
-
-    The following algorithm defines a way to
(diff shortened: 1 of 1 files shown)


annevk:
It looks like the IDL build setup isn't as strict as the WHATWG build setup. string/convert currently fails to link as it's not yet indexed, but everything suggests there's no issue here.

PR #850, opened 2020-03-11
Meta: use Makefile from deploy.sh

annevk:
Avoids duplication and hopefully keeps strictness consistent.

@tobie @domenic @Ms2ger any reason these are not in sync? I noticed the build script not failing with https://github.com/heycam/webidl/pull/848 while it was failing locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 SHELL=/bin/bash
 
 bs_installed  := $(shell command -v bikeshed 2> /dev/null)
+node_installed := $(shell command -v node 2> /dev/null)
 npm_installed := $(shell command -v npm 2> /dev/null)
 pp_webidl_installed := $(shell npm ls webidl-grammar-post-processor --parseable --depth=0 2> /dev/null)
 
@@ -27,6 +28,11 @@ else
 		rm -f index.html; \
 		exit 22 \
 	);
+endif
+ifdef node_installed
+	node ./check-grammar.js index.html
+else
+	@echo You need node for grammer checking.
 endif
 ifdef pp_webidl_installed
 	npm run pp-webidl -- --input index.html
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,15 @@ index.html : index.bs
 ifdef bs_installed
 	bikeshed spec --die-on=warning index.bs
 else
+ifndef $(TRAVIS)
 	@echo Can\'t find a local version of Bikeshed. To install it, visit:
 	@echo
 	@echo https://github.com/tabatkins/bikeshed/blob/master/docs/install.md
 	@echo
 	@echo Trying to build the spec using the online API at: https://api.csswg.org/bikeshed/
 	@echo This will fail if you are not connected to the network.
+	@echo
+endif
 	@ (HTTP_STATUS=$$(curl https://api.csswg.org/bikeshed/ \
 	                       --output index.html \
 	                       --write-out "%{http_code}" \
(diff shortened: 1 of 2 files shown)


annevk:
Hmm, perhaps `node ./check-grammar.js` should move into the `Makefile` as well as I guess it happens at the wrong time now?

Makefile
```diff
@@ -28,6 +29,11 @@ else
 		exit 22 \
 	);
 endif
+ifdef node_installed
+	node ./check-grammar.js index.html
+else
+	@echo You need node for grammer checking.
```
Ms2ger: One comment, though... Do we want this to not happen silently in automation if something's off with the setup, or should it fail loudly?

annevk:
It seems to work well now, but I guess a second check from @domenic wouldn't hurt.

Makefile
annevk: That's a good point, not sure. Perhaps it's easier to continue to duplicate...

domenic: You could check if one of https://docs.travis-ci.com/user/environment-variables/#default-environment-variables is set. Figuring out how to do that would require figuring out how Makefile syntax works though -_-.

domenic: Similarly that could help avoid CI echoing the whole "Bikeshed is not locally installed" thing above by wrapping that in an if statement.

annevk: @sideshowbarker I suspect this is up your alley?

sideshowbarker: > You could check if one of https://docs.travis-ci.com/user/environment-variables/#default-environment-variables is set. Figuring out how to do that would require figuring out how Makefile syntax works though -_-.
> @sideshowbarker I suspect this is up your alley?

Yeah, I’ll take a look

sideshowbarker: > You could check if one of https://docs.travis-ci.com/user/environment-variables/#default-environment-variables is set. Figuring out how to do that would require figuring out how Makefile syntax works though -_-.

So I see that makefile is already using some `ifdef …`. The makefile syntax for checking environment variables is to just use, e.g. `ifdef $(TRAVIS)` — that’d seem to be the simplest way to identify whether we’re being run in the Travis environment. There’s also `ifndef` for the negation of that, as well as `ifeq` and `ifneq` for any case where we’d want to actually check the value of an environment variable (rather that just checking if it’s defined.)

(I’d push a change for this to the PR branch myself, but I don’t have push perms here.)

sideshowbarker:
```diff
commit 4c9f51e332cb6522f98c34ed1f5f00b2eb62034d
Author: Michael[tm] Smith <mike@w3.org>
Date:   Fri Mar 13 16:20:43 2020 +0900

    Add some 'ifndef $(TRAVIS)'

diff --git a/Makefile b/Makefile
index e310d6a..77542fe 100644
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ index.html : index.bs
 ifdef bs_installed
 	bikeshed spec --die-on=warning index.bs
 else
+ifndef $(TRAVIS)
 	@echo Can\'t find a local version of Bikeshed. To install it, visit:
 	@echo
 	@echo https://github.com/tabatkins/bikeshed/blob/master/docs/install.md
@@ -29,11 +30,14 @@ else
 		exit 22 \
 	);
 endif
+endif
 ifdef node_installed
 	node ./check-grammar.js index.html
 else
+ifndef $(TRAVIS)
 	@echo You need node for grammer checking.
 endif
+endif
 ifdef pp_webidl_installed
 	npm run pp-webidl -- --input index.html
 else ifdef npm_installed
```

PR #856, opened 2020-03-17
Normative: Rename numeric types to match ECMAScript typed arrays

ExE-Boss:
Since there was support for <https://github.com/heycam/webidl/issues/843>, I decided to go ahead with opening a PR for this.

I decided to go with `i*` for signed integers because it’s what browser engines and the **Rust** programming language use and due to **ECMAScript** providing implicit support for 32‑bit signed integers using the binary **OR** operator, but none for unsigned integers:

```js
let numI32 = (num | 0); // Converts `num` to a 32-bit signed integer
```

**ECMAScript** `Array`s also restrict the length to a positive 32‑bit signed integer.

## Depends on:

- [ ] <https://github.com/heycam/webidl/pull/857>

---

Resolves #843

🔗 [**PR Preview**](https://pr-preview.s3.amazonaws.com/EB-Forks/webidl/pull/856.html)

***
<a href="https://pr-preview.s3.amazonaws.com/EB-Forks/webidl/pull/856.html" title="Last updated on Sep 6, 2020, 7:42 AM UTC (46cef88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/856/691b8d2...EB-Forks:46cef88.html" title="Last updated on Sep 6, 2020, 7:42 AM UTC (46cef88)">Diff</a>

Issue #843, opened 2020-02-14
Consider renaming number types to match the equivalent WebAssembly types


ExE-Boss:
Currently, WebIDL&nbsp;uses **C**&nbsp;names for&nbsp;number&nbsp;types.

This&nbsp;doesn’t&nbsp;match either&nbsp;**ECMAScript** or&nbsp;**WebAssembly**, where&nbsp;the&nbsp;former just&nbsp;uses the&nbsp;generic **IEEE&nbsp;754** 64‑bit&nbsp;`number`&nbsp;type, and&nbsp;the&nbsp;latter&nbsp;uses `i*`&nbsp;for&nbsp;signed&nbsp;integers, `u*`&nbsp;for&nbsp;unsigned&nbsp;integers and&nbsp;`f*`&nbsp;for&nbsp;**IEEE&nbsp;754** floating&nbsp;point&nbsp;numbers.

I&nbsp;suggest&nbsp;applying the&nbsp;following&nbsp;renaming:

| Before | After |
| ------: | -----: |
| `byte` | `i8` |
| `octet` | `u8` |
| `short` | `i16` |
| <code>unsigned&nbsp;short</code> | `u16` |
| `long` | `i32` |
| <code>unsigned&nbsp;long</code> | `u32` |
| <code>long&nbsp;long</code> | `i64` |
| <code>unsigned&nbsp;long&nbsp;long</code> | `u64` |
| `float` | `f32` |
| <code>unrestricted&nbsp;float</code> | <code>unrestricted&nbsp;f32</code> |
| `double` | `f64` |
| <code>unrestricted&nbsp;double</code> | <code>unrestricted&nbsp;f64</code> |

bzbarsky:
Do the IDL types actually have the same semantics as the WebASM types?  Because if not, naming them the same thing is a bit dangerous.

I do agree that it would be nice if the size were just explicit.  But whether changing every single web spec for this is worthwhile is less clear.

One option would be to add the sized typedefs explicitly and see whether people actually use them...

ExE-Boss:
Correction: **WebAssembly**&nbsp;uses&nbsp;`s*` for&nbsp;signed&nbsp;integers, `i*`&nbsp;is&nbsp;for&nbsp;uninterpreted&nbsp;integers.

[...]

Issue #33, opened 2014-12-05
Numeric type reform strawperson


domenic:
Inspired by [bug 26901](https://www.w3.org/Bugs/Public/show_bug.cgi?id=26901) and @marcoscaceres's #14, I somehow ended up writing a detailed proposal for reforming the numeric types.

# WebIDL Numeric Type Reform Strawperson

## Motivation

The existing WebIDL numeric types are problematic in a number of ways.

First, they create a misleading parallel with numeric types in other languages, where there is actually a proper numeric type system. This encourages people to use them in situations where they might in other languages, creating non-JavaScript-idiomatic APIs. Most JavaScript APIs will want one of a few things: any numeric value at all; any finite numeric value (possibly restricted to integers); or possibly more complicated validation or clamping logic within a given range. WebIDL does provide some facilities for the latter, but only if your range is based on powers of two; ranges like 0 to 100, -180 to +180, or similar are not supported and require prose.

Second, they have the usual WebIDL problem of using the type system for two different purposes: coercions from JavaScript values, as in the cases of parameter lists and dictionaries, and documentation, as in the case of return values and constants. For the documentation cases, the proliferation of numeric types is simply confusing; declaring a return type as an `unsigned long` is meaningless given that it will be exposed as a normal JavaScript number (i.e. double-precision floating point value). It would be better to use a generic `number` type for those purposes, and create types that emphasize the potential coercion or validation strategies for use in those scenarios.

Finally, the spec is bloated with repetitive and spread-out text for performing the different numeric type coercions. The attributes `[EnforceRange]` and `[Clamp]` modifying the steps so that the types come to have drastically different behaviors when they are applied.

My proposed solution is to try to give tools that are both more aligned with spec-authoring use cases and clearer about what they are accomplishing. The complete functionality of the current types is preserved, and in fact further use cases (such as custom ranges) are enabled.

The proposal consists of removing all existing numeric types, as well as the `[Clamp]` and `[EnforceRange]` extended attributes, and replacing them with:

- A new `number` type, along with a `[EnforceFinite]` extended attribute that can apply to it, for documentation cases and for the most generic floating-point number processing.
- Two new parametrized types, `IntEnforceWithin<x, y>` and `IntClampRound<x, y>`, which are expected to be broadly useful for defining integer inputs that must stay within a specific range.
- Two new parametrized types, `IntMod<x, y>` and `UintMod<y>`, which are expected to be less useful and are mainly kept to enable matching legacy semantics.
[...]

PR #857, opened 2020-03-18
BREAKING: Convert the `unrestricted` keyword into an extended attribute

ExE-Boss:
Pre‑requisite for <https://github.com/heycam/webidl/issues/843> (PR <https://github.com/heycam/webidl/pull/856>).

~~I decided to go with `[UnrestrictedFloat]` instead of `[LegacyUnrestricted]`, as the latter implies that it applies to non‑floating‑point types as well.~~

That and there are valid non‑legacy reasons to allow non‑finite floating point values, specifically for Math APIs.

After <https://github.com/heycam/webidl/pull/857#issuecomment-600776106>, I’ve changed it to just be `[Unrestricted]`.

## Depends on:

- [x] <https://github.com/heycam/webidl/pull/884> (merged)

***
<a href="https://pr-preview.s3.amazonaws.com/EB-Forks/webidl/pull/857.html" title="Last updated on May 18, 2020, 6:48 AM UTC (e79993c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/857/4a3c045...EB-Forks:e79993c.html" title="Last updated on May 18, 2020, 6:48 AM UTC (e79993c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -705,7 +705,6 @@ underscore.
         "static"
         "stringifier"
         "typedef"
-        "unrestricted"
 </pre>
 
 If an <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> token is used, then the
@@ -1699,12 +1698,12 @@ lie outside the valid range of values for its type, as given in
     1.  Let |result| be the Mathematical Value that would be obtained if
         |S| were parsed as an ECMAScript <emu-nt>[=NumericLiteral=]</emu-nt>.
     1.  If the <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token is being
-        used as the value for a {{float}} or {{unrestricted float}}, then
-        the value of the <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token
+        used as the value for a {{float}}, then the value of the
+        <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token
         is the IEEE 754 single-precision floating point number closest to |result|.
     1.  Otherwise, the <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token is being
-        used as the value for a {{double}} or {{unrestricted double}}, and
-        the value of the <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token
+        used as the value for a {{double}}, then the value of the
+        <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t> token
         is the IEEE 754 double-precision floating point number closest to |result|. [[!IEEE-754]]
 </div>
 
@@ -1716,17 +1715,17 @@ constant, dictionary member or optional argument is is being used as the
 value for:
 
 <dl class="switch">
-     :  Type {{unrestricted float}}, constant value <emu-t>Infinity</emu-t>
+     :  Type [{{Unrestricted}}] {{float}}, constant value <emu-t>Infinity</emu-t>
      :: The value is the IEEE 754 single-precision positive infinity value.
-     :  Type {{unrestricted double}}, constant value <emu-t>Infinity</emu-t>
+     :  Type [{{Unrestricted}}] {{double}}, constant value <emu-t>Infinity</emu-t>
      :: The value is the IEEE 754 double-precision positive infinity value.
-     :  Type {{unrestricted float}}, constant value <emu-t>-Infinity</emu-t>
+     :  Type [{{Unrestricted}}] {{float}}, constant value <emu-t>-Infinity</emu-t>
      :: The value is the IEEE 754 single-precision negative infinity value.
-     :  Type {{unrestricted double}}, constant value <emu-t>-Infinity</emu-t>
+     :  Type [{{Unrestricted}}] {{double}}, constant value <emu-t>-Infinity</emu-t>
      :: The value is the IEEE 754 double-precision negative infinity value.
-     :  Type {{unrestricted float}}, constant value <emu-t>NaN</emu-t>
+     :  Type [{{Unrestricted}}] {{float}}, constant value <emu-t>NaN</emu-t>
      :: The value is the IEEE 754 single-precision NaN value with the bit pattern 0x7fc00000.
-     :  Type {{unrestricted double}}, constant value <emu-t>NaN</emu-t>
+     :  Type [{{Unrestricted}}] {{double}}, constant value <emu-t>NaN</emu-t>
      :: The value is the IEEE 754 double-precision NaN value with the bit pattern 0x7ff8000000000000.
 </dl>
 
@@ -5728,12 +5727,13 @@ The following types are known as <dfn id="dfn-integer-type" export>integer types
 {{long long}} and
 {{unsigned long long}}.
 
+The following types are known as <dfn id="dfn-float-type" export>float types</dfn>:
+{{float}} and
+{{double}}.
+
 The following types are known as <dfn id="dfn-numeric-type" export>numeric types</dfn>:
-the [=integer types=],
-{{float}},
-{{unrestricted float}},
-{{double}} and
-{{unrestricted double}}.
+the [=integer types=], and
+the [=float types=].
 
 The <dfn id="dfn-primitive-type" export>primitive types</dfn> are
 {{bigint}}, {{boolean}} and the [=numeric types=].
@@ -5837,7 +5837,7 @@ type.
 <pre class="grammar" id="prod-PrimitiveType">
     PrimitiveType :
         UnsignedIntegerType
-        UnrestrictedFloatType
+        FloatType
         "undefined"
         "boolean"
         "byte"
@@ -5845,13 +5845,7 @@ type.
         "bigint"
 </pre>
 
-<pre class="grammar" id="prod-UnrestrictedFloatType">
-    UnrestrictedFloatType :
-        "unrestricted" FloatType
-        FloatType
-</pre>
-
-<pre class="grammar" id="prod-FloatType">
+<pre class="grammar" oldids="prod-UnrestrictedFloatType" id="prod-FloatType">
     FloatType :
         "float"
         "double"
@@ -6061,6 +6055,10 @@ The {{float}} type is a floating point numeric
 type that corresponds to the set of finite single-precision 32 bit
 IEEE 754 floating point numbers. [[!IEEE-754]]
 
+When defined with the [{{Unrestricted}}] [=extended attribute=], the {{float}} type
+additionally corresponds to the single-precision 32 bit IEEE 754 non-finite,
+and special "not a number" values (NaNs). [[!IEEE-754]]
+
 {{float}} constant values in IDL are
 represented with <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t>
 tokens.
@@ -6077,25 +6075,15 @@ The [=type name=] of the
 </p>
 
 
-<h4 oldids="dom-unrestrictedfloat" id="idl-unrestricted-float" interface>unrestricted float</h4>
-
-The {{unrestricted float}} type is a floating point numeric
-type that corresponds to the set of all possible single-precision 32 bit
-IEEE 754 floating point numbers, finite, non-finite, and special "not a number" values (NaNs). [[!IEEE-754]]
-
-{{unrestricted float}} constant values in IDL are
-represented with <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t>
-tokens.
-
-The [=type name=] of the
-{{unrestricted float}} type is "<code>UnrestrictedFloat</code>".
-
-
 <h4 oldids="dom-double" id="idl-double" interface>double</h4>
 
 The {{double}} type is a floating point numeric
 type that corresponds to the set of finite double-precision 64 bit
 IEEE 754 floating point numbers. [[!IEEE-754]]
+
+When defined with the [{{Unrestricted}}] [=extended attribute=], the {{double}} type
+additionally corresponds to the double-precision 64 bit IEEE 754 non-finite,
+and special "not a number" values (NaNs). [[!IEEE-754]]
 
 {{double}} constant values in IDL are
 represented with <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t>
@@ -6103,20 +6091,6 @@ tokens.
 
 The [=type name=] of the
 {{double}} type is "<code>Double</code>".
-
-
-<h4 oldids="dom-unrestricteddouble" id="idl-unrestricted-double" interface>unrestricted double</h4>
-
-The {{unrestricted double}} type is a floating point numeric
-type that corresponds to the set of all possible double-precision 64 bit
-IEEE 754 floating point numbers, finite, non-finite, and special "not a number" values (NaNs). [[!IEEE-754]]
-
-{{unrestricted double}} constant values in IDL are
-represented with <emu-t class="regex"><a href="#prod-decimal">decimal</a></emu-t>
-tokens.
-
-The [=type name=] of the
-{{unrestricted double}} type is "<code>UnrestrictedDouble</code>".
 
 
 <h4 id="idl-bigint" interface>bigint</h4>
@@ -6636,8 +6610,9 @@ annotate are called <dfn export for="annotated types" lt="inner type">inner type
 The following extended attributes are <dfn for="extended attributes">applicable to types</dfn>:
 [{{AllowShared}}],
 [{{Clamp}}],
-[{{EnforceRange}}], and
-[{{LegacyNullToEmptyString}}].
+[{{EnforceRange}}],
+[{{LegacyNullToEmptyString}}], and
+[{{Unrestricted}}].
 
 <div algorithm>
     The <dfn for="IDL type" lt="extended attribute associated with|extended attributes associated with">extended attributes associated with</dfn>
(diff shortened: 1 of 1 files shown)


annevk:
I don't think that rationale holds. `[AllowShared]` only applies to certain types. I think we should just call this `[Unrestricted]` and only have it apply to float types.

index.bs
```diff
@@ -7454,15 +7428,22 @@ In effect, where <var ignore>x</var> is a Number value,
 </div>
 
 
-<h4 id="es-float">float</h4>
+<h4 id="es-decimal-types">Decimal types</h4>
```
annevk: Floats are not decimal types. Any reason not to stick with the existing structure here?

index.bs
```diff
@@ -7454,15 +7428,22 @@ In effect, where <var ignore>x</var> is a Number value,
...
+
+<h5 oldids="es-unrestricted-float" id="es-float">float</h5 id="es-unrestricted-float">
```
annevk: Closing tags should not have attributes.

ExE-Boss: That’s caused by a bug in **Visual Studio Code**’s new [**HTML mirror cursor**](https://code.visualstudio.com/updates/v1_41#_html-mirror-cursor) feature: <https://github.com/microsoft/vscode/issues/87067>.

It will hopefully be fixed by <https://github.com/microsoft/vscode/issues/88424#issuecomment-600840619>.

index.bs
ExE-Boss: It’s to allow `[Unrestricted]` to refer to the `float` and `double` section as a whole.

annevk: Can't we just say float or double? We don't do this for integers types.

ExE-Boss: Actually, [`[Clamp]`](https://heycam.github.io/webidl/#Clamp) does do that.

annevk: Fair, so let's add "float types" in a similar fashion, but not establish a new outline for the document.

annevk:
As a high-level comment, I suspect that before merging this we should also have a PR against HTML and perhaps some infrastructure PRs? I'm not entirely sure what will break if we make this kind of change. Are unknown extended attributes generally fine?

saschanaz:
>Are unknown extended attributes generally fine?

ReSpec will throw when unknown extattrs appear.

domenic:
Should we get buy-in from various bindings teams that they are interested in this churn?

annevk:
We should. If it helps, I think it's worth doing as the main reason we have unrestricted is because extended attributes were not type-bound back in the day. Moving to JavaScript-aligned type names for floats and integers will be similar churn, but also well worth doing readability-wise. (E.g., that IDL's byte is not Infra's byte is super confusing.)

index.bs
```diff
@@ -10722,6 +10654,64 @@ for the specific requirements that the use of
...
+If the [{{Unrestricted}}] [=extended attribute=] appears on one of the [=float types=],
+it creates a new IDL type such that that when an ECMAScript Number is converted to the IDL type,
+IEEE 754 non-finite, and special "not a number" values (NaNs) are allowed. [[!IEEE-754]]
+
+The [{{Unrestricted}}]
+extended attribute must
+[=takes no arguments|take no arguments=].
+
+A type annotated with the [{{Unrestricted}}] extended attribute must not appear
+in a [=read only=] attribute. A type that is not a [=float type=] must not
```
EdgarChen: Is there any reason we don't allow `[Unrestricted]` appeared in a read only attribute? There is such usage in current spec, like https://html.spec.whatwg.org/multipage/media.html#htmlmediaelement

ExE-Boss: Because the conversion from `float` and `double` to an **ECMAScript** `number` doesn’t require it.

annevk: But the prose says that only if the extended attribute is specified it can contain the additional values.

saschanaz: Specifically the HTMLMediaElement case returns NaN which requires Unrestricted.

index.bs
```diff
@@ -5977,6 +5971,9 @@ The {{float}} type is a floating point numeric
 type that corresponds to the set of finite single-precision 32 bit
 IEEE 754 floating point numbers. [[!IEEE-754]]
 
+When defined with the [{{Unrestricted}}] [=extended attribute=], the {{float}} type
+additionally corresponds to IEEE 754 non-finite, and special "not a number" values (NaNs). [[!IEEE-754]]
```
TimothyGu: IEEE 754 specifies several versions of non-finite and not-a-number values: single-precision, double-precision, etc. You should say single-precision 32-bit to be consistent with above.

index.bs
```diff
@@ -7151,8 +7124,9 @@ ECMAScript value type.
     1.  If <a abstract-op>Type</a>(|V|) is Boolean, then
         return the {{boolean}} value that represents the same truth value.
     1.  If <a abstract-op>Type</a>(|V|) is Number, then
-        return the result of <a href="#es-to-unrestricted-double">converting</a> |V|
-        to an {{unrestricted double}}.
+        return the result of <a href="#es-to-double">converting</a> |V|
+        to an {{double}} [=extended attribute associated with|associated with=]
```
TimothyGu: "a" double, not "an" double.

Also, _extended attributes_ are associated with _types_: not vice versa, as you have here. You should say "double with an [associated](https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with) [Unrestricted] extended attribute." I realize that unfortunately what you have is the norm in the spec right now, but doesn't hurt to be correct when we get a chance to.

index.bs
```diff
@@ -10722,6 +10660,64 @@ for the specific requirements that the use of
...
+
+    <pre highlight="webidl">
+        [Exposed=Window]
+        interface GraphicsContext {
+          void setColorFloat(double red, double green, double blue);
+          void setColorUnrestrictedFloat([Unrestricted] double red, [Unrestricted] double green, [Unrestricted] double blue);
+        };
+    </pre>
+
+    In an ECMAScript implementation of the IDL, a call to setColor3f with
```
TimothyGu: What's `setColor3f`?

ExE-Boss: A leftover from what I used to call the `double`‑taking `setColor` methods before I settled on `setColorDouble`.

annevk:
So in terms of implementer interest, I guess it would be good to hear from @yuki3 (Chrome), @EdgarChen (Firefox), @plinss (widlparser, somewhat critical for a lot of standards), @cdumez (WebKit; not entirely sure Chris is the right person, but he'd know the right person), @saschanaz (webidl2.js, somewhat critical for WPT).

Both for this change and the proposed renaming of numeric types in #843.

Then, if there's implementer support, before making this particular change I think we should at least patch HTML and do a quick grep to see what other specifications are impacted (WebGL?).

ExE-Boss:
https://github.com/tidoust/reffy-reports seems to indicate that 29 specs use the `unrestricted` keyword, out of the 119 specs that use floating‑point number types.

- <https://github.com/tidoust/reffy-reports/search?q=%22unrestricted+double%22+OR+%22unrestricted+float%22+path%3Aed%2Fidl+path%3Atr%2Fidl>
- <https://github.com/tidoust/reffy-reports/search?q=double+OR+float+path%3Aed%2Fidl+path%3Atr%2Fidl>

annevk:
Thanks, I guess for specifications we really need @saschanaz onboard then as their bot can update specifications automatically (well, up to the point of review).

saschanaz:
I'll write an update when we get a consensus to land this.

yuki3:
Personally I like the both ideas, but I'd like to see the entire type system before moving on.  For example, what do we do for `USVString`?  `[USV] DOMString`?  IMHO, the type system should be designed as a whole on top of the consistent concepts.  Depending on what concept model of type system we have, I expect that we can discuss whether `Int32Array` or `ArrayView<i32>` is desirable for example.

In other words, I don't yet understand the (ultimate) goal of these proposals.  If we'll keep `USVString` and `DOMString` separate, I don't understand why we'd like to turn `unrestricted float` into `[UnrestrictedFloat] float`.

ExE-Boss:
We turn `unrestricted float` into `[Unrestricted] float` to simplify the parser.

domenic:
It's not clear there's anything "simplifying" about moving from working code to new code.

saschanaz:
I assume this is mainly to ease migration for #843, as we currently can't define a typedef with a space e.g. `unrestricted f32`.

ExE-Boss:
It’s not just that we can’t define a `typedef` with a space, it’s that `unrestricted float32` and `unrestricted float64` are syntax errors, so we have to use `[Unrestricted] float32` and `[Unrestricted] float64`.

saschanaz:
> it’s that `unrestricted float32` and `unrestricted float64` are syntax errors

I think that specific problem can be solved by modifying existing syntax, though?

yuki3:
As a thought experiment, we could turn everything into type-annotation:
```
                long →           [Integer32Bit] type
  unrestricted float → [UnrestrictedFloat32Bit] type
           USVString →              [USVString] type
```
As another thought experiment, we can turn all numeric types into a single type:
```
                long →           [Integer32Bit] numeric
  unrestricted float → [UnrestrictedFloat32Bit] numeric
           USVString →                    [USV] string
```

Why don't we do these things?  Intuitively bad?

I expect that we can explain why we define `X` as a separate type and why we define `[ExtAttr] X` as an annotated type (not a separate type `Y`) _based on clear guideline, policy, concept model, and/or rationale_.  At this moment, I can tell why `unrestricted float` should be `[Unrestricted] float` only from my intuitiveness, which I think we shouldn't rely on too much.

If this proposal were just a patchwork fix, I'm not convinced.  Maybe it's better to fix typedef.  Maybe it's expected that you cannot write `unrestricted TypedefName` because `unrestricted` is not a type modifier (as of now).

As long as we have consensus about the type system and it's clear what should be type and what should be annotated type, we should be able to discuss whether this proposal is aligned to the guideline/policy/etc or not.

Note that I'm not objecting to this proposal.  My intuitiveness supports this way.  But I also see lack of justification/rationale behind the proposal.

domenic:
Very well put @yuki3.

First, I think it's worth separating primitives from all other types, for this discussion. I am reasonably happy with our non-primitive type system, e.g. dictionaries, enums, callbacks, sequences, records, promises, etc. So I think this is mostly about primitives. In particular: numeric types, `boolean`, `DOMString`, `ByteString`, `USVString`, `symbol`, and maybe `any` and `object`.

Of those, numeric types and string types are basically what we're discussing. The rest are straightforward mappings to JS primitives.

One potential framework we could work toward is saying that types should correspond to JavaScript types, and extended attributes should modify the conversion algorithms. This would argue for turning all numeric types into a single type. This would have a benefit of corresponding IDL better to JS-ecosystem things like TypeScript or MDN documentation. However, for numeric types especially, this is more cumbersome for spec authors and implementers.

Another potential framework is to try to match some sort of consensus-type-system of C++/Rust, i.e. implementer languages. This is fuzzier, both because of differences between C++ and Rust, and because of differences in C++ dialects (e.g. different string types). For numeric types, I think the result would be similar to #843. For strings, its unclear which direction this would lean.

One issue with a C++/Rust type approach is that those languages don't have an idiomatic type for "double/float but not Infinity or NaN", and it's currently an explicit goal of Web IDL to make the easy path for spec authors not require thinking about those. That is, unrestricted behavior is opt-in in Web IDL, both currently and with this PR. Whereas if we took a principled C++/Rust approach, it would be opt-out.

A third possibility would be something in between. For example, we could try to be "JS first", but add conveniences or concessions to make spec-writing easier. It's super-common to want to accept only nonnegative integers, so forcing people to write `[EnforceRange] [NonNegative] [Integer] number` or something like that is not great. It's also historically common to add C++-like wrapping behavior for numbers by using `unsigned long` or `long`, even if [this is not great practice](https://w3ctag.github.io/design-principles/#numeric-types). Maybe we just make some typedefs for the common cases, e.g. `uint64` and `int32`, but we get rid of the rarer cases (like `octet` or `byte` or `short`), requiring those to be written long-hand or typedefed in the using spec.

(As an aside, people may enjoy my 2014-era proposal in #33 for reforming the numeric types. I don't really like it any more, but it illustrates somewhat similar thinking, of trying to be "JavaScript first" and not treating C++ boundaries like 2<sup>32</sup> or 0 as special. https://github.com/heycam/webidl/issues/33#issuecomment-468076275 is a 2019-era update on my thinking to be more spec-author friendly, but it is definitely not prinicipled in the way @yuki3 describes.)

saschanaz:
>Whereas if we took a principled C++/Rust approach, it would be opt-out.

IMO this part should be decided before landing #843, as it will be way harder to migrate from `float32 / [Unrestricted] float32` to `[Restricted] float32 / float32` while it should be relatively straightforward to migrate from `float / unrestricted float`.

domenic:
Agreed @saschanaz. Which reminds me that I kind of forgot a conclusion to my post. Unfortunately the conclusion is basically pessimistic: I cannot think of any principled framework for this space which accomplishes the goals of:

- Grounding our primitive type system in an existing one (either JS, or C++/Rust)
- Avoiding an increase in spec authors/implementer usage complexity
  - Any JS-based primitive type system causes a great increase in such complexity
  - Any C++/Rust-based primitive type system increases complexity by requiring folks to remember to prohibit or handle NaN/Infinities.

Maybe there is some other guiding principle besides grounding our primitive type system in an existing one?

saschanaz:
BTW, my intuition says an extended attribute for a type is for slightly modifying the condition, e.g. `[AllowShared] BufferSource` still receives a buffer source but allows shared state. `[Unrestricted] float` meets this description, while `[Integer] number` is quite a lot restriction so doesn't meet it.

yuki3:
For numeric types, I'm thinking about the following suggestion from Domenic.

> Another potential framework is to try to match some sort of consensus-type-system of C++/Rust, i.e. implementer languages.

1. Define numeric types based on its memory size.
int8, int16, int32, and int64 should be different types.
2. float32 and int32, float64 and int64 should be different types because it's very different how to interpret their bit patterns.
3. int32 and unsigned int32 should be different types because sign extension, equality comparison, etc. are different even if their bit patterns are the same.

It looks good/reasonable to me to put memory layout and bit pattern interpretation as key principles of the type system of primitive types.  I agree that [Restricted] looks more natural and consistent with others.  (We should choose a better name for [Restricted], though.  For example, [DisallowNanAndInf] looks clearer to me.)

annevk:
I tried to come up with some things I value in IDL that I think largely overlap with prior points made here:

* Types in implementations. If a systems language might want a different type, I think that's a good hint for IDL. (This might argue for splitting ArrayBuffer and SharedArrayBuffer, though not sure about typed arrays. It also defends the current decisions around strings.)
* Clarity and readability. For example, I think the current `float` fails here a bit as the size is not clear and more importantly it's not clear that it's very different from other languages (by not accepting infinity and not-a-number). `float32` and `restricted-float64` (I'm not sure about restricted, but also not sure about basic/numeric alternatives) seem preferable. IDL is read a lot, the clearer it is at a glance what will happen the better.
* Alignment with type terminology of Infra, JavaScript, and WebAssembly. (DOMString -> String or even string?) And also alignment with type terminology from other languages to the extent possible.
* Ease-of-use. Common things should be straightforward. I think in specifications we also get a lot of easy-of-use through "automatic" mapping between IDL types and Infra. We haven't quite figured out what the story for numbers should be there, but Infra should probably not have as many types as IDL. (I'm thinking "floats" and "integers" for prose and the IDL boundary takes care of conversion.)

In the end some of these decisions will have to be tradeoffs, but as long as try to maximize the benefit for these different audiences we'll be fine I think.

TimothyGu:
There was a lot of good discussion happening, but I think it's clear that the idea needs some more time to bake. I'll close this PR for now.

bathos:
Very late to this, but in case it were to return:

Currently extended attributes are described like this:

> ...used to control how language bindings will handle those constructs...

And the extended attributes defined by Web IDL are described like this:

> This specification defines a number of extended attributes that are applicable to the ECMAScript language binding, which are described in § 3.3 Extended attributes.

That description doesn’t outright preclude the possibility that the “built-in” EAs could also be applicable to other language bindings — but it’s notable that if so, it would not be due to anything the spec states on the matter. Certain EAs seem like they surely ought to have significance to other bindings, but the current definitions of EAs stay true to the text quoted above: all of their effects and associated algorithms are specific to ECMAScript. CrossOriginIsolated, for example, is not defined in terms of its abstract effects but its concrete effects on the initialization of Web IDL constructs in a new ECMAScript realm.

Most non-exposure-related EAS don’t have an “intuitive” applicability outside of JS though. Looking at the existing applicable-to-numeric-types EAs, would there be any reason to believe that [Clamp] or [EnforceRange] should have significance for other language bindings? Their defined behavior appears pretty specific to ES’s Number type. In new bindings for a language with a native u64 type, it would be a bit odd if the setting a `[Clamp] long long` attribute to 0x20_0000_0000_0000 would mean it first gets converted it to 0x1F_FFFF_FFFF_FFFF (assuming a generous interpretation of all of the algorithm steps that wouldn’t be coherent anyway).

To me this makes it seem a lot like `unrestricted double` isn’t `double`-but-annotated. The Web IDL types proper have different value ranges — it’s not just about the behaviors realized at the boundary between ES and Web IDL.

This PR does (or would have) addressed this since this sentence is unqualified by anything specific to ES:

> When defined with the [Unrestricted] extended attribute, the float type additionally corresponds to the single-precision 32 bit IEEE 754 non-finite, and special "not a number" values (NaNs).

If this is still bubbling, I’d suggest that having a clearer model of the intended scope of EA effects first would be helpful, and that the EA introduction would need to be updated to make it explicit that some, like [Unrestricted], are _not_ binding-specific. However I also suspect this wouldn’t be a positive change and that it’s actually desirable for EAs, at least those applicable to types, to remain binding-specific.

PR #859, opened 2020-03-18
Clarifications to dictionary semantics

### LINKED ISSUE #524: Change dictionary section to encourage Infra wording
Per https://github.com/whatwg/dom/issues/574 we no longer need to use 

>_dict_'s member is present 

and can instead use 

> _dict_["member"] exists

The dictionary section should probably encourage the latter style and only define "present" for as long as it takes everyone to migrate.

### LINKED ISSUE #852: Can a dictionary with default values ever have a not present member?
Consider

```webidl
dictionary D {
  DOMString a;
  DOMString b = "";
};

interface I {
  D output();
};
```

with the spec text

> output():
> 1. Return a new `D`.

If JavaScript code calls the `output()` method, do they get `{}`, or do they get `{ b: "" }`?

---

My understanding was that they would get `{}`. From what I can tell, default values only apply when using dictionaries as inputs. This is circumstantially supported by text such as

> In the ECMAScript binding, a value of undefined is treated as not present, or will trigger the default value where applicable.

(only talks about the input, not the output) or

> Dictionary members can also optionally have a default value, which is the value to use for the dictionary member when passing a value to a platform object that does not have a specified value.

which doesn't make a lot of sense, but again is only talking about output, not input, or

> On a given dictionary value, the presence of each dictionary member is optional, unless that member is specified as required

which only mentions the presence being guaranteed if the value is `required`, and doesn't say anything about default values.

And finally, the normative algorithms in https://heycam.github.io/webidl/#es-dictionary where "An ECMAScript value esDict is converted to an IDL dictionary type value" takes into account default values but "An IDL dictionary value V is converted to an ECMAScript Object value" does not mention them.

---

However, there is one sentence which implies that maybe it should be `{ b: "" }`:

> Dictionary members with default values are always considered to be present.

If this is the case, then  "An IDL dictionary value V is converted to an ECMAScript Object value" step 3.1.2 "If the dictionary member named key is present in V" would trigger, and we would define the `b` property.

It's not anywhere formally defined that the value of _V_[_key_] is the default value given in the IDL definition, which is a bit suspicious, but maybe that was just an oversight.

So... we're unsure what to do in Chromium here.

domenic:
This rewrites some of the text around dictionaries to clear up the confusion in #852, as well as generally try to smooth out the reading experience and definitions. Concrete changes:

* Rewrote "dictionaries are always passed by value" to be more specific and clear.
* Moved "dictionaries must not be used as the type of an attribute or constant" near the top of the dictionary section, commensurate with its importance.
* Removed the terms "present" and "not present"; instead we can use Infra's "exists" for maps.
* Fixed several cases where the overload resolution algorithm discussed arguments being "not present", even though that was only defined for dictionaries. Instead it now correctly talks about "missing".
* Rewrote the discussion of required dictionary members and dictionary members with default values to make it clear how they contribute to the dictionary's entries.
* Added a note to the IDL-to-ES conversion algorithm for dictionaries explicitly pointing out that default-value dictionary members will always be present, and thus always show up in the output.

Closes #852. Closes #524.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/859.html" title="Last updated on Mar 25, 2020, 7:40 PM UTC (92a21a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/859/691b8d2...92a21a0.html" title="Last updated on Mar 25, 2020, 7:40 PM UTC (92a21a0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4787,7 +4787,7 @@ Note that specifying [=dictionary members=] as [=dictionary member/required=] on
 an observable effect when converting other representations of dictionaries (like an ECMAScript value
 supplied as an argument to an [=operation=]) to an IDL dictionary. Specification authors should
 leave the members [=dictionary member/optional=] in all other cases, including when a dictionary
-type is used soley as the [=return type=] of [=operations=].
+type is used solely as the [=return type=] of [=operations=].
 
 [=dictionary member/Optional=] [=dictionary members=] can also be specified as having a
 <dfn id="dfn-dictionary-member-default-value" for="dictionary member" export>default value</dfn>,
--- a/index.bs
+++ b/index.bs
@@ -4803,7 +4803,7 @@ optional, and their entries might or might not [=map/exist=] in the dictionary v
 
 <p class="advisement">
     As with [=optional argument/default value|operation argument default values=], it is strongly
-    suggested not to use <emu-val>true</emu-val> as the [=dictionary member/default value=] for
+    encouraged not to use <emu-val>true</emu-val> as the [=dictionary member/default value=] for
     {{boolean}}-typed [=dictionary members=], as this can be confusing for authors who might
     otherwise expect the default conversion of <emu-val>undefined</emu-val> to be used (i.e.,
     <emu-val>false</emu-val>).
--- a/index.bs
+++ b/index.bs
@@ -2279,7 +2279,7 @@ corresponding argument omitted.
 If the type of an argument is a [=dictionary type=] or a [=union type=] that has
 a [=dictionary type=] as one
 of its [=flattened member types=], and that dictionary type and its ancestors have no
-[=required dictionary member|required members=], and the argument is either the final argument or is
+[=dictionary member/required=] [=dictionary member|members=], and the argument is either the final argument or is
 followed only by [=optional arguments=], then the argument must be specified as
 optional and have a default value provided.
 
@@ -3877,8 +3877,8 @@ defined with [=optional arguments=] and merged into one,
     };
 </pre>
 
-the [=overload resolution algorithm=] would treat the <var ignore>path</var> argument as [=not
-present=] given the same call <code>stroke(undefined)</code>, and not throw any exceptions.
+the [=overload resolution algorithm=] would treat the <var ignore>path</var> argument as missing
+given the same call <code>stroke(undefined)</code>, and not throw any exceptions.
 
 Note: For this particular example, the latter behavior is actually what Web developers would
 generally expect. If {{CanvasDrawPath}} were to be designed today, [=optional arguments=] would be
@@ -4062,8 +4062,7 @@ determine what Web IDL language feature to use:
     class="idl">foo(|arg|)</code> operation, with |arg| set to null, while <code>foo()</code> alone
     would go to the first overload. This can be a surprising behavior for many API users. Instead,
     specification authors are encouraged to use an [=optional argument=], which would categorize
-    both <code>foo()</code> and <code>foo(undefined)</code> as "|arg| is [=not
-    present=]".
+    both <code>foo()</code> and <code>foo(undefined)</code> as "|arg| is missing".
 
     <pre highlight="webidl">
         interface A {
@@ -4741,9 +4740,15 @@ where [=map/keys=] are strings and [=map/values=] are of a particular type speci
     };
 </pre>
 
-Dictionaries are always passed by value.  In language bindings where a dictionary is represented by an object of some kind, passing a
-dictionary to a [=platform object=] will not result in a reference to the dictionary being kept by that object.
-Similarly, any dictionary returned from a platform object will be a copy and modifications made to it will not be visible to the platform object.
+Dictionaries are always passed by value: the dictionary does not retain a reference to its
+language-specific representation (e.g., the corresponding ECMAScript object). So for example,
+returning a dictionary from an [=operation=] will result in a new ECMAScript object being created
+from the current values of the dictionary. And, an operation that accepts a dictionary as an
+argument will perform a one-time conversion from the given ECMAScript value into the dictionary,
+based on the current properties of the ECMAScript object. Modifications to the dictionary will not
+be reflected in the corresponding ECMAScript object, and vice-versa.
+
+Dictionaries must not be used as the type of an [=attribute=] or [=constant=].
 
 A dictionary can be defined to <dfn id="dfn-inherit-dictionary" for="dictionary" export>inherit</dfn> from another dictionary.
 If the identifier of the dictionary is followed by a colon and a [=identifier=],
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4741,9 +4740,15 @@ where [=map/keys=] are strings and [=map/values=] are of a particular type speci
     };
 </pre>
 
-Dictionaries are always passed by value.  In language bindings where a dictionary is represented by an object of some kind, passing a
-dictionary to a [=platform object=] will not result in a reference to the dictionary being kept by that object.
-Similarly, any dictionary returned from a platform object will be a copy and modifications made to it will not be visible to the platform object.
+Dictionaries are always passed by value: the dictionary does not retain a reference to its
+language-specific representation (e.g., the corresponding ECMAScript object). So for example,
+returning a dictionary from an [=operation=] will result in a new ECMAScript object being created
```
marcoscaceres: I don't think the returning part is what should be clarified, as returning dictionaries is rare... it's more the passing in a dictionary in from JS to a method that is "deeply copied" (passed by value) that should be clarified... most JS programmers won't know what "passed by value" means, but it's second nature to all c++ and Rust programmers.

domenic: Did you see the next sentence?

(BTW it's shallowly copied, not deeply.)

marcoscaceres: yep, next sentence is good, nm. 

I thought: 

```
x.method({a: {b: {c: "d"}}})
```

copied the whole thing "deeply"? Or did I misunderstand?

domenic: Well, OK, if the inner things are dictionaries, then _their_ copy semantics get applied, so yeah, that particular example is deep. But `x.method{ a: window })` does not copy `window` deeply.

domenic:
@EdgarChen and/or @annevk, would one of you be able to review?

index.bs
marcoscaceres: sorry, yeah, I was thinking exclusively about passing dictionaries.

index.bs
```diff
@@ -4772,30 +4777,42 @@ from another dictionary, then the set is empty.  Otherwise, the set
...
-not have a specified value.  Dictionary members with default values are
-always considered to be present.
+[=Dictionary members=] can be specified as
+<dfn id="required-dictionary-member" export for="dictionary member">required</dfn>, meaning that
+converting a language-specific value to a dictionary requires providing a value for that member.
+They can also be specified as having a
+<dfn id="dfn-dictionary-member-default-value" for="dictionary member" export>default value</dfn>,
+which is the value used by default when author code or specification text does not provide a value
+for that member. Note that [=dictionary member/required=] dictionary members only have meaning for
+dictionaries used as operation arguments; members should be left optional if a dictionary is only
```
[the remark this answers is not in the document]
domenic: This is new. It's just a should saying not to add pointless stuff. It's not likely to be easily enforceable on an implementation level because it would require whole-corpus analysis to find all the dictionary usage sites.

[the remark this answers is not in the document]
annevk: It's not entirely clear to me if it's pointless as it's basically a specification-level assert. But since it can't be used consistently (unless you always have distinct output dictionaries, which also isn't great), I'm inclined to agree it's better not to add it.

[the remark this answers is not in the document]
TimothyGu: I feel the "only have meaning" part isn't clear: _of course_ the word "required" has meaning in general!

How about

Note that specifying dictionary members as required only has an observable effect when converting other representations of dictionaries (like an ECMAScript value supplied as an argument to an [operation](https://heycam.github.io/webidl/#dfn-operation)) to an IDL dictionary. Specification authors should leave members optional in all other cases, including when a dictionary type is used solely as the [return type](https://heycam.github.io/webidl/#dfn-return-type) of [operations](https://heycam.github.io/webidl/#dfn-operation).

We should also define "optional" as "not required" in the context of dictionary members.

index.bs
```diff
@@ -4772,30 +4777,42 @@ from another dictionary, then the set is empty.  Otherwise, the set
...
+    [=dictionary member/default value=] if one is present, or will result in no [=map/entry=]
+    existing in the dictionary value otherwise.
+</p>
+
+<p class="advisement">
+    As with [=optional argument/default value|operation argument default values=], it is strongly
+    encouraged not to use <emu-val>true</emu-val> as the [=dictionary member/default value=] for
+    {{boolean}}-typed [=dictionary members=], as this can be confusing for authors who might
+    otherwise expect the default conversion of <emu-val>undefined</emu-val> to be used (i.e.,
+    <emu-val>false</emu-val>).
```
TimothyGu: Cite https://w3ctag.github.io/design-principles/ (https://w3ctag.github.io/design-principles/#prefer-dict-to-bool in particular)?

index.bs
```diff
@@ -4772,30 +4777,42 @@ from another dictionary, then the set is empty.  Otherwise, the set
...
 </p>
 
 An [=ordered map=] with string [=map/keys=] can be implicitly treated as a dictionary value of a
 specific dictionary |D| if all of its [=map/entries=] correspond to [=dictionary members=], in the
-correct order and with the correct types, and with appropriate [=map/entries=] for any required
-dictionary members.
+correct order and with the correct types, and with appropriate [=map/entries=] for any required or
```
TimothyGu: Is the order requirement really important? Also, what order is this? I assume it's the order members appear in the actual IDL source. However, this interpretation runs into trouble later as the ES-to-IDL conversion algorithm seems to use lexicographic order within each inherited dictionary.

Maybe we could just get rid of the order requirement?

index.bs
```diff
@@ -7756,9 +7758,7 @@ the object (or its prototype chain) correspond to [=dictionary members=].
     running the following algorithm (where |D| is the [=dictionary type=]):
 
     1.  If <a abstract-op>Type</a>(|esDict|) is not Undefined, Null or Object, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  Let |idlDict| be an empty dictionary value of type |D|;
-        every [=dictionary member=]
-        is initially considered to be [=not present=].
+    1.  Let |idlDict| be an empty [=ordered map=].
```
TimothyGu: I think having the "representing a dictionary of type _D_" information somewhere would still be beneficial.

index.bs
```diff
@@ -4741,9 +4740,15 @@ where [=map/keys=] are strings and [=map/values=] are of a particular type speci
     };
 </pre>
 
-Dictionaries are always passed by value.  In language bindings where a dictionary is represented by an object of some kind, passing a
-dictionary to a [=platform object=] will not result in a reference to the dictionary being kept by that object.
-Similarly, any dictionary returned from a platform object will be a copy and modifications made to it will not be visible to the platform object.
+Dictionaries are always passed by value: the dictionary does not retain a reference to its
+language-specific representation (e.g., the corresponding ECMAScript object). So for example,
```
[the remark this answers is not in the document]
domenic: I'm not sure what you mean by "will not be observable directly outside of it", except for the idea that when the dictionary is converted to ES, it will not have such a link. So they seem like the same before and after the colon to me.

That said, I'd be happy to just remove "Dictionaries are always passed by value". "Pass by value" is a notoriously confusing concept and it might be better to just explicitly state what we mean, as the parts after the colon try to do.

index.bs
domenic: Well, Infra maps are ordered. But yes, perhaps it's OK to just assume any order works.

index.bs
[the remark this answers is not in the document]
domenic: This section isn't talking about the word "required" in general. It's say "required dictionary members only have meaning...". So I think it's still fairly clear. Your rephrasing is OK too though.

index.bs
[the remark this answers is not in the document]
TimothyGu: Removal sounds good to me.

I understood "pass by value" as passing the IDL dictionaries passed around different spec-level algorithms by value, while you understood it as passing ES objects around into IDL operations by value. See if this helps with understanding.

index.bs
domenic: Let's do that in a followup if desired, for both dictionaries and boolean arguments in general.

index.bs
```diff
@@ -4779,19 +4779,26 @@ from and all of |E|’s [=inherited dictionaries=].
...
-They can also be specified as having a
+converting a language-specific value to a dictionary requires providing a value for that member. Any
+dictionary member that is not [=dictionary member/required=] is
+<dfn export for="dictionary member">optional</dfn>.
+
+Note that specifying [=dictionary members=] as [=dictionary member/required=] only has
+an observable effect when converting other representations of dictionaries (like an ECMAScript value
+supplied as an argument to an [=operation=]) to an IDL dictionary. Specification authors should
+leave the members [=dictionary member/optional=] in all other cases, including when a dictionary
+type is used soley as the [=return type=] of [=operations=].
```
TimothyGu:
```suggestion
type is used solely as the [=return type=] of [=operations=].
```

PR #860, opened 2020-03-19
fix: Only install `@@toStringTag` on the prototype

ExE-Boss:
Supersedes and closes #357

---

This matches the behaviour of **Chromium**, **WebIDL2JS** and the pre‑**ES2015** `[[Class]]` internal slot, which is still used in **Firefox** and **WebKit**.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11516,8 +11516,7 @@ Issue: Define those properties imperatively instead.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


domenic:
We'll continue using #357 for this. It is still blocked on Chromium updating to match the proposed spec.

PR #808, opened 2019-09-27
Add async value iterators and async iterator arguments

domenic:
The last commit here (the rest are #802) is a start at addressing #800. It is missing the glue code to pass arguments. I'll discuss over in #800 some of the things I ran in to, but having this here is a nice reference.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/808.html" title="Last updated on Apr 21, 2020, 1:42 PM UTC (13670e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/808/42a9d4b...13670e8.html" title="Last updated on Apr 21, 2020, 1:42 PM UTC (13670e8)">Diff</a>

Issue #467, opened 2017-10-17
Make creation of all initial objects/properties/methods/etc. imperative


domenic:
We've been doing this piecemeal. For example, https://heycam.github.io/webidl/#namespace-object provides a procedure for creating namespace objects given a namespace and realm. We did the same for interface object: https://heycam.github.io/webidl/#interface-object , but did not do it for interface *members* (like we did for namespace members). This will be a great clarity improvement, consolidating various sprinkled prose which tells you what global properties exist and what their property descriptors are and so on.

Once this is done completely for all initial objects, we can replace

> Each ECMAScript global environment must have its own unique set of each of the initial objects, created before control enters any ECMAScript execution context associated with the environment, but after the global object for that environment is created.

with a procedure that is given a realm and ... a set of all IDL definitions on the web platform? ... and loops over all exposed definitions and creates them on that realm's global object.

Then, I suppose, we can call that procedure from places like https://html.spec.whatwg.org/#creating-a-new-browsing-context as a customization for InitializeHostDefinedRealm(). Although that might be overkill, as the hand-wavey manner in which things currently work allows you to "automatically" get the right set of globals without having to add an explicit step when creating your global. (Thus making it easier to define new globals.) We can cross that bridge when we get there.

tobie:
See also: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27355

annevk:
FWIW, I think we actually want to define when new globals are created as we can then also define when agents and agent clusters are created, and how those all relate. That is rather indirect at the moment (in part due to how hard it is to sort out navigation properly).

domenic:
Class string (cf. #357) is a small thing that would be easy to make imperative. However, doing so in a non-confusing way is blocked on first making iterator prototype objects and async iterator prototype objects imperative, since for those two there's currently no algorithm to slot the "create a `@@toStringTag` property" into.

PR #862, opened 2020-03-24
Editorial: make more iterator-related stuff imperative

domenic:
As part of #467, this moves the installation of @@iterator/@@asyncIterator/forEach/entries/keys/values to an imperative style, for the cases of pair iterators, value iterators, async iterators, and indexed property getters.

This does not yet make imperative the same methods for maplikes and setlikes, nor does it make the construction of the iterator-related objects and prototype objects themselves imperative.

---

This will help me do #808 more properly. I think it's also a big clarity improvement; in particular how each of the four cases mentioned (pair iterators, value iterators, async iterators, and indexed property getters) is now easy to read off from the algorithm instead of being scattered throughout the section for each method.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/862.html" title="Last updated on Mar 25, 2020, 6:44 PM UTC (9c2d5ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/862/31871e8...9c2d5ef.html" title="Last updated on Mar 25, 2020, 6:44 PM UTC (9c2d5ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11466,6 +11466,7 @@ with the [{{NoInterfaceObject}}] [=extended attribute=].
     1.  If |interface| is not declared with the [{{Global}}] [=extended attribute=], then:
         1.  [=Define the regular attributes=] of |interface| on |interfaceProtoObj| given |realm|.
         1.  [=Define the regular operations=] of |interface| on |interfaceProtoObj| given |realm|.
+        1.  [=Define the iterability=] of |interface| on |interfaceProtoObj| given |realm|.
     1.  [=Define the constants=] of |interface| on |interfaceProtoObj| given |realm|.
     1.  If the [{{NoInterfaceObject}}] [=extended attribute=] was not specified on |interface|, then:
         1.  Let |constructor| be the [=interface object=] of |interface| in |realm|.
@@ -12179,18 +12180,12 @@ then there must exist a property with the following characteristics:
     property is the String value "<code>toString</code>".
 
 
-<h4 id="es-iterators">Common iterator behavior</h4>
+<h4 id="es-iterators">Maplike and setlike iterator behavior</h4>
 
 
 <h5 id="es-iterator">@@iterator</h5>
 
-If the [=interface=] has any of the following:
-
-*   an [=iterable declaration=]
-*   an [=indexed property getter=]
-*   a [=maplike declaration=]
-*   a [=setlike declaration=]
-
+If the [=interface=] has a [=maplike declaration=] or a [=setlike declaration=],
 then a property must exist
 whose name is the {{@@iterator}} symbol, with attributes
 { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
@@ -12201,29 +12196,6 @@ The location of the property is determined as follows:
 *   If the interface was declared with the [{{Global}}] [=extended attribute=],
     then the property exists on the single object that [=implements=] the interface.
 *   Otherwise, the property exists solely on the interface’s [=interface prototype object=].
-
-If the interface defines an [=indexed property getter=],
-then the [=function object=] is {{%ArrayProto_values%}}.
-
-<div algorithm="to invoke the @@iterator property of interfaces with a pair iterator">
-
-    If the interface has a [=pair iterator=], then the [=function object=] is a [=built-in function object=] that,
-    when invoked, must behave as follows:
-
-    1.  Let |object| be the result of calling <a abstract-op>ToObject</a> on the <emu-val>this</emu-val> value.
-    1.  If |object| [=is a platform object=],
-        then [=perform a security check=], passing:
-        *   the platform object |object|,
-        *   the identifier "<code>@@iterator</code>", and
-        *   the type "<code>method</code>".
-    1.  Let |interface| be the [=interface=]
-        the [=iterable declaration=] is on.
-    1.  If |object| does not [=implement=] |interface|,
-        then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-    1.  Let |iterator| be a newly created [=default iterator object=]
-        for |interface| with |object| as its target and iterator kind "<code>key+value</code>".
-    1.  Return |iterator|.
-</div>
 
 <div algorithm="to invoke the @@iterator property of interfaces with maplike or setlike declarations">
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12403,158 +12294,157 @@ property is the String value "<code>forEach</code>".
...
+        1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>entries</code>",
+            {{%ArrayProto_values%}}).
+        1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>keys</code>",
+            {{%ArrayProto_keys%}}).
+        1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>values</code>",
+            {{%ArrayProto_values%}}).
+        1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>forEach</code>",
+            {{%ArrayProto_forEach%}}).
+    1.  If |definition| has a [=pair iterator=], then:
+        1.  Define the {{@@iterator}} and <code>entries</code> methods:
```
TimothyGu: Is this "define the … methods" thing defined anywhere? Other places are using CreateBuiltinFunction which seems to be what this should do too.

domenic: They are just headers grouping a bunch of steps together. The CreateBuiltinFunction calls are nested under these.

index.bs
```diff
@@ -12403,158 +12294,157 @@ property is the String value "<code>forEach</code>".
...
-        {{ECMAScript/TypeError}}.
-    1.  Let |iterator| be a newly created [=default asynchronous iterator object=] for |interface|
-        with |object| as its [=default asynchronous iterator object/target=],
-        "<code>value</code>" as its [=default asynchronous iterator object/kind=], and
-        [=default asynchronous iterator object/is finished=] set to false.
-    1.  Run the [=asynchronous iterator initialization steps=] for |interface| with |object| and
-        |iterator|, if any such steps exist.
-    1.  Return |iterator|.
+<div algorithm>
+    To <dfn oldids="es-iterable-entries,es-iterable-keys,es-iterable-values">define the iterability</dfn>
```
TimothyGu: I tried to get used to the name of this algorithm but I don't think I can. "Iterability" is just too different from the likes of "regular operations", etc. How about "to define the iteration methods"?

index.bs
```diff
@@ -12403,158 +12294,157 @@ property is the String value "<code>forEach</code>".
...
+    of [=interface=] |definition| on |target|, given [=Realm=] |realm|, run the following steps:
+
+    1.  If |definition| has an [=indexed property getter=], then:
```
TimothyGu: I feel like the flow here could be enhanced a bit to make clear the implication and mutual exclusivity relations of the conditions, like:

1. If _definition_ has an indexed property getter, then:
   1. …
   1. If _definition_ has a value iterator, then …
1. Otherwise, if _definition_ has a pair iterator, then …
1. Otherwise, if _definition_ has an asynchronously iterable declaration, then …

You could also consider a switch-like syntax.

index.bs
TimothyGu: Ah got it. This makes me think that there's value in the readability and navigability. The whole algorithm creates seven different functions, without anchors for each. In ES, this would correspond to seven sections; while here, it's one section. I think the optimal value is somewhere in between. (Putting everything into a single algorithm/section would be even worse for maplike/setlike: each one of those declarations creates 6–7 methods.)

Maybe something more similar to the original section layout (except the `entries`, `keys`, and `values` sections are renamed to "`entries` for pair iterators" and so on), and in the root section do something like

1. Let _steps_ be the algorithm steps defined in §&nbsp;3.7.1 `entries` for pair iterators.
2. Let _F_ be ! CreateBuiltinFunction(_steps_, « », _realm_).

TimothyGu: For async iterable declarations, I feel like they should be in a separate heading from §&nbsp;3.6.9 Iterable declarations anyway. They are not that similar.

domenic: This turns out to be difficult because the steps close over _definition_. We can smuggle it in as an internal slot on the function but that's a lot of effort.

Ultimately I'm not sure giving these sections is that helpful:

- The steps are fairly repetitive, with the non-forEach ones varying only in key/value/key+value.
- You still have 4-5 lines per method of CreateBuiltinFunction, SetFunctionName, SetFunctionLength, and CreateDataProperty/CreateMethodProperty. That is, if we move the steps out, then we just have 17 steps next to each other repeating calls to those five abstract ops. Which is harder to read, IMO, than the current version.

domenic: We could split up the iterable and async iterable sections, though; that seems reasonable. I guess then we'd have "define the iteration methods" and "define the asynchronous iteration methods".

TimothyGu: I think this is good now that a big chunk has been moved out of this algorithm.

index.bs
```diff
@@ -12403,158 +12296,95 @@ property is the String value "<code>forEach</code>".
...
+<div algorithm>
+    To <dfn oldids="es-iterable-entries,es-iterable-keys,es-iterable-values">define the iteration methods</dfn>
+    of [=interface=] |definition| on |target|, given [=Realm=] |realm|, run the following steps:
+
+    1.  If |definition| has an [=indexed property getter=], then:
+        1.  Perform [=!=] [$CreateMethodProperty$](|target|, {{@@iterator}},
+            {{%ArrayProto_values%}}).
+        1.  If |definition| has a [=value iterator=], then:
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>entries</code>",
+                {{%ArrayProto_values%}}).
```
TimothyGu:
```suggestion
                {{%ArrayProto_entries%}}).
```

index.bs
```diff
@@ -12403,158 +12296,95 @@ property is the String value "<code>forEach</code>".
...
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>keys</code>",
+                {{%ArrayProto_keys%}}).
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>values</code>",
+                {{%ArrayProto_values%}}).
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>forEach</code>",
+                {{%ArrayProto_forEach%}}).
+    1.  Otherwise, if |definition| has a [=pair iterator=], then:
+        1.  Define the {{@@iterator}} and <code>entries</code> methods:
```
TimothyGu: nit: `<code class="idl">entries</code>`; here and below.

Essentially all `<code>` that refers to an identifier and not a string should have `class=idl`.

domenic:
@annevk / @EdgarChen want to do a final pass? @TimothyGu's sign-off is a pretty good sign IMO, and this is just editorial, so it should be an easy review :)

index.bs
```diff
@@ -12403,158 +12296,95 @@ property is the String value "<code>forEach</code>".
...
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>entries</code>",
+                {{%ArrayProto_entries%}}).
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>keys</code>",
+                {{%ArrayProto_keys%}}).
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>values</code>",
+                {{%ArrayProto_values%}}).
+            1.  Perform [=!=] [$CreateDataProperty$](|target|, "<code>forEach</code>",
+                {{%ArrayProto_forEach%}}).
+    1.  Otherwise, if |definition| has a [=pair iterator=], then:
+        1.  Define the {{@@iterator}} and <code class="idl">entries</code> methods:
```
annevk: I find this informative style a little weird, but I guess it's clear enough.

PR #873, opened 2020-04-25
Install @@toStringTag on namespace objects

shvaikalesh:
Follow-up of #357.

This change aligns WebIDL [namespace objects](https://heycam.github.io/webidl/#namespace-object) with ECMA-262 ones like [`Math`](https://tc39.es/ecma262/#sec-math-@@tostringtag) and [`Atomics`](https://tc39.es/ecma262/#sec-atomics-@@tostringtag).

Chrome (84.0.4126.0) already implements this PR:
```js
WebAssembly[Symbol.toStringTag] // => "WebAssembly"
```

_cc_ @domenic

***
<a href="https://pr-preview.s3.amazonaws.com/shvaikalesh/webidl/pull/873.html" title="Last updated on Jul 21, 2020, 1:48 PM UTC (083d283)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/873/350fb2a...shvaikalesh:083d283.html" title="Last updated on Jul 21, 2020, 1:48 PM UTC (083d283)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14042,7 +14042,7 @@ The characteristics of a namespace object are described in [[#namespace-object]]
     1.  Return |namespaceObject|.
 </div>
 
-The [=class string=] of a [=namespace object=] is the [=interface=]’s [=identifier=].
+The [=class string=] of a [=namespace object=] is the [=namespace=]’s [=identifier=].
 
 
 <h3 id="es-exceptions">Exceptions</h3>
--- a/index.bs
+++ b/index.bs
@@ -14042,6 +14042,8 @@ The characteristics of a namespace object are described in [[#namespace-object]]
     1.  Return |namespaceObject|.
 </div>
 
+The [=class string=] of a [=namespace object=] is the [=interface=]’s [=identifier=].
+
 
 <h3 id="es-exceptions">Exceptions</h3>
 


domenic:
Interesting. I don't think we should do this; instead we can fix Chrome. Namespace objects here are patterned after those in the JS spec, such as `Math` and `JSON`, which do not have such toStringTags.

Note also that the other namespace objects on the platform, `CSS` and `Math`, do not have such toStringTags in Chrome. So I think this is just a Chrome bug.

domenic:
Oh, I didn't click through your link. I see the JS spec is inconsistent, and `Reflect` has no toStringTag, but all the others do.

Hmm, so in that case probably we should align with the JS spec...

domenic:
I filed https://github.com/tc39/ecma262/issues/1970 to get some clarity here. We should probably wait for that to settle before going either way.

shvaikalesh:
Thank you for rapid feedback! I've removed `WebAssembly[@@toStringTag]` from WebKit's patch.

index.bs
```diff
@@ -14042,6 +14042,8 @@ The characteristics of a namespace object are described in [[#namespace-object]]
     1.  Return |namespaceObject|.
 </div>
 
+The [=class string=] of a [=namespace object=] is the [=interface=]’s [=identifier=].
```
ExE-Boss, line 14045:
```suggestion
The [=class string=] of a [=namespace object=] is the [=namespace=]’s [=identifier=].
```

domenic:
@shvaikalesh it looks like the plan is to have toStringTag on namespace objects after all, per https://github.com/tc39/ecma262/pull/2057#issuecomment-661259830. Would you be up for finishing this PR, and also adding web platform tests? (For web platform tests we should test `WebAssembly`, `console`, and `CSS`.)

ExE-Boss:
When it comes to `console`, should the `%Symbol.toStringTag%` property be installed on `console.[[Prototype]]` or on `console`?

domenic:
It should follow the normal rules for Web IDL.

shvaikalesh:
@domenic Thank you for moving this forward! The tests are at https://github.com/web-platform-tests/wpt/pull/24717. Is there something I could do to finish this PR?

Issue #875, opened 2020-04-29
Consider adding a `[CrossOriginIsolated]` extended attribute.


mikewest:
We have a [`[SecureContext]`](https://heycam.github.io/webidl/#SecureContext) extended attribute that governs whether or not a given construct is exposed within a given context. I'd suggest that something similar would be helpful as we begin restricting access to APIs based on whether or not a given context is cross-origin isolated. 

The [Securer Contexts](https://github.com/mikewest/securer-contexts) proposal talks about this attribute in terms of `[SecureContext=Isolated]`, but I now think it's probably a better idea to create a new attribute (`[CrossOriginIsolated]`) for this purpose, as it would give us the opportunity to reformulate `[SecureContext]` in the future to something more specifically related to the aspect of security it actually governs.

That is, for upcoming APIs like [`performance.measureMemory()`](https://github.com/WICG/performance-measure-memory), I'd like them to be able to define the API as limited to isolated and secure contexts in a standardized way. Perhaps along the lines of:

```
partial interface Performance {
  [SecureContext, CrossOriginIsolated] Promise<MemoryMeasurements> measureMemory();
}
```

That would have the effect of exposing the method only when the context is cross-origin isolated via COOP and COEP (as defined in https://github.com/whatwg/html/issues/4732#issuecomment-570223375).

WDYT?

@annevk @camillelamy @domenic

annevk:
Is https://github.com/whatwg/html/issues/5435 still something that is being pursued by Google? (It would be doable in Firefox, but I'm not sure we want to lead with shipping that as well.)

Note that if you have `[CrossOriginIsolated]` there should be no need to list `[SecureContext]`. It's effectively a subset.

mikewest:
> Is whatwg/html#5435 still something that is being pursued by Google?

I think so? Either @arturjanc or @ulan pinged me about it earlier this week, and it seems like a reasonable thing to do that this would layer on top of.

> Note that if you have `[CrossOriginIsolated]` there should be no need to list `[SecureContext]`. It's effectively a subset.

If we keep `[SecureContext]`, I think I'd suggest that folks list both for clarity.

(As part of Securer Contexts, I'd secretly like to deprecate `[SecureContext]` by making it the default, and annotating all unrestricted APIs with `[UnsafelyExposedToNetworkAttackers]` or something similar... but that's a bit off topic for this particular issue).

ExE-Boss:
[...]

Issue #876, opened 2020-04-29
Consider inverting the `[SecureContext]` model.


mikewest:
By default, IDL constructs are exposed to both secure and non-secure contexts, subject only to the `Exposed` attribute. Developers can opt-into stricter checks by adding `[SecureContext]` to their constructs, which limits exposure to secure contexts. This setup made a lot of sense when we introduced the concept, as secure contexts were rare and most APIs wouldn't want to restrict themselves. Given recent trends, these assumptions no longer hold: secure contexts are plentiful given user browsing habits, and many (most?) new features require secure contexts.

With this in mind, it seems reasonable to me to flip the way we talk about things in IDL around, such that constructs are limited to secure contexts by default, and only those which need to be exposed to non-secure contexts need to annotate themselves.

That is, given the following IDL:

```
[Exposed=Window]
interface NonSecureThing {
  ...
}

[SecureContext,Exposed=Window]
interface SecureThing {
  ...
}
```

I'd suggest that we instead ask ourselves to write something like the following with the same meaning:

```
[UnsafelyExposedToNetworkAttackers, Exposed=Window]
interface NonSecureThing {
  ...
}

[Exposed=Window]
interface SecureThing {
  ...
}
```

Making this change would involve a lot of churn in specs, which is unfortunate. But it seems to me to be the right direction for the future.

WDYT?

/cc @annevk (also @domenic and @wanderview, who I think are skeptical of this proposal)

annevk:
We also have #471 (require one of SecureContext and SecureContext=None).

PR #883, opened 2020-05-11
Define the `[CrossOriginIsolated]` extended attribute.

mikewest:
WebIDL currently defines a `[SecureContext]` extended attribute that
governs whether or not a given construct is exposed within a given
context. This patch defines a similar `[CrossOriginIsolated]` attribute
to govern exposure based on cross-origin isolation.

This supports the broader Securer Contexts proposal
(https://github.com/mikewest/securer-contexts), which aims to guide
spec authors to combat threats we've started paying more attention to
over the last few years.

Closes https://github.com/heycam/webidl/issues/875.

***
<a href="https://pr-preview.s3.amazonaws.com/mikewest/webidl/pull/883.html" title="Last updated on Jun 25, 2020, 10:39 AM UTC (4a46b22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/883/9a739b5...mikewest:4a46b22.html" title="Last updated on Jun 25, 2020, 10:39 AM UTC (4a46b22)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10056,46 +10056,44 @@ that does specify [{{SecureContext}}].
 
     <pre highlight="webidl">
         [Exposed=Window]
-        interface PowerfulFeature {
+        interface ExampleFeature {
           // This call will succeed in all contexts.
           Promise &lt;Result&gt; calculateNotSoSecretResult();
 
           // This operation will not be exposed to a non-secure context. In such a context,
-          // there will be no "calculateSecretResult" property on PowerfulFeature.prototype.
+          // there will be no "calculateSecretResult" property on ExampleFeature.prototype.
           [SecureContext] Promise&lt;Result&gt; calculateSecretResult();
 
           // The same applies here: the attribute will not be exposed to a non-secure context,
           // and in a non-secure context there will be no "secretBoolean" property on
-          // PowerfulFeature.prototype.
+          // ExampleFeature.prototype.
           [SecureContext] readonly attribute boolean secretBoolean;
         };
 
         // HeartbeatSensor will not be exposed in a non-secure context, nor will its members.
         // In such a context, there will be no "HeartbeatSensor" property on Window.
-        [SecureContext]
+        [Exposed=Window, SecureContext]
         interface HeartbeatSensor {
           Promise&lt;float&gt; getHeartbeatsPerMinute();
         };
 
         // The interface mixin members defined below will never be exposed in a non-secure context,
-        // regardless of whether the interface that includes them is.
-        // In a non-secure context, there will be no "snap" property on
-        // PowerfulFeature.prototype.
+        // regardless of whether the interface that includes them is. That is, in a non-secure
+        // context, there will be no "snap" property on ExampleFeature.prototype.
         [SecureContext]
         interface mixin Snapshotable {
           Promise&lt;boolean&gt; snap();
         };
-        PowerfulFeature includes Snapshotable;
+        ExampleFeature includes Snapshotable;
 
-        // On the other hand, the following interface mixin members will be exposed
-        // to a non-secure context when included by a host interface
-        // that doesn't have the [SecureContext] extended attribute.
-        // In a non-secure context, there will be a "log" property on
-        // PowerfulFeatures.prototype.
+        // On the other hand, the following interface mixin members will be exposed to a non-secure
+        // context when included by a host interface that doesn't have the [SecureContext] extended
+        // attribute. That is, in a non-secure context, there will be a "log" property on
+        // ExampleFeature.prototype.
         interface mixin Loggable {
           Promise&lt;boolean&gt; log();
         };
-        PowerfulFeatures includes Loggable;
+        ExampleFeature includes Loggable;
     </pre>
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1008,13 +1008,15 @@ The following extended attributes are applicable to interfaces:
 [{{LegacyWindowAlias}}],
 [{{LegacyFactoryFunction}}],
 [{{LegacyNoInterfaceObject}}],
-[{{LegacyOverrideBuiltIns}}], and
-[{{SecureContext}}].
+[{{LegacyOverrideBuiltIns}}],
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 The following extended attributes are applicable to [=partial interfaces=]:
 [{{Exposed}}],
-[{{LegacyOverrideBuiltIns}}], and
-[{{SecureContext}}].
+[{{LegacyOverrideBuiltIns}}],
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 [=Interfaces=] which are not annotated
 with a [{{LegacyNoInterfaceObject}}] [=extended attribute=]
@@ -1256,7 +1258,7 @@ in the <a href="#es-namespaces">ECMAScript binding</a>.
 Note that unlike [=interfaces=] or [=dictionaries=], [=interface mixins=] do not create types.
 
 Of the extended attributes defined in this specification,
-only the [{{Exposed}}] and [{{SecureContext}}] extended attributes
+only the [{{Exposed}}], [{{SecureContext}}], and [{{CrossOriginIsolated}}] extended attributes
 are applicable to [=interface mixins=].
 
 An <dfn>includes statement</dfn> is a definition
@@ -1782,7 +1784,8 @@ on which they appear.  It is language binding specific whether
 
 The following extended attributes are applicable to constants:
 [{{Exposed}}],
-[{{SecureContext}}].
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 <pre class="grammar" id="prod-Const">
     Const :
@@ -1964,7 +1967,8 @@ The following [=extended attributes=]
 are applicable to regular and static attributes:
 [{{Exposed}}],
 [{{SameObject}}],
-[{{SecureContext}}].
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 The following [=extended attributes=]
 are applicable only to regular attributes:
@@ -2418,6 +2422,7 @@ The following extended attributes are applicable to operations:
 [{{Exposed}}],
 [{{NewObject}}],
 [{{SecureContext}}],
+[{{CrossOriginIsolated}}], and
 [{{LegacyUnforgeable}}].
 
 The <dfn>method steps</dfn> of an operation |operation| should be introduced using text of the form
@@ -4278,7 +4283,8 @@ must not have a
 
 The following extended attributes are applicable to [=iterable declarations=]:
 [{{Exposed}}],
-[{{SecureContext}}].
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 <pre class="grammar" id="prod-Iterable">
     Iterable :
@@ -4468,7 +4474,8 @@ must not have a [=maplike declaration=], [=setlike declaration=], or [=iterable 
 
 The following extended attributes are applicable to [=asynchronously iterable declarations=]:
 [{{Exposed}}],
-[{{SecureContext}}].
+[{{SecureContext}}], and
+[{{CrossOriginIsolated}}].
 
 Issue: these [=extended attributes=] are not currently taken into account.
 When they are, the effect will be as you would expect.
@@ -4718,7 +4725,7 @@ The order that members appear in has significance for property enumeration in th
 
 Note that unlike interfaces or dictionaries, namespaces do not create types.
 
-Of the extended attributes defined in this specification, only the [{{Exposed}}] and [{{SecureContext}}] extended attributes are applicable to namespaces.
+Of the extended attributes defined in this specification, only the [{{Exposed}}], [{{SecureContext}}], and [{{CrossOriginIsolated}}] extended attributes are applicable to namespaces.
 
 [=Namespaces=] must be annotated with the [{{Exposed}}] [=extended attribute=].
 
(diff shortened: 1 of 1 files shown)


mikewest:
Hey folks! This can't actually land as-is, given that "cross-origin isolated" is still floating around in a set of PRs against HTML, but I'm hopeful that y'all can give me some direction about the mechanics of defining the `[CrossOriginIsolated]` attribute in the meantime.

This patch does the simplest thing possible, by copy/pasting `[SecureContext]` and munging it a bit to define exposure restrictions based upon the "cross-origin isolated" concept rather than the "secure context" concept. This is possibly fine? I considered pulling out the common bits into an "exposure attribute" concept, which would make it simpler to define these two concepts, as well as an attribute related to XSS mitigation that I haven't written up in detail yet (what https://github.com/mikewest/securer-contexts currently calls `[SecureContext=Injection]`).

WDYT about the approach? Would it be worth going back to making "exposure attribute" a generic concept? Or is a bit of duplication better for clarity?

/cc @annevk @camillelamy @domenic

domenic:
My take is that at least the "available only in cross-origin isolated contexts" / "available only in secure contexts" bits should be deduped into a single parametrized algorithm. (I noticed one [SecureContext] was not changed to [CrossOriginIsolated] in the former ^_^.) The duplication for the rest seems more clarity-enhancing, since it is kind of general "extended attribute boilerplate" that you expect to see for all of them.

Another nit: extended attributes are alphabetized :).

annevk:
I miss a statement that SecureContext and CrossOriginIsolated are mutually exclusive. I think like elsewhere, we should not allow duplicate requirements.

mikewest:
Hrm. I noted my preference for the opposite in https://github.com/heycam/webidl/issues/875#issuecomment-621162997. :)

But, since I kinda want to throw `[SecureContext]` away in any event, I can live with either. Would you prefer that we make `[SecureContext, CrossOriginIsolated]` an error, or is noting that "`[CrossOriginIsolated]` implies the `[SecureContext]` restrictions as a prerequisite" enough?

annevk:
I think it should be an error.

mikewest:
> I think it should be an error.

I can write something up that makes `[SecureContext, CrossOriginIsolated]` an error (e.g. "The `[SecureContext]` extended attribute must not appear on a construct that also specifies `[CrossOriginIsolated]`." Do you also want to do some sort of inheritance checking to ensure that it doesn't appear on e.g. a method inside an interface that defines `CrossOriginIsolated`? If so, are there any examples of that kind of check in the text that I could cargo-cult? :)

annevk:
SecureContext has some of those checks specific to only SecureContext, e.g., "The [SecureContext] extended attribute must not be specified both on an interface member and its interface or partial interface." I guess you could make "The [SecureContext] extended attribute must not be specified on an interface member if its interface or partial interface has the [CrossOriginIsolated] extended attribute." from that. This is not my area of expertise though.

@saschanaz might have better ideas.

saschanaz:
Sounds like it needs two rules, one to prevent descendants to use `[CrossOriginIsolated]` when there is `[SecureContext]` and vise versa.

I wonder we could rename `[SecureContext]` to `[Context=Secure]` and `[CrossOriginIsolated]` to `[Context=CrossOriginIsolated]` so that we can have only one check: whether descendants have `[Context]` or not.

mikewest:
> Sounds like it needs two rules, one to prevent descendants to use `[CrossOriginIsolated]` when there is `[SecureContext]` and vise versa.

I think we'd have one rule for `[CrossOriginIsolated]` and another for `[SecureContext]` along the lines of what @annevk suggested.

> I wonder we could rename `[SecureContext]` to `[Context=Secure]` and `[CrossOriginIsolated]` to `[Context=CrossOriginIsolated]` so that we can have only one check: whether descendants have `[Context]` or not.

This is more or less the spelling of the proposal in https://github.com/mikewest/securer-contexts#a-proposal (`[SecureContext=(Transport,Isolation,Injection)]`). That's a fine approach if it's what we end up wanting. I think I'd prefer distinct attributes, however, as that would allow us to invert the `[SecureContext]` model entirely (https://github.com/heycam/webidl/issues/876).

saschanaz:
> I think I'd prefer distinct attributes, however, as that would allow us to invert the [SecureContext] model entirely (#876).

I think my suggestion still allows inversion: `[Context=Insecure]` or such.

mikewest:
> I think my suggestion still allows inversion: `[Context=Insecure]` or such.

That's fair. I think the inversion is clearer as distinct attributes, but it certainly works as `[Context=UnsafelyExposedToNetworkAttackers]` or something similar. I can certainly live with that spelling.

That said, let's go back to your suggestion that:

> ... we can have only one check: whether descendants have `[Context]` or not.

This works for secure transport and cross-origin isolation, as the latter implies the former. Injection mitigation is currently orthogonal, evaluating whether a CSP is present, not whether the context is securely delivered. Perhaps we'd say that any contextual exposure would require secure contexts?

mikewest:
@annevk: Any objections to something like `[ExposureConditions=(type, type, type)]`?

EdgarChen:
> @annevk: Any objections to something like `[ExposureConditions=(type, type, type)]`?

I lean to this, but it probably should only allow one argument, given that there is no use case to have multiple arguments so far?

And maybe we could also consider having `LegacyNone` or such as the default exposure condition if there is no exposure condition specified explicitly. If we decide to move to #876 model in the future, we could then switch the default value to `SecureContext`.

EdgarChen:
> > ... we can have only one check: whether descendants have `[Context]` or not.
> 
> This works for secure transport and cross-origin isolation, as the latter implies the former. Injection mitigation is currently orthogonal, evaluating whether a CSP is present, not whether the context is securely delivered. Perhaps we'd say that any contextual exposure would require secure contexts?

Another point is that do we allow a interface with `SecureContext` to contain a member with `CrossOriginIsolated` or a interface with `CrossOriginIsolated` inherits from another interface with `SecureContext`? It seems they are valid use cases to me.

mikewest:
> I lean to this, but it probably should only allow one argument, given that there is no use case to have multiple arguments so far?

As hand-waved at above, I do think we'll end up with multiple arguments. I have concrete plans for at least `CrossOriginIsolated` and `SomethingSomethingInjectionMitigation` which have non-overlapping requirements. I'm happy to consider both as requiring secure transport, but the one doesn't otherwise imply the other.

> And maybe we could also consider having LegacyNone or such as the default exposure condition if there is no exposure condition specified explicitly.

I'd be a bit more aggressive with the naming; `UnsafelyExposedToNetworkAttackers` wasn't really a joke. :)

> Another point is that do we allow a interface with `SecureContext` to contain a member with `CrossOriginIsolated` or a interface with `CrossOriginIsolated` inherits from another interface with `SecureContext`? It seems they are valid use cases to me.

Yes. We would want to support this.

saschanaz:
> > Another point is that do we allow a interface with `SecureContext` to contain a member with `CrossOriginIsolated` or a interface with `CrossOriginIsolated` inherits from another interface with `SecureContext`? It seems they are valid use cases to me.
> 
> Yes. We would want to support this.

So we want them to be *not* mutually exclusive?

EdgarChen:
> As hand-waved at above, I do think we'll end up with multiple arguments. I have concrete plans for at least `CrossOriginIsolated` and `SomethingSomethingInjectionMitigation` which have non-overlapping requirements. I'm happy to consider both as requiring secure transport, but the one doesn't otherwise imply the other.

I see, so if two conditions have non-overlapping requirements, then (`Condition1`, `Condition2`) could have two meanings, *OR* condition or *AND* condition. If we would want to support both combinations, we need a way to differentiate them.

mikewest:
> I see, so if two conditions have non-overlapping requirements, then (Condition1, Condition2) could have two meanings, OR condition or AND condition. If we would want to support both combinations, we need a way to differentiate them.

I think _AND_ is the only combination that makes sense.

mikewest:
> > > Another point is that do we allow a interface with SecureContext to contain a member with CrossOriginIsolated or a interface with CrossOriginIsolated inherits from another interface with SecureContext? It seems they are valid use cases to me.
>
> > Yes. We would want to support this.
>
> So we want them to be not mutually exclusive?

I want the following to be legal so that we can restrict new things added to old interfaces (I'm thinking of the `Performance` interface in particular, which isn't currently locked to secure contexts, but wouldn't it be nice if it was?).

```
[SecureContext]
interface VerySecureThing {
  Promise<float> secureNumber();
  [CrossOriginIsolated] Promise<boolean> verySecureBoolean(); 
}
```

@annevk does not want the following to be legal:

```
[SecureContext,CrossOriginIsolated]
interface extraSuperSecureThing {
  ...
};

[CrossOriginIsolated]
interface SecureThing {
  ...
  [SecureContext] bool verySecureBoolean();
};

interface RegularThing {
  ...
  [SecureContext,CrossOriginIsolated] bool verySecureBoolean();
};
```

So, `SecureContext` should not be specified when `CrossOriginIsolated` applies to a construct. The inverse is possible, as `CrossOriginIsolated` is more restrictive than `SecureContext`.

(Sorry this is confusing.)

domenic:
This looks pretty close to landing, and https://github.com/WICG/performance-measure-memory/ would like to use it. I'm happy to carry this over the finish line, but can someone confirm for me what's left to do? My read is:

- Update to cross-link to the HTML spec definition of "cross-origin isolated", now that it exists. Since it lives on an agent cluster I guess we'll need to indirect a bit; something like global -> relevant agent -> agent cluster -> cross-origin isolated.

- Specify the restrictions @annevk prefers on doubling this up with [SecureContext]. In particular, I think it suffices to say something like "[SecureContext] must not be specified when a construct is conditionally exposed on [CrossOriginIsolated]; doing so would be redundant, since every environment which is cross-origin isolated is also a secure context".

mikewest:
I would love for you to pick this up, @domenic. I have not had time to work out the right way to address your second bullet above, which seems like the only complicated bit left. If simple prose assertions make other folks on the thread happy, I'm certainly fine with them. :)

domenic:
OK, I think this is ready then. Instead of agent cluster's cross-origin isolated, I went with environment settings object's cross-origin isolated capability, since we want the exposure here to be controlled by `allow="cross-origin-isolated"`.

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+that does specify [{{CrossOriginIsolated}}].
+
+<div class="example">
+
+    The following [=IDL fragment=] defines an interface with one [=operation=] that is executable
+    from all contexts, and two which are executable only from cross-origin isolated contexts.
+
+    <pre highlight="webidl">
+        [Exposed=Window]
+        interface PowerfulFeature {
```
annevk, line 9160:
```suggestion
        interface ExampleFeature {
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+          // This call will succeed in all contexts.
+          Promise &lt;Result&gt; calculateNotSoSecretResult();
+
+          // This operation will not be exposed to a non-isolated context. In such a context,
+          // there will be no "calculateSecretResult" property on PowerfulFeature.prototype.
```
annevk, line 9165:
```suggestion
          // there will be no "calculateSecretResult" property on ExampleFeature.prototype.
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+          [CrossOriginIsolated] Promise&lt;Result&gt; calculateSecretResult();
+
+          // The same applies here: the attribute will not be exposed to a non-isolated context,
+          // and in a non-secure context there will be no "secretBoolean" property on
+          // PowerfulFeature.prototype.
```
annevk, line 9170:
```suggestion
          // ExampleFeature.prototype.
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+        // In such a context, there will be no "HighResolutionTimer" property on Window.
+        [CrossOriginIsolated]
+        interface HighResolutionTimer {
+          DOMHighResTimeStamp getHighResolutionTime();
+        };
+
+        // The interface mixin members defined below will never be exposed in a non-isolated context,
+        // regardless of whether the interface that includes them is.
+        // In a non-isolated context, there will be no "snap" property on
+        // PowerfulFeature.prototype.
```
annevk, line 9184:
```suggestion
        // ExampleFeature.prototype.
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+        [CrossOriginIsolated]
+        interface mixin Snapshotable {
+          Promise&lt;boolean&gt; snap();
+        };
+        PowerfulFeature includes Snapshotable;
```
annevk, line 9189:
```suggestion
        ExampleFeature includes Snapshotable;
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+
+        // On the other hand, the following interface mixin members will be exposed
+        // to a non-isolated context when included by a host interface
+        // that doesn't have the [CrossOriginIsolated] extended attribute.
+        // In a non-secure context, there will be a "log" property on
+        // PowerfulFeatures.prototype.
```
annevk, line 9195:
```suggestion
        // ExampleFeatures.prototype.
```

index.bs
```diff
@@ -9107,6 +9114,93 @@ for the specific requirements that the use of
...
+        interface mixin Loggable {
+          Promise&lt;boolean&gt; log();
+        };
+        PowerfulFeatures includes Loggable;
```
annevk, line 9199:
```suggestion
        ExampleFeatures includes Loggable;
```

annevk (approved):
Thanks this looks good. I rather not call things PowerfulFeature as that tends to create confusion.

Are the 4 newlines before a `<h4>` intended?

PR #885, opened 2020-05-20
Use an "end of iteration" signal instead of undefined

domenic:
The use of undefined as a sentinel for the end of iteration was problematic for value iterators, where the possible value space could contain undefined.

@TimothyGu, could you take a look?

There's another issue lurking here which is that promises should only be resolved with either Web IDL values (if these are Web IDL `Promise<T>`s) or JS values (if these are JS promises). I considered adding a note explaining that end-of-iteration could be represented as a unique Web IDL/JS object with identity, and a pair could be represented as an appropriate Web IDL dictionary/JS null-prototype object, but I couldn't decide what type of promise these were. I guess we can figure that out as a followup, if we really want to.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/885.html" title="Last updated on May 21, 2020, 4:24 PM UTC (f9bab58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/885/4a3c045...f9bab58.html" title="Last updated on May 21, 2020, 4:24 PM UTC (f9bab58)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4354,8 +4354,8 @@ Prose accompanying an [=interface=] with an [=asynchronously iterable declaratio
 <dfn id="dfn-get-the-next-iteration-result" export>get the next iteration result</dfn> algorithm.
 This algorithm receives the instance of the [=interface=] that is being iterated, as well as the
 async iterator itself (which can be useful for storing state).
-It must return a {{Promise}} that either rejects, resolves with <dfn export>end of iteration</dfn>
-to signal the end of the iteration, or resolves with one of the following:
+It must return a {{Promise}} that either rejects, resolves with a special <dfn export>end of
+iteration</dfn> value to signal the end of the iteration, or resolves with one of the following:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4354,13 +4354,13 @@ Prose accompanying an [=interface=] with an [=asynchronously iterable declaratio
 <dfn id="dfn-get-the-next-iteration-result" export>get the next iteration result</dfn> algorithm.
 This algorithm receives the instance of the [=interface=] that is being iterated, as well as the
 async iterator itself (which can be useful for storing state).
-It must return a {{Promise}} that either rejects, resolves with undefined to signal the end of the
-iteration, or resolves with one of the following:
+It must return a {{Promise}} that either rejects, resolves with <dfn export>end of iteration</dfn>
```
TimothyGu, line 4357: Maybe this could make it clearer? "a special **end of iteration** value"

PR #898, opened 2020-06-18
add links for dotted intrinsic form

### LINKED ISSUE #897: Parsing updated ecma262 intrinsic notation
Hi!

In the ecma262 spec, we've added a [new intrinsic notation form](https://github.com/tc39/ecma262/pull/1376) that will avoid unbounded growth of our list of intrinsics.

Specifically, things like `%ArrayPrototype%` and `%ObjProto_valueOf%` should now be represented as `%Array.prototype%` and `%Object.prototype.valueOf%`, respectively.

Our intention is to delete the "legacy" forms entirely in favor of the dotted forms. I wanted to file a PR on https://github.com/whatwg/html, but then realized I might need to file one on https://github.com/heycam/webidl/blob/master/index.bs#L88-L110 first (I filed https://github.com/w3c/webidl2.js/issues/480 before this, and was directed here).

I'm happy to do the work to update spec text (and for HTML, I have a branch prepared already), but I doubt I'm qualified to do the work here. Could someone help give me pointers as to what the next steps should be?

In particular, one approach that occurred to me (that I could handle!) would be to add to the current "hardcoded legacy intrinsics mentioned in HTML" the list of "hardcoded dotted intrinsics mentioned in HTML". That seems simplest, but has the downside (one which already exists now) that a PR would still have to be made here to support autolinking a new intrinsic. The ideal approach (which perhaps could be made later) would be one that figures out from any `%A.b%` which section it should link to.

ljharb:
In the ecma262 spec, we've added a [new intrinsic notation form](https://github.com/tc39/ecma262/pull/1376) that will avoid unbounded growth of our list of intrinsics.

Specifically, things like `%ArrayPrototype%` and `%ObjProto_valueOf%` should now be represented as `%Array.prototype%` and `%Object.prototype.valueOf%`, respectively.

Our intention is to delete the "legacy" forms entirely in favor of the dotted forms.

Closes #897.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -86,28 +86,26 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: Set; url: sec-set-objects
         text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
         text: %AsyncIteratorPrototype%; url: sec-asynciteratorprototype
-        text: %ArrayPrototype%; url: sec-properties-of-the-array-prototype-object
-        text: %ErrorPrototype%; url: sec-properties-of-the-error-prototype-object
-        text: %FunctionPrototype%; url: sec-properties-of-the-function-prototype-object
+        text: %Array.prototype%; url: sec-properties-of-the-array-prototype-object
+        text: %Error.prototype%; url: sec-properties-of-the-error-prototype-object
+        text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
         text: %IteratorPrototype%; url: sec-%iteratorprototype%-object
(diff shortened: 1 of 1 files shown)


ljharb:
My assumption was that I'd land this first, and then I'd be able to land the PR in HTML, but if they can go in parallel I'll file that one now.

domenic:
Hmm, I wasn't discussing anything in HTML in my above comment; I was talking purely about how to change the links that show up in e.g. https://heycam.github.io/webidl/#es-iterable

ljharb:
aha, thanks, that helps. updating this now.

PR #902, opened 2020-07-09
Propagate active script to callbacks

domenic:
See discussion in https://github.com/tc39/ecma262/pull/2086#issuecomment-656257061.

This ensures that in cases like

```html
<!-- index.html -->
<script src="sub/path/foo.js"></script>
```

```js
// sub/path/foo.js
setTimeout(eval, 0, 'import(`./example.mjs`)');
```

the resulting imported module is `sub/path/example.mjs`, and not `example.mjs`. I.e. it ensures that `import()` resolves relative to the active script, even when it is being called via a Web IDL callback.

I will write web platform tests for this too.

/cc @hiroshige-g as he was involved in very similar discussions for cases like `` setTimeout("import(`./example.mjs`)") `` and @yuki3 as the Chromium binding person working on incumbents.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,39 +40,7 @@ spec: ecma-262; type: dfn; for: /; text:realm
 </pre>
 
 <pre class="anchors">
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_astc/; spec: WEBGL_compressed_texture_astc
-    type: interface; text: WEBGL_compressed_texture_astc
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_s3tc_srgb/; spec: WEBGL_compressed_texture_s3tc_srgb
-    type: interface; text: WEBGL_compressed_texture_s3tc_srgb
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_draw_buffers/; spec: WEBGL_draw_buffers
-    type: interface; text: WEBGL_draw_buffers
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_lose_context/; spec: WEBGL_lose_context
-    type: interface; text: WEBGL_lose_context
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/ANGLE_instanced_arrays/; spec: ANGLE_instanced_arrays
-    type: interface; text: ANGLE_instanced_arrays
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/EXT_blend_minmax/; spec: EXT_blend_minmax
-    type: interface; text: EXT_blend_minmax
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/; spec: EXT_color_buffer_float
-    type: interface; text: EXT_color_buffer_float
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/EXT_disjoint_timer_query/; spec: EXT_disjoint_timer_query
-    type: interface; text: EXT_disjoint_timer_query
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/OES_standard_derivatives/; spec: OES_standard_derivatives
-    type: interface; text: OES_standard_derivatives
-urlPrefix: https://www.khronos.org/registry/webgl/extensions/OES_vertex_array_object/; spec: OES_vertex_array_object
-    type: interface; text: OES_vertex_array_object
-urlPrefix: https://www.w3.org/TR/geolocation-API/; spec: GEOLOCATION-API
-    type: interface
-        text: Geolocation; url: geolocation;
-        text: Coordinates; url: coordinates;
-        text: Position; url: position;
-        text: PositionError; url: position-error;
-urlPrefix: https://w3c.github.io/deviceorientation/spec-source-orientation.html; spec: ORIENTATION-EVENT
-    type: interface
-        text: DeviceRotationRate; url: device_rotation_rate;
-        text: DeviceAcceleration; url: device_acceleration;
-urlPrefix: https://w3c.github.io/mediacapture-main/; spec: MEDIACAPTURE-STREAMS
-    type: interface
-        text: ConstrainablePattern; url: dom-constrainablepattern;
+url: https://html.spec.whatwg.org/multipage/webappapis.html#active-script; spec: HTML; type: dfn; text: active script
 urlPrefix: http://www.unicode.org/glossary/; spec: UNICODE
     type: dfn
         text: Unicode scalar value; url: unicode_scalar_value
@@ -6253,8 +6221,8 @@ about the execution context at the time the language binding specific object ref
 converted to an IDL value.
 
 Note: For ECMAScript objects, the [=callback context=] is used to hold a reference to the
-[=incumbent settings object=] at the time the Object value is converted to an IDL callback
-interface type value. See [[#es-callback-interface]].
+[=incumbent settings object=] and [=active script=] at the time the Object value is converted to an
+IDL callback interface type value. See [[#es-callback-interface]].
 
 There is no way to represent a constant object reference value for a particular
 [=callback interface type=] in IDL.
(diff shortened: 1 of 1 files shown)


yuki3:
Quite interesting.  Thanks for letting me know.  :)

domenic:
Based on the test results at https://github.com/web-platform-tests/wpt/pull/24535, nobody seems to implement this yet. Even Firefox, who does great on [the existing tests](https://wpt.fyi/results/html/semantics/scripting-1/the-script-element/module/dynamic-import?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=%2Fhtml%2Fsemantics%2Fscripting-1%2Fthe-script-element%2Fmodule%2Fdynamic-import%2F).

However, I think this is definitely the right thing to do. Can you agree on Chrome's behalf, @yuki3 and/or @hiroshige-g? It's fine if it's lower priority, but I think we should make sure this happens...

yuki3:
I don't fully understand about 'import', but I totally agree that 'eval's incumbent must be the realm of the caller of 'setTimeout'.  Thus, I agree with the test.

AFAIK, in Chromium, setTimeout + IDL callback function already supports the backup incumbent stack, so probably this is @hiroshige-g's area, I think?

domenic:
Basically, this PR is adding another "incumbent-like thing" that gets propagated in the same way as incumbent: namely, the active script. In particular, the active script's base URL is used for `import()`.

So the example `setTimeout(eval, 0, 'import(`./example.mjs`)');` is about propagating the active script, not about propagating the incumbent. (The incumbent would come into play with an example like `setTimeout(eval, 0, 'postMessage(...)')`.)

So indeed, this is a bit more in @hiroshige-g's area. But, it will likely impact the bindings code, because it will need propagation through Web IDL callback conversion + calling in the same way incumbent currently does.

yuki3:
Oh my.  I'm sorry.  Now I read this PR a little more carefully.  IIUC, `active script` is not always the script or module of the `incumbent realm`, so we have to track `active script` in addition to `incumbent`.  It means another runtime cost.  Is this correct?

If so, it will be 25% object size increase of all IDL callback functions and IDL callback interfaces (in Blink).  Probably it's okay, but if there is a guarantee that, in case of IDL callbacks, the incumbent settings object's execution context's ScriptOrModule never be null, that's great.  (Anyway, I think IDL callbacks will be okay.)

I think it's better to notify those who will implement Promise callbacks and WeakRef callbacks of Chromium.  They will need to implement not only the incumbent but also the active script.

Summary:
1. I think we can support active script in IDL callback functions/interfaces.  But not sure about Promise and WeakRef callbacks.
2. Q: Is there a case that IDL callback's incumbent's execution context's ScriptOrModule is null?

domenic:
Yes, I was a bit worried about the runtime costs. I'm glad you think that cost will be OK, but don't hesitate to push back if you think it is too expensive. Although I think this change is a good one, I don't think it's infinitely good; if it has high costs then we should reevaluate.

> I think it's better to notify those who will implement Promise callbacks and WeakRef callbacks of Chromium. They will need to implement not only the incumbent but also the active script.

Yes, @syg is helping with that, and that is actually what prompted this discussion. Currently promises are specced to propagate active script, and they do in Chromium; we were working on generalizing that to WeakRef, but noticed that there was an inconsistency with Web IDL callbacks.

> Q: Is there a case that IDL callback's incumbent's execution context's ScriptOrModule is null?

I think it is never null. By "incumbent's execution context" you probably mean the execution context in step 1 of https://html.spec.whatwg.org/#incumbent-settings-object? In that case then it is definitely never null.

This is very interesting. I guess your idea is that Chromium would store the "topmost script-having execution context" (i.e. "incumbent's execution context") instead of storing the (incumbent settings object, active script) tuple? Then it could infer active settings object using the algorithm in https://html.spec.whatwg.org/#incumbent-settings-object, and active script using the ScriptOrModule component.

I think that works, and means no extra runtime cost. (At least, in spec terms; I am not sure how it maps to Chromium implementation primitives.)

yuki3:
> I think it is never null. By "incumbent's execution context" you probably mean the execution context in step 1 of https://html.spec.whatwg.org/#incumbent-settings-object? In that case then it is definitely never null.

Yes, I meant it, and I had the same idea with yours.  That's a good news that we can avoid increase of memory consumption about IDL callbacks.  :)  Then, I have no concern on this PR.

domenic:
Excellent. I will probably update this PR to use that technique, then, to make it more obvious for implementers.

syg:
I'll update https://github.com/whatwg/html/pull/5722 to follow suit so we have a common mechanism.

domenic:
Hmm, now I am worried about the case where the topmost script-having execution context is null. In that case we still need to capture the incumbent settings object as per the backup stack.

I think this could still be done, but we would have to convert the backup incumbent settings object stack into a backup incumbent _execution context_ stack.

That is a larger change. So I would prefer to land this plus whatwg/html#5722 first using the mechanism from this PR, i.e. an (incumbent settings object, active script) tuple. Then I can follow up with a refactoring to consolidate everything to use execution contexts.

hiroshige-g:
+1 for making active script the ScriptOrModule component of the incumbent.
Active script and incumbent have similar intention, and the concept of incumbent is already complicated (e.g. https://github.com/whatwg/html/issues/1430). It's better to merge active script into incumbent and maintain the incumbent spec/impl as the single source of truth.

I also suspect the current browsers implementation is more like referring the incumbent rather than the current specification of active script, according to cases where a script triggers event firing synchronously.
WPT: https://github.com/web-platform-tests/wpt/pull/24305
In such cases, https://html.spec.whatwg.org/multipage/webappapis.html#skip-when-determining-incumbent-counter makes the incumbent's script and the active script different.

(Thanks @yuki3, the chat we had today was quite helpful for me!)

domenic:
> +1 for making active script the ScriptOrModule component of the incumbent.

This change will require more 262-side changes, @syg. In particular https://tc39.es/ecma262/#sec-getactivescriptormodule would not longer be the correct algorithm; we'd need instead to skip execution contexts whose skip-when-determining-incumbent counter is non-zero.

syg:
Oh boy. That makes me very wary to support such a unification...

domenic:
Yeah, I can understand that. Here is my proposal:

1. Land https://github.com/whatwg/html/pull/5722 (tracking active script/incumbent for promises correctly)
1. Land https://github.com/whatwg/html/pull/4571 on top of that (weak references, with proper active script/incumbent tracking)
1. Land this PR and its associated tests, in basically this form. (Tracking active script for Web IDL callbacks)
1. Investigate unification, i.e. no longer tracking active script and incumbent separately, but instead tracking incumbent execution context, and deriving active script from that.
  - This could involve tests like @hiroshige-g's above. Or even simplifications of incumbent to make it more like active script.

I will set aside my (local on-disk) unification pull request, so we can proceed incrementally. So at this point I think the ball is in other folks' courts. In particular, @syg is working on those HTML pull requests (and their 262 dependencies), and it would be good to get Mozilla or Apple input on the question posed by this PR and its associated tests about

```js
// sub/path/foo.js
setTimeout(eval, 0, 'import(`./example.mjs`)');
```

annevk:
I discussed it briefly with @jonco3 and neither of us feel strongly.

littledan:
This change feels "correct" to me, but I'm curious, does it affect any cases besides when `eval` (or possibly other kinds of functions implemented by the platform which parse JavaScript) is used *directly* as a WebIDL callback? This feels like a relatively narrow case; it doesn't seem so bad to fall back to the document's base URL in this edge case.

I'm wondering how much harm the current semantics of falling back to `import()` being relative to the document's base URL is causing for web developers. Have we seen any bug reports from web developers confused about these semantics, for example?

domenic:
I'm starting to come around to @littledan's point of view, that while this is the correct thing to do, but is not worth the implementation effort.

> does it affect any cases besides when eval (or possibly other kinds of functions implemented by the platform which parse JavaScript) is used directly as a WebIDL callback?

It affects any case where the platform function is dependent on the active script. I think in practice that's only going to be things like `eval` and `setTimeout`, because the only thing that currently depends on the active script is `import()` path resolution, and the only thing that triggers `import()` is author code.

We may want to remove active script tracking from job callbacks as well, since @hiroshige-g is discovering that no browsers seem to do that either?

hiroshige-g:
FYI The test that confirms there seem no active scripts for `Promise.resolve('import(...)').then(eval)` on browsers is https://github.com/web-platform-tests/wpt/pull/30241.

Issue #905, opened 2020-08-04
Getting rid of optional


annevk:
If we do #60, should we get rid of optional and instead use defaulting as sole mechanism?
```
  [NewObject] Document createHTMLDocument(optional DOMString title);
```
would become
```
  [NewObject] Document createHTMLDocument(DOMString title = void);
```
and
```
  void observe(Node target, optional MutationObserverInit options = {});
```
would become
```
  void observe(Node target, MutationObserverInit options = {});
```
(And if you need to distinguish between not given and undefined you use overloading, as XMLHttpRequest does for `open()`.)

cc @saschanaz @tabatkins

saschanaz:
Sounds just right, no objection here.

domenic:
This is kind of nice. And argues more strongly for `undefined` instead of `void`.

tabatkins:
I don't have a strong opinion here; I like the `optional` notation, but I'm fine with getting rid of it in favor of just being more JS-y.

farukrana:
All good

saschanaz:
A late question:

```
void foo(boolean abc = false); // okay, `false` is a boolean
void foo(float abc = 1.0); // okay, 1.0 is a float, sure.

void foo(DOMString abc = undefined);
// well, is `undefined` a DOMString? 🤔
```

I'm also a bit worried that the implementation would be different while the syntax is similar: `DOMString bar = "str"` is always a string while `DOMString abc = undefined` is something like `Optional<DOMString>`.

domenic:
I agree that this is not like other defaults. But I think that's OK.

Another way to think about it is to realize how weird the current `optional` keyword is. In `foo(boolean abc = false)`, the `abc` argument is "optional": the web developer doesn't have to pass it in. But we don't use the `optional` keyword for that---because it's redundant with the default value. Similarly, if we can use `undefined` as a default value, then adding the `optional` keyword becomes redundant for the `foo(DOMString abc = undefined)` case.

In other words, it is good to align the case of "allows no argument and gets a default value of `false`" with "allows no argument and gets a default value of `undefined`".

saschanaz:
> But we don't use the `optional` keyword for that---because it's redundant with the default value.

Currently we do use `optional` keyword for every optional argument, and specifying a default value without `optional` is a syntax error.

```
ArgumentRest ::
    optional TypeWithExtendedAttributes ArgumentName Default
    Type Ellipsis ArgumentName
```

domenic:
[...]

PR #906, opened 2020-08-05
void -> undefined, and allow it anywhere.

### LINKED ISSUE #60: add a type to represent the undefined value
It's too awkward to deal with the undefined value by having to use `any`. Let's introduce a new type (`undefined`?) to represent that one value. Some care will be needed to make sure we handle optional arguments, distinguishability, etc. properly.

tabatkins:
Fixes #60

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -557,9 +557,9 @@ in [[#es-extended-attributes]].
 
           attribute Paint currentPaint;
 
-          void drawRectangle(double x, double y, double width, double height);
+          undefined drawRectangle(double x, double y, double width, double height);
 
-          void drawText(double x, double y, DOMString text);
+          undefined drawText(double x, double y, DOMString text);
         };
     </pre>
 
@@ -768,7 +768,7 @@ across [=IDL fragments=].
     <pre highlight="webidl">
         [Exposed=Window]
         interface B : A {
-          void f(SequenceOfLongs x);
+          undefined f(SequenceOfLongs x);
         };
 
         [Exposed=Window]
@@ -882,14 +882,14 @@ accessed on the object.
     <pre highlight="webidl">
         [Exposed=Window]
         interface A {
-          void f();
-          void g();
+          undefined f();
+          undefined g();
         };
 
         [Exposed=Window]
         interface B : A {
-          void f();
-          void g(DOMString x);
+          undefined f();
+          undefined g(DOMString x);
         };
     </pre>
 
@@ -1151,11 +1151,11 @@ The <dfn>qualified name</dfn> of an [=interface=] |interface| is defined as foll
           readonly attribute DOMString nodeName;
           readonly attribute Node? parentNode;
           Node appendChild(Node newChild);
-          void addEventListener(DOMString type, EventListener listener);
+          undefined addEventListener(DOMString type, EventListener listener);
         };
 
         callback interface EventListener {
-          void handleEvent(Event event);
+          undefined handleEvent(Event event);
         };
     </pre>
 
@@ -1318,7 +1318,7 @@ No [=extended attributes=] defined in this specification are applicable to [=inc
         };
 
         interface mixin Observable {
-          void addEventListener(DOMString type,
+          undefined addEventListener(DOMString type,
                                 EventListener listener,
                                 boolean useCapture);
           // ...
@@ -1791,9 +1791,15 @@ The following extended attributes are applicable to constants:
 
 <pre class="grammar" id="prod-ConstValue">
     ConstValue :
+        UndefinedLiteral
         BooleanLiteral
         FloatLiteral
         integer
+</pre>
+
+<pre class="grammar" id="prod-UndefinedLiteral">
+    UndefinedLiteral :
+        "undefined"
 </pre>
 
 <pre class="grammar" id="prod-BooleanLiteral">
@@ -2120,7 +2126,6 @@ defined on the same [=interface=].
 The <dfn id="dfn-return-type" export>return type</dfn> of the operation is given
 by the type (matching <emu-nt><a href="#prod-ReturnType">ReturnType</a></emu-nt>)
 that appears before the operation’s optional [=identifier=].
-A return type of {{void}} indicates that the operation returns no value.
 If the return type is an
 [=identifier=] followed by <emu-t>?</emu-t>,
 then the identifier must
@@ -2190,8 +2195,8 @@ language bindings.
           boolean isMouseOver();
 
           // Overloaded operations.
-          void setDimensions(Dimensions size);
-          void setDimensions(unsigned long width, unsigned long height);
+          undefined setDimensions(Dimensions size);
+          undefined setDimensions(unsigned long width, unsigned long height);
         };
     </pre>
 </div>
@@ -2231,8 +2236,8 @@ when the <emu-t>...</emu-t> token is used in their argument lists.
         interface IntegerSet {
           readonly attribute unsigned long cardinality;
 
-          void union(long... ints);
-          void intersection(long... ints);
+          undefined union(long... ints);
+          undefined intersection(long... ints);
         };
     </pre>
 
@@ -2520,7 +2525,6 @@ steps are:” followed by a list, or “The <code>|operation|(<var ignore>arg1</
 <pre class="grammar" id="prod-ReturnType">
     ReturnType :
         Type
-        "void"
 </pre>
 
 <h5 id="idl-tojson-operation">toJSON</h5>
@@ -2770,7 +2774,7 @@ will not be such functionality.
           readonly attribute unsigned long propertyCount;
 
           getter double (DOMString propertyName);
-          setter void (DOMString propertyName, double propertyValue);
+          setter undefined (DOMString propertyName, double propertyValue);
         };
     </pre>
 
@@ -2795,7 +2799,7 @@ simplify [=method steps=] of an interface’s operations.
           readonly attribute unsigned long propertyCount;
 
           getter double getProperty(DOMString propertyName);
-          setter void setProperty(DOMString propertyName, double propertyValue);
+          setter undefined setProperty(DOMString propertyName, double propertyValue);
         };
     </pre>
     <pre highlight="webidl">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2520,7 +2525,6 @@ steps are:” followed by a list, or “The <code>|operation|(<var ignore>arg1</
 <pre class="grammar" id="prod-ReturnType">
     ReturnType :
         Type
```
saschanaz, line 2527: Is this needed anymore?

tabatkins: Not really, but I left it in because maybe it's useful to distinguish that in the future? I can just fold it away if y'all think that would be better.

index.bs
```diff
@@ -14547,7 +14549,7 @@ are passed to it or what kind of value is returned from it.
 
 <h3 id="VoidFunction" callback>VoidFunction</h3>
 
-<pre class="idl">callback VoidFunction = void ();</pre>
+<pre class="idl">callback VoidFunction = undefined ();</pre>
```
styfle, line 14552: Should this be `UndefinedFunction?

tabatkins: I didn't think it was worth changing the name there, and requiring all users of VoidFunction to update. It's still a meaningful name post-update.

saschanaz: Note that 4 specs in reffy-reports that uses VoidFunction: CSS Layout, CSS Paint, HTML, and WebRTC.

index.bs
domenic: Let's fold it away; I think a big win of this PR is unifying the two value spaces.

saschanaz:
> * @saschanaz, can you get your bot ready to pull request all the specs?

Sure, but the code currently does not support HTML and WebGL, as those specs manually decorate their IDL while others have pure IDL. Patching those two should thus be manual.

tabatkins:
> I guess the actual change, e.g. allowing it in unions, falls out of changing the grammar?

Yup, just had to make it a PrimitiveType, rather than a special ReturnType only.

> Let's fold in Type into ReturnType.

Two votes for, so new commit coming shortly.

tabatkins:
> File issues on all the places located in [...]

Will get to that shortly, thanks for the pointer.

index.bs
```diff
@@ -1791,11 +1791,17 @@ The following extended attributes are applicable to constants:
 
 <pre class="grammar" id="prod-ConstValue">
     ConstValue :
+        UndefinedLiteral
```
saschanaz, line 1794: Is this for #905?

tabatkins: No, unrelated. It just looked correct, since I was adding a new literal, that it should be added to the Const productions as well.

saschanaz: Does this PR uses the new literal elsewhere? I think this is the only use.

saschanaz:
Does an argument typed `undefined` make sense? Should it be prevented?

index.bs
tabatkins: > Does this PR uses the new literal elsewhere? I think this is the only use.

No, this is it. But the other primitives all have a literal, except null (which is a little weird, but it's handled as a type annotation rather than a type on its own, so eh).

I could remove this if necessary, it just seemed nice.

tabatkins:
> Does an argument typed undefined make sense? Should it be prevented?

By itself, like `DOMString foo(undefined bar)`? Probably not. In a union, tho, definitely yes. I don't think it's worth trying to grammar-finesse it out of existence; doing so wouldn't be avoiding any footguns, just complexifying the grammar.

saschanaz:
> Probably not. In a union, tho, definitely yes.

I think `optional` or `= undefined` should be used instead of making a union in an argument, unless I missed something. Probably return type is the only place that a undefined union makes sense.

index.bs
domenic: It'd probably be better to avoid unnecessary indirections in the grammar.

domenic:
> I think `optional` or `= undefined` should be used instead of making a union in an argument, unless I missed something. Probably return type is the only place that a undefined union makes sense.

Good point... Hmm. Is that something that the grammar can prevent? Otherwise we should add a prose restriction.

tabatkins:
It is, but it would require adding `ReturnType` back in *and* making it more complex, duplicating the `Union` grammars into `Return*` variants that allow `undefined`.

Like I said, tho, I don't think this is a footgun that we actually need to protect anyone against. We can put in a prose restriction if y'all feel it's absolutely necessary, but I *really* don't want to enforce this in the grammar.

domenic:
I think this is pretty important to enforce in the grammar.

saschanaz:
Agreed, it would be very verbose to force it in the grammar. It should probably be okay to just write a prose, as WPT has a validator that should prevent authors to do that.

domenic:
I don't understand why it's important to avoid verbosity in the grammar here. What else is the grammar for, if not to enforce restrictions of this sort?

tabatkins:
Oh dang, upon review it would actually be even worse than I thought to handle in the grammar - I'd have to split the *typedef* production as well so that typedefs with an `undefined` in them are only usable in return types. :(
 Basically every single place where a type is mentioned needs to split into a Return* variant to enforce this.

> I don't understand why it's important to avoid verbosity in the grammar here.

Overall, simple grammars are better because they're easier to read and understand, and easier to maintain over time without mistakes or oversights. Grammar simplicity is mostly a "spec author" convenience in the priority of constituencies, slightly a "UA author" issue. 

> What else is the grammar for, if not to enforce restrictions of this sort?

"Of this sort", sure. My objection is that this isn't actually a restriction worth enforcing in the first place, so it's not worth adding a bunch of complexity for. We should add restrictions when we think spec authors will actually make a mistake that they'd appreciate being informed about.

If we're really concerned with people doing `= undefined` defaults before we make a final decision about `optional`, I'll just remove that part of the patch. But banning `undefined` from argument types seems like something we really don't need to care about, imo.

domenic:
I think this is a very important that people not use `undefined` in argument types, for the reasons listed above. If we can't enforce that via grammar, fine, but we need a prose restriction at the very least.

tabatkins:
Ah yeah, and on further consideration, even if I split the grammar it wouldn't help, because doing so would also strongly break the LL(1) nature of the grammar, instead now requiring infinite lookahead to parse (because an `undefined` in the very last token of a very large union would affect which grammar you're parsing with). I'm pretty sure that's a no-go right away, then.

I'll remove the const bit and add a prose restriction, one sec.

saschanaz:
I guess it might not make sense on dictionaries either. I wonder we should allow it only at return type position.

tabatkins:
Same deal there - we literally can't restrict `undefined` in those ways if we want to allow it in unions *at all*, because it'll suddenly make webidl require infinite lookahead, rather than the finite (single-token, in fact) lookahead it currently uses.

We'd need to introduce a whole new syntax construct that immediately distinguishes "unions that allow undefined" from "unions without undefined".

tabatkins:
(Sorry, deleted previous comment, I was misreading.)

> I guess it might not make sense on dictionaries either.

Hm, because dictionary entries can be marked as "optional" too?

saschanaz:
> Hm, because dictionary entries can be marked as "optional" too?

Yeah, the entries are optional by default so `undefined` means nothing. And attributes, we probably don't want to define attributes that return `undefined`. Could we narrow the scope here?

domenic:
Well, we definitely want attributes which return undefined OR something else. But I agree restricting dictionaries is a good idea.

saschanaz:
> Well, we definitely want attributes which return undefined OR something else.

Interesting, do we have an existing use case?

domenic:
I mean, the majority of nullable attributes would be reasonable if expressed that way.

saschanaz:
I don't exactly disagree but that sounds like a significant convention change.

tabatkins:
> I don't exactly disagree but that sounds like a significant convention change.

While a lot of DOM uses null as the missing value, things interacting more with JS would use undefined. It's no more of a convention change than methods returning undefined instead of null. (IOW, yeah, most methods/attributes won't, but it's important for some.)

Anyway, dictionaries disallowed now too.

annevk:
We could change dictionaries as part of #905 to use `= undefined` too and then we'd no longer need `required`. Probably better as a separate change though and maybe not worth it given how common optional dictionary members are.

foolip:
> /cc @foolip and @stephenmcgruer who has been doing some IDL stuff, I think related to reffy reports. (Maybe reffy should also go in that list of places to file issues?)

I've filed https://github.com/heycam/webidl/pull/907 for that. Although Reffy uses webidl2.js, it is still useful to have a heads up about changes that will be happening ecosystem-wide. That tends to uncover when some spec is no longer being scraped and updated correctly.

tabatkins:
> File issues on all the places

I'm filing issues today (sorry for the delay) - this is a "breaking change", and thus needs issues on the browsers as well as the parsers, right?

domenic:
Yep!

tabatkins:
* [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1659158)
* [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=215514)
* [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1116522)
* [WebIDL bug](https://github.com/web-platform-tests/wpt/issues/25014)
* [IdlHarness bug](https://github.com/web-platform-tests/wpt/issues/25015)
* [Reffy issue](https://github.com/tidoust/reffy/issues/375)
* [WidlParser issue](https://github.com/plinss/widlparser/issues/60)
* [WebIdl2.js issue](https://github.com/w3c/webidl2.js/issues/487)
* [TSJS-lib generator issue](https://github.com/microsoft/TSJS-lib-generator/issues/896)

emilio:
I get this, though as a C++ person this makes WebIDL harder to parse for me, fwiw :)

domenic:
LGTM. Will merge Monday to give any others time to comment. (E.g. @saschanaz has been particularly helpful in spotting things I've missed :).)

Then, the great spec-update-ificiation begins... I can take HTML and Web GL, and @saschanaz's bot should be able to take the rest.

domenic:
Unleash the bots!

saschanaz:
(Waiting for https://github.com/w3c/webidl2.js/pull/486 to merge)

marcoscaceres:
hehe, you guys just like to watch the world burn 🌎🔥

PR #914, opened 2020-09-03
Normative: Match ECMA‑262 function property enumeration order

ExE-Boss:
This makes the order of the `"length"`, `"name"` and `"prototype"` properties on **WebIDL** function objects match web reality (at least **Firefox**’s behaviour) and **ECMA‑262** as of <https://github.com/tc39/ecma262/pull/1490>.

I’ve also moved the computation of the function’s `name` and `length` to before the [CreateBuiltinFunction] call in preparation for <https://github.com/tc39/ecma262/pull/2116>.

[CreateBuiltinFunction]: https://tc39.es/ecma262/#sec-createbuiltinfunction


ExE-Boss:
<https://github.com/tc39/ecma262/pull/2116> has now been merged, so this PR has been updated to pass the <code><var>length</var></code> and <code><var>name</var></code> parameters to [CreateBuiltinFunction](https://tc39.es/ecma262/#sec-createbuiltinfunction) directly.

annevk:
Great work @ExE-Boss! I'll leave this to others to review, but this looks like a really nice cleanup. Do we have some tests for this as well?

index.bs
```diff
@@ -11794,10 +11788,8 @@ in which case they are exposed on every object that [=implements=] the interface
         1.  If |attribute|'s type is a [=promise type=], then return
             [=!=] <a abstract-op>Call</a>({{%Promise.reject%}}, {{%Promise%}}, «|E|»).
         1.  Otherwise, end these steps and allow the exception to propagate.
-    1.  Let |F| be [=!=] <a abstract-op>CreateBuiltinFunction</a>(|steps|, « », |realm|).
-    1.  Let |name| be the string "<code>get </code>" prepended to |attribute|'s [=identifier=].
-    1.  Perform [=!=] <a abstract-op>SetFunctionName</a>(|F|, |name|).
-    1.  Perform [=!=] <a abstract-op>SetFunctionLength</a>(|F|, 0).
+    1.  Let |name| be |attribute|'s [=identifier=].
+    1.  Let |F| be [=!=] [$CreateBuiltinFunction$](|steps|, 0, |name|, « », |realm|, {{%Function.prototype%}}, "<code>get</code>").
     1.  Return |F|.
```
Ms2ger, line 11793:
```suggestion
    1.  Return [=!=] [$CreateBuiltinFunction$](|steps|, 0, |name|, « », |realm|, {{%Function.prototype%}}, "<code>get</code>").
```

index.bs
```diff
@@ -11880,10 +11872,8 @@ in which case they are exposed on every object that [=implements=] the interface
         1.  Perform the [=setter steps=] of |attribute|,
             with |idlObject| as [=this=] and |idlValue| as [=the given value=].
         1.  Return <emu-val>undefined</emu-val>
-    1.  Let |F| be [=!=] <a abstract-op>CreateBuiltinFunction</a>(|steps|, « », |realm|).
-    1.  Let |name| be the string "<code>set </code>" prepended to |id|.
-    1.  Perform [=!=] <a abstract-op>SetFunctionName</a>(|F|, |name|).
-    1.  Perform [=!=] <a abstract-op>SetFunctionLength</a>(|F|, 1).
+    1.  Let |name| be |attribute|'s [=identifier=].
+    1.  Let |F| be [=!=] [$CreateBuiltinFunction$](|steps|, 1, |name|, « », |realm|, {{%Function.prototype%}}, "<code>set</code>").
     1. Return |F|.
```
Ms2ger, line 11877:
```suggestion
    1.  Return [=!=] [$CreateBuiltinFunction$](|steps|, 1, |name|, « », |realm|, {{%Function.prototype%}}, "<code>set</code>").
```

index.bs
```diff
@@ -11998,13 +11988,11 @@ in which case they are exposed on every object that [=implements=] the interface
...
         1.  Otherwise, end these steps and allow the exception to propagate.
-    1.  Let |F| be [=!=] <a abstract-op>CreateBuiltinFunction</a>(|steps|, « », |realm|).
-    1.  Perform [=!=] <a abstract-op>SetFunctionName</a>(|F|, |id|).
     1.  [=Compute the effective overload set=]  for [=regular operations=] (if |op| is a regular
         operation) or for [=static operations=] (if |op| is a static operation) with [=identifier=]
         |id| on |target| and with argument count 0, and let |S| be the result.
     1.  Let |length| be the length of the shortest argument list in the entries in |S|.
-    1.  Perform [=!=] <a abstract-op>SetFunctionLength</a>(|F|, |length|).
+    1.  Let |F| be [=!=] [$CreateBuiltinFunction$](|steps|, |length|, |id|, « », |realm|).
     1.  Return |F|.
```
Ms2ger, line 11996:
```suggestion
    1.  Return [=!=] [$CreateBuiltinFunction$](|steps|, |length|, |id|, « », |realm|).
```

index.bs
```diff
@@ -12748,7 +12722,7 @@ The \[[Prototype]] [=internal slot=] of an [=asynchronous iterator prototype obj
 
     1.  If |ongoingPromise| is not null, then:
         1.  Let |afterOngoingPromiseCapability| be [=!=] [$NewPromiseCapability$]({{%Promise%}}).
-        1.  Let |onSettled| be [=!=] [$CreateBuiltinFunction$](|nextSteps|, « »).
+        1.  Let |onSettled| be [=!=] [$CreateBuiltinFunction$](|nextSteps|, 0, "", « »).
```
Ms2ger, line 12725: Are these observable? Should the length be one like the others?

Ms2ger (approved):
Looks good as far as the spec goes, but each of these calls needs to be tested. The values are probably mostly covered by idlharness, but the order of iteration probably isn't.

index.bs
ExE-Boss: > Are these observable?

I don’t think so, since it’s passed directly to [PerformPromiseThen](https://tc39.es/ecma262/#sec-performpromisethen), rather than going through <code>[%Promise.prototype%.then](https://tc39.es/ecma262/#sec-promise.prototype.then)</code>.

----

> Should the length be one like the others?

Well, <var>nextSteps</var> doesn’t take any parameters, so it’s similar to:
```js
const nextSteps = () => { ... };
```
Which has a length of `0`.

index.bs
```diff
@@ -11792,11 +11786,8 @@ in which case they are exposed on every object that [=implements=] the interface
         1.  If |attribute|'s type is a [=promise type=], then return
             [=!=] <a abstract-op>Call</a>({{%Promise.reject%}}, {{%Promise%}}, «|E|»).
         1.  Otherwise, end these steps and allow the exception to propagate.
-    1.  Let |F| be [=!=] <a abstract-op>CreateBuiltinFunction</a>(|steps|, « », |realm|).
-    1.  Let |name| be the string "<code>get </code>" prepended to |attribute|'s [=identifier=].
-    1.  Perform [=!=] <a abstract-op>SetFunctionName</a>(|F|, |name|).
-    1.  Perform [=!=] <a abstract-op>SetFunctionLength</a>(|F|, 0).
-    1.  Return |F|.
+    1.  Let |name| be |attribute|'s [=identifier=].
+    1.  Return [=!=] [$CreateBuiltinFunction$](|steps|, 0, |name|, « », |realm|, {{%Function.prototype%}}, "<code>get</code>").
```
TimothyGu, line 11790:
```suggestion
    1.  Return [=!=] [$CreateBuiltinFunction$](|steps|, 0, |name|, « », |realm|,
        |realm|.\[[Intrinsics]].\[[{{%Function.prototype%}}]], "<code>get</code>").
```

We need to use the %Function.prototype% from _realm_, since when this operation is called there could well be no current realm.

index.bs
```diff
@@ -11878,11 +11869,8 @@ in which case they are exposed on every object that [=implements=] the interface
         1.  Perform the [=setter steps=] of |attribute|,
             with |idlObject| as [=this=] and |idlValue| as [=the given value=].
         1.  Return <emu-val>undefined</emu-val>
-    1.  Let |F| be [=!=] <a abstract-op>CreateBuiltinFunction</a>(|steps|, « », |realm|).
-    1.  Let |name| be the string "<code>set </code>" prepended to |id|.
-    1.  Perform [=!=] <a abstract-op>SetFunctionName</a>(|F|, |name|).
-    1.  Perform [=!=] <a abstract-op>SetFunctionLength</a>(|F|, 1).
-    1. Return |F|.
+    1.  Let |name| be |attribute|'s [=identifier=].
+    1.  Return be [=!=] [$CreateBuiltinFunction$](|steps|, 1, |name|, « », |realm|, {{%Function.prototype%}}, "<code>set</code>").
```
TimothyGu, line 11873: Ditto. Explicitly get %Function.prototype% from _realm_.

index.bs
```diff
@@ -13793,21 +13764,21 @@ those of normal <code>Array</code> instances:
...
     1.  Perform [=!=] [$CreateDataProperty$](|handler|, "<code>ownKeys</code>", |ownKeys|).
-    1.  Let |preventExtensions| be [=!=] [$CreateBuiltinFunction$](the steps from [[#es-observable-array-preventExtensions]], « », |realm|).
+    1.  Let |preventExtensions| be [=!=] [$CreateBuiltinFunction$](the steps from [[#es-observable-array-preventExtensions]], the number of non-optional parameters from [[#es-observable-array-preventExtensions]], "<code>preventExtensions</code>", « », |realm|).
     1.  Perform [=!=] [$CreateDataProperty$](|handler|, "<code>preventExtensions</code>", |preventExtensions|).
-    1.  Let |set| be [=!=] [$CreateBuiltinFunction$](the steps from [[#es-observable-array-set]], « », |realm|).
+    1.  Let |set| be [=!=] [$CreateBuiltinFunction$](the steps from [[#es-observable-array-set]], the number of non-optional parameters from [[#es-observable-array-set]], "<code>set</code>", « », |realm|).
     1.  Perform [=!=] [$CreateDataProperty$](|handler|, "<code>set</code>", |set|).
```
TimothyGu, line 13782: For readability, may I suggest a tabular format like this:

6. For each row in the following table:

   Name | Section | Length
   ---|---|---
   "`defineProperty`" | [§ 3.10.1 defineProperty](https://heycam.github.io/webidl/#es-observable-array-defineProperty) | 3
   "`deleteProperty`" | [§ 3.10.2 deleteProperty](https://heycam.github.io/webidl/#es-observable-array-deleteProperty) | 2
   … | … | …

   1. Let (_name_, _section_, _length_) be the entries of the current row.
   2. Let _F_ be ! CreateBuiltinFunction(the steps from _section_, _length_, _name_, « », _realm_).
   3. Perform ! CreateDataProperty(_handler_, _name_, _F_).

This helps make length changes more visible to implementors and folks writing tests, and also seems to be generally more readable.

TimothyGu (commented):
Just wondering @ExE-Boss, have you had any progress on writing tests for this PR?

index.bs
TimothyGu: @domenic could you take a look at this part of the PR if have some time?

domenic: Hmm. I kind of like the way the PR currently does it for lengths, in that it doesn't require syncing the length column with the steps like your proposal would.

Since the proxy handlers are an unobservable internal implementation detail anyway, I think we could just set the length to 0 and the name to "" for all of them, with a `<p class="note">` explaining?

The table/loop format would improve the section in general, but I'm not sure it needs to be done in this PR.

ExE-Boss:
Tests are now written in <https://github.com/web-platform-tests/wpt/pull/30333>.

domenic:
Based on https://github.com/web-platform-tests/wpt/pull/30333 it looks like this matches Firefox but mismatches Chrome and Safari (for constructors). So we should make sure at least one of those browsers is interested in changing. @yuki3 @shvaikalesh what do you think?

bakkot:
`Reflect.ownKeys(Blob)`:

- Chrome: ["length", "name", "arguments", "caller", "prototype"]
- Firefox: [ "length", "name", "prototype", Symbol("Symbol.hasInstance") ]
- Safari: ["prototype", "name", "length"]

So, Chrome has additional `arguments` and `caller` properties, which surprises me. (They appear to be non-functional, at least - `(function f(){ new Blob([], { get type(){ console.log(Blob.caller, Blob.arguments) } }) })()` prints `null` twice.)

But Chrome does match the enumeration order for the `length` and `name` properties, which I _think_ is the only change actually implied by this PR. In particular, ecma262 specifically refrains from specifying the full enumeration order, including of `prototype`; it only specifies that `length` appears before `name`. (The [test262 tests](https://github.com/EB-Forks/test262/blob/7fc0484aafa601cc411c2434cfac062559a136fe/test/built-ins/Function/property-order.js) for the corresponding [ecma262 PR](https://github.com/tc39/ecma262/pull/2116) are much more permissive than the WPT linked above.)

So, possibly the WPT is just over-zealous for the actual change implied by this PR, and could be changed to match the ecma262 PR. Then both Chrome and Firefox would pass.

domenic:
@bakkot to check my understanding, it seems like the correct test would be that name appears _right after_ length, right? So the test262 test is not as fully strict as it could be.

I think Web IDL, being more imperative than the ES spec, actually gives a full ordering between length/name/prototype that matches the tests written. But in terms of ensuring implementer interest for this particular change, I agree we're probably OK.

ExE-Boss:
> it seems like the correct test would be that name appears _right after_ length, right? So the test262 test is not as fully strict as it could be.

Actually, [the test262 test](https://github.com/tc39/test262/blob/7fc0484aafa601cc411c2434cfac062559a136fe/test/built-ins/Function/property-order.js#L14) checks exactly that using: `nameIndex === lengthIndex + 1`
```js
assert(lengthIndex >= 0 && nameIndex === lengthIndex + 1,
	"The `length` property comes before the `name` property on built-in functions");
```

domenic:
Sorry, I totally skimmed the test and misread it. My bad.

ExE-Boss:
@bakkot
> So, possibly the WPT is just over-zealous for the actual change implied by this PR, and could be changed to match the test262 tests. Then both Chrome and Firefox would pass.

With this PR, the `prototype` property is the first property that’s added to constructors after the `CreateBuiltinFunction` call in [create an interface object](https://pr-preview.s3.amazonaws.com/EB-Forks/webidl/pull/914.html#create-an-interface-object).

bakkot:
> it seems like the correct test would be that name appears _right after_ length, right?

Right.

> I think Web IDL, being more imperative than the ES spec, actually gives a full ordering between length/name/prototype that matches the tests written.

Ah, so it does.

shvaikalesh:
I like this change, both editorial-wise and the way it improves enumeration order consistency. Please use [this bug](https://bugs.webkit.org/show_bug.cgi?id=230584) to follow WebKit implementation.

As for `caller` and `arguments`, there is [a proposal](https://github.com/claudepache/es-legacy-function-reflection) to move them to `Function.prototype`, which was [recently implemented in WebKit](https://trac.webkit.org/changeset/280289) to match Firefox.

yuki3:
> Based on [web-platform-tests/wpt#30333](https://github.com/web-platform-tests/wpt/pull/30333) it looks like this matches Firefox but mismatches Chrome and Safari (for constructors). So we should make sure at least one of those browsers is interested in changing. @yuki3 @shvaikalesh what do you think?

I'm very sorry that I somehow overlooked the mention.  I'm fine with the change to make the order of property enumeration consistent.  However, I'm afraid that I cannot help much on this.  IIUC, the order of these standard(ish?) properties on function objects is controlled by V8.  We'd better ask V8 team to make the change.  @verwaest

index.bs
domenic: @TimothyGu should we merge as-is? Or go for the 0/"" solution maybe?

verwaest:
The proposed changes sound alright. Making this more uniform makes sense.

It seems like we're invalidly adding special accessors for caller/arguments but will always return null for native functions. We can change that too.

As for moving caller/arguments to the prototype chain: Currently v8 returns the caller of the "holder" on which the property is installed. With this I mean that if you put a function with a .caller on the prototype chain of e.g., a sloppy-mode arrow function, we won't return the caller of the arrow function, but the caller of the function on the prototype chain. What are the expected semantics there? (The change itself sounds like an improvement though!)

shvaikalesh:
@verwaest The proposal sets up `Function.prototype.{arguments,caller}` as getters ([spec](https://github.com/claudepache/es-legacy-function-reflection/blob/master/spec.md#get-functionprototypecaller)), so they can return `null` or real caller or even throw a `TypeError` depending on a holder (receiver).

One important bit of the proposal I kinda hope V8 would implement is returning `null` for (async) generators to match with JSC: we can't return real caller for them and it seems web-compatible.

ExE-Boss:
The “needs tests” label is wrong, as <https://github.com/web-platform-tests/wpt/pull/30333> has been merged.

Ms2ger:
Argh, apologies for merging the tests ahead of time. Would be good to get this landed anyway, though.

annevk:
That's blocked on one of two things:

1. @ExE-Boss updates https://github.com/whatwg/participant-data/blob/main/individuals.json with their real name.
2. Someone else makes this contribution in a new PR.

Issue #649, opened 2019-02-23
Interaction of extended attributes and typedefs is weird


bzbarsky:
Consider this IDL:
```
    typedef [Clamp] long ClampedLong;

    interface Bar {
      void func(ClampedLong arg);
    };
```
Is this valid IDL?  Per https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with step 6, the extended attributes "associated with" `ClampedLong` include `[Clamp]`.  But https://heycam.github.io/webidl/#Clamp says:

> A type that is not an integer type must not be associated with the [Clamp] extended attribute.

and `ClampedLong` is not an "integer type" in the sense linked here, as far as I can tell...

Presumably the intent is to allow this, right?  That should be made clearer.

bzbarsky:
@domenic

bzbarsky:
Also, is this allowed?

    typedef long Foo;

    interface Bar {
      void func([Clamp] Foo arg);
    };

https://heycam.github.io/webidl/#idl-typedefs says no extended attributes apply to typedefs, but presumably that means the actual typedef statement, right?  So in this case, does the arg get clamping behavior?

domenic:
Dang, I'd hoped given that one of the points in my OP of https://github.com/heycam/webidl/pull/286 was about interaction with typedefs, that we'd nailed this the first time around. OK, let me try to reload this into my brain...

> and `ClampedLong` is not an "integer type" in the sense linked here, as far as I can tell...

It seems like it should be an integer type. Simpler, it seems to me that given `typedef long Long`, then `Long` should be an integer type.

> https://heycam.github.io/webidl/#idl-typedefs says no extended attributes apply to typedefs, but presumably that means the actual typedef statement, right?

Right, that was definitely the intent there. I guess it should instead say something like

> Although it is allowed in the grammar, no extended attributes apply to typedef declarations themselves. (But, the type being given a new name might itself include extended attributes.)

> So in this case, does the arg get clamping behavior?

It seems like it should, right? The typedef should be transparent, ideally. https://heycam.github.io/webidl/#idl-type-extended-attribute-associated-with seems to do the right thing, so I guess this is just about whether `Foo` is an integer type again?

bzbarsky:
> It seems like it should be an integer type

That would be ideal; it just doesn't seem to be what the spec says right now.

> The typedef should be transparent, ideally

I agree.

> so I guess this is just about whether `Foo` is an integer type again

Hmm.  Yes, I think so...

PR #915, opened 2020-09-03
Allow to use LegacyNullToEmptyString on USVString.

emilio:
This is needed for the CSSOM spec, as that allows to use USVString as
the underlying string type.

Given the USVString conversion to JS values is already calling into the
DOMString conversion, there's no need to modify that section even, which
is great.

cc https://github.com/w3c/csswg-drafts/issues/5010


domenic (approved):
As sad as I am about the interop issue that is CSSOMString, this does seem like approximately the right fix. (E.g., the idea of only applying it to the DOMString typedef would be weird, as then the behavior would differ in browsers which use DOMString vs. USVString.)

There's a separate issue where it's not clear that extended attributes properly flow across typedefs; see #649. But the intent is clear enough, and that can be fixed orthogonally.

PR #882, opened 2020-05-09
Define definition conventions

annevk:
...

Helps with #730.

***
<a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/pull/882.html" title="Last updated on May 11, 2020, 4:57 PM UTC (e89870a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/882/9a739b5...e89870a.html" title="Last updated on May 11, 2020, 4:57 PM UTC (e89870a)">Diff</a>

PR #917, opened 2020-09-11
Specify value in setter steps

jakearchibald:
I think this was missing from #882. cc @annevk @domenic.


jakearchibald:
Oh wait, is this what "the given value" is for?

jakearchibald:
Yes, it looks like it is

PR #920, opened 2020-09-17
Require [Exposed] even when marked as [LegacyNoInterfaceObject]

### LINKED ISSUE #919: What does [Exposed] mean on [LegacyNoInterfaceObject] interfaces?
Currently the spec allows `[Exposed]` and `[LegacyNoInterfaceObject]` coexist [with Example 74](https://heycam.github.io/webidl/#LegacyNoInterfaceObject), while `[Exposed]` is optional in such case.

What does it mean to have both, though?

saschanaz:
Fixes #919

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1016,9 +1016,7 @@ The following extended attributes are applicable to [=partial interfaces=]:
 [{{LegacyOverrideBuiltIns}}], and
 [{{SecureContext}}].
 
-[=Interfaces=] which are not annotated
-with a [{{LegacyNoInterfaceObject}}] [=extended attribute=]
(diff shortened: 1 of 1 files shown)


annevk (approved):

annevk (changes requested):
Hmm, reading #423 again I might have missed something. I think we allowed this for mixin scenarios (which we used to use LegacyNoInterfaceObject for), but I'm not sure if we still have those type of mixins left or if all the LegacyNoInterfaceObject usage are interfaces that ought to be exposed but aren't.

annevk (approved):
I think I have convinced myself that changing this is good as if there are any interfaces like that left they should be converted to mixins.

domenic (approved):

PR #922, opened 2020-09-17
Fix "react to a promise"'s default handlers

### LINKED ISSUE #921: "React to a promise" is not equivalent to Promise.prototype.then with missing handlers
The "react to a promise" algorithm seems to be a spec-level version of `Promise.prototype.then`, except that when it's called with a missing handler, a default handler is substituted which simply resolves the promise with `undefined`.

In other words, you would expect "react to a promise" to be equivalent to:
```js
// promise: Promise<T>
// onFulfillment?: (T) => U
// onRejection?: (any) => U
promise.then(onFulfillment, onRejection) // Promise<U>
```
and it's instead:
```js
promise.then(onFulfillment ?? () => undefined, onRejection ?? () => undefined) // Promise<U | undefined>
```

I understand that, even when there's no fulfillment handler given, the conversion to IDL of the resolved value must happen, so the promise can reject if the conversion throws (see also #782). But even in that case, the return value shouldn't be `undefined`. And for the rejection handler even that is unnecessary, since converting an ECMAScript value to `any` cannot fail.

The fact that this is the intended behavior is borne out by the definitions of "upon fulfillment" and "upon rejection", as well as by the usage of these operations in the Fetch, Streams, Background Fetch and Service Worker specs.

andreubotella:
"React to a promise"'s default handlers used to return undefined rather than propagating the resolved value or the rejection reason to the returned promise. This change fixes that, and changes "upon fulfillment" and "upon rejection" to return the resulting promise.

Closes #921.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8427,19 +8427,17 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=] 
     1.  Let |onFulfilledSteps| be the following steps given argument |V|:
         1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
             value of type |T|.
-        1.  If there are no steps that are required to be run if the promise was fulfilled, then
-            return <emu-val>undefined</emu-val>.
-        1.  Let |result| be the result of performing any steps that were required to be run if
-            the promise was fulfilled, given |value| if |T| is not {{undefined}}.
+        1.  If there is a set of steps to be run if the promise was fulfilled, then let |result| be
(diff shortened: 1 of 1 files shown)


TimothyGu (approved):

domenic (approved):
LGTM too. As noted in IRC, I think originally the intention was that "Upon fulfillment" and "Upon rejection" not return anything, and that people use "reacting" if they want to return. But looking back on it, that was kind of dumb. It's just less powerful for no reason, and causes spec authors pain when they don't realize the intention. So the change here to make them return is good.

PR #929, opened 2020-10-07
Add [Historical] extended attribute

### LINKED ISSUE #910: Syntactic way to mark as historical/deprecated
There are several "historical" things that can't be removed for compatibility but not recommended for modern web platform, for example `KeyboardEvent#charCode`.

So far spec authors write prose or comments to describe those as historical, but it's hard for TSJS-lib-generator to track such things. A syntactic way would be helpful here, for the above example something like:

```webidl
partial dictionary KeyboardEventInit {
  [Historical] unsigned long charCode = 0;
  [Historical] unsigned long keyCode = 0;
};

// or 

[Historical]
partial dictionary KeyboardEventInit {
  unsigned long charCode = 0;
  unsigned long keyCode = 0;
};
```

It probably won't mean anything for browser implementations except a suggestion to add use counters.

saschanaz:
Closes #910 

This intentionally applies it to member constructs rather than interfaces to keep it simple.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/webidl/pull/929.html" title="Last updated on Oct 7, 2020, 1:22 PM UTC (9823b6d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/929/accdef1...saschanaz:9823b6d.html" title="Last updated on Oct 7, 2020, 1:22 PM UTC (9823b6d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9725,6 +9725,31 @@ corresponding to [=interface members=].
 </div>
 
 
+<h4 id="Historical" extended-attribute lt="Historical">[Historical]</h4>
+
+When the [{{Historical}}] [=extended attribute=] appears on an individual
+[=dictionary member=], [=interface member=], [=interface mixin member=], or
+[=namespace member=], it indicates that the construct exists for historical
+reason.
+
+The [{{Historical}}] extended attribute must
+[=takes no arguments|take no arguments=].
+
+<div class="example">
+    There are Web APIs that have been replaced by better new APIs and are
+    generally discouraged to use. Code editors can consider this extended
+    attribute as a hint to warn users where such APIs are used. Browser
+    implementations may also consider this as a suggestion to add use counters.
+
+    <pre highlight="webidl">
+        partial interface KeyboardEvent {
+            [Historical] attribute DOMString charCode;
+            [Historical] attribute DOMString keyCode;
+        }
+    </pre>
+</div>
+
+
 <h4 id="NewObject" extended-attribute lt="NewObject">[NewObject]</h4>
 
 If the [{{NewObject}}]
--- a/index.bs
+++ b/index.bs
@@ -9738,8 +9738,7 @@ The [{{Historical}}] extended attribute must
 <div class="example">
     There are web APIs that have been replaced by better new APIs and are
     generally discouraged to use. Code editors can consider this extended
-    attribute as a hint to warn users where such APIs are used. Browser
-    implementations may also consider this as a suggestion to add use counters.
+    attribute as a hint to warn users where such APIs are used.
 
     <pre highlight="webidl">
         partial interface KeyboardEvent {
--- a/index.bs
+++ b/index.bs
@@ -9745,7 +9745,7 @@ The [{{Historical}}] extended attribute must
         partial interface KeyboardEvent {
             [Historical] attribute DOMString charCode;
             [Historical] attribute DOMString keyCode;
-        }
+        };
     </pre>
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -9729,14 +9729,14 @@ corresponding to [=interface members=].
 
 When the [{{Historical}}] [=extended attribute=] appears on an individual
 [=dictionary member=], [=interface member=], [=interface mixin member=], or
-[=namespace member=], it indicates that the construct exists for historical
+[=namespace member=], it indicates that the construct exists for a historical
 reason.
 
 The [{{Historical}}] extended attribute must
 [=takes no arguments|take no arguments=].
 
 <div class="example">
-    There are Web APIs that have been replaced by better new APIs and are
+    There are web APIs that have been replaced by better new APIs and are
     generally discouraged to use. Code editors can consider this extended
     attribute as a hint to warn users where such APIs are used. Browser
     implementations may also consider this as a suggestion to add use counters.


index.bs
```diff
@@ -9725,6 +9725,31 @@ corresponding to [=interface members=].
 </div>
 
 
+<h4 id="Historical" extended-attribute lt="Historical">[Historical]</h4>
+
+When the [{{Historical}}] [=extended attribute=] appears on an individual
+[=dictionary member=], [=interface member=], [=interface mixin member=], or
+[=namespace member=], it indicates that the construct exists for historical
```
annevk, line 9732: a historical reason

index.bs
```diff
@@ -9725,6 +9725,31 @@ corresponding to [=interface members=].
...
+reason.
+
+The [{{Historical}}] extended attribute must
+[=takes no arguments|take no arguments=].
+
+<div class="example">
+    There are Web APIs that have been replaced by better new APIs and are
```
annevk, line 9739: Generally we keep web lowercase as per https://whatwg.org/style-guide (unless it's part of a title or some such) though it seems IDL is somewhat inconsistent on this.

index.bs
```diff
@@ -9725,6 +9725,31 @@ corresponding to [=interface members=].
...
+    generally discouraged to use. Code editors can consider this extended
+    attribute as a hint to warn users where such APIs are used. Browser
+    implementations may also consider this as a suggestion to add use counters.
```
annevk, line 9742: I actually don't think it's worth the effort of removing these APIs. They are not harmful and removing them might break the end user experience on sites, causing frustration for most parties involved.

annevk (commented):
Somewhat confusingly we also use the term historical in web-platform-tests to test for non-support of obsolete features. (And HTML uses obsolete to mean obsolete for web developers, but not fur user agents.)

Suffice to say it's all a bit of a mess, but I'd like us to agree on a way forward. Are we going to use historical for discouraged features that stick around? (And name new non-support tests differently?)

index.bs
saschanaz: Yes, quite a lot `Web` are in this spec, probably time to fix them.

domenic:
I remain -1 on this. Comments are the appropriate way to add information which does not impact bindings generation.

index.bs
saschanaz: I added this as I thought Gecko already had use counters for them but it seems not. I'm okay without this.

saschanaz:
>Comments are the appropriate way to add information

If only humans are reading them, yes. If any machines want to detect such things then I think a machine-readable way is more preferable. Call it MDN, TypeScript-DOM-lib-generator, or whatever.

domenic:
I think those machines are better off using a different source of data, so Web IDL can stay focused as a language designed for web browser and spec bindings generators.

saschanaz:
Maybe yes if people were not already mixing it into IDL as comments, unlike HTML events. My opinion is, why a different source when IDL already includes the info?

Looking at real world IDLs I can see some major use cases for comments inside IDL (locally searched `//` inside [here](https://github.com/microsoft/TypeScript-DOM-lib-generator/tree/master/inputfiles/idl)):

1. Sectioning (e.g. Logging/Counting/etc. in Console)
2. Marking as historical
3. Showing default value for attributes (e.g. in HTML Canvas) - Maybe also something to discuss, but not totally sure
4. Others (proses)

Which shows it's not a rare convention.

domenic:
> My opinion is, why a different source when IDL already includes the info?

I understand that's your opinion. Mine is the opposite. The ad-hoc, per-editor usage of commenting out things that that particular editor feels are historical, should not be enshrined into IDL. It's just a comment from the editor to the reader explaining something interesting, like "why are there three attributes that do the same thing". It's not a reliable source of data for historical-ness.

Instead, different sources should use their own judgment to collate their own lists of historical things.

saschanaz:
> It's just a comment from the editor to the reader explaining something interesting, like "why are there three attributes that do the same thing". It's not a reliable source of data for historical-ness.

Probably because there is no real web-platform-wide definition of being historical. Maybe a chance to define it here if that's the case, as it might remove some potentially confusing comments. (I don't have any specific examples, could you provide one?)

domenic:
I don't think settling on a platform-wide definition of historical is possible. Different people will consider different things historical. For example, I would consider `appendChild()` historical; `append()` is much better. Going and marking everything that I consider as historical as such, and then debating that with other stakeholders, is not a good use of anyone's time. It's not what specs are for.

Instead, if MDN wants to mark things as historical, it can do so independently, using the MDN decision makers' judgment.

saschanaz:
The definition can start with an intersection of what people thinks to prevent needless debates. I don't think it has to cover everything, if it can cover HTML/DOM cases that already marked as obsolete/historical, that should already be a good start.

If we can never get a consensus then I think we shouldn't have an "obsolete" section in HTML, because well, does everyone in every browser vendor really thinks they are obsolete? (If the answer is yes then we have a consensus.)

domenic:
At this point it feels like we are going in circles, so I'll stop repeating myself. I think I've made it clear though, that bringing this level of "social information" into a bindings-generator language is not appropriate, in my view.

saschanaz:
> Are we going to use historical for discouraged features that stick around? (And name new non-support tests differently?)

Or maybe just `[Discouraged]`.

saschanaz:
My use case is covered by BCD so I have no interest anymore.

PR #934, opened 2020-10-29
Normative: Match ECMA‑262 order of the `constructor` property

ExE-Boss:
In **ECMA‑262**, the `"constructor"` property is the first property that’s added to the prototype object after its creation.

---

This matches **V8**’s implementation of built‑in objects, including ones implemented using **WebIDL**.


TimothyGu:
Are there tests available for this?

PR #940, opened 2020-11-22
Specify that dictionaries can't be the type of attr/constant

saschanaz:
The same sentence is already in [Dictionary section](https://heycam.github.io/webidl/#idl-dictionaries), and this change copy-pastes it to the [Types - Dictionary types](https://heycam.github.io/webidl/#idl-dictionary) for consistency with other Types sections.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6321,8 +6321,9 @@ implicitly understood from context that the map is being treated as an instance 
 dictionary type. However, there is no way to represent a constant dictionary value inside IDL
 fragments.
 
-The [=type name=] of a dictionary type
-is the [=identifier=] of the dictionary.
+Dictionaries must not be used as the type of an [=attribute=] or [=constant=].
+
+The [=type name=] of a dictionary type is the [=identifier=] of the dictionary.
 
 
 <h4 id="idl-enumeration" dfn export>Enumeration types</h4>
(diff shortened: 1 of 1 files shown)


TimothyGu:
It looks like the same could be done for callback function types, which cannot be the type of any constant.

index.bs
```diff
@@ -5595,8 +5595,7 @@ and the return type and argument list (matching <emu-nt><a href="#prod-Type">Typ
 and <emu-nt><a href="#prod-ArgumentList">ArgumentList</a></emu-nt>) on the right side of the equals
 sign gives the signature of the callback function type.
 
-[=Callback functions=] must not
-be used as the type of a [=constant=].
+[=Callback functions=] must not be used as the type of a [=constant=].
```
TimothyGu, line 5598: This duplicates the statement on line 6367. I think this should be removed, as the sentence is better fit in the types section.

TimothyGu: Wait never mind, I misunderstood the PR.

TimothyGu (approved):

bathos:
Just a recap so it's more clear what seems odd to me - please correct me if this analysis is incorrect:

- There are twenty one sections for types which cannot be the type of a constant (as established by the section that defines constants).
- Currently only one of those sections (callback functions) explicitly states that it cannot be the type of a constant. (It's not clear to me why this one was previously singled out.)
- This PR would add the explicit "cannot be the type of a constant" to three more (dictionaries, sequences, records).
- That will leave seventeen other type sections that continue to not say "cannot be the type of a constant" even though that would also be true for them.

bathos:
Ah, got it. Sorry for the confusing comments. I was searching with an overspecific string. I see now that the 19 vs four (I think the fifth stems from appearing in relation callback function twice?) are a pre-existing quirk.

PR #941, opened 2020-11-22
Drop stringifier operation longhand

### LINKED ISSUE #816: Is the longer form of stringifier operations a thing?
Can't find any example of `stringifier DOMString customIdentifier();`, is there any known use?

saschanaz:
A search by `stringifier\s+DOMString` in [webref](https://github.com/w3c/webref) says [Selection API](https://w3c.github.io/selection-api/#selection-interface) is the only spec using it in the form of `stringifier DOMString ();`, which is just same as `stringifier;`.

Closes #816

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/webidl/pull/941.html" title="Last updated on Nov 22, 2020, 2:36 PM UTC (6d4427a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/941/ccef980...saschanaz:6d4427a.html" title="Last updated on Nov 22, 2020, 2:36 PM UTC (6d4427a)">Diff</a>


saschanaz:
https://github.com/w3c/selection-api/pull/131 dropped the only case.

domenic:
This is great! But I think it needs a bit more work, since it still talks about stringifiers being a type of operation. Thus there's a sentence like "Stringifier operations must take zero arguments and return a {{DOMString}}." which doesn't really make sense.

Instead I think stringifiers should be pulled out of the special operations section and given their own section, "Stringifiers". (I thought at first "Stringifier declarations", but I think that's not correct because of the attribute form; the section needs to cover both the attribute form and the declaration form.)

WDYT?

saschanaz:
Tried pulling the section out and splitting a few references.

TimothyGu:
I'm a little tempted to keep the spec as is. Though the longhand syntax is not used right now, I like the consistency we currently have between `stringifier` and `getter`/`setter`. Also, without the longhand notation we can't exactly express "this stringifier always returns a USVString", which would be desirable if we make `URL`'s [stringifier](https://url.spec.whatwg.org/#URL-stringification-behavior) an actual operation rather than attribute – though it does look like the spec only allows DOMString return type on long-form stringifiers. Hmm…

domenic:
My take is that this is a good incremental improvement that prevents us from using the longhand syntax on future APIs.

Longer-term, I think it'd be best to move stringifiers to just be operations the same we did for serializer -> toJSON(). That is, remove the stringifier concept, and just let people add their own `toString()` operations. We could add something like `[ForwardsTo=href]` if we want to allow omitting the method steps, but I'm not sure whether that's necessary.

That longer-term change requires a bit of spec ecosystem churn so I'm not sure how worthwhile it is.

index.bs
```diff
@@ -2708,6 +2715,121 @@ See [[#interface-object]] for details on how a [=constructor operation=] is to b
 <div data-fill-with="grammar-ArgumentNameKeyword"></div>
 
 
+<h4 id="idl-stringifiers">Stringifiers</h4>
+
+When an [=interface=] has a <dfn id="dfn-stringifier">stringifier</dfn>, it indicates that objects
+that implement the interface have a non-default conversion to a string.  As mentioned above,
+stringifiers can be specified using a <emu-t>stringifier</emu-t> keyword.
```
domenic, line 2722: Maybe add "This creates a stringifier operation"?

domenic (approved):
LGTM with minor suggestion

index.bs
saschanaz, line 2722: Removing "As mentioned above" since I think it refers to the table in special operations section.

```suggestion
that implement the interface have a non-default conversion to a string.  Stringifiers can be
specified using a <emu-t>stringifier</emu-t> keyword, which creates a stringifier operation when
used alone.
```

PR #946, opened 2021-01-15
Remove disused entries from error names table (2.8.1.)

kt3k:
No major browser implements these 3 error-name-to-error-code mappings in DOMException constructor. So this PR suggests removing these 3 entries from the table.

We found this while updating web platform test at https://github.com/web-platform-tests/wpt/pull/27181.

cc @domenic

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5291,11 +5291,6 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
             <td><strong>Deprecated.</strong> Use {{RangeError}} instead.</td></td>
             <td><dfn id="dom-domexception-index_size_err" for="DOMException" const export><code>INDEX_SIZE_ERR</code></dfn>&nbsp;(1)</td>
         </tr>
-        <tr class="deprecated">
-            <td>"<dfn id="domstringsizeerror" exception export><code>DOMStringSizeError</code></dfn>"</td>
-            <td><strong>Deprecated.</strong> Use {{RangeError}} instead.</td>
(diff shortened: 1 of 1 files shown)


annevk:
I'm not sure this is correct, e.g., `DOMException.DOMSTRING_SIZE_ERR` returns 2 for me in Firefox.

kt3k:
> DOMException.DOMSTRING_SIZE_ERR returns 2 for me in Firefox.

That's right, but this PR doesn't suggest removing that constant. What the table (2.8.1.) defines is the mapping from the error name to error code. e.g. `new DOMException('foo', 'DOMStringSizeError').code` should be 2 (=== DOMException.DOMSTRING_SIZE_ERR) according to the table, but `new DOMException('foo', 'DOMStringSizeError').code` is actually 0 in all 4 major browsers.

I checked these mappings (of deprecated ones) by the following expression:
```js
[
  new DOMException('foo', 'IndexSizeError').code,
  new DOMException('foo', 'DOMStringSizeError').code,
  new DOMException('foo', 'ValidationError').code,
  new DOMException('foo', 'InvalidAccessError').code,
  new DOMException('foo', 'NoDataAllowedError').code,
  new DOMException('foo', 'TypeMismatchError').code
]
```
The major 4 browsers all returns `[ 1, 0, 0, 15, 0, 17 ]`, which means they don't implement mapping of `DOMStringSizeError` => 2, `ValidationError` => 6, `NoDataAllowedError` => 16.

domenic (approved):

PR #955, opened 2021-02-06
fix: Remove raw percentage character from ID

nschonni:
Came across this trying to run some HTML validation upstream in the Bikeshed repo

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -89,7 +89,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %Array.prototype%; url: sec-properties-of-the-array-prototype-object
         text: %Error.prototype%; url: sec-properties-of-the-error-prototype-object
         text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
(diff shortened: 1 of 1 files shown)


nschonni:
Looks like this needs an upstream fix in tc39/ecma262

mathiasbynens:
Seems like the fix should be to URL-encode the `%` sign as `%25` instead of changing the upstream ECMA-262 document.

index.bs
```diff
@@ -89,7 +89,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: %Array.prototype%; url: sec-properties-of-the-array-prototype-object
         text: %Error.prototype%; url: sec-properties-of-the-error-prototype-object
         text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
-        text: %IteratorPrototype%; url: sec-%iteratorprototype%-object
+        text: %IteratorPrototype%; url: sec-iteratorprototype-object
```
ExE-Boss, line 92:
```suggestion
        text: %IteratorPrototype%; url: sec-%25iteratorprototype%25-object
```

PR #963, opened 2021-03-11
Named properties object: allow symbols in [[DefineOwnProperty]] and [[Delete]]

shvaikalesh:
Given that `X` is [named properties object](https://heycam.github.io/webidl/#named-properties-object) for `Window`, the following code:

```js
Object.defineProperty(X, Symbol.toStringTag,
  Object.getOwnPropertyDescriptor(X, Symbol.toStringTag));
```

should throw `TypeError` per current spec. While being compliant with [invariants of EIM](https://tc39.es/ecma262/#sec-invariants-of-the-essential-internal-methods), this behavior is very weird.

The same snippet for `X` that is [module namespace object](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects) doesn't throw, even though the namespace is non-extensible and its `Symbol.toStringTag` is non-configurable.

While we could just special-case `Symbol.toStringTag` or own properties that are symbols, I suggest we treat symbols in [`[[DefineOwnProperty]]`](https://heycam.github.io/webidl/#named-properties-object-defineownproperty) and [`[[Delete]]`](https://heycam.github.io/webidl/#named-properties-object-delete) the same way as ordinary methods do. To prevent clashes with [supported property names](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object), it's critical for `WindowProperties` to prohibit expando properties that are strings, but not symbols.

Proposed change feels like an expected behavior, reducing divergence with ordinary methods, and aligns [named properties object](https://heycam.github.io/webidl/#named-properties-object) with [legacy platforms objects](https://heycam.github.io/webidl/#legacy-platform-object-defineownproperty) & [module namespace object](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-defineownproperty-p-desc). It is already implemented in **Blink** and also, with [bug #222918](https://bugs.webkit.org/show_bug.cgi?id=222918), in **WebKit**.

Currently, no browser features spec-perfect `WindowProperties` object, so this change can be implemented along with [fixing current spec incompatibilities](https://github.com/web-platform-tests/wpt/pull/27970#issue-588774871).

Tests: https://github.com/web-platform-tests/wpt/pull/27970.

_cc_ @domenic @evilpie

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -157,6 +157,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: NormalCompletion; url: sec-normalcompletion
         text: OrdinaryObjectCreate; url: sec-ordinaryobjectcreate
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
+        text: OrdinaryDelete; url: sec-ordinarydelete
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
         text: OrdinarySetWithOwnDescriptor; url: sec-ordinarysetwithowndescriptor
@@ -11632,20 +11633,23 @@ is the concatenation of the [=interface=]’s
 
 <div algorithm="to invoke the [[DefineOwnProperty]] internal method of named properties object">
 
-    When the \[[DefineOwnProperty]] internal method of a [=named properties object=] is called,
+    When the \[[DefineOwnProperty]] internal method of a [=named properties object=] |O|
+    is called with property key |P| and [=Property Descriptor=] |Desc|,
     the following steps are taken:
 
-    1.  Return <emu-val>false</emu-val>.
+    1.  If [$Type$](|P|) is String, then return <emu-val>false</emu-val>.
+    1.  Otherwise, return [=?=] [$OrdinaryDefineOwnProperty$](|O|, |P|, |Desc|).
 </div>
 
 <h5 id="named-properties-object-delete">\[[Delete]]</h5>
 
 <div algorithm="to invoke the [[Delete]] internal method of named properties object">
 
-    When the \[[Delete]] internal method of a [=named properties object=] is called,
-    the following steps are taken:
+    When the \[[Delete]] internal method of a [=named properties object=] |O|
+    is called with property key |P|, the following steps are taken:
 
-    1.  Return <emu-val>false</emu-val>.
+    1.  If [$Type$](|P|) is String, then return <emu-val>false</emu-val>.
+    1.  Otherwise, return [=?=] [$OrdinaryDelete$](|O|, |P|).
 </div>
 
 


index.bs
```diff
@@ -11632,20 +11633,23 @@ is the concatenation of the [=interface=]’s
...
 <div algorithm="to invoke the [[Delete]] internal method of named properties object">
 
-    When the \[[Delete]] internal method of a [=named properties object=] is called,
-    the following steps are taken:
+    When the \[[Delete]] internal method of a [=named properties object=] |O|
+    is called with property key |P|, the following steps are taken:
 
-    1.  Return <emu-val>false</emu-val>.
+    1.  If [$Type$](|P|) is String, then return <emu-val>false</emu-val>.
+    1.  Otherwise, return [=?=] [$OrdinaryDelete$](|O|, |P|).
```
shvaikalesh, line 11652: Unlike in [legacy platform objects](https://heycam.github.io/webidl/#legacy-platform-object-delete), we can call [`OrdinaryDelete`](https://tc39.es/ecma262/#sec-ordinarydelete) instead of reimplementing it, as it's reachable only for own symbol properties. I mildly prefer calling ordinary methods so the implementations can match the spec more closely.

evilpie:
@annevk do we want this in Firefox? I think this would mean creating an expando object for WindowNamedProperties, which we currently don't have.

annevk:
I would be fine keeping this as-is personally. Roughly what I care about for these legacy objects is:

1. Not breaking compat or security.
2. Interop.
3. Agreement with ECMAScript on the object model.

And although this is a bit of 2, that could also be achieved by Chromium changing.

@yuki3 was this intentional on the side of Chromium?

shvaikalesh:
> I think this would mean creating an expando object for WindowNamedProperties, which we currently don't have.

Creating an expando object for `WindowNamedProperties` can be avoided if we leave `[[Delete]]` and `[[DefineOwnProperty]]` as is, but change `Symbol.toStringTag` to be non-configurable like it's done for [module namespace object](https://tc39.es/ecma262/#sec-@@tostringtag).

> 3. Agreement with ECMAScript on the object model.

It feels a bit of 3 to me.

yuki3:
> @yuki3 was this intentional on the side of Chromium?

About named properties and indexed properties, we're struggling to implement them correctly due to unique characteristics of V8 APIs (which sometimes do not have a clear mapping to ECMAScript spec).  Clearly not all behaviors are implemented on purpose.

I don't fully understand the issue and cannot say what's the best.  If necessary, I'm happy to change the Chromium's behaviors.

annevk:
@shvaikalesh could you elaborate on why the specified semantics are weird with respect to the invariants?

evilpie:
I like the idea of making `Symbol.toStringTag` non-configurable. Right now it's non-writable, but configurable in Firefox. Making it non-configurable would make it more consistent with the always throwing [[DefineOwnProperty]].

shvaikalesh:
Given the following example:

```html
<!DOCTYPE html>
<form name=foo></form>
<script>
const wp = Object.getPrototypeOf(Window.prototype);
wp.foo = 1;
</script>
```

@yuki3 Is it possible so the `[[Set]]` is ignored? Currently it is an interop issue: Firefox throws (it should't since no `"use strict"` there, but ok), while Chrome creates expando `foo` property that shadows `<form name=foo>`.

@annevk current spec does comply to invariants, but makes `Object.defineProperty` unnecessarily weird. Module namespace object's `[[DefineOwnProperty]]` used to [return `false` unconditionally in ES6](https://262.ecma-international.org/6.0/#sec-module-namespace-exotic-objects-defineownproperty-p-desc), but was [later relaxed](https://github.com/tc39/ecma262/pull/858) to be inline with ordinary counterpart.

In perfect world, it would be nice to do the same for `WindowProperties`. I don't really have hard opinion on whether symbols should be allowed or what descriptors should they have.

On implementation side, WebCore uses JavaScriptCore classes directly, without the API, so it's possible to implement anything that can be spec-ed.

yuki3:
> @yuki3 Is it possible to make the `[[Set]]` ignored? Currently it is an interop issue: Firefox throws (it should't since no `"use strict"` there, but ok), while Chrome creates expando `foo` property that shadows `<form name=foo>`.

To my best understanding, this is simply a bug.  We're forbidding `Object.defineProperty(wp, 'foo', {value: 1})` on purpose, but somehow assignments are mis-implemented.  I'm now fixing this issue.

IIUC, `[[Set]]` in this case will fallback to [`[[DefineOwnProperty]]`](https://heycam.github.io/webidl/#named-properties-object-defineownproperty) and the property won't be defined.  Correct me if I'm wrong.

petervanderbeken:
> Firefox throws (it should't since no `"use strict"` there, but ok)

The throwing without `use strict` is just a bug I think, we'll fix that.

> In perfect world, it would be nice to do the same for `WindowProperties`. I don't really have hard opinion on whether symbols should be allowed or what descriptors should they have.

I don't feel great about fiddling with some of these legacy objects, but it doesn't seem too complicated for us to change this in Gecko I think (limited to symbols). Making @@toStringTag non-configurable is fine by me too.

yuki3:
> > @yuki3 Is it possible to make the `[[Set]]` ignored? Currently it is an interop issue: Firefox throws (it should't since no `"use strict"` there, but ok), while Chrome creates expando `foo` property that shadows `<form name=foo>`.
> 
> To my best understanding, this is simply a bug. We're forbidding `Object.defineProperty(wp, 'foo', {value: 1})` on purpose, but somehow assignments are mis-implemented. I'm now fixing this issue.

I've landed a fix.  Chromium (tip of tree) is fixed.  Canary (nightly) version containing the fix will be shipped in a day or two.

shvaikalesh:
Thank you @yuki3, it's a great fix that covers all `[[Set]]` edge cases (like setters upper in prototype chain).

> IIUC, `[[Set]]` in this case will fallback to `[[DefineOwnProperty]]` and the property won't be defined.

That's right, per step 3 of [`OrdinarySet`](https://tc39.es/ecma262/#sec-ordinarysetwithowndescriptor).

---

I've decided not to pursue this change: if we make `@@toStringTag` non-configurable, named properties' `[[DefineOwnProperty]]` would still need to be updated to accept descriptors like:
* `{}`
* `{ value: "WindowProperties" }`
* `{ configurable: false }`

And even if we fix symbols, `[[DefineOwnProperty]]` would still remain weird (IMO) for existing string properties, while `[[Delete]]` would still fail for non-existing string properties (in perfect world, it shouldn't).

Instead, let's reach interop with the current spec. I've updated the tests and reported bugs on remaining minor issues:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1712180
* https://bugs.chromium.org/p/chromium/issues/detail?id=1211447

PR #964, opened 2021-03-11
Named properties object: enumerate supported properties in [[OwnPropertyKeys]]

shvaikalesh:
Currently, Gecko implements proposed `[[OwnPropertyKeys]]` semantics, but Blink and WebKit do not.

This PR feels like an expected behavior and aligns [named properties object](https://heycam.github.io/webidl/#named-properties-object) with [legacy platform objects](https://heycam.github.io/webidl/#legacy-platform-object-ownpropertykeys) on enumerating supported property names.

Since only `Window` has  [named properties object](https://heycam.github.io/webidl/#named-properties-object), and it also has a [`LegacyUnenumerableNamedProperties`](https://heycam.github.io/webidl/#LegacyUnenumerableNamedProperties) extended attribute, this change has no performance impact on `for (var k in window) {}`.

If this is feasible to implement in Blink, I will contribute extensive WPT coverage and author a patch for WebKit. Once #963 and `[[OwnPropertyKeys]]` discrepancies are resolved, `WindowProperties` will be 100% compatible across all implementations!

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -159,6 +159,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
+        text: OrdinaryOwnPropertyKeys; url: sec-ordinaryownpropertykeys
         text: OrdinarySetWithOwnDescriptor; url: sec-ordinarysetwithowndescriptor
         text: PerformPromiseThen; url: sec-performpromisethen
         text: ProxyCreate; url: sec-proxycreate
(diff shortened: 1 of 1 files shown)


annevk:
How would `for (var k in window) {}` end up invoking this, exactly? Wouldn't that end up invoking https://html.spec.whatwg.org/#windowproxy-ownpropertykeys?

shvaikalesh:
@annevk `for/in` indeed calls [`WindowProxy`'s `[[OwnPropertyKeys]]`](https://html.spec.whatwg.org/#windowproxy-ownpropertykeys) first, but then [goes up the prototype chain](https://tc39.es/ecma262/#sec-for-in-iterator-objects) for same-origin windows.

WebKit, and I believe other runtimes as well, accept additional parameter when implementing `[[OwnPropertyKeys]]` so non-enumerable property names (for `Object.keys` and `for/in`) can be rejected without calling `[[GetOwnProperty]]`.

annevk:
Thank you! (I sometimes wonder if we should import more of this `Window`-specific logic into HTML, but that's also quite an ordeal.)

shvaikalesh:
I've decided not to pursue this change: there is no clear benefit / use case to justify adding the method into 2 implementations, and we aren't changing [other methods of named properties](https://github.com/heycam/webidl/pull/963#issuecomment-845368437) either way.

Instead, to ensure interop, https://github.com/web-platform-tests/wpt/pull/27970 adds a test that named properties objects has [ordinary `[[OwnPropertyKeys]]` method](https://tc39.es/ecma262/#sec-ordinaryownpropertykeys).

Issue #962, opened 2021-03-09
undefined can't be a member of a union type


petervanderbeken:
I think one of the goals of #60 was to allow union types with undefined as a member type for return values. It did so by just adding undefined to the grammar for Type. Union types currently require member types to be distinguishable (_Each pair of flattened member types in a union type, T and U, must be distinguishable_). I don't see any changes to that in #906, or any changes to add rules for undefined in the distinguishability algorithm, so right now undefined cannot actually be a member type of a union type.

The prose for union type conversion to an ECMAScript value also still contains the following sentence: _If the value is an object reference to a special object that represents an ECMAScript undefined value, then it is converted to the ECMAScript undefined value_. I'm not sure we should keep that, there is a [rule](https://heycam.github.io/webidl/#es-undefined) for conversion of the undefined type, so it should be possible to rely on the generic algorithm there.

petervanderbeken:
It might work to make undefined distinguishable from all other types.

annevk:
cc @tabatkins

tabatkins:
Oooh, I didn't realize additional spec work had to be done there.

Yes, both of these changes are reasonable, I'll write a PR.

tabatkins:
Looking over the remaining spots where `undefined` is mentioned, I'm not entirely sure what to do about the "convert an ES value to an IDL union type" part, and I think it raises questions about distinguishability too:

> If V is null or undefined, then:
> 
> * If types includes a dictionary type, then return the result of converting V to that dictionary type.

So this implies that the current behavior is that a JS `undefined` can always be converted into a dictionary type. I presume this is fallout from the fact that dictionary *arguments* are required to be optional. Does this imply that `undefined` and dictionary types should not be distinguishable, tho?  Or is it okay for them to remain distinguishable, so I'll add a new first step that checks if the union type contains `undefined`, and then the dictionary conversion is a fallback?

(Also, hm, all of the 3.2.24 algo confuses me, since it's composed of nested conditionals where the inner conditionals don't have "else" clauses. Is the implication that you continue to next top-level step if you fail all the conditions?)

annevk:
It might be okay for them to be distinguishable in theory, but I don't think we'd ever want to have APIs use that.

And yeah, if the algorithm doesn't return, it continues, ultimately throwing.

petervanderbeken:
[...]

PR #965, opened 2021-03-11
More fixes for `undefined`

tabatkins:
* Added `undefined` to the distinguishability table
* Removed text from any/union conversion that handled `undefined` specially, now that it's just an ordinary IDL value.
* Editorially, cleaned up the markup for the distinguishability table, and changed one unrelated code example to use double-quotes on a few strings, as it was inconsistent with quote usage elsewhere in the same example, and the empty string messed up Bikeshed's syntax-highlighting for the rest of the spec from that point on.

Currently a draft because there's an open question in <https://github.com/heycam/webidl/issues/962#issuecomment-796938662>, don't merge until that's answered and likely has another commit addressing it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3608,6 +3608,7 @@ the following algorithm returns <i>true</i>.
         <thead>
             <tr>
                 <th class="corner"></th>
+                <th><div><span>undefined</span></div>
                 <th><div><span>boolean</span></div>
                 <th><div><span>numeric types</span></div>
                 <th><div><span>bigint</span></div>
@@ -3620,7 +3621,21 @@ the following algorithm returns <i>true</i>.
                 <th><div><span>sequence-like</span></div>
         </thead>
         <tr>
+            <th>undefined
+            <td>
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+            <td>●
+        <tr>
             <th>boolean
+            <td class="belowdiagonal">
             <td>
             <td>●
             <td>●
@@ -3633,6 +3648,7 @@ the following algorithm returns <i>true</i>.
             <td>●
         <tr>
             <th>numeric types
+            <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td>
             <td>
@@ -3647,6 +3663,7 @@ the following algorithm returns <i>true</i>.
             <th>bigint
             <td class="belowdiagonal">
             <td class="belowdiagonal">
+            <td class="belowdiagonal">
             <td>(b)
             <td>●
             <td>●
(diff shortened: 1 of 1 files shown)


tabatkins:
Ready for review now!

Based on discussion in #962:

* `undefined` is *not* distinguishable from dictionary types; `(undefined or MyDict)` is invalid
* `undefined` *is* nullable; `undefined?` is valid (will return an IDL null for a JS null, or IDL undefined for anything else)
* `undefined` is *technically* implicitly nullable, since it converts *anything* into itself, per previous spec text and `void` behavior

index.bs
```diff
@@ -8304,9 +8303,9 @@ ECMAScript Array values.
         // Each element will be converted to a double by first calling ToNumber().
         // So the following call is equivalent to the previous one, except that
         // "hi" will be alerted before drawPolygon() returns.
-        a = [false, '',
-             { valueOf: function() { alert('hi'); return 100; } }, 0,
-             '50', new Number(62.5)];
+        a = [false, "",
+             { valueOf: function() { alert("hi); return 100; } }, 0,
```
ExE-Boss, line 8307: This is missing a closing quote `"`:
```suggestion
             { valueOf: function() { alert("hi"); return 100; } }, 0,
```

index.bs
```diff
@@ -3605,169 +3605,164 @@ the following algorithm returns <i>true</i>.
...
+                <th><div><span>numeric types</span></div>
+                <th><div><span>bigint</span></div>
+                <th><div><span>string types</span></div>
+                <th><div><span>object</span></div>
+                <th><div><span>symbol</span></div>
+                <th><div><span>interface-like</span></div>
+                <th><div><span>callback function</span></div>
+                <th><div><span>dictionary-like</span></div>
+                <th><div><span>sequence-like</span></div>
+        </thead>
```
ExE-Boss, line 3622: Note that every other table in this specification includes the closing `</t[d|h|r]>` tags.

tabatkins: I mean, I'm happy to remove those from all the other tables too, it just wasn't relevant for this PR. I can do it in a followup.

ExE-Boss: Well, doing it in a separate PR would make this diff cleaner.

tabatkins: Just to close the loop: are you saying that removing the reformatting of this table is a requested change for an R+ here? Or is it just a drive-by preference, and a commitment to filing a clean-up PR afterwards reformatting the rest of the tables accordingly would be sufficient?

ExE-Boss: It’s mainly a drive‑by preference.

tabatkins: In that case, can you r+? Or else tell me that you won't r+ until I make the change?

domenic: There might be some confusion here; @ExE-Boss is not an editor for this spec and is just doing a drive-by review.

domenic (approved):
Sorry this has been lagging so long. This is really great work in all respects. Undefined being nullable is a little wonky but seems harmless.

PR #966, opened 2021-03-11
Adjust sample code so it makes more sense

travisleithead:
In order for the following line, pop(), to return "employeeB", employeeB needs to be on the "top" of the stack, meaning the splice should remove employeeC at index 2. :)


TimothyGu:
@domenic could you take a look?

domenic (approved):
Yep, just a classic off-by-one error I believe...

PR #977, opened 2021-04-19
Allow constants on namespaces

### LINKED ISSUE #440: We may want to suport constants on namespaces
I don't see an obvious reason to not support them, and they can be useful in some cases, I'd think.

domenic:
Closes #440.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4722,9 +4722,9 @@ associated behaviors.
 A namespace is a specification of a set of
 <dfn id="dfn-namespace-member" export lt="namespace member">namespace members</dfn>
 (matching <emu-nt><a href="#prod-NamespaceMembers">NamespaceMembers</a></emu-nt>), which are the
-[=regular operations=] and [=read only=] [=regular attributes=] that appear between the braces in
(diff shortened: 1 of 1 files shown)


annevk (approved):
The change looks good, but is there implementer support and some API that needs this? (I guess we've given up on steering APIs away from constants?)

domenic:
I did this because of the WebGPU case mentioned in https://github.com/heycam/webidl/issues/440, which is currently using interfaces: https://github.com/gpuweb/gpuweb/issues/304, e.g. https://gpuweb.github.io/gpuweb/#gpubufferusage

They're using enums in most places. However they use constants for bitmasks, which seems reasonable: it's relatively tricky to express the equivalent of `GPUBufferUsage.MAP_READ | GPUBufferUsage.MAP_WRITE | GPUBufferUsage.VERTEX` otherwise.

They also discuss this in https://gpuweb.github.io/gpuweb/explainer/#bitflags

kainino0x:
This will be really nice for WebGPU. Thank you Domenic!

Issue #865, opened 2020-03-27
How to return a SharedArrayBuffer object


annevk:
As far as I can tell a JS `SharedArrayBuffer` can enter IDL through an IDL `ArrayBuffer` that's effectively a pointer to the JS object.

Now assuming #342 is fixed and you can use `[AllowShared] ArrayBuffer` as return value, what language do you use to return a `SharedArrayBuffer`? Do you need to construct the JS object, then make an IDL `ArrayBuffer` object that points to it, and then return the IDL object?

TimothyGu:
Since Web IDL never seems to say an IDL ArrayBuffer has to be an actual JavaScript ArrayBuffer, couldn't you just use `ArrayBuffer` as the return type in IDL? I.e., the [AllowShared] part in the return type isn't actually strictly necessary.

domenic:
Sure, but how do you write spec prose saying that the IDL `ArrayBuffer` is actually a JavaScript SharedArrayBuffer? @annevk's last sentence points to one route, but it's kind of icky, as we'd like IDL to be an intermediary layer so that specs don't have to touch JS objects so directly.

PR #987, opened 2021-06-02
Add and improve operations on BufferSources

domenic:
* Introduces algorithms for creating ArrayBuffers and ArrayBufferViews from byte sequences
* Introduces algorithms for writing into ArrayBuffers and ArrayBufferViews, replacing "get a reference to the bytes held by the buffer source".
* Introduces "transfer" for an ArrayBuffer.
* Introduces "byte length", "underlying buffer", and "detached" for introspecting a BufferSource.
* Improves the specification for "get a copy of the bytes held by a buffer source" and "detach".

Updates to dependent specs:

* https://github.com/WebAudio/web-audio-api/pull/2358
* https://github.com/whatwg/streams/pull/1130
* https://github.com/whatwg/encoding/pull/262

I'd like review here from @syg and/or other ES editors, especially for the "write" operation.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9010,6 +9010,14 @@ a reference to the same object that the IDL value represents.
         false, Unordered).
 </div>
 
+<div algorithm>
+    The <dfn export for="BufferSource">byte length</dfn> of a {{BufferSource}} |bufferSource| is
+    given by the following algorithm:
+
+    1.  If |bufferSource| is an {{ArrayBufferView}}, then return |bufferSource|.\[[ByteLength]].
+    1.  Return |bufferSource|.\[[ArrayBufferByteLength]].
+</div>
+
 <div algorithm="ArrayBuffer/write">
     To <dfn export for="ArrayBuffer">write</dfn> a [=byte sequence=] |bytes| into an {{ArrayBuffer}}
     |arrayBuffer|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn>
--- a/index.bs
+++ b/index.bs
@@ -76,6 +76,11 @@ urlPrefix: https://w3c.github.io/mediacapture-main/; spec: MEDIACAPTURE-STREAMS
 urlPrefix: http://www.unicode.org/glossary/; spec: UNICODE
     type: dfn
         text: Unicode scalar value; url: unicode_scalar_value
+urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASM-JS-API-1
+    type: interface
+        text: Memory; url: #memory
+    type: attribute
+        text: buffer; for: Memory; url: #dom-memory-buffer
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
     type: interface; for: ECMAScript
         text: Array; url: sec-array-objects
@@ -142,6 +147,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: GetFunctionRealm; url: sec-getfunctionrealm
         text: GetIterator; url: sec-getiterator
         text: GetMethod; url: sec-getmethod
+        text: GetValueFromBuffer; url: sec-getvaluefrombuffer
         text: IfAbruptRejectPromise; url: sec-ifabruptrejectpromise
         text: IsArray; url: sec-isarray
         text: IsAccessorDescriptor; url: sec-isaccessordescriptor
@@ -157,9 +163,10 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: NewModuleEnvironment; url: sec-newmoduleenvironment
         text: NewPromiseCapability; url: sec-newpromisecapability
         text: NormalCompletion; url: sec-normalcompletion
-        text: OrdinaryObjectCreate; url: sec-ordinaryobjectcreate
+        text: OrdinaryCreateFromConstructor; sec: sec-ordinarycreatefromconstructor
         text: OrdinaryDefineOwnProperty; url: sec-ordinarydefineownproperty
         text: OrdinaryGetOwnProperty; url: sec-ordinarygetownproperty
+        text: OrdinaryObjectCreate; url: sec-ordinaryobjectcreate
         text: OrdinaryPreventExtensions; url: sec-ordinarypreventextensions
         text: OrdinarySetWithOwnDescriptor; url: sec-ordinarysetwithowndescriptor
         text: PerformPromiseThen; url: sec-performpromisethen
@@ -6779,51 +6786,9 @@ There is no way to represent a constant value of any of these types in IDL.
 The [=type name=] of all
 of these types is the name of the type itself.
 
-At the specification prose level, IDL [=buffer source types=]
-are simply references to objects.  To inspect or manipulate the bytes inside the buffer,
-specification prose must first either
-<dfn id="dfn-get-buffer-source-reference" export lt="get a reference to the buffer source">get a reference to the bytes held by the buffer source</dfn>
-or <dfn id="dfn-get-buffer-source-copy" export lt="get a copy of the buffer source">get a copy of the bytes held by the buffer source</dfn>.
-With a reference to the buffer source’s bytes, specification prose can get or set individual
-byte values using that reference.
-
-<div class="advisement">
-
-    Extreme care must be taken when writing specification text that gets a reference
-    to the bytes held by a buffer source, as the underlying data can easily be changed
-    by the script author or other APIs at unpredictable times.  If you are using a buffer source type
-    as an operation argument to obtain a chunk of binary data that will not be modified,
-    it is strongly recommended to get a copy of the buffer source’s bytes at the beginning
-    of the prose defining the operation.
-
-    Requiring prose to explicitly get a reference to or copy of the bytes is intended to
-    help specification reviewers look for problematic uses of these buffer source types.
-
-</div>
-
-<div class="note">
-
-    When designing APIs that take a buffer, it is recommended to use the
-    {{BufferSource}} typedef rather than {{ArrayBuffer}}
-    or any of the view types.
-
-    When designing APIs that create and return a buffer, it is recommended
-    to use the {{ArrayBuffer}} type rather than
-    {{Uint8Array}}.
-
-</div>
-
-Attempting to [=get a reference to the buffer source|get a reference to=] or
-[=get a copy of the buffer source|get a copy of the bytes held by a buffer source=]
-when the {{ArrayBuffer}} has been [=ArrayBuffer/detached=]
-will fail in a language binding-specific manner.
-
-Note: See [[#es-buffer-source-types]] below for
-how interacting with buffer source types works in the ECMAScript language binding.
-
-<p class="issue">
-    We should include an example of specification text that uses these types and terms.
-</p>
+At the specification prose level, IDL [=buffer source types=] are simply references to objects. To
+inspect or manipulate the bytes inside the buffer, specification prose needs to use the algorithms
+in [[#es-buffer-source-types]].
 
 <pre class="grammar" id="prod-BufferRelatedType">
     BufferRelatedType :
@@ -8999,55 +8964,17 @@ an IDL value of any [=buffer source type=]
 to an ECMAScript value is the Object value that represents
 a reference to the same object that the IDL value represents.
 
-<div algorithm="get a reference to a buffer source">
-
-    When [=get a reference to the buffer source|getting a reference to=]
-    or [=get a copy of the buffer source|getting a copy of the bytes held by a buffer source=]
-    that is an ECMAScript {{ECMAScript/ArrayBuffer}}, {{ECMAScript/DataView}}
-    or typed array object, these steps must be followed:
-
-    1.  Let |O| be the ECMAScript object that is the buffer source.
-    1.  Initialize |arrayBuffer| to |O|.
-    1.  Initialize |offset| to 0.
-    1.  Initialize |length| to 0.
-    1.  If |O| has a \[[ViewedArrayBuffer]] [=internal slot=], then:
-        1.  Set |arrayBuffer| to the value of |O|’s \[[ViewedArrayBuffer]] [=internal slot=].
-        1.  Set |offset| to the value of |O|’s \[[ByteOffset]] [=internal slot=].
-        1.  Set |length| to the value of |O|’s \[[ByteLength]] [=internal slot=].
-    1.  Otherwise, set |length| to the value of |O|’s \[[ArrayBufferByteLength]] [=internal slot=].
-    1.  If <a abstract-op>IsDetachedBuffer</a>(|arrayBuffer|) is <emu-val>true</emu-val>, then
-        return the empty byte sequence.
-    1.  Let |data| be the value of |O|’s \[[ArrayBufferData]] [=internal slot=].
-    1.  Return a reference to or copy of (as required) the |length| bytes in |data|
-        starting at byte offset |offset|.
-</div>
-
-<div algorithm>
-
-    To <dfn id="dfn-detach" for="ArrayBuffer" export>detach</dfn> an {{ArrayBuffer}} |ab|:
-
-    1.  Perform [=!=] <a abstract-op>DetachArrayBuffer</a>(|ab|).
-</div>
-
-<div algorithm="ArrayBuffer/write">
-    To <dfn export for="ArrayBuffer">write</dfn> a [=byte sequence=] |bytes| into an {{ArrayBuffer}}
-    |ab|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn> (default 0):
-
-    1.  Assert: |ab|.\[[ArrayBufferByteLength]] ≥ |startingOffset| + |bytes|'s
-        [=byte sequence/length=].
-    1.  For |i| in [=the range=] |startingOffset| to |startingOffset| + |bytes|'s [=byte
-        sequence/length=] &minus; 1, inclusive, perform [=!=] [$SetValueInBuffer$](|ab|, |i|, Uint8,
-        |bytes|[|i| - |startingOffset|], false, Unordered).
-</div>
+<hr>
 
 <div algorithm>
     To <dfn export for="ArrayBuffer">create</dfn> an {{ArrayBuffer}} from a [=byte sequence=]
     |bytes| in a [=Realm=] |realm|:
 
-    1.  Let |ab| be [=?=] [$AllocateArrayBuffer$](|realm|.\[[Intrinsics]].[[{{%ArrayBuffer%}}]],
-        |bytes|'s [=byte sequence/length=]).
-    1. [=ArrayBuffer/Write=] |bytes| into |ab|.
-    1. Return |ab|.
+    1.  Let |arrayBuffer| be [=?=]
+        [$AllocateArrayBuffer$](|realm|.\[[Intrinsics]].[[{{%ArrayBuffer%}}]], |bytes|'s
+        [=byte sequence/length=]).
+    1. [=ArrayBuffer/Write=] |bytes| into |arrayBuffer|.
+    1. Return |arrayBuffer|.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9019,10 +9023,40 @@ a reference to the same object that the IDL value represents.
...
+</div>
+
+<div algorithm="ArrayBuffer/write">
+    To <dfn export for="ArrayBuffer">write</dfn> a [=byte sequence=] |bytes| into an {{ArrayBuffer}}
+    |ab|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn> (default 0):
+
+    1.  Assert: |ab|.\[[ByteLength]] ≥ |startingOffset| + |bytes|'s [=byte sequence/length=].
+    1.  For |i| in [=the range=] |startingOffset| to |startingOffset| + |bytes|'s [=byte
+        sequence/length=] &minus; 1, inclusive, perform [=!=] [$SetValueInBuffer$](|ab|, |i|, Uint8,
+        |bytes|[|i|], false, Unordered).
```
syg, line 9038: This should be `bytes[i-startingOffset]`, I think. Judging by the length check above, `startingOffset` is only meant to apply to the destination `ab`.

index.bs
```diff
@@ -9019,10 +9023,40 @@ a reference to the same object that the IDL value represents.
...
-    1.  Let |O| be the ECMAScript object that is the {{ArrayBuffer}}.
-    1.  Perform [=!=] <a abstract-op>DetachArrayBuffer</a>(|O|).
+    1.  Perform [=!=] <a abstract-op>DetachArrayBuffer</a>(|ab|).
+</div>
+
+<div algorithm="ArrayBuffer/write">
+    To <dfn export for="ArrayBuffer">write</dfn> a [=byte sequence=] |bytes| into an {{ArrayBuffer}}
+    |ab|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn> (default 0):
+
+    1.  Assert: |ab|.\[[ByteLength]] ≥ |startingOffset| + |bytes|'s [=byte sequence/length=].
```
syg, line 9035: This should be [[ArrayBufferByteLength]]. [[ByteLength]] is for integer-indexed exotic objects (TypedArrays).

index.bs
```diff
@@ -9019,10 +9023,40 @@ a reference to the same object that the IDL value represents.
...
+
+<div algorithm>
+    To <dfn export for="ArrayBufferView">create</dfn> one of the {{ArrayBufferView}} types from a
+    [=byte sequence=] |bytes| in a [=Realm=] |realm|:
+
+    1.  Let |ab| be the result of [=ArrayBuffer/creating=] an {{ArrayBuffer}} from |bytes| in
+        |realm|.
+    1.  Let |constructor| be the appropriate constructor from |realm|.\[[Intrinsics]] for the type
+        of {{ArrayBufferView}} being created.
+    1.  Return [=!=] [$TypedArrayCreate$](|constructor|, « |ab| »).
```
syg, line 9059: This should be a ? TypedArrayCreate, unless the precondition of this algorithm is that `byte` is of a length that's a multiple of the appropriate element size.

syg (commented):
Glad to see ABs and TAs having more solid foundations here!

index.bs
annevk: I think we should require that (and assert it).

annevk (commented):
Nice!

I wonder how I would use this for https://encoding.spec.whatwg.org/#dom-textencoder-encodeinto.

Part of the problem I guess is https://heycam.github.io/webidl/#dfn-get-buffer-source-reference not being very clear. Perhaps we should aim to replace that completely by having a copy operation that returns a byte sequence? And instead of obtaining a reference you'd use the algorithms you have here (though we'd need to add write for views).

domenic:
Oh yeah, I like the idea of getting rid of "get a reference" in favor of having people write directly into it. And then making the "get a copy" operation more detailed using this same sort of primitive. I'll investigate.

domenic:
OK, I expanded and updated this PR's scope a bit. The OP has been edited with more details. A re-review would be lovely!

I will work on an Encoding spec PR to show how the "write" concepts would be used.

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+    [=byte sequence=] |bytes| in a [=Realm=] |realm|:
 
-    To <dfn id="dfn-detach" for="ArrayBuffer" export>detach</dfn> an {{ArrayBuffer}}, these steps must be followed:
+    1.  Assert: if the type is not {{DataView}}, then |bytes|'s [=byte sequence/length=] [=modulo=]
+        the [=element size=] of that type is 0.
+    1.  Let |arrayBuffer| be the result of [=ArrayBuffer/creating=] an {{ArrayBuffer}} from |bytes|
+        in |realm|.
+    1.  Let |constructor| be the appropriate constructor from |realm|.\[[Intrinsics]] for the type
+        of {{ArrayBufferView}} being created.
+    1.  Return [=!=] [$TypedArrayCreate$](|constructor|, « |arrayBuffer| »).
```
syg, line 8990: I had not realized that ArrayBufferView also parameterizes over DataView. TypedArrayCreate doesn't work for DataView. I guess you need a branch here that calls the DataView constructor.

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+    |arrayBuffer|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn>
+    (default 0):
 
-    1.  Let |O| be the ECMAScript object that is the {{ArrayBuffer}}.
-    1.  Perform [=!=] <a abstract-op>DetachArrayBuffer</a>(|O|).
+    1.  Assert: |bytes|'s [=byte sequence/length=] ≤ |arrayBuffer|.\[[ArrayBufferByteLength]]
+        &minus; |startingOffset|.
+    1.  For |i| in [=the range=] |startingOffset| to |startingOffset| + |bytes|'s [=byte
+        sequence/length=] &minus; 1, inclusive, perform [=!=] [$SetValueInBuffer$](|arrayBuffer|,
+        |i|, Uint8, |bytes|[|i| - |startingOffset|], false, Unordered).
```
syg, line 9030: Same as above, I'd pass true instead of false here.

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+    1.  If |bufferSource| has a \[[ViewedArrayBuffer]] [=internal slot=], then:
+        1.  Set |arrayBuffer| to |bufferSource|.\[[ViewedArrayBuffer]].
+        1.  Set |offset| to |bufferSource|.\[[ByteOffset]].
+        1.  Set |length| to |bufferSource|.\[[ByteLength]].
+    1.  Otherwise, set |length| to |bufferSource|.\[[ArrayBufferByteLength]].
+    1.  If [=!=] [$IsDetachedBuffer$](|arrayBuffer|) is true, then return the empty
+        [=byte sequence=].
+    1.  Let |bytes| be a new [=byte sequence=] of [=byte sequence/length=] equal to |length|.
+    1.  For |i| in [=the range=] |offset| to |offset| + |length| &minus; 1, inclusive, set
+        |bytes|[|i| &minus; |offset|] to [=!=] [$GetValueFromBuffer$](|arrayBuffer|, |i|, Uint8,
```
syg, line 9009: On second read, I'd pass true instead of false here, which basically means "aligned"; see this note for an explanation. For byte-wise writes I don't think it's actually observable, but it's editorially clearer IMO.

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+easily be changed by the script author or other APIs at unpredictable times. This is especially true
+if the [{{AllowShared}}] [=extended attribute=] is involved. For the non-shared cases, it is
+recommended to [=ArrayBuffer/transfer=] the {{ArrayBuffer}} first if possible, to ensure the writes
+cannot overlap with other modifications, and then give the new {{ArrayBuffer}} instance to author
+code as necessary.
+
+<div algorithm>
+    To <dfn id="dfn-detach" for="ArrayBuffer" export>detach</dfn> an {{ArrayBuffer}} |arrayBuffer|:
+
+    1.  Asssert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
```
syg, line 9058:
```suggestion
    1.  Assert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
```

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+    <p class="note">This will throw an exception if |arrayBuffer| has an \[[ArrayBufferDetachKey]]
+    that is not undefined, such as a {{Memory|WebAssembly.Memory}}'s {{Memory/buffer}}.
+    [[WASM-JS-API-1]]
+</div>
+
+<div algorithm>
+    To <dfn for="ArrayBuffer" export>transfer</dfn> an {{ArrayBuffer}} |arrayBuffer|, optionally
+    given a [=Realm=] |realm|:
+
+    1.  Asssert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
```
syg, line 9070:
```suggestion
    1.  Assert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
```

index.bs
```diff
@@ -8994,37 +8964,126 @@ an IDL value of any [=buffer source type=]
...
+    1.  Assert: [=!=] [$IsDetachedBuffer$](|arrayBuffer|) is false.
+    1.  Let |arrayBufferData| be |arrayBuffer|.\[[ArrayBufferData]].
+    1.  Let |arrayBufferByteLength| be |arrayBuffer|.\[[ArrayBufferByteLength]].
+    1.  Perform [=?=] [$DetachArrayBuffer$](|arrayBuffer|).
+    1.  If |realm| is not given, let |realm| be |arrayBuffer|.\[[Realm]].
```
syg, line 9075: I don't think ABs have a [[Realm]], only functions do.

syg (approved):
lgtm modulo the [[Realm]] question for ABs.

index.bs
domenic: I guess we can just use Construct(); TypedArrayCreate doesn't buy us much.

index.bs
domenic: Hmm, right. I guess that is why streams uses the current Realm. Probably that's the right fallback here.

annevk: They do have a realm per IDL, see https://heycam.github.io/webidl/#dfn-associated-realm. See also https://github.com/tc39/ecma262/issues/1333.

domenic:
Can I get an editor review? I'm not sure merging this will work given the travis-ci.org shutdown, but I think we should at least try.

index.bs
```diff
@@ -8994,37 +8963,126 @@ an IDL value of any [=buffer source type=]
...
+    1.  Assert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
+    1.  Assert: [=!=] [$IsDetachedBuffer$](|arrayBuffer|) is false.
+    1.  Let |arrayBufferData| be |arrayBuffer|.\[[ArrayBufferData]].
+    1.  Let |arrayBufferByteLength| be |arrayBuffer|.\[[ArrayBufferByteLength]].
+    1.  Perform [=?=] [$DetachArrayBuffer$](|arrayBuffer|).
+    1.  If |realm| is not given, let |realm| be the [=current Realm=].
+    1.  Let |transferred| be [=!=]
+        [$OrdinaryCreateFromConstructor$](|realm|.\[[Intrinsics]].[[{{%ArrayBuffer%}}]],
+        "%ArrayBuffer.prototype%", « \[[ArrayBufferData]], \[[ArrayBufferByteLength]],
+        \[[ArrayBufferDetachKey]] »).
```
TimothyGu, line 9078: Having a list of internal slots here seems relatively brittle. Can we say

> Let _transferred_ be ? [AllocateArrayBuffer](https://tc39.es/ecma262/#sec-allocatearraybuffer)(_realm_.[[Intrinsics]].[[%ArrayBuffer%]], 0).

index.bs
```diff
@@ -8994,37 +8963,126 @@ an IDL value of any [=buffer source type=]
...
+cannot overlap with other modifications, and then give the new {{ArrayBuffer}} instance to author
+code as necessary.
+
+<div algorithm>
+    To <dfn id="dfn-detach" for="ArrayBuffer" export>detach</dfn> an {{ArrayBuffer}} |arrayBuffer|:
+
+    1.  Assert: [=!=] [$IsSharedArrayBuffer$](|arrayBuffer|) is false.
+    1.  Perform [=?=] [$DetachArrayBuffer$](|arrayBuffer|).
+
+    <p class="note">This will throw an exception if |arrayBuffer| has an \[[ArrayBufferDetachKey]]
```
TimothyGu, line 9060: Might be good to note that detaching an ArrayBuffer that's already detached is a no-op.

TimothyGu (commented):
Suggestions for a few more concepts to define:

1. BufferSource/is detached: basically an IDL version of IsDetachedBuffer that also supports views.
2. ArrayBufferView/get the underlying ArrayBuffer: spares specs from having to step into ES land in order to transfer the input buffer.
3. (possibly) ArrayBufferView/transfer

TimothyGu:
Is "get a reference to the bytes held by the buffer source" used anywhere?

index.bs
```diff
@@ -8994,37 +8963,126 @@ an IDL value of any [=buffer source type=]
...
+
+<div algorithm>
+    To <dfn export for="ArrayBuffer">create</dfn> an {{ArrayBuffer}} from a [=byte sequence=]
+    |bytes| in a [=Realm=] |realm|:
+
+    1.  Let |arrayBuffer| be [=?=]
+        [$AllocateArrayBuffer$](|realm|.\[[Intrinsics]].[[{{%ArrayBuffer%}}]], |bytes|'s
+        [=byte sequence/length=]).
+    1. [=ArrayBuffer/Write=] |bytes| into |arrayBuffer|.
+    1. Return |arrayBuffer|.
```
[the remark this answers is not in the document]
domenic: I often conflate the two in other specs, but I guess in this particular spec we want to be clear.

index.bs
```diff
@@ -14285,7 +14343,7 @@ A {{DOMException}} is represented by a
 
 <div algorithm="to create a simple exception or DOMException">
 
-    To [=create=] a [=simple exception=] or {{DOMException}} |E|, with a string giving the
+    To create a [=simple exception=] or {{DOMException}} |E|, with a string giving the
```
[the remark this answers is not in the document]
domenic: The "create" term is now ambiguous and this didn't feel like it was referencing the existing "create" to me so I deleted it instead of disambiguating.

index.bs
domenic: The downside is that this creates a temporary data block, but I guess specs shouldn't worry about such details.

domenic: I think I'm going to keep the ! on the assumption that zero-byte allocation should never fail.

index.bs
```diff
@@ -8994,37 +8963,126 @@ an IDL value of any [=buffer source type=]
...
+    1.  [=ArrayBuffer/Write=] |bytes| into |view|.\[[ViewedArrayBuffer]] with
+        <i>[=ArrayBuffer/write/startingOffset=]</i> set to |view|.\[[ByteOffset]] +
+        |startingOffset|.
+</div>
+
+<p class="advisement">Extreme care must be taken when writing specification text that
+[=ArrayBuffer/writes=] into an {{ArrayBuffer}} or {{ArrayBufferView}}, as the underlying data can
+easily be changed by the script author or other APIs at unpredictable times. This is especially true
+if the [{{AllowShared}}] [=extended attribute=] is involved. For the non-shared cases, it is
+recommended to [=ArrayBuffer/transfer=] the {{ArrayBuffer}} first if possible, to ensure the writes
```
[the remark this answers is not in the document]
domenic: Done, although I'm not 100% sure on the wording; let me know what you think.

domenic:
I added BufferSource/is detached for now.

On the others, it occured to me that we have two directions here: super-polymorphism, where everything is defined on BufferSources, or something more true to the model where ArrayBuffers are special. I was curious for your thoughts.

Super-polymorphism:
  - Maybe merge ArrayBuffer/create and ArrayBufferView/create into BufferSource/create (with ArrayBuffer just being one of the types you can specify)
  - Merge ArrayBuffer/write and ArrayBufferView/write
  - Replace ArrayBuffer/detach with BufferSource/detach
  - Replace ArrayBuffer/transfer BufferSource/transfer

ArrayBuffers are special:

  - Define ArrayBufferView/get the underlying ArrayBuffer
  - Don't define ArrayBufferView/transfer or ArrayBufferView/detach; make it clear that you're reaching below the view level and messing with the buffer beneath
  - Maybe change BufferSource/is detached to just ArrayBuffer/is detached, for similar reasons (although it's less important for introspection)
  - Probably keep ArrayBuffer/write and ArrayBufferView/write separate, to make it clear when you're writing into a view that you only have influence over the view's range

> Is "get a reference to the bytes held by the buffer source" used anywhere?

The one place I was aware of was Encoding. But I just thought to check a bit more and [Web Audio](https://webaudio.github.io/web-audio-api/) is a heavy consumer. I should prepare a PR for them before we merge this.

annevk:
I think it's fine to to go with "super-polymorphism" as long as those algorithms are appropriately detailed. I don't really see a downside in terms of what is possible (especially since we can provide a way to get the underlying buffer once we identify a need for that) and it makes it easier to write prose. It also shouldn't get in the way of #865 which is still something we need to do I think.

index.bs
```diff
@@ -8994,37 +8963,167 @@ an IDL value of any [=buffer source type=]
...
+
+    <p class="note">This will throw an exception if |esArrayBuffer| has an \[[ArrayBufferDetachKey]]
+    that is not undefined, such as is the case with the value of {{Memory|WebAssembly.Memory}}'s
+    {{Memory/buffer}} attribute. [[WASM-JS-API-1]]
+
+    <p class="note">Detaching a buffer that is already detached is a no-op.
+</div>
+
+<div algorithm>
+    A {{BufferSource}} |bufferSource| <dfn for="BufferSource" export>is detached</dfn> if the
```
TimothyGu, line 9095:
```suggestion
    A {{BufferSource}} |bufferSource| is <dfn for="BufferSource" export>detached</dfn> if the
```

index.bs
```diff
@@ -8994,37 +8963,167 @@ an IDL value of any [=buffer source type=]
...
+    1.  Let |esArrayBuffer| be be the result of [=converted to an ECMAScript value|converting=]
+        |arrayBuffer| to an ECMAScript value.
+    1.  Assert: [=!=] [$IsSharedArrayBuffer$](|esArrayBuffer|) is false.
+    1.  Perform [=?=] [$DetachArrayBuffer$](|esArrayBuffer|).
+
+    <p class="note">This will throw an exception if |esArrayBuffer| has an \[[ArrayBufferDetachKey]]
+    that is not undefined, such as is the case with the value of {{Memory|WebAssembly.Memory}}'s
+    {{Memory/buffer}} attribute. [[WASM-JS-API-1]]
+
+    <p class="note">Detaching a buffer that is already detached is a no-op.
```
TimothyGu, line 9091:
```suggestion
    <p class="note">Detaching a buffer that is already [=BufferSource/detached=] is a no-op.
```

TimothyGu (approved):
> On the others, it occured to me that we have two directions here: super-polymorphism, where everything is defined on BufferSources, or something more true to the model where ArrayBuffers are special. I was curious for your thoughts.

Here's an alternative: super-polymorphism for read-like operations (byte length, is detached, get a copy), ArrayBuffers are special for write-like operations (create, write, detach, transfer). The name "BufferSource" is fairly explicit about being the _source_, so I don't think it's worthwhile to make write operations work for any BufferSource.

Come to think of it, this layout is what you have already!

annevk:
We need write for all BufferSource types, no? How else would you write into a Uint8Array respecting the length and such?

TimothyGu:
@annevk Yeah, it’s more like we define it separately from ArrayBuffer so we’d have ArrayBuffer/write and ArrayBufferView/write, rather than a combined BufferSource/write.

index.bs
```diff
@@ -76,6 +76,11 @@ urlPrefix: https://w3c.github.io/mediacapture-main/; spec: MEDIACAPTURE-STREAMS
 urlPrefix: http://www.unicode.org/glossary/; spec: UNICODE
     type: dfn
         text: Unicode scalar value; url: unicode_scalar_value
+urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASM-JS-API-1
+    type: interface
+        text: Memory; url: #memory
```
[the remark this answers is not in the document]
domenic: Actually this can now be deleted since https://github.com/tabatkins/bikeshed/issues/2061 got merged.

domenic:
Alright, we'll merge, but TBD if it'll actually deploy given the Travis situation...

bathos:
> > Is "get a reference to the bytes held by the buffer source" used anywhere?
> 
> The one place I was aware of was Encoding. But I just thought to check a bit more and [Web Audio](https://webaudio.github.io/web-audio-api/) is a heavy consumer. I should prepare a PR for them before we merge this.

Another is in WebAuthn: https://w3c.github.io/webauthn/#sctn-createCredential

PR #988, opened 2021-06-02
Meta: fix linking to "incumbent settings object"

domenic:
Although it was un-exported from HTML recently, Web IDL is one of the very few specs which has legitimate reason to link to it.


annevk (approved):
Should we add some kind of source comment about this? This kinda makes me think we should still rename the term as not exporting it might not dissuade folks from not using it.

domenic:
I'm hopeful it's clear enough that Web IDL is special... we should keep our eyes peeled though.

PR #992, opened 2021-06-22
Meta: move to GitHub Actions

domenic:


TimothyGu:
LGTM if it works. I don't see Actions running on this PR though?

domenic:
Yeah, that is a bit worrying. Maybe it's a function of the rights I have to the repo? :/

domenic:
Yeah, so, we merged #987 but it's not deploying.

I'm somewhat hopeful that even if GitHub won't run first-time actions for PRs, it'll start running them after we merge this? So it might be worth trying to merge this and seeing what happens?

annevk:
@heycam maybe you have to turn this on somewhere as admin?

domenic:
Ahah, I had the wrong branch name. When that is corrected it ran. Can I get a green checkmark?

TimothyGu (approved):

PR #1006, opened 2021-07-30
More fixes for observable array type

EdgarChen:
Per https://heycam.github.io/webidl/#es-observable-array, 
> Instead of the usual conversion algorithms, observable array types have special handling as part of the [attribute getter]((https://heycam.github.io/webidl/#dfn-attribute-getter) and [attribute setter](https://heycam.github.io/webidl/#dfn-attribute-setter) algorithms.

So I think using observable array type as the inner type of an observable array type, e.g. `ObservableArray<ObservableArray<boolean>>`, or a nullable type, e.g. `ObservableArray<boolean>?`, should be disallowed. Otherwise, spec isn't clear how to handle/get the observable array exotic objects for the inner types.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6368,9 +6368,10 @@ character after an existing type.
 The [=nullable types/inner type=] must not be:
 
 *   {{any}},
-*   a [=Promise type=],
+*   a [=promise type=],
+*   an [=observable array type=],
 *   another nullable type, or
-*   a [=union type=] that itself [=includes a nullable type=]
+*   an [=union type=] that itself [=includes a nullable type=]
     or has a dictionary type as one of its [=flattened member types=].
 
 Note: Although dictionary types can in general be nullable,
@@ -6825,10 +6826,12 @@ An <dfn id="dfn-observable-array-type" export>observable array type</dfn> is a p
 whose values are references to a combination of a mutable list of objects of type |T|, as well as
 behavior to perform when author code modifies the contents of the list.
 
-The parameterized type must not be a [=dictionary type=], [=sequence type=], or [=record type=].
+The parameterized type must not be a [=dictionary type=], [=sequence type=], [=record type=], or [=observable array type=].
 
 Similar to [=sequence types=] and [=frozen array types=], observable array types wrap around
 ECMAScript array types, imposing additional semantics on their usage.
+
+Observable array types are non-nullable, but T may be nullable.
 
 Observable array types must only be used as the type of [=regular attributes=] defined on an [=interface=].
 
--- a/index.bs
+++ b/index.bs
@@ -6371,7 +6371,7 @@ The [=nullable types/inner type=] must not be:
 *   a [=promise type=],
 *   an [=observable array type=],
 *   another nullable type, or
-*   an [=union type=] that itself [=includes a nullable type=]
+*   a [=union type=] that itself [=includes a nullable type=]
     or has a dictionary type as one of its [=flattened member types=].
 
 Note: Although dictionary types can in general be nullable,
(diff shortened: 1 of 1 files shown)


EdgarChen:
Please let me know if I miss something. Thanks!

index.bs
```diff
@@ -6368,9 +6368,10 @@ character after an existing type.
 The [=nullable types/inner type=] must not be:
 
 *   {{any}},
-*   a [=Promise type=],
+*   a [=promise type=],
+*   an [=observable array type=],
 *   another nullable type, or
-*   a [=union type=] that itself [=includes a nullable type=]
+*   an [=union type=] that itself [=includes a nullable type=]
```
domenic, line 6374: Nit: this should stay "a", see https://www.writersdigest.com/write-better-fiction/a-before-consonants-and-an-before-vowels-is-not-the-rule

index.bs
```diff
@@ -6825,11 +6826,13 @@ An <dfn id="dfn-observable-array-type" export>observable array type</dfn> is a p
 whose values are references to a combination of a mutable list of objects of type |T|, as well as
 behavior to perform when author code modifies the contents of the list.
 
-The parameterized type must not be a [=dictionary type=], [=sequence type=], or [=record type=].
+The parameterized type must not be a [=dictionary type=], [=sequence type=], [=record type=], or [=observable array type=].
 
 Similar to [=sequence types=] and [=frozen array types=], observable array types wrap around
 ECMAScript array types, imposing additional semantics on their usage.
 
+Observable array types are non-nullable, but T may be nullable.
```
domenic, line 6834:
```suggestion
Observable array types are non-nullable, but |T| may be nullable.
```

domenic (commented):
Great catch.

We should probably also disallow `sequence<ObservableArray<T>>`, `record<DOMString, ObservableArray<T>>`, `FrozenArray<ObservableArray<T>>`, and `Promise<ObservableArray<T>>`.

index.bs
```diff
@@ -6825,11 +6826,13 @@ An <dfn id="dfn-observable-array-type" export>observable array type</dfn> is a p
 whose values are references to a combination of a mutable list of objects of type |T|, as well as
 behavior to perform when author code modifies the contents of the list.
 
-The parameterized type must not be a [=dictionary type=], [=sequence type=], or [=record type=].
+The parameterized type must not be a [=dictionary type=], [=sequence type=], [=record type=], or [=observable array type=].
```
TimothyGu, line 6829:
```suggestion
The parameterized type |T| must not be a [=dictionary type=], [=sequence type=], [=record type=],
or [=observable array type=]. However, |T| may be nullable.
```

index.bs
TimothyGu, line 6834: I'd suggest removing this sentence, and have the nullability of |T| moved up. It's not that observable array types themselves are non-nullable (every type other than nullable types are non-nullable), but more like they cannot be the inner type of a nullable type.

TimothyGu (commented):
Some nits, but LGTM overall.

TimothyGu:
> We should probably also disallow `sequence<ObservableArray<T>>`, `record<DOMString, ObservableArray<T>>`, `FrozenArray<ObservableArray<T>>`, and `Promise<ObservableArray<T>>`.

Do we have to say this, considering we already restrict ObservableArray to only be used for interface attributes?

domenic:
> Do we have to say this, considering we already restrict ObservableArray to only be used for interface attributes?

Oh, good point, you are right. OK then this PR looks good modulo nits.

index.bs
TimothyGu: Actually, I see the same sentence is used for promise types. Consider this optional then.

EdgarChen: Yeah, I saw the same sentence is used for promise types, so I added this.
But I think you are right. And the nullable type section already mentions inner type could not be an observable array type, it should be clear enough.

domenic (approved):

Issue #1016, opened 2021-08-30
Move to WHATWG


annevk:
I've started thinking about the steps needed to move this document to the WHATWG and I would appreciate some review (feel free to edit this comment as needed):

- [x] Create `webidl.spec.whatwg.org` and ensure it can be deployed to. @foolip?
- [x] Create a logo. Anyone ideas? https://github.com/whatwg/whatwg.org/pull/377
- [x] Ensure the Twitter account is associated with the Twitter bot.
- [x] Create a PR that updates `index.bs` to follow WHATWG conventions. See #1018.
- [x] Align wpt directory name with shortname: https://github.com/web-platform-tests/wpt/pull/30327.
- [x] @heycam to move the repository to the WHATWG GitHub organization.
- [x] @heycam to set up a redirect for `heycam.github.io/webidl/`. Presumably as per https://gist.github.com/domenic/1f286d415559b56d725bee51a62c24a7.
- [x] Update the various repository settings, such as URL, description, branches, etc.
- [x] Update the labels.
- [x] Rename the master branch to main.
- [x] Merge https://github.com/whatwg/sg/pull/167. (This should probably not happen too soon as some tooling uses this database.)
- [x] Merge the PR for `index.bs`.
- [x] Run https://github.com/whatwg/spec-factory and merge the resulting PR. (We wanted to update spec-factory as well for this if I recall correctly. We could decide to not block on that and instead modify the resulting PR as needed to keep some of the existing modifications.) See https://github.com/whatwg/spec-factory/pull/31. The output is already included in the `index.bs` PR.
- [x] Check the box in https://github.com/w3c/whatwg-coord/issues/3.
- [x] Update wpt readme to point to the spec: https://github.com/web-platform-tests/wpt/pull/30355.
- [x] Ensure participate.whatwg.org is updated: https://github.com/whatwg/sg/issues/171
- [x] Make changes to specref https://github.com/tobie/specref/pull/686
- [x] Changes to Shepherd?
- [x] Publish a RD draft at the appropriate time (the first one needs some help)

cc @heycam @TimothyGu @domenic @EdgarChen @foolip

domenic:
> Create a logo. Anyone ideas?

Curly brackets, e.g. `{}`, perhaps? Not sure the font is great but here's a hacked-together idea of what it would look like: https://jsbin.com/jefomakuma/edit?html,output

bathos:
[...]

PR #1018, opened 2021-08-31
Turn Web IDL into a Living Standard

annevk:
Helps with #1016.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 3, 2021, 2:15 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fheycam%2Fwebidl%2Fb20c4900899e028d5e123659c6ab21c71f1d2e87%2Findex.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 2.
FATAL ERROR: Found unmatched text macro [COMMIT-SHA]. Correct the macro, or escape it with a leading backslash.
 ✘  Did not generate, due to fatal errors
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20heycam/webidl%231018.)._
</details>


annevk:
A thing I did not realize is that Web IDL has some Node.js-based post processing as well. It's not clear to me if spec-factory's `post_build_step` is sufficient for that. I suppose it might be...

annevk:
I essentially added
```
deploy: index.bs
	curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh
	POST_BUILD_STEP='npm run pp-webidl -- --input index.html' \
	bash ./deploy.sh
```
to the `Makefile` to see what kind of errors I would run into. There were a lot. 😟

I ended up suppressing RFC2119 warnings. Given that "optional" is a term that's frequently used that ended up generating too much noise (I suspect "required" as well, but I didn't get that far).

index.bs
```diff
@@ -2807,7 +2753,7 @@ On a given [=interface=], there must exist at most one stringifier.
         };
     </pre>
 
-    Thus, prose is required to explain the stringification behavior.
+    Thus, prose needs to explain the stringification behavior.
```
andreubotella, line 2756:
```suggestion
    Thus, there needs to be prose to explain the stringification behavior.
```

I read this as an existential: "_some_ prose is required to explain the stringification behavior". The changed version doesn't allow that reading, and so doesn't sound very idiomatic.

andreubotella:
The default WHATWG boilerplate includes an entry in the spec metadata section that links to the WPT tests, and I noticed that for Web IDL, the WPT folder it links to is `webidl`, in lowercase, rather than the correct `WebIDL`. This results in a 404 when you visit the link. The Bikeshed boilerplate should probably include a `WPTDIRNAME` macro that defaults to the shortname.

annevk:
Really appreciate your help with this @andreubotella!

@domenic want to take a look? @TimothyGu you might also be interested?

.github/workflows/build.yml
```diff
@@ -2,25 +2,29 @@ name: Build
 on:
   pull_request:
     branches:
-    - master
+    - main
```
domenic, line 5: We might want to do the branch rename separately?

Makefile
```diff
@@ -1,62 +1,28 @@
...
-ci:
-	mkdir -p out
-	make index.html
-	mv index.html out/index.html
+local: index.bs
+	bikeshed spec index.bs index.html --md-Text-Macro="COMMIT-SHA LOCAL COPY"
+
+deploy: index.bs
+	curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh
+	POST_BUILD_STEP='node ./check-grammar.js "$$DIR/index.html" && npm run pp-webidl -- --input "$$DIR/index.html"' \
```
domenic, line 23: Where does `$$DIR` come from?

annevk: It's passed in as an argument. Encoding does the same thing.

.github/workflows/build.yml
annevk: I was thinking that we do the repo move, branch rename, and then attempt to run and land this. Is there any value in separating it?

domenic: I think that sequence would work.

index.bs
```diff
@@ -532,7 +476,7 @@ in [[#es-extended-attributes]].
         IncludesStatement
 </pre>
 
-<div class="example">
+<div class="example" id="example-25f60a7c">
```
domenic, line 479: Can we give these examples real IDs? I care about this a good bit so if you can't be bothered I guess I'll volunteer :)

annevk: These are the current IDs, for better or worse. I guess I don't feel too strongly about keeping these IDs working, but folks might be linking to them. (I haven't checked, but either this is disabled or Bikeshed must generate a lot of warnings for the document in its current state.)

domenic: Oh, that turns this from a "we should fix before merging" to a "we should maybe fix eventually".

foolip (approved):

foolip:
Alright, the first deploy worked!
https://webidl.spec.whatwg.org/
https://webidl.spec.whatwg.org/commit-snapshots/37c0e6fe29c50d9c19842135f3f2b265ddcbe932/

PR #1028, opened 2021-10-05
Editorial: Web IDL is with a space

annevk:


foolip (approved):
I had the same commit locally, but hadn't finished writing the commit message 😆

PR #1030, opened 2021-10-05
Normative: Rename integer types to match ECMAScript typed arrays

ExE-Boss:
I’ve separated out the integer types rename from <https://github.com/heycam/webidl/pull/856>, as it has greater support and can be done without waiting for parser changes (see also <https://github.com/whatwg/webidl/issues/477>).

---

This renames the integer types to match **ECMAScript** typed arrays in order to resolve the conflict between [the **Infra** byte](infra.spec.whatwg.org/#bytes) and [the **IDL** `byte`](https://webidl.spec.whatwg.org/#idl-byte); refs <https://github.com/whatwg/webidl/issues/843#issuecomment-586203060>:
> (One problem it would resolve is that IDL and Infra conflict on byte. What Infra calls byte IDL currently calls octet.)

Floating‑point types were left for a future **PR** because it’s not possible to simply rename `unrestricted float`/`unrestricted double` to `unrestricted float32`/`unrestricted float64` without the parser being updated first, which is why that depends on something like <https://github.com/whatwg/webidl/pull/857>.

---

I decided to go with `int*` for signed integers since it’s what **ECMAScript** typed arrays use, and because **ECMAScript** provides built‑in support for 32&#x2011;bit signed integers using the bitwise **OR** operator, but none for unsigned integers:

```js
let numInt32 = (num | 0); // Converts `num` to a 32-bit signed integer
```

**ECMAScript** `Array`s also restrict the length to a positive 32&#x2011;bit signed integer.

---

- **Part of:** <https://github.com/whatwg/webidl/issues/843>
- **Extracted from:** <https://github.com/whatwg/webidl/pull/856>
</ul>

	* …
	* …
	* …
	* Chrome: …
	* Firefox: …
	* Safari: …


TimothyGu:
I think this is a good change if we can pull it off, with regards to the tooling and downstream review impact.

index.bs
```diff
@@ -5837,108 +5845,164 @@ The [=type name=] of the
...
 
 The [=type name=] of the
-{{unsigned short}} type is "<code>UnsignedShort</code>".
+{{uint16}} type is "<code>Uint16</code>".
+
+For legacy compatibility, <code>unsigned short</code> is to be treated
+as an alias for {{uint16}}, similar to:
+
+<pre>
+    typedef (uint16) unsigned short;
```
saschanaz, line 5924: I believe this causes syntax error, and not sure it's really needed. Each parser implementation can support both form while autofixes by @autokagami land.

index.bs
```diff
@@ -7512,7 +7584,7 @@ In effect, where <var ignore>x</var> is a Number value,
         1.  If |signedness| is "<code>unsigned</code>", then let |lowerBound| be 0.
         1.  Otherwise let |lowerBound| be −2<sup>53</sup> + 1.
 
-            Note: this ensures {{long long}} types
+            Note: this ensures {{int64}} types
```
bathos, line 7587: This note pertained to both unsigned long long and long long, which already wasn’t very clear as written, but with “int64,” it’s even less clear that it means both variants (the upper bound clips both). Seems like listing both int64 and uint64 explicitly would fix that (or, given it’s a note, it could probably be worded like “64-bit integer types”).

index.bs
```diff
@@ -689,7 +689,7 @@ across [=IDL fragments=].
         interface A {
         };
 
-        typedef sequence&lt;long&gt; SequenceOfLongs;
+        typedef sequence&lt;int32&gt; SequenceOfLongs;
```
bathos, line 692:
```suggestion
        typedef sequence&lt;int32&gt; SequenceOfInt32s;
```

(or something. just seems weird to leave “longs” here.)

index.bs
```diff
@@ -1694,7 +1694,7 @@ on which they appear.  It is language binding specific whether
     <pre highlight="webidl">
         [Exposed=Window]
         interface A {
-          const short rambaldi = 47;
+          const uint16 rambaldi = 47;
```
bathos, line 1697:
```suggestion
          const int16 rambaldi = 47;
```

This one was signed before. Almost certainly doesn’t matter here for the example, but keeping them the same may help avoid tripping reviewers.

index.bs
bathos: The “similar to” phrasing may have intended to acknowledge that — but it still seems confusing to have pseudo-idl in the idl spec itself (not very friendly for folks not already familiar enough to know it’s not meant to be valid).

Given the original productions are still explicitly part of the grammar in this PR, I don’t think this framing is needed anyway.

saschanaz:
Thanks for removing the pseudo IDL blocks but I mean the aliases are probably not needed at all, and parsers can support both in the reasonably short term while @autokagami sends patches. It does depend on how quickly maintainers will respond (and plus, on whether other people here are convinced), though.

PR #1090, opened 2022-01-26
Introduce "mark as handled" for promises

### LINKED ISSUE #829: Define an operation to make a promise as handled
Split from https://github.com/whatwg/fetch/pull/959.

Streams and Fetch sometimes manipulate promise.\[[PromiseIsHandled]] to suppress unhandled rejection error reports. We'd appreciate if this is defined here so that they don't need to manipulate the internal property directly.

@annevk, @domenic, @ricea

domenic:
This helps specifications avoid reaching into internal slots. Closes #829.

Also includes some editorial touchups to nearby promise definitions.

***
<a href="https://whatpr.org/webidl/1090.html" title="Last updated on Feb 2, 2022, 10:17 PM UTC (c930213)">Preview</a> | <a href="https://whatpr.org/webidl/1090/ae0de45...c930213.html" title="Last updated on Feb 2, 2022, 10:17 PM UTC (c930213)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8390,7 +8390,7 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=] 
         |newCapability|).
 
     Note: This algorithm will behave in a very similar way to the
-    {{%Promise.prototype.then%|Promise.then()}} method.
+    {{%Promise.prototype.then%|promise.then()}} method.
     In particular, if the steps return a value of type |U| or
     <code><a interface>Promise</a>&lt;|U|&gt;</code>, this algorithm returns a
     <code><a interface>Promise</a>&lt;|U|&gt;</code> as well.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8465,11 +8465,24 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=]
...
 
-This phrase is useful when you wish to aggregate the results of multiple promises, and then produce
-another promise from them, in the same way that {{Promise/all()|Promise.all()}} functions for
-JavaScript code.
+<div algorithm>
+    To <dfn export lt="mark a promise as handled|mark as handled">mark as handled</dfn> a
+    <code><a interface>Promise</a>&lt;<var ignore>T</var>&gt;</code> |promise|, set
+    |promise|.\[[Promise]].\[[PromiseIsHandled]] to true.
+
+    <p class="note">This definition is useful for promises which you expect rejections to often be
```
annevk, line 8480: for which you expect?

index.bs
```diff
@@ -8465,11 +8465,24 @@ IDL [=promise type=] values are represented by ECMAScript [=PromiseCapability=]
...
+    ignored; it ensures such promises do not cause {{Window/unhandledrejection}} events. The most
+    common use case is for promise properties, which the web developer might or might not consult.
+    An example is the {{WritableStreamDefaultWriter/closed|writableStreamWriter.closed}} promise,
+    which is [=marked as handled=] immediately after it is [=rejected=].
```
annevk, line 8484: Would it make sense to offer an operation that rejects and marks as handled as a combined operation? Or are there cases where they need to be disjoint?

domenic: So what I've found recently (in app history) is that it's actually simplest to just mark the promise as handled upon creation. What streams does is perhaps unnecessarily complicated, and I wonder if maybe I should remove this example or make it more vague.

The advantage of this is that you can centralize your "this promise is special" logic at creation time, instead of all the different places you reject it. So if anything maybe a "new already-handled promise" algorithm would be useful? But for now my instinct is to just expose the primitive, since so far use cases are pretty rare.

domenic: For now I just removed "which is marked as handled immediately after it is rejected" to make it a bit more vague.

annevk (approved):

PR #1098, opened 2022-02-07
Allow same identifier in static and regular ops

### LINKED ISSUE #1097: Allow static operations to have the same name as regular operations
WebIDL currently specifies that:

> The identifier of a static operation also must not be the same as the identifier of a regular operation defined on the same interface.

In light of https://github.com/whatwg/fetch/issues/1389, I'd like to remove this limitation. The limitation was originally added in https://github.com/whatwg/webidl/commit/082298c0f7e4d231d121823778f820f7d852ca55 by @heycam. Does anyone remember why?

Is anyone opposed to this change?

lucacasonato:
Closes #1097

   * Deno
   * Cloudflare Workers
   * Chromium
   * Mozilla
   N/A
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1305399
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1765992
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=239672
   * Deno: already supported
   * Node.js: https://github.com/nodejs/node/issues/42831

***
<a href="https://whatpr.org/webidl/1098.html" title="Last updated on May 10, 2022, 10:13 AM UTC (2915f6b)">Preview</a> | <a href="https://whatpr.org/webidl/1098/ae0de45...2915f6b.html" title="Last updated on May 10, 2022, 10:13 AM UTC (2915f6b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2037,7 +2037,7 @@ Note: The identifier can be the same
 as that of another operation on the interface, however.
 This is how operation overloading is specified.
 
-The [=identifier=] of a [=static operation=]
+Note: The [=identifier=] of a [=static operation=]
 may be the same as the identifier
 of a [=regular operation=]
 defined on the same [=interface=].
--- a/index.bs
+++ b/index.bs
@@ -2038,7 +2038,7 @@ as that of another operation on the interface, however.
 This is how operation overloading is specified.
 
 Note: The [=identifier=] of a [=static operation=]
-may be the same as the identifier
+can be the same as the identifier
 of a [=regular operation=]
 defined on the same [=interface=].
 
--- a/index.bs
+++ b/index.bs
@@ -2038,7 +2038,7 @@ as that of another operation on the interface, however.
 This is how operation overloading is specified.
 
 The [=identifier=] of a [=static operation=]
-also must not be the same as the identifier
+may be the same as the identifier
 of a [=regular operation=]
 defined on the same [=interface=].
 


index.bs
```diff
@@ -2038,7 +2038,7 @@ as that of another operation on the interface, however.
 This is how operation overloading is specified.
 
 The [=identifier=] of a [=static operation=]
-also must not be the same as the identifier
+may be the same as the identifier
```
andreubotella, line 2041: This paragraph used to be a normative restriction, but this change turns it into a clarification, so it should probably be a note.

domenic: Don't use "may" in a non-normative note; probably "can" is better here. https://infra.spec.whatwg.org/#conformance

Okiamgood (commented):
Screenshot_20220305-200443
Screenshot_20220305-233720~2

lucacasonato:
@TimothyGu / @EdgarChen would you mind taking a look? This is a prerequisite for https://github.com/whatwg/fetch/pull/1392.

EdgarChen (approved):
This looks good to me.
I checked how we handle the overloaded operations, [there are separated overload set for regular and static operation](https://webidl.spec.whatwg.org/#compute-the-effective-overload-set), so I think it is fine to have same identifier in regular and static operations.

EdgarChen:
@TimothyGu, WDYT?

EdgarChen:
BTW, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1765992 for Gecko.

lucacasonato:
Thanks @EdgarChen. Added the link to the PR description. I'll open a WebKit issue in a moment.

Can I count Gecko and Chromium as interested implementers for the working mode requirements?

jasnell:
+1 to this. Definitely useful

lucacasonato:
cc @tabatkins, as this may affect bikeshed IDL handling too.

tabatkins:
Yeah, thanks for the ping. I will indeed need to make a change in how Bikeshed handles these. (The change makes sense tho, so Bikeshed shouldn't be a blocker here.)

lucacasonato:
Support for this was merged in Chromium (https://chromium-review.googlesource.com/c/chromium/src/+/3623640), so we have two "interested implementers" now. @EdgarChen think we can land this?

foolip:
I think this will impact MDN documentation, because static and instance members aren’t distinguished in the URL convention. Does anyone have a plan for that?

domenic:
From what I understand MDN has had some consensus to move away from their current pretend-all-methods-are-static style for some time, but I haven't seen any actual updates: https://github.com/mdn/content/discussions/5121 and some other discussions.

foolip:
It's been discussed, but there weren't any plans for anyone to work on it. I've commented on https://github.com/mdn/content/discussions/5121 and filed https://github.com/openwebdocs/project/issues/104 to put it on the radar, but I would expect `Response.json` to not be documented (or in BCD) until that's resolved.

PR #1130, opened 2022-04-26
Export getter/setter steps, and note the tools you have access to when defining them.

tabatkins:
I was surprised and glad to see that getters/setters have defined terminology to work from now, but unfortunately the dfns weren't exported.

I also initially wrote my algos explicitly taking in a |this| (and a |val| for the setters) and only coincidentally remembered that [=this=] exists, which is also when I learned that [=the given value=] exists. To help myself and others in the future, I've dropped a Note here to remind of their existence, since the terms are defined elsewhere in the spec and not linked anywhere nearby.

***
<a href="https://whatpr.org/webidl/1130.html" title="Last updated on Apr 26, 2022, 5:50 PM UTC (4643902)">Preview</a> | <a href="https://whatpr.org/webidl/1130/c2b9c12...4643902.html" title="Last updated on Apr 26, 2022, 5:50 PM UTC (4643902)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2351,10 +2351,12 @@ The following extended attributes are applicable to operations:
 [{{NewObject}}], and
 [{{SecureContext}}].
 
-The <dfn>method steps</dfn> of an operation |operation| should be introduced using text of the form
+The <dfn export>method steps</dfn> of an operation |operation| should be introduced using text of the form
 “The <code>|operation|(<var ignore>arg1</var>, <var ignore>arg2</var>, ...)</code> method
 steps are:” followed by a list, or “The <code>|operation|(<var ignore>arg1</var>,
 <var ignore>arg2</var>, ...)</code> method steps are to” followed by an inline description.
+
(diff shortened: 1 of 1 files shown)


domenic:
I guess so far we haven't exported these because we find it stylistically a bit verbose to keep linking them at every usage site. But that's probably not a good reason.

We should probably also export, and add a similar note, to "method steps"?

tabatkins:
I mean, you're already *saying* them at every usage site (at least, if you follow the recommended wording). Adding a `[==]` to the source, and an underline to the output, is basically nil in comparison, and in return you get a nice link making it clear that this is a term of art, and exactly what it means.

And yeah, doing the same for method steps makes sense, I just hadn't seen that yet. I'll throw them in, one sec.

tabatkins:
Okay, exported method and constructor steps, and added note about [=this=] to method steps. Didn't add it to constructor steps, as [=this=] is mentioned a few times in the nearby text already.

domenic (approved):

PR #1135, opened 2022-04-29
Make it clearer that static and namespace attrs don't get to have state. Fixes #1133

### LINKED ISSUE #1133: "attribute getter" steps skip security checks on namespaces
The [attribute getter algo](https://webidl.spec.whatwg.org/#dfn-attribute-getter) can be called with an interface or a namespace. However, step 1.1.2 only checks if the object is an interface; if it is, it does some security checks, and actually grabs the IDL value, so step 1.1.3 can invoke the getter steps for the attribute on it.

A namespace will fail that check, meaning "idlObject" is left as `null` (set in step 1.1.1), and then 1.1.3 invokes the getter steps for the namespace's attribute on `null`. This'll break spec algos that expect the `[=this=]` value to actually be the namespace object the attribute is defined on.

Is this intentional? Are namespaces not intended to be able to carry any internal state that an attribute getter might look at?  (And intended to be accessible cross-origin, since the security check is bypassed? Well, I suppose getting the namespace object from `Window` in the first place would fail, since none of the allowed [cross-origin properties](https://html.spec.whatwg.org/multipage/browsers.html#crossoriginproperties-(-o-)) are namespaces.)  If so, would you mind if I PR'd some editorial fixes that make it clear this is intentional rather than it just being a behavior that falls out of the algo?

tabatkins:
Per #1133, interface objects themselves, and namespaces, aren't intended to have complex internal state, so the algo creating their attribute getters (aka statics for interfaces, or regulars for namespaces) didn't give the "getter steps" access to the object themselves and instead called them with null.

This was somewhat implicit in the algo, so this is a light editorial restructuring that makes it more explicit, and adds a short note explaining it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11880,7 +11880,6 @@ in which case they are exposed on every object that [=implements=] the interface
 
     1.  Let |steps| be the following series of steps:
         1.  Try running the following steps:
-            1.  Let |idlObject| be null.
             1.  If |target| is an [=interface=], and |attribute| is a [=regular attribute=]:
                 1.  Let |esValue| be the <emu-val>this</emu-val> value, if it is not
                     <emu-val>null</emu-val> or <emu-val>undefined</emu-val>, or |realm|'s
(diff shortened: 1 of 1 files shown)


annevk (commented):
I think this introduces some scoping issues. _R_ isn't accessible outside the inner steps.

_idlObject_ also isn't initialized with "Let" anymore.

tabatkins:
Good catches, fixed.

PR #1138, opened 2022-05-02
WIP: switching maplike/setlike to use infra maps and sets

### LINKED ISSUE #254: Are operations on the `[[BackingMap]]` and `[[BackingSet]]` page-hookable?  And if not, why not?
I was sure I had filed this before, but I can't find it...

Consider this testcase:

    <script>
    Object.defineProperty(Set.prototype, "size", { value: 42 });
    var set = document.fonts;
    alert(set.size);
    </script>

What should get alerted?  Per spec as currently written we end up at https://heycam.github.io/webidl/#es-set-size we end up (in step 5) doing:

> Return the result of calling the [[Get]] internal method of _set_ passing “size” and _set_ as arguments.

where _set_ is the `[[BackingSet]]` of the object.  And the `[[BackingSet]]` is defined like so:

> Any object that implements an interface that has a setlike declaration must have a [[BackingSet]] internal slot, which is initially set to a newly created Set object.

So what should get alerted?  Browsers alert 0 if there are no font-face rules.  As an implementor, I would prefer to alert 0 in general.  But if the `[[BackingSet]]` is created in the same Realm as the object itself, then you ought to get 42, right?

We should specify how this actually works.  I see three possible options:

1.  Specify that the `[[BackingSet]]` is created in the global of the IDL object, leave everything as is.  Then this testcase should alert 42, which doesn't match browsers.
2.  Specify that the `[[BackingSet]]` is created in a separate global, leave everything as is.  Then this testcase should alert 0.
3.  Specify that the `size` getter invokes the original value of the `Set.prototype.size` getter.  Then this testcase should alert 0.  Same with the various other forwarding getters.

Options 2 and 3 are _not_ black-box identical.  The reason is that `entries` will return you an iterator, and you can tell which global the `Set`'s internal `entries` function comes from by looking at the prototype of that iterator.  In fact, you can distinguish not only options 2 and 3 from each other but also different variants of option 2 (e.g. a single separate global for all interfaces, one separate global per interface, one separate global per object, etc).

My personal feeling is that we should go with option 3, with the `[[BackingSet]]` created in the IDL object's global.  That's certainly how maplike/setlike are implemented in Gecko.  Note that FontFaceSet is NOT implemented as a setlike in Gecko for the moment, though.  It looks like we ship no actual setlikes so far, and only one maplike: RTCStatsReport.  Unfortunately, I have yet to figure out how to get my hands on an RTCStatsReport to see how it behaves in other browsers...

Doing black-box testing of FontFaceSet in Blink, it looks like the iterator it returns from `entries()` doesn't have `%SetIteratorPrototype%` (from any global) as its prototype.  So Blink is not following the spec for this part either so far.  Grepping Blink's idl, this is the only setlike, and the only maplikes are the MIDIInput/OutputMaps.  Those don't use an object with `%MapIteratorPrototype%` as its proto either.  That makes it hard to tell whether Blink is doing option 2 or option 3, of course, since at least for `entries` it's not doing either one.

@tobie @domenic @heycam @annevk

### LINKED ISSUE #824: make hooking into maplike and setlike methods easier
Currently the spec allows `clear`, `delete`, and `set` on a maplike interfaces (and `add`, `clear`, and `delete` on setlike interfaces) so that spec authors can add validation of inserted keys and values, and to respond to changes in the contents of the backing map or set.

The prose required to do this is a bit awkward.  For example, to override `set` to add syntax checking for the key argument, you would need to write something like this:

> The **set(key, value)** method on the Foo interface, when invoked, must run these steps:
> 1. If _key_ does not contain a hyphen, then throw a TypeError.
> 2. Let _keyArg_ be the result of converting key to an IDL value.
> 3. Let _valueArg_ be the result of converting key to an IDL value.
> 4. Let _result_ be the result of running the steps for a built-in maplike set function, with the context object as the **this** object, and with arguments _keyArg_ and _valueArg_.
> 5. Return _result_.

I think it may be better to disallow overriding these methods, but instead to have some spec hooks that are invoked from the built-in `add`, `clear`, `delete`, and `set` functions.  That way the spec author can remain in IDL value land and doesn't need to deal with invoking the built-in steps.

cc @frivoal

tabatkins:
Fixes #254, fixes #824.

This PR switches maplike and setlike interfaces from using a [[BackingMap]]/[[BackingSet]] ES Map/Set to use an Infra map/set instead. As explained in #254 and #824, this makes the interfaces vastly easier to work with in spec-ese (you can directly manipulate an Infra map/set, rather than having to carefully perform the ES algo dance) , and removes some underdefined behavior (as currently defined, the algorithms look up and utilize the equivalent methods on Map.prototype and Set.prototype, which can be author-supplied).

Unless the author is relying on the "replace the Map.prototype methods" behavior (I haven't tested if impls actually allow this, in any case), this change should be undetectable, and just result in specs using maplike or setlike interfaces having an easier, better-defined time manipulating them.

Progress:
* [x] Switched maplike and setlike to use an Infra map/set instead of a [[BackingMap]]/[[BackingSet]]
* [x] Switched @@iterator to go over the infra map/set (but still return an iterator with the right prototype)
* [x] Switch the rest of the Map methods to use the infra map
* [x] Switch the rest of the Set methods to use the Infra set
* [x] Moved the maplike/setlike @@iterator and forEach methods to live with the rest of their methods, rather than being in their own section
* [x] <s>Add verbage for "map setting steps"/etc for specs to use when defining custom behavior, a la "getter steps"</s>

***
<a href="https://whatpr.org/webidl/1138.html" title="Last updated on May 4, 2022, 8:20 PM UTC (2cb3b55)">Preview</a> | <a href="https://whatpr.org/webidl/1138/c2b9c12...2cb3b55.html" title="Last updated on May 4, 2022, 8:20 PM UTC (2cb3b55)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12915,7 +12915,7 @@ the <code>entries</code> property.
 
 <h5 id="es-map-entries">entries</h5>
 
-There must exist a <code class="idl">entries</code> data property on
+There must exist an <code class="idl">entries</code> data property on
 |A|’s [=interface prototype object=]
 with the following characteristics:
 
@@ -13201,7 +13201,7 @@ the <code class="idl">values</code> property.
 
 <h5 id="es-set-entries" oldids="es-set-entries-keys">entries</h5>
 
-There must exist a <code class="idl">entries</code> data property on
+There must exist an <code class="idl">entries</code> data property on
 |A|’s [=interface prototype object=]
 with the following characteristics:
 
--- a/index.bs
+++ b/index.bs
@@ -4465,7 +4465,9 @@ that have [=attributes=] or [=constants=] with these names.
 Note: Read-write maplike interfaces <em>can</em> have <em>operations</em>
 named "<code>clear</code>", "<code>delete</code>", or "<code>set</code>",
 which will override the default implementation of those methods
-(defined in [[#es-maplike]]).
+(defined in [[#es-maplike]]). If such operations are defined, they must
+match the input and output expectations of each method, defined in their
+default implementation sections.
 
 An interface must not have more than one
 [=maplike declaration=].
@@ -13083,6 +13085,10 @@ The value of the [=function object=]’s <code class="idl">length</code> propert
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>set</code>".
 
+Note: If the interface does declare a <code class=idl>set</code> method,
+it should similarly map a -0 key to +0,
+and must return the value that was set.
+
 
 <h5 id="es-map-delete">delete</h5>
 
@@ -13113,6 +13119,10 @@ The value of the [=function object=]’s <code class="idl">length</code> propert
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>delete</code>".
 
+Note: If the interface does declare a <code class=idl>delete</code> method,
+it should similarly map a -0 key to +0,
+and must return a {{boolean}} indicating whether the key was present or not.
+
 
 <h5 id="es-map-clear">clear</h5>
 
@@ -13139,6 +13149,10 @@ with the following characteristics:
 The value of the [=function object=]’s <code class="idl">length</code> property is the Number value <emu-val>0</emu-val>.
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>clear</code>".
+
+Note: If the interface does declare a <code class=idl>clear</code> method,
+it must preserve the [=map entries=] object (rather than generating a new one)
+and must return {{undefined}}.
 
 
 <h4 id="es-setlike">Setlike declarations</h4>
@@ -13333,6 +13347,10 @@ The value of the [=function object=]’s <code class="idl">length</code> propert
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>add</code>".
 
+Note: If the interface does declare an <code class=idl>add</code> method,
+it should similarly map a -0 value to +0,
+and must return the value that was set.
+
 
 <h5 id="es-set-delete">delete</h5>
 
@@ -13362,6 +13380,10 @@ The value of the [=function object=]’s <code class="idl">length</code> propert
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>delete</code>".
 
+Note: If the interface does declare a <code class=idl>delete</code> method,
+it should similarly map a -0 value to +0,
+and must return a {{boolean}} indicating whether the value was present or not.
+
 
 <h5 id="es-set-clear">clear</h5>
 
@@ -13388,6 +13410,10 @@ with the following characteristics:
 The value of the [=function object=]’s <code class="idl">length</code> property is the Number value <emu-val>0</emu-val>.
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>clear</code>".
+
+Note: If the interface does declare a <code class=idl>clear</code> method,
+it must preserve the [=set entries=] object (rather than generating a new one)
+and must return {{undefined}}.
 
 
 <h3 id="es-platform-objects">Platform objects implementing interfaces</h3>
--- a/index.bs
+++ b/index.bs
@@ -72,11 +72,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
         text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
         text: %IteratorPrototype%; url: sec-%25iteratorprototype%25-object
         text: %Map.prototype%; url: sec-properties-of-the-map-prototype-object
-        text: %MapIteratorPrototype%; url: sec-%mapiteratorprototype%-object
+        text: %MapIteratorPrototype%; url: sec-%25mapiteratorprototype%25-object
         text: %Object.prototype%; url: sec-properties-of-the-object-prototype-object
         text: %Promise%; url: sec-promise-constructor
         text: %Set.prototype%; url: sec-properties-of-the-set-prototype-object
-        text: %SetIteratorPrototype%; url: sec-%setiteratorprototype%-object
+        text: %SetIteratorPrototype%; url: sec-%25setiteratorprototype%25-object
     type: exception; for: ECMAScript
         text: Error; url: sec-error-objects
         text: SyntaxError; url: sec-native-error-types-used-in-this-standard-syntaxerror
--- a/index.bs
+++ b/index.bs
@@ -72,9 +72,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
         text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
         text: %IteratorPrototype%; url: sec-%25iteratorprototype%25-object
         text: %Map.prototype%; url: sec-properties-of-the-map-prototype-object
+        text: %MapIteratorPrototype%; url: sec-%mapiteratorprototype%-object
         text: %Object.prototype%; url: sec-properties-of-the-object-prototype-object
         text: %Promise%; url: sec-promise-constructor
         text: %Set.prototype%; url: sec-properties-of-the-set-prototype-object
+        text: %SetIteratorPrototype%; url: sec-%setiteratorprototype%-object
     type: exception; for: ECMAScript
         text: Error; url: sec-error-objects
         text: SyntaxError; url: sec-native-error-types-used-in-this-standard-syntaxerror
@@ -107,6 +109,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
         text: CreateArrayIterator; url: sec-createarrayiterator
         text: CreateBuiltinFunction; url: sec-createbuiltinfunction
         text: CreateDataProperty; url: sec-createdataproperty
+        text: CreateIteratorFromClosure; url: sec-createiteratorfromclosure
         text: CreateIterResultObject; url: sec-createiterresultobject
         text: CreateMapIterator; url: sec-createmapiterator
         text: CreateMethodProperty; url: sec-createmethodproperty
@@ -118,6 +121,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
             text: CreateMutableBinding
             text: InitializeBinding
         text: FromPropertyDescriptor; url: sec-frompropertydescriptor
+        text: GeneratorYield; url: sec-generatoryield
         text: Get; url: sec-get-o-p
         text: GetFunctionRealm; url: sec-getfunctionrealm
         text: GetIterator; url: sec-getiterator
@@ -163,11 +167,13 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
         text: ToUint16; url: sec-touint16
         text: ToUint32; url: sec-touint32
         text: Type; url: sec-ecmascript-data-types-and-values
+        text: Yield; url: sec-yield
         text: abs; url: eqn-abs
         text: floor; url: eqn-floor
         text: max; url: eqn-max
         text: min; url: eqn-min
-    type: dfn
+    type: dfn;
+        text: Abstract Closure; url: sec-abstract-closure
         text: NumericLiteral; url: sec-literals-numeric-literals
         text: modulo; url: eqn-modulo
         url: sec-returnifabrupt-shorthands
@@ -12355,8 +12361,10 @@ The location of the property is determined as follows:
         then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
     1.  If the interface has a [=maplike declaration=], then
         [=create a map iterator=] from |thisValue|'s [=map entries=]
+        with <var ignore>kind</var> set to "key+value"
         and return the result.
     1.  Otherwise, [=create a set iterator=] from |thisValue|'s [=set entries=]
+        with <var ignore>kind</var> set to "value"
         and return the result.
 </div>
 
@@ -12371,36 +12379,52 @@ if the interface has a [=setlike declaration=].
 
 <div algorithm>
     To <dfn>create a map iterator</dfn>
-    from a [=map=] |map|
+    from a [=/map=] |map|
     and a |kind| ("key+value", "key", or "value"):
 
-1. Let |closure| be a new [=ECMAScript/Abstract Closure=] with no parameters
-    that captures |map| and |kind| and performs the following steps when called:
-    1. Let |index| be 0.
-    1. Let |numEntries| be the [=map/size=] of |map|.
-    1. Repeat, while |index| < |numEntries| and |index| < |map|'s [=map/size=]:
-        1. Let |key| and |value| be the [=map/key=] and [=map/value=] of the
-            |index|th [=map/entry=] of |map|,
-            each [=converted toECMAScript values=].
-        1. Set |index| to |index| + 1.
-        1. If |kind| is "key", let |result| be |key|.
-        1. Else if |kind| is "value", let |result| be |value|.
-        1. Else, let |result| be [$CreateArrayFromList$](« |key|, |value| »)
-        1. Perform [=?=] [$GeneratorYield$]([$CreaterIterResultObject$](|result|, <emu-value>false</emu-value>)).
+    1. Let |closure| be a new [=/Abstract Closure=] with no parameters
+        that captures |map| and |kind| and performs the following steps when called:
+        1. Let |index| be 0.
+        1. Let |numEntries| be the [=map/size=] of |map|.
+        1. Repeat, while |index| < |numEntries| and |index| < |map|'s [=map/size=]:
+            1. Let |key| and |value| be the [=map/key=] and [=map/value=] of the
+                |index|th [=map/entry=] of |map|,
+                each [=converted to ECMAScript values=].
+            1. Set |index| to |index| + 1.
+            1. If |kind| is "key", let |result| be |key|.
+            1. Else if |kind| is "value", let |result| be |value|.
+            1. Else, let |result| be [$CreateArrayFromList$](« |key|, |value| »)
+            1. Perform [=?=] [$GeneratorYield$]([$CreateIterResultObject$](|result|, <emu-value>false</emu-value>)).
 
-        Note: The [=size=] of |map|, and the order of its entries,
-        might have changed while execution of this abstract operation
-        was paused by [$Yield$].
-    1. Return <emu-value>undefined</emu>.
-1. Return [$CreateIteratorFromClosure$](|closure|, "%MapIteratorPrototype%", {{%MapIteratorPrototype%}}).
+            Note: The [=map/size=] of |map|, and the order of its entries,
+            might have changed while execution of this abstract operation
+            was paused by [$Yield$].
+        1. Return <emu-value>undefined</emu>.
+    1. Return [$CreateIteratorFromClosure$](|closure|, "%MapIteratorPrototype%", {{%MapIteratorPrototype%}}).
 </div>
 
 <div algorithm>
     To <dfn>create a set iterator</dfn>
-    from a [=set=] |set|
+    from a [=/set=] |set|
     and a |kind| ("key+value" or "value"):
 
-    Issue: ...
+    1. Let |closure| be a new [=/Abstract Closure=] with no parameters
+        that captures |set| and |kind| and performs the following steps when called:
+        1. Let |index| be 0.
+        1. Let |numEntries| be the [=set/size=] of |set|.
+        1. Repeat, while |index| < |numEntries| and |index| < |set|'s [=set/size=]:
+            1. Let |entry| be the |index|th [=set/item=] of |set|,
+                [=converted to an ECMAScript value=].
+            1. Set |index| to |index| + 1.
+            1. If |kind| is "value", let |result| be |entry|.
+            1. Else, let |result| be [$CreateArrayFromList$](« |entry|, |entry| »)
+            1. Perform [=?=] [$GeneratorYield$]([$CreateIterResultObject$](|result|, <emu-value>false</emu-value>)).
+
+            Note: The [=set/size=] of |set|, and the order of its entries,
+            might have changed while execution of this abstract operation
+            was paused by [$Yield$].
+        1. Return <emu-value>undefined</emu>.
+    1. Return [$CreateIteratorFromClosure$](|closure|, "%SetIteratorPrototype%", {{%SetIteratorPrototype%}}).
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -741,7 +741,7 @@ across [=IDL fragments=].
     on the <code class="idl">TextField</code> [=interface=]
     need not have been escaped with an underscore (because "<code>value</code>" is
     not a keyword in the IDL grammar), it is still unescaped
-    to obtain the attribute’s [=identifier=].
+    to obtain the attribute's [=identifier=].
 
 </div>
 
@@ -839,7 +839,7 @@ a given interface |A| is the set of all interfaces that |A|
 inherits from, directly or indirectly.  If |A| does not [=interface/inherit=]
 from another interface, then the set is empty.  Otherwise, the set
 includes the interface |B| that |A| [=interface/inherits=]
-from and all of |B|’s [=inherited interfaces=].
+from and all of |B|'s [=inherited interfaces=].
 
 An interface must not be declared such that
 its inheritance hierarchy has a cycle.  That is, an interface
@@ -862,7 +862,7 @@ interface |B| that inherits from |A|, and so on.
 Note that general multiple inheritance of interfaces is not supported, and
 objects also cannot implement arbitrary sets of interfaces.
 Objects can be defined to implement a single given interface |A|,
-which means that it also implements all of |A|’s [=inherited interfaces=].
+which means that it also implements all of |A|'s [=inherited interfaces=].
 In addition, an [=includes statement=] can be used
 to define that objects implementing an [=interface=] |A|
 will always also include the [=interface mixin member|members=]
@@ -2051,14 +2051,14 @@ defined on the same [=interface=].
 
 The <dfn id="dfn-return-type" export>return type</dfn> of the operation is given
 by the type (matching <emu-nt><a href="#prod-Type">Type</a></emu-nt>)
-that appears before the operation’s optional [=identifier=].
+that appears before the operation's optional [=identifier=].
 If the return type is an
 [=identifier=] followed by <emu-t>?</emu-t>,
 then the identifier must
 identify an [=interface=], [=dictionary=], [=enumeration=],
 [=callback function=], [=callback interface=] or [=typedef=].
 
-An operation’s arguments (matching <emu-nt><a href="#prod-ArgumentList">ArgumentList</a></emu-nt>)
+An operation's arguments (matching <emu-nt><a href="#prod-ArgumentList">ArgumentList</a></emu-nt>)
 are given between the parentheses in the declaration.  Each individual argument is specified
 as a type (matching <emu-nt><a href="#prod-Type">Type</a></emu-nt>) followed by an [=identifier=]
 (matching <emu-nt><a href="#prod-ArgumentName">ArgumentName</a></emu-nt>).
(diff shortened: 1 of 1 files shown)


tabatkins:
Ugh, I didn't push all my latest commits from my work machine. This is *even more WIP than intended*; you'll have to wait for tomorrow to *actually* see my current progress.

tabatkins:
Okay, current progress is up and now matches the task list as I have represented it. Review would be nice, but I'll continue working on this without waiting.

index.bs
```diff
@@ -12945,40 +12978,28 @@ the result of concatenating the [=identifier=] of the [=interface=] and the stri
...
-the object’s [=map entries=].
-
 If an [=interface=] |A| is declared with
 a [=maplike declaration=], then
 there exists a number of additional properties on |A|’s
 [=interface prototype object=].
 These additional properties are described in the sub-sections below.
 
 <div algorithm>
+    To <dfn local-lt="coherency-check">coherency-check an object</dfn> |this|
```
domenic, line 12988: IMO using |this| as a variable name is very confusing.

domenic, line 12988: The algorithm name doesn't give me a lot of insight into what this is doing. Also, it says it's checking an object but the first step is ToObject. Maybe "security-and-implements check"?

index.bs
```diff
@@ -13019,66 +13034,121 @@ with attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-
...
-            *   an identifier equal to |name|, and
-            *   the type "<code>method</code>".
-        1.  If |O| does not [=implement=] <var ignore>A</var>, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-        1.  Let |map| be the {{ECMAScript/Map}} object that is the value of |O|’s \[[BackingMap]] [=internal slot=].
+The value of the [=function object=]’s <code class="idl">length</code> properties is the Number value <emu-val>1</emu-val>.
+
+The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>get</code>".
+
+
+<h5 id="es-map-has">get</h5>
```
domenic, line 13105:
```suggestion
<h5 id="es-map-has">has</h5>
```

domenic (commented):
Generally looks like it's on the right track; very nice work.

I only left a few substantive comments. I think there are some editorial things but we can do those in a final pass. One more-pervasive editorial thing you might want to get done earlier is to use "`string`" instead of "string" for strings.

tabatkins:
Okay, finished the work I wanted to do, and addressed @domenic's editorial concerns they've raised so far. This PR is ready for proper review now.

I'm putting off defining "map setting steps"/etc; the text already defines that specs that want to provide their own set/delete/etc should just provide normal IDL methods of those names, so that should suffice.

I'd like to, probably in a follow-on issue, define all these built-ins in terms of the existing ["create an operation function"](https://webidl.spec.whatwg.org/#dfn-create-operation-function) algo, as that does all the nice setup that I'm currently doing by hand in each of the definitions, but that probably requires rejiggering the whole thing into imperative style. (I was gonna write an abstraction algo that folded away most of the boilerplate, but it turns out that "create an operation function" *is* that abstraction algo.) The current style is how the spec currently defines these operations, so keeping it as-is will at least not make things worse for now.

index.bs
```diff
@@ -4511,15 +4514,14 @@ An [=interface=] can be declared to be
...
+known as its <dfn id="dfn-set-entries" export>set entries</dfn>.
 The type of the values is given in the angle
 brackets of the setlike declaration.  Values are required to be unique.
 
-The [=set entries=] of an object
-implementing a [=setlike=] interface is empty at
-the of the object’s creation.  Prose accompanying the interface can
-describe how the [=set entries=]
-of an object change.
+Specification authors can modify the contents of the [=map entries=],
```
domenic, line 4524:
```suggestion
Specification authors can modify the contents of the [=set entries=],
```

index.bs
```diff
@@ -11422,6 +11418,32 @@ and its value is an object called a <dfn id="dfn-legacy-factory-function" export
 which allows creation of objects that implement the interface.
 The characteristics of a legacy factory function are described in [[#legacy-factory-functions]].
 
+<div algorithm>
+    Some interface operations are defined directly over ECMAScript values,
```
domenic, line 11424: I don't think "interface operations" is right here, because both "interface" and "operation" have specific meanings in Web IDL which do not match the maplike/setlike functions.

I think an appropriate introduction would be something like

> Some ECMAScript methods defined in this section will perform an implementation check in their opening steps, to ensure they're being called on the correct kind of object and that the method is valid to call from the current context.

index.bs
```diff
@@ -4415,15 +4421,14 @@ An [=interface=] can be declared to be
 </pre>
 
 Objects implementing an interface that is declared to be maplike
-represent an ordered list of key–value pairs known as its <dfn id="dfn-map-entries" export>map entries</dfn>.
+represent an [=ordered map=] of key–value pairs, initially empty,
+known as its <dfn id="dfn-map-entries" export>map entries</dfn>.
```
domenic, line 4425:
```suggestion
known as that object's <dfn id="dfn-map-entries" export>map entries</dfn>.
```

here and below. (To make it clear _each platform object implementing a maplike interface_ owns a map entries, not the maplike interface itself.)

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
 |A|’s [=interface prototype object=]
 with attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
 and whose value is the [=function object=] that is the value of
-the {{@@iterator}} property.
+the <code>entries</code> property.
+
+<div algorithm>
+    To <dfn>create a map iterator</dfn>
+    from a [=/map=] |map|
+    and a |kind| ("key+value", "key", or "value"):
```
domenic, line 12899: String markup. Also I find the parens a bit confusing so I'd say ``|kind| which is either "`key+value`", "`key`", or "`value`"``

String markup issue recurs 3 more times in this algorithm.

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+
+    1. Let |closure| be a new [=/Abstract Closure=] with no parameters
+        that captures |map| and |kind| and performs the following steps when called:
+        1. [=map/For each=] |key|→|value| of |map|'s [=map entries=]:
```
domenic, line 12903:
```suggestion
        1. [=map/For each=] |key| → |value| of |map|'s [=map entries=]:
```

domenic, line 12903: |map| is a map, which does not have map entries. I think you just want to iterate over |map|?

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+            1. Set |key| and |value| to each [=converted to an ECMAScript value=].
+            1. If |kind| is "key", let |result| be |key|.
+            1. Else if |kind| is "value", let |result| be |value|.
+            1. Else, let |result| be [$CreateArrayFromList$](« |key|, |value| »)
```
domenic, line 12907:
```suggestion
            1. Else, let |result| be [$CreateArrayFromList$](« |key|, |value| »).
```

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
 
 *   The property has attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
-*   The value of the property is a [=built-in function object=] that [=forwards to the internal map object|forwards that name to the internal map object=].
+*   The value of the property is a [=built-in function object=]
+    whose behavior when invoked is as follows:
+
+    <div algorithm="to invoke the entries method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>entries</code>" and type "<code>method</code>".
+        1. Let |map| be |O|'s [=map entries=].
+        1. Return the result of [=creating a map iterator=] from |map| with kind "key+value".
```
domenic, line 12931: String markup

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+*   The value of the property is a [=built-in function object=]
+    whose behavior when invoked is as follows:
 
-*   The property has attributes
-    { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
-*   <div algorithm="to invoke the get and has methods of Maps">
+    <div algorithm="to invoke the keys method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>keys</code>" and type "<code>method</code>".
+        1. Let |map| be |O|'s [=map entries=].
+        1. Return the result of [=creating a map iterator=] from |map| with kind "key".
```
domenic, line 12951: String markup

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+with the following characteristics:
+
+*   The property has attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }.
+*   The value of the property is a [=built-in function object=]
+    whose behavior when invoked is as follows:
+
+    <div algorithm="to invoke the values method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>values</code>" and type "<code>method</code>".
+        1. Let |map| be |O|'s [=map entries=].
+        1. Return the result of [=creating a map iterator=] from |map| with kind "value".
```
domenic, line 12972: String markup

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+
+    <div algorithm="to invoke the forEach method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>forEach</code>" and type "<code>method</code>".
+        1. Let |map| be |O|'s [=map entries=].
+        1.  Let |callbackFn| be the first argument passed to the function, or <emu-val>undefined</emu-val> if not supplied.
+        1.  If [$IsCallable$](|callbackFn|) is <emu-val>false</emu-val>, [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+        1.  Let |thisArg| be the second argument passed to the function, or <emu-val>undefined</emu-val> if not supplied.
+        1. [=map/For each=] |key|→|value| of |map|:
+            1. Let |esKey| and |esValue| be |key| and |value| [=converted to an ECMAScript value=].
+            1. Perform [=?=] [$Call$](|callbackFn|, |thisArg|, « |esValue|, |esKey|, |O| »)
```
domenic, line 12998:
```suggestion
            1. Perform [=?=] [$Call$](|callbackFn|, |thisArg|, « |esValue|, |esKey|, |O| »).
```

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+        1.  Let |map| be |O|'s [=map entries=].
         1.  Let |keyType| be the key type specified in the [=maplike declaration=].
-        1.  Let |function| be [=!=] <a abstract-op>Get</a>(|map|, "<code>delete</code>").
         1.  Let |keyArg| be the first argument passed to this function, or <emu-val>undefined</emu-val> if not supplied.
-        1.  Let |keyIDL| be the result of [=converted to an IDL value|converting=] |keyArg| to an IDL value of type |keyType|.
-        1.  Let |key| be the result of [=converted to ECMAScript values|converting=] |keyIDL| to an ECMAScript value.
-        1.  Return [=?=] <a abstract-op>Call</a>(|function|, |map|, «|key|»).
+        1.  Let |key| be |keyArg| [=converted to an IDL value=] of type |keyType|.
+        1.  If |key| is -0, set |key| to +0.
+        1.  If |map|[|key|] [=map/exists=], then return <emu-val>true</emu-val>,
```
domenic, line 13050:
```suggestion
        1.  If |map|[|key|] [=map/exists=], then return <emu-val>true</emu-val>;
```

index.bs
```diff
@@ -12995,207 +12873,279 @@ with the following characteristics:
...
+*   The value of the property is a [=built-in function object=]
+    whose behavior when invoked is as follows:
+
+    <div algorithm="to invoke the entries method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>entries</code>" and type "<code>method</code>".
+        1. Let |map| be |O|'s [=map entries=].
+        1. Return the result of [=creating a map iterator=] from |map| with kind "key+value".
+    </div>
 
 The value of the [=function objects=]’ <code class="idl">length</code> properties is the Number value <emu-val>0</emu-val>.
```
domenic, line 12934: Here and elsewhere, it should be

```suggestion
The value of the [=function object=]'s <code class="idl">length</code> property is the Number value <emu-val>0</emu-val>.
```

This unfortunately got copied into a bunch of other places instead of the other places' singular version getting copied here :)

index.bs
```diff
@@ -12995,207 +12875,291 @@ with the following characteristics:
...
         1.  Let |valueArg| be the second argument passed to this function, or <emu-val>undefined</emu-val> if not supplied.
-        1.  Let |keyIDL| be the result of [=converted to an IDL value|converting=] |keyArg| to an IDL value of type |keyType|.
-        1.  Let |valueIDL| be the result of [=converted to an IDL value|converting=] |valueArg| to an IDL value of type |valueType|.
-        1.  Let |key| be the result of [=converted to ECMAScript values|converting=] |keyIDL| to an ECMAScript value.
-        1.  Let |value| be the result of [=converted to ECMAScript values|converting=] |valueIDL| to an ECMAScript value.
-        1.  Perform [=?=] <a abstract-op>Call</a>(|function|, |map|, «|key|, |value|»).
-        1.  Return |O|.
+        1.  Let |value| be |valueArg| [=converted to an IDL value=] of type |valueType|.
+        1.  [=map/Set=] |map|[|key|] to |value|.
+        1.  Return |value| [=converted to an ECMAScript value=].
```
domenic, line 13081: This looks like an unintentional normative change. You need to return O.

index.bs
```diff
@@ -11422,6 +11420,32 @@ and its value is an object called a <dfn id="dfn-legacy-factory-function" export
...
+    to ensure they're being called on the correct kind of object
+    and that the operation is valid to call from the given script.
+
+    To <dfn local-lt="implementation check|implementation-check">implementation-check an object</dfn> |esValue|
+    against the interface |interface|,
+    with the identifier |name|
+    and the type |type|:
+
+    1. Let |object| to [=?=] [$ToObject$](|esValue|).
+    1. If |object| is a [=platform object=],
```
domenic, line 11437: So, |object| is an ECMAScript object right now, so it will never be a platform object. You need to instead convert it to an ECMAScript object before doing "is a platform object" and "implements" checks. That can replace the ToObject() check.

This helps most of your call sites because then you have the platform object, which has a map entries. (Right now your call sites are looking at the map entries of the ECMAScript object, which does not have any map entries.)

However it hurts call sites where you need to pass the value back to ECMAScript code, such as the return value of `set()`. For those you'd either need to convert back or preserve the original ECMAScript value.

index.bs
```diff
@@ -12995,207 +12875,291 @@ with the following characteristics:
...
+        1.  Return |value| [=converted to an ECMAScript value=].
     </div>
 
 The value of the [=function object=]’s <code class="idl">length</code> property is the Number value <emu-val>2</emu-val>.
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>set</code>".
 
+Note: If the interface does declare a <code class=idl>set</code> method,
+it should similarly map a -0 key to +0,
+and must return the value that was set.
```
domenic, line 13090: Nah it should return [=this=].

index.bs
```diff
@@ -72,9 +72,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
         text: %Function.prototype%; url: sec-properties-of-the-function-prototype-object
         text: %IteratorPrototype%; url: sec-%25iteratorprototype%25-object
         text: %Map.prototype%; url: sec-properties-of-the-map-prototype-object
+        text: %MapIteratorPrototype%; url: sec-%mapiteratorprototype%-object
```
domenic: Need %25 here and below to avoid conformance errors, it looks like.

index.bs
```diff
@@ -13210,134 +13174,247 @@ with the following characteristics:
...
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>has</code>".
 
 
-<h5 id="es-add-delete">add and delete</h5>
+<h5 id="es-set-add" oldids="es-add-delete">add</h5>
 
-For both of <code class="idl">add</code> and <code class="idl">delete</code>, if:
+If |A| does not declare a [=member=] with identifier "<code>add</code>",
+and |A| was declared with a read–write maplike declaration,
```
domenic, line 13326:
```suggestion
and |A| was declared with a read–write setlike declaration,
```

index.bs
```diff
@@ -13210,134 +13174,247 @@ with the following characteristics:
...
-        1.  Let |idlValue| be the result of [=converted to an IDL value|converting=] |arg| to an IDL value of type |type|.
-        1.  Let |value| be the result of [=converted to ECMAScript values|converting=] |idlValue| to an ECMAScript value.
-        1.  Let |result| be [=?=] <a abstract-op>Call</a>(|function|, |set|, «|value|»).
-        1.  If |name| is "delete", then return |result|.
-        1.  Otherwise, return |O|.
+
+<h5 id="es-set-delete">delete</h5>
+
+If |A| does not declare a [=member=] with identifier "<code>delete</code>",
+and |A| was declared with a read–write maplike declaration,
```
domenic, line 13358:
```suggestion
and |A| was declared with a read–write setlike declaration,
```

index.bs
```diff
@@ -13210,134 +13174,247 @@ with the following characteristics:
...
 <h5 id="es-set-clear">clear</h5>
 
-If |A| does not declare a [=member=]
-with a matching identifier, and
-|A| was declared with a read–write setlike declaration,
-then a <code class="idl">clear</code> data property with the following characteristics
-must exist on |A|’s
-[=interface prototype object=]:
+If |A| does not declare a [=member=] with identifier "<code>clear</code>",
+and |A| was declared with a read–write maplike declaration,
```
domenic, line 13391:
```suggestion
and |A| was declared with a read–write setlike declaration,
```

domenic (commented):
I didn't do a full review of setlike but there might be some stuff to copy over.

index.bs
tabatkins: Ah, right, that's a bad mixing of levels. Your suggested intro sounds great.

index.bs
tabatkins: Ah right, forgot about those strings.

The paren usage is copied from the ES algos, but happy to reword.

index.bs
tabatkins: Whoops, indeed, not sure why I thought it returned the value. I know I wrote that on purpose, huh.

index.bs
tabatkins: > So, |object| is an ECMAScript object right now, so it will never be a platform object. You need to instead convert it to an ECMAScript object

One of these "ECMAScript object"s are meant to be "IDL object", right? Otherwise I don't know what you're saying.

In either case, tho, I believe my text is right in this regard - the value is definitely an ES object at this point, since I'm defining the behavior of an ES function object. And "is a platform object" *is* invoked on ES objects (as is the "implements" algo).

> (Right now your call sites are looking at the map entries of the ECMAScript object, which does not have any map entries.)

Yeah, this is a weakness. I hadn't found how to properly invoke the notion of the IDL value it's paired to.

...ah, found it, "Set idlObject to the IDL [interface type](https://github.com/whatwg/webidl/pull/1138#idl-interface) value that represents a reference to esValue." (from step 2.1.2.4 of ["create an operation function"](file:///usr/local/google/home/tabatkins/specs/webidl/index.html#dfn-create-operation-function))

domenic: > One of these "ECMAScript object"s are meant to be "IDL object", right? Otherwise I don't know what you're saying.

Yes, the latter.

So you're right that "implements" is used on interfaces. "is a platform object" is more ambiguous, but I guess you're right there too, looking at usages.

> I hadn't found how to properly invoke the notion of the IDL value it's paired to.

Usually we just convert the ES value to an IDL object. The verbiage in that algorithm is similar: https://whatpr.org/webidl/1138.html#es-interface

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
-        1.  Let |O| be the <emu-val>this</emu-val> value.
-        1.  If |O| [=is a platform object=],
-            then [=perform a security check=], passing:
-            *   the platform object |O|,
-            *   the identifier "<code>size</code>", and
-            *   the type "<code>getter</code>".
-        1.  If |O| does not [=implement=] <var ignore>A</var>, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
-        1.  Let |map| be the {{ECMAScript/Map}} object that is the value of |O|’s \[[BackingMap]] [=internal slot=].
-        1.  Return [=!=] <a abstract-op>Get</a>(|map|, "<code>size</code>").
+        1.  Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>size</code>" and type "<code>getter</code>".
```
domenic, line 12875:
```suggestion
        1.  Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>A</var> with identifier "<code>size</code>" and type "<code>getter</code>".
```

here and below

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
+        1.  Let |map| be the [=map entries=] of the IDL value
+            that represents a reference to |O|.
+        1.  Return |map|'s [=map/size=], [=converted to an ECMAScript value=].
     </div>
```
domenic, line 12879: There appears to be a stray `<div>` here. If we put a `<div algorithm>` around the entire section then we could remove the `<var ignore>`.... or we could just delete this.

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
 |A|’s [=interface prototype object=]
 with attributes { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>false</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
 and whose value is the [=function object=] that is the value of
-the {{@@iterator}} property.
+the <code>entries</code> property.
 
+<div algorithm>
+    To <dfn>create a map iterator</dfn>
+    from a [=/map=] |map|
+    and a |kind| which is either "<code>key+value</code>", "<code>key</code>", or "<code>value</code>"):
```
domenic, line 12897:
```suggestion
    and a |kind| which is either "<code>key+value</code>", "<code>key</code>", or "<code>value</code>":
```

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
+
+    1. Let |closure| be a new [=/Abstract Closure=] with no parameters
+        that captures |map| and |kind| and performs the following steps when called:
+        1. [=map/For each=] |key| → |value| of |map|:
+            1. Set |key| and |value| to each [=converted to an ECMAScript value=].
+            1. If |kind| is "<code>key</code>", let |result| be |key|.
+            1. Else if |kind| is "<code>value</code>", let |result| be |value|.
+            1. Else, let |result| be [$CreateArrayFromList$](« |key|, |value| »).
+            1. Perform [=?=] [$GeneratorYield$]([$CreateIterResultObject$](|result|, <emu-value>false</emu-value>)).
```
domenic, line 12906:
```suggestion
            1. Perform [=?=] [$GeneratorYield$]([$CreateIterResultObject$](|result|, <emu-val>false</emu-val>)).
```

I found three other instances, so a general Ctrl+F for "emu-value" would be worthwhile.

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
+
+            Note: The [=map/size=] of |map|, and the order of its entries,
+            might have changed while execution of this abstract operation
+            was paused by [$Yield$].
+        1. Return <emu-value>undefined</emu>.
+    1. Return [$CreateIteratorFromClosure$](|closure|, "%MapIteratorPrototype%", {{%MapIteratorPrototype%}}).
```
domenic, line 12912:
```suggestion
    1. Return [$CreateIteratorFromClosure$](|closure|, "<code>%MapIteratorPrototype%</code>", {{%MapIteratorPrototype%}}).
```

here and for the set counterpart

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
+
+    <div algorithm="to invoke the entries method of Maps">
+        1. Let |O| be the <emu-val>this</emu-val> value, [=implementation-checked=] against <var ignore>a</var> with identifier "<code>entries</code>" and type "<code>method</code>".
+        1.  Let |map| be the [=map entries=] of the IDL value
+            that represents a reference to |O|.
+        1. Return the result of [=creating a map iterator=] from |map| with kind "<code>key+value</code>".
+    </div>
 
-The value of the [=function objects=]’ <code class="idl">length</code> properties is the Number value <emu-val>0</emu-val>.
+The value of the [=function object=]'s <code class="idl">length</code> properties is the Number value <emu-val>0</emu-val>.
```
domenic, line 12933: properties -> property, here and below

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
 
-The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>keys</code>" or "<code>values</code>", correspondingly.
+The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>entries</code>".
```
domenic, line 12935: It isn't really your mess to clean up but now we have inconsistent smart vs. not-smart quotes on these clauses... if you wanted to make them consistent (probably not-smart?) then that would make things slightly nicer.

index.bs
```diff
@@ -12995,207 +12872,301 @@ with the following characteristics:
...
+        1.  Let |value| be |valueArg| [=converted to an IDL value=] of type |valueType|.
+        1.  [=map/Set=] |map|[|key|] to |value|.
         1.  Return |O|.
     </div>
 
 The value of the [=function object=]’s <code class="idl">length</code> property is the Number value <emu-val>2</emu-val>.
 
 The value of the [=function object=]’s <code class="idl">name</code> property is the String value "<code>set</code>".
 
+Note: If the interface does declare a <code class=idl>set</code> method,
```
domenic, line 13093:
```suggestion
If the interface does declare a <code class=idl>set</code> method,
```

(don't use normative language "should"/"must" in non-normative notes)

Here and below

domenic (commented):
This round of editorial fixes seems likely final-ish...

index.bs
tabatkins: Argh, my earlier regex had the "preserve case" option on by accident and I caught that it gave me `<Var` but not that it had dropped the case of the `A`. :/

index.bs
tabatkins: Unsure what you mean here - this is the same pattern used for the algos on all the methods.

index.bs
tabatkins: wtf, i also had two `</emu>`

index.bs
domenic: This was bogus; I could not find the opening `<div>` for this `</div>` but it's because GitHub diffs are broken for large files.

index.bs
tabatkins: Eh just took a few minutes to review all the existing smart quotes and replace them with apostrophes. They were indeed wildly inconsistent.

tabatkins:
Oof, sorry about the lingering copy-paste errors I still had.

domenic (approved):
LGTM. @EdgarChen do you want to take a look? I feel like I understand the map/set/infra parts of the Web IDL spec pretty well but as editor you should definitely get a chance to weigh in if you want :)

EdgarChen:
I took a look and it looks good to me. Thanks for all your works!

Liamolucko:
This PR changed the setlike `add` method to return the value that was added to the setlike object instead of the object itself. Was that an accident?

TimothyGu:
@tabatkins Can you check the comment above?

annevk:
When PRs or issues are closed, please always file a follow-up. I've taken care of it this time around: #1268.

PR #1140, opened 2022-05-04
Exclude undefined from the PrimitiveType in grammar

### LINKED ISSUE #1139: Grammar and prose are inconsistent about whether undefined is a primitive type
The grammar has (see https://webidl.spec.whatwg.org/#index-prod-PrimitiveType):
```
PrimitiveType ::
    UnsignedIntegerType
    UnrestrictedFloatType
    undefined
    boolean
    byte
    octet
    bigint
```
which seems to imply that `undefined` is a primitive type.

Primitive types are defined as (see https://webidl.spec.whatwg.org/#dfn-primitive-type): `bigint, boolean and the numeric types`, which doesn't include `undefined`.

I think it makes sense to exclude `undefined` from the primitive types, and the simplest solution would be to move it in the grammar from `PrimitiveType` to `DistinguishableType`.

EdgarChen:
Closes #1139.

***
<a href="https://whatpr.org/webidl/1140.html" title="Last updated on May 5, 2022, 10:45 AM UTC (ef8513f)">Preview</a> | <a href="https://whatpr.org/webidl/1140/271cff8...ef8513f.html" title="Last updated on May 5, 2022, 10:45 AM UTC (ef8513f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2247,6 +2247,10 @@ the three special floating point literal values (<emu-t>Infinity</emu-t>,
 [=optional argument/default value=],
 it is interpreted in the same way as for a [=constant=].
 
+When the <emu-t>undefined</emu-t> token is used as the
+[=optional argument/default value=],
+the value is the IDL {{undefined}} value.
+
 <div id="string-literal" algorithm="value of string literal tokens">
     Optional argument default values can also be specified using a
     <emu-t class="regex"><a href="#prod-string">string</a></emu-t>
@@ -2365,6 +2369,7 @@ Note: When defining [=method steps=], you implicitly have access to [=this=].
         "[" "]"
         "{" "}"
         "null"
+        "undefined"
 </pre>
 
 <pre class="grammar" id="prod-Operation">
--- a/index.bs
+++ b/index.bs
@@ -5718,6 +5718,7 @@ are known as <dfn id="dfn-object-type" export>object types</dfn>.
         "FrozenArray" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         "ObservableArray" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         RecordType Null
+        "undefined" Null
 </pre>
 
 <div data-fill-with="grammar-ConstType"></div>
@@ -5726,7 +5727,6 @@ are known as <dfn id="dfn-object-type" export>object types</dfn>.
     PrimitiveType :
         UnsignedIntegerType
         UnrestrictedFloatType
-        "undefined"
         "boolean"
         "byte"
         "octet"


domenic:
This also removes undefined from ConstType, which I think is incorrect.

EdgarChen:
but

> The type of a constant (matching [ConstType](https://webidl.spec.whatwg.org/#prod-ConstType)) must not be any type other than a [primitive type](https://webidl.spec.whatwg.org/#dfn-primitive-type). If an [identifier](https://webidl.spec.whatwg.org/#dfn-identifier) is used, it must reference a [typedef](https://webidl.spec.whatwg.org/#dfn-typedef) whose type is a primitive type.

and undefined is not a primitive type in

> The primitive types are [bigint](https://webidl.spec.whatwg.org/#idl-bigint), [boolean](https://webidl.spec.whatwg.org/#idl-boolean) and the [numeric types](https://webidl.spec.whatwg.org/#dfn-numeric-type).

So could a constant type be `undefined`?

domenic:
It seems the intention was to allow that, per

> [undefined](https://webidl.spec.whatwg.org/#idl-undefined) constant values in IDL are represented with the undefined token.

Probably we should just make undefined a primitive type? It's certainly a primitive type in JavaScript, and I'm not sure why we'd depart.

petervanderbeken:
> It seems the intention was to allow that, per
> 
> > [undefined](https://webidl.spec.whatwg.org/#idl-undefined) constant values in IDL are represented with the undefined token.

And yet https://webidl.spec.whatwg.org/#prod-ConstValue has:

```
ConstValue ::
     BooleanLiteral
     FloatLiteral
     integer
```

petervanderbeken:
Also, are there actual use cases for constants with undefined as the value? 

> It's certainly a primitive type in JavaScript, and I'm not sure why we'd depart.

So are String and Symbol, but unless there's a clear use case I don't think we would add those either?

domenic:
Ah yeah. Actually allowing `const undefined = undefined` seems pretty dumb.

What's not dumb is allowing `undefined` as a `DefaultValue`.

petervanderbeken:
Ah, yes, good point. So let's add it to DefaultValue (which actually has string too ;-)).

petervanderbeken:
Note that `DefaultValue` is also used for dictionary members, which can't be undefined, so we might need some prose excluding it. Although `undefined` is not explicitly in:

> When a boolean literal token (true or false), the null token, an [integer](https://webidl.spec.whatwg.org/#prod-integer) token, a [decimal](https://webidl.spec.whatwg.org/#prod-decimal) token, one of the three special floating point literal values (Infinity, -Infinity or NaN), a [string](https://webidl.spec.whatwg.org/#prod-string) token, the two token sequence [], or the two token sequence {} is used as the [default value](https://webidl.spec.whatwg.org/#dfn-dictionary-member-default-value), it is interpreted in the same way as for an [operation](https://webidl.spec.whatwg.org/#dfn-operation)’s [optional argument default value](https://webidl.spec.whatwg.org/#dfn-optional-argument-default-value).

so maybe that's good enough?

EdgarChen:
I add `undefined` into DefaultValue (thanks for the patch from @petervanderbeken).

domenic (approved):

PR #1156, opened 2022-06-15
Meta: list webidl2.js and widlparser in PR template

### LINKED ISSUE #1147: Meta: downstream issues
We'll file issues against Deno and Node.js, but presumably we should also file an issue against https://github.com/w3c/webidl2.js. And potentially against https://github.com/web-platform-tests/wpt as idlharness.js does not use the new webidl2.js automatically. Is https://github.com/plinss/widlparser by @plinss what @tabatkins's https://github.com/tabatkins/bikeshed uses? Do both of those require an issue or just the former?

Once we have a list of places to file issues against I can make the necessary changes to whatwg/spec-factory and whatwg/meta.

cc @saschanaz @dontcallmedom @foolip

annevk:
Also pull in other changes from spec-factory.

Together with https://github.com/whatwg/meta/pull/247 and https://github.com/whatwg/spec-factory/pull/37 this closes #1147.


annevk:
@domenic it seems that https://github.com/whatwg/spec-factory/commit/51ce9d2c2b59b11b6b558ef01ee8ca3cb74707e8 might have been problematic after all. Maybe play around with this PR before merging to see what values we want?

.github/workflows/build.yml
```diff
@@ -11,16 +11,16 @@ jobs:
...
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 2
     # Note: `python` will also be this version, which various scripts depend on.
-    - uses: actions/setup-python@v2
+    - uses: actions/setup-python@v3
       with:
-        python-version: 3.8
-    - uses: actions/setup-node@v2
+        python-version: 3.10
```
saschanaz, line 20: >Run actions/setup-python@v3
>Version 3.1 was not found in the local cache
>Error: Version 3.1 with arch x64 not found


Might want to try "3.10" instead of 3.10?

annevk:
Thanks @saschanaz, looks like that was it. I'll PR upstream.

domenic (approved):

f56id:
Sann448

PR #1179, opened 2022-08-13
Add DOMException cause

legendecas:
Add cause support to DOMException.

Fixes https://github.com/whatwg/webidl/issues/969

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …


index.bs
```diff
@@ -14559,10 +14559,16 @@ The {{DOMException}} type is an [=interface type=] defined by the following IDL
 fragment:
 
 <pre class="idl">
+
+dictionary DOMExceptionOptions {
+  any cause;
+  DOMString name;
```
annevk, line 14565:
```suggestion
  DOMString name = "Error";
```

index.bs
```diff
@@ -14559,10 +14559,16 @@ The {{DOMException}} type is an [=interface type=] defined by the following IDL
...
+};
+
 [Exposed=(Window,Worker),
  Serializable]
 interface DOMException { // but see below note about ECMAScript binding
-  constructor(optional DOMString message = "", optional DOMString name = "Error");
+  constructor(optional DOMString message = "", optional (DOMString or DOMExceptionOptions) nameOrOptions = "Error");
```
annevk, line 14571:
```suggestion
  constructor(optional DOMString message = "", optional (DOMString or DOMExceptionOptions) options = {});
```

index.bs
```diff
@@ -14602,11 +14608,19 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
...
+<dfn constructor for="DOMException" lt="DOMException(message, nameOrOptions)">
+ <code>new DOMException(|message|, |nameOrOptions|)</code>
+</dfn>
 constructor steps are:
 
-1. Set [=this=]'s [=DOMException/name=] to |name|.
-1. Set [=this=]'s [=DOMException/message=] to |message|.
+1.  Set [=this=]'s [=DOMException/message=] to |message|.
+1.  If |nameOrOptions| is a String, set [=this=]'s [=DOMException/name=] to
+    |nameOrOptions|.
```
annevk, line 14618:
```suggestion
1.  If |options| is a string, then set [=this=]'s [=DOMException/name=] to |options|.
```

index.bs
```diff
@@ -14602,11 +14608,19 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
...
+1.  Else,
```
annevk, line 14619:
```suggestion
1.  Otherwise:
```

index.bs
```diff
@@ -14602,11 +14608,19 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
...
+    1.  If |nameOrOptions|'name is present, set [=this=]'s [=DOMException/name=]
+        to |nameOrOptions|'s name, else set [=this=]'s [=DOMException/name=]
+        to "Error".
+    1.  Perform [=?=] <a abstract-op>InstallErrorCause</a>([=this=], |nameOrOptions|).
```
annevk, line 14623:
```suggestion
    1.  Set [=this=]'s [=DOMException/name=] to |options|["{{DOMExceptionOptions/name}}"].
    1.  Perform [=?=] <a abstract-op>InstallErrorCause</a>([=this=], |options|).
```

annevk (commented):
I have some editorial suggestions. This also aligns the IDL with the style we used for `addEventListener()`. Essentially downplaying the role of the legacy argument type which also simplifies the resulting algorithm a bit.

legendecas:
@annevk Thank you for the suggestions! Updated.

index.bs
```diff
@@ -14602,11 +14608,16 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
 <dfn for="DOMException">message</dfn>, both [=strings=].
 
 The
-<dfn constructor for="DOMException" lt="DOMException(message, name)"><code>new DOMException(|message|, |name|)</code></dfn>
+<dfn constructor for="DOMException" lt="DOMException(message, nameOrOptions)">
+ <code>new DOMException(|message|, |nameOrOptions|)</code>
+</dfn>
```
annevk, line 14613: Forgot about this one, sorry.
```suggestion
<dfn constructor for="DOMException" lt="DOMException(message, options)"><code>new DOMException(|message|, |options|)</code></dfn>
```

legendecas: Sorry about missing this. Fixed.

index.bs
```diff
@@ -14602,11 +14609,14 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
...
+<dfn constructor for="DOMException" lt="DOMException(message, options)"><code>new DOMException(|message|, |options|)</code></dfn>
 constructor steps are:
 
-1. Set [=this=]'s [=DOMException/name=] to |name|.
-1. Set [=this=]'s [=DOMException/message=] to |message|.
+1.  Set [=this=]'s [=DOMException/message=] to |message|.
+1.  If |options| is a string, then set [=this=]'s [=DOMException/name=] to |options|.
+1.  Otherwise,
+    1.  Set [=this=]'s [=DOMException/name=] to |options|["{{DOMExceptionOptions/name}}"].
+    1.  Perform [=?=] <a abstract-op>InstallErrorCause</a>([=this=], |options|).
```
domenic, line 14619: It's not good that InstallErrorCause() performs HasProperty() and Get(), which the Web IDL bindings layer has already done.

It's also not really valid spec text to pass |options|, which is a Web IDL dictionary at this point, to InstallErrorCause(), which expects a JavaScript object. (Although, we do often fudge these boundaries; I've certainly done it myself.)

I think we should instead duplicate the InstallErrorCause logic here:

2. If |options|["cause"] [=map/exists=], then perform [CreateNonEnumerableDataPropertyOrThrow](https://tc39.es/ecma262/#sec-createnonenumerabledatapropertyorthrow)(O, "cause", |options|["cause"]).

domenic (changes requested):

index.bs
legendecas: Thank you for the suggestion!

AFAICT, `O` is not defined in this context. Should we perform abstract-op CreateNonEnumerableDataPropertyOrThrow on `[=this=]` instead? I'm not sure if `[=this=]` here refers to the WebIDL type or the JavaScript `this` 😅

domenic: You can use `[=this=]` and assume it refers to an ES object. Technically it is defined to refer to an ES object, but in general we have let it be used both ways in some cases like this. (My personal rule is: if the IDL value type is `any` or `object`, then you can assume implicit conversions, since the conversions are trivial. That is why I am less comfortable with `|options|`, since it is a dictionary type and dictionary conversions are not trivial.)

index.bs
```diff
@@ -14602,11 +14609,14 @@ Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn
...
+<dfn constructor for="DOMException" lt="DOMException(message, options)"><code>new DOMException(|message|, |options|)</code></dfn>
 constructor steps are:
 
-1. Set [=this=]'s [=DOMException/name=] to |name|.
-1. Set [=this=]'s [=DOMException/message=] to |message|.
+1.  Set [=this=]'s [=DOMException/message=] to |message|.
+1.  If |options| is a string, then set [=this=]'s [=DOMException/name=] to |options|.
+1.  Otherwise,
+    1.  Set [=this=]'s [=DOMException/name=] to |options|["{{DOMExceptionOptions/name}}"].
+    1.  If |options|["cause"] [=map/exists=], then perform <a abstract-op>CreateNonEnumerableDataPropertyOrThrow</a>([=this=], "cause", |options|["cause"]).
```
domenic, line 14619:
```suggestion
    1.  If |options|["cause"] [=map/exists=], then perform <a abstract-op>CreateNonEnumerableDataPropertyOrThrow</a>([=this=], "`cause`", |options|["{{DOMExceptionOptions/cause}}"]).
```

domenic (commented):
This looks good with a suggested nit.

However I had another doubt about the general shape here. Which is, should we really be installing an own data property? That's what `Error` does, but we've already decided to depart from `Error` for `name` and `message`, making those getters instead.

I lean slightly toward consistency with `DOMException`'s existing properties over consistency with `Error`, given how in the past I tried to encourage consistency with `Error` and failed (see https://github.com/whatwg/webidl/pull/378).

But this brings us to my next point, which is, it seems like it's time to ask for implementer support for this change. So let's try pinging: @yuki3, @EdgarChen, @petervanderbeken, and @shvaikalesh and see what they say, both on the general proposal and on whether we should do a data property or a readonly attribute (getter) for `cause`.

petervanderbeken:
This looks fine to me, but I'd also like it to be consistent with the other DOMException properties (with a getter). I also think the serialization should include this value.

yuki3:
+1 to make it an accessor property to be consistent.

jasnell:
As far as implementor support is concerned, I will implement support in Cloudflare Workers also.

index.bs
```diff
@@ -14601,19 +14608,28 @@ requirements beyond the normal ones for [=interface types=].
 Each {{DOMException}} object has an associated <dfn for="DOMException">name</dfn> and
 <dfn for="DOMException">message</dfn>, both [=strings=].
 
+Each {{DOMException}} object has an associated <dfn for="DOMException">cause</dfn>, which
+is a JavaScript value. It is {{undefined}} unless specified otherwise.
```
domenic, line 14612:
```suggestion
is a JavaScript value. It is <emu-val>undefined</emu-val> unless specified otherwise.
```

index.bs
```diff
@@ -14601,19 +14608,28 @@ requirements beyond the normal ones for [=interface types=].
...
+<dfn constructor for="DOMException" lt="DOMException(message, options)"><code>new DOMException(|message|, |options|)</code></dfn>
 constructor steps are:
 
-1. Set [=this=]'s [=DOMException/name=] to |name|.
-1. Set [=this=]'s [=DOMException/message=] to |message|.
+1.  Set [=this=]'s [=DOMException/message=] to |message|.
+1.  If |options| is a string, then set [=this=]'s [=DOMException/name=] to |options|.
+1.  Otherwise,
+    1.  Set [=this=]'s [=DOMException/name=] to |options|["{{DOMExceptionOptions/name}}"].
+    1.  Set [=this=]'s [=DOMException/cause=] to |options|["{{DOMExceptionOptions/cause}}"].
```
domenic, line 14622: You have to handle the case where `|options|["{{DOMExceptionOptions/cause}}"]` does not `[=map/exist=]`.

A proposal to make this easier is https://github.com/whatwg/webidl/issues/983 but it is not available yet.

domenic (commented):
Looking good with small issues.

legendecas:
As people seem to be more comfortable with being consistent with IDL attributes on the `cause`, I've updated the text to define the `cause` as an IDL attribute on `DOMException`.

Also, I've updated the steps to include the `cause` in serialization/deserialization.

index.bs
legendecas: Thanks, fixed.

ljharb:
I'm confused - `'cause' in error` is very explicitly designed to determine if an error has a cause or not.

Making it an accessor on the prototype breaks this highly intentional design decision in ecma262. It's meant to be an own property that is *completely absent* by default.

bathos:
I was in favor of the accessor for consistency with `message` (and `stack` in SpiderMonkey) + the unusualness of Web IDL constructs defining own data properties, but @ljharb’s explanation changed my mind. More reliable API contracts for _“whatever value was just thrown at me”_ is worth a lot more. DOMException is already a lil freaky anyway.

Also tho I sincerely hope nobody ends up needing to spot the distinction between `"cause" in new Error("", { cause: undefined })`  and `"cause" in new Error`, if they did need to and couldn’t, that’d be worse. Given `cause` would typically be a value the cause-bearing-error thrower is just passing along from possibly unrelated code, this might be your only signal that something more spooky’s going down.

domenic:
Yeah, I think that's a decision we just won't support on the web platform. It seems fine to not support undefined causes as distinct from absent causes. I gave this feedback back when the proposal was in TC39 and it was rejected. But for IDL and the web platform generally we'll use the web-platform-standard idioms of accessors on the prototype and undefined by default.

ljharb:
Google reps were perfectly able to prevent advancing the language proposal - it seems pretty unfortunate to just callously choose to ignore the result of a standards process just because you don’t like the result.

domenic:
We're not ignoring the result of the standards process. The standards process you're discussing was about Error. We have our own standards process on the web platform for DOMException.

ljharb:
… which inherits from Error, and thus creates user confusion when it deviates from it.

jasnell:
We discussed this today on the WinterCG call. Most of the attendees did not have strong opinions one way or the other with regard to `cause` as an accessor or `cause` as a data property. However, those that did have strong opinions were in favor of matching the behavior specified by TC-39 for `Error` objects.

Specifically from the Cloudflare Workers point of view, because we are aligned with WebIDL, it is easier for us to implement `cause` as using an accessor, and in our typical use case we would treat a missing `cause` and a `cause` that is always `undefined` as being semantically equivalent, however there is a valid argument to make that a cause that is explicitly `undefined` vs. a `cause` that just was never provided are very different things and should be treated as such. Unless there's a strong technical argument to make in favor of the accessor option beyond "that's what we do for `name` and `message` so let's keep it consistent", I'd prefer that `cause` on `DOMException` was defined as an own data property.

andreubotella:
In the WinterCG meeting it was also mentioned that this might go against the [priority of constituencies](https://w3ctag.github.io/design-principles/#priority-of-constituencies), since the `cause` property in `Error` instances can be tested for using `"cause" in error`, and authors would expect that to also work for `DOMException`. This would override any implementability or spec concerns.

domenic:
Thanks for the feedback! Those sound like reasons server-side runtimes might want to deviate from what we do in browsers, but in browsers we'll be following web platform patterns and using a prototype accessor.

jasnell:
For workers, we'll implement cause as a data property, and I will be encouraging Node.js to do the same. If users ask why browsers work differently, I'll refer them to this thread.

Do browsers intend to diverge from the language spec for `cause` in any other way we should be concerned about?

ljharb:
The priority of constituencies should be why browsers do a data property as well - "webIDL" is irrelevant compared to user expectations, as I understand it, and it's not like the web has no other own data properties, even on Errors (`stack`).

domenic:
Browsers intend to follow the Web IDL spec, which will specify cause as an accessor property like the other 10,000+ web platform object properties.

ljharb:
`stack` is a web platform object data property, it's just not on a web platform object.

Web IDL isn't for web developers, it's for web spec authors. Why is following Web IDL more important than following the pattern that the core JS error objects follow?

domenic:
Following the patterns of the web platform is more important for web platform objects, than following the pattern of (Chrome and Safari's implementation of) `stack`.

ljharb:
Ok, so you expect web users to understand the difference between language objects and web platform objects? That’s great, i was under the impression that browsers’ position was that there wasn’t a distinction in the majority of user’s minds.

domenic:
No.

ljharb:
I’m confused then - it seems like either users understand the difference between web and language, and the consistency of web platform objects matters to them more than consistency with the language - or, it’s all just JavaScript, and consistency solely within the web platform is irrelevant compared to consistency with the union of the web and the language (and non-web environments, as well), since it’s all one big bucket.

If I’m making a false dichotomy I’d love to understand more about how.

saschanaz:
Just curious, do we have a clue that `"cause" in err` already a well established pattern in the field?

ljharb:
@saschanaz nope, i think `cause` is new enough that there are no established patterns - but the only other alternatives are `Object.hasOwn(err, 'cause')` - which wouldn't work with this accessor approach either - or the incomplete approach of `typeof err.cause !== 'undefined'`, which fails to distinguish between a causeless error and an undefined cause error.

saschanaz:
Thanks for the quick answer! That was my impression too.

>which fails to distinguish between a causeless error and an undefined cause error.

Sounds like it, but does it / should it practically matter? I think Web IDL generally doesn't try hard to distinguish between explicit `undefined` and nonexistence (e.g. dictionaries treat explicit `undefined` as nonexistent), and I believe nobody complained with it.

bathos:
> I think Web IDL generally doesn't try hard to distinguish between explicit undefined and nonexistence (e.g. dictionaries treat explicit undefined as nonexistent), and I believe nobody complained with it.

Whether property absence & property access evaluating to undefined are semantically equivalent or distinct is a characteristic of an API contract, not of properties/objects inherently, and I’d figure the decision is typically informed by the value space that needs to be expressed. In the Error `cause` contract (and a few other places in ECMA-262), property absence is used to express a +1 language value space: “all ES values or none”. It’s the least painful mechanism the language has for expressing this `empty` notion: a “third null” (cursed lol — not-a-value `NaV`?) couldn’t do the same because when reified as a (public) ES value, it’s “an ES value” and you’re back at square one. The next best thing is pretty gross: a second property (`causeIsEmpty: true`) for the extra information bit.

Given Web IDL dictionaries don’t need to express this meta value space, it makes sense that they make both cases equivalent (it seems not far off from "0" / 0 / -0 being semantically equivalent when the dictionary member type is `unsigned long`: smaller value space = more conversion & conflation desirable). The contract in this case though is that of a non-Web IDL API though from a user POV, the API of the (at least hypothetical) “super class” (and its other subclasses, notably?). I’m not sure the question of whether the semantic distinction _should have_ been made ought to determine what’s done here. Perhaps the empty/undefined distinction might never help anybody debug anything, but I’m pretty sure `cause`-always-present will create some. People will end up using this check regardless and writing tests for whether some catch logic behaves right for each native/platform Error subclass would be unusual.

While I don’t have a strong opinion on whether the distinction was worth making, I haven’t understood why it’d be so off-the-table to — in intuitive, not literal terms — “just pass it to `super()`”. Is it a pain for implementators?

---

Aside re: _“Web IDL Web IDL generally doesn't try hard to distinguish between explicit undefined and nonexistence,”_  I’ve often wished it would try even _less_ hard :)

Chromium console showing history.pushState(1) throwing due to arity — the undefined/absence distinction — while history.pushState(1, undefined) succeeds

annevk:
> I’ve often wished it would try even less hard :)

Somewhat off-topic, but please file an issue on that against whatwg/html? Seems worth fixing one way or another.

bathos:
Possibly notable:

```js
console.log("cause" in new WebAssembly.CompileError("xxx", {}));
console.log("cause" in new WebAssembly.CompileError("xxx", { cause: 1 }));
```

already prints `false` and `true` in both Chromium & FF.

annevk:
https://webassembly.github.io/spec/js-api/#error-objects

> Note: It is not currently possible to define this behavior using Web IDL.

(WebAssembly being partially defined in Web IDL, but not actually being implemented in terms of Web IDL apparently has a couple of other twists to it.)

jasnell:
@bathos
> “just pass it to super()”. Is it a pain for implementors?

Conceptually, no, unfortunately the reality is a bit different. For those of us who implement `DOMException` in C++ using v8's APIs -- which do not provide any mechanism at the C++ level to invoke the `Error` constructor easily -- we can't simply "pass it to super()"... but that's an implementation problem.

The key issue here for me is that `DOMException` taking the getter/setter route when all other standardized Error types use the own property route means that users will be forced to either special case `DOMException` in error handling logic or otherwise be forced to always handle things the way `DOMException` does, eroding the language level guarantee guarantees.

For example,

```
try {
  doSomething();
} catch (err) {
  // It's now completely ambiguous what 'cause' in err means.

  // I either have to...

  if (err instanceof DOMException) {
    // 'cause' in err is meaningless.
  } else {
    // 'cause' in err has semantic value
  }

  // or always assume 'cause' in error is meaningless.
}
```

This difference does not just apply to non-browser runtimes. Browser developers will face this same issue. As the other runtimes also implement this, the fact that DOMException does it differently means that developers in general will not be able to rely on it.

Now, that said it's absolutely true that JavaScript allows you to throw literally anything so it can be argued that the mechanism is already unreliable, but since `DOMException` extends from `Error`, it really ought to follow the language specified contract for `Error` or it otherwise should not claim to *be* an `Error`.

```js
const ex = new DOMException();
ex instanceof Error;  // true
```

jasnell:
Alternatively, if `DOMException` is not going to follow the same semantics with `cause` as the language spec and other `Error`s, then DOMException should use a different name for the property so that it does not intentionally introduce a conflict: 
 `domException.whatwgCause` for instance.

domenic:
The boat has already sailed with message and name.

jasnell:
`message` and `name` are not relevant here. And it's too late for `DOMException` to not claim to be an `Error`, so the option remaining here is that if `DOMException` is not going to expose `cause` in a way consistent with `Error`, it should use a different property name, anything other than `cause` so that it does not intentionally introduce an inconsistency. Users will still have to special case `DOMException`, unfortunately but at least it wouldn't cause a conflict.

jasnell:
And it's too late for `DOMException` to not claim to be an `Error`, so the option remaining here is that if `DOMException` is not going to expose `cause` in a way consistent with `Error`, it should use a different property name, anything other than `cause` so that it does not intentionally introduce an inconsistency. Users will still have to special case `DOMException`, unfortunately, but

annevk:
`DOMException` is not alone. There are other exceptions that build on it, e.g., `RTCError`. Which I suppose raises the question as to what should be done about their constructors.

And I think saying that making `cause` a getter introduces an inconsistency is too broad, as it would also introduce an internal inconsistency for it to not do so.

Also, if we manage to get rid of optional (#905) the natural thing here would be that `cause` is at least `undefined`. I'm not even sure what the current binding layer makes of this with dictionaries. It might well already be `undefined` when missing (or more likely missing when `undefined`).

bathos:
> I'm not even sure what the current binding layer makes of this with dictionaries. It might well already be undefined when missing (or more likely missing when undefined).

Even though ES → WebIDL treats undefined-for-any-reason as absence (i.e. default-or-absent-or-throw-if-required),  WebIDL → ES makes a distinction between “Web IDL absence” and “Web IDL undefined”. No property is added if the member is absent but a property is added if the member is present & its value is undefined.

littledan:
This PR looks good to me as is. I don’t see it as important for JS developers to be able to distinguish between undefined and missing cause—how frequently do you want to check whether there is a cause at all?

I think it would be good if this PR could land so that web error classes have the significant benefit of this cause feature. This utility is much greater than the edge case being discussed above.

For some context for people coming from TC39: it is extremely useful for web specs to stick to WebIDL conventions. Previous attempts to deviate from this pattern (eg streams) were found to be highly complex without developer benefit and were later reverted.

I just don’t see why JS developers should care about this case, so I don’t see why it makes sense to invoke priority of constituencies. This is neutral for JS developers, and helpful for implementers. (It is also helpful for spec writers, but they are lower down on the totem pole.)

Ms2ger (approved):
This looks sensible enough; perhaps https://webidl.spec.whatwg.org/#es-creating-throwing-exceptions should be amended as well.

Using an own property seems like it would make specification and implementation more complicated and bugs more likely for little gain.

What is holding this up at this point?

annevk:
[OP](https://github.com/whatwg/webidl/pull/1179#issue-1337844223) needs to be filled out and @domenic needs to re-review as he explicitly requested changes.

Let me hereby declare that you can consider WebKit as a supportive implementer. You'll need to get an okay from Chromium or Gecko as well.

ljharb:
The gain is spec alignment and matching the design philosophy of the feature, which reduces the likelihood of bugs. If web IDL is getting in the way of that, perhaps that should be addressed?

saschanaz:
> The gain is spec alignment and matching the design philosophy of the feature, which reduces the likelihood of bugs. If web IDL is getting in the way of that, perhaps that should be addressed?

And the loss is violating the principle that is applied to all other attributes (as mentioned in https://github.com/whatwg/webidl/pull/1179#issuecomment-1241298885), which is used for feature detection for example. I think there's no golden answer here, and I'd rather follow the web platform principle as this is a web platform interface.

petervanderbeken:
> Let me hereby declare that you can consider WebKit as a supportive implementer. You'll need to get an okay from Chromium or Gecko as well.

Gecko will support this (https://github.com/whatwg/webidl/pull/1179#issuecomment-1236772927).

index.bs
```diff
@@ -14679,6 +14694,7 @@ Their [=serialization steps=], given <var>value</var> and <var>serialized</var>,
 <ol>
     <li>Set <var>serialized</var>.\[[Name]] to <var>value</var>'s [=DOMException/name=].</li>
     <li>Set <var>serialized</var>.\[[Message]] to <var>value</var>'s [=DOMException/message=].</li>
+    <li>Set <var>serialized</var>.\[[Cause]] to the [=sub-serialization=] of the value of <var>value</var>'s [=DOMException/cause=].</li>
```
domenic, line 14697:
```suggestion
    <li>Set <var>serialized</var>.\[[Cause]] to the [=sub-serialization=] of <var>value</var>'s [=DOMException/cause=].</li>
```

index.bs
```diff
@@ -14689,6 +14705,7 @@ Their [=deserialization steps=], given <var>value</var> and <var>serialized</var
 <ol>
     <li>Set <var>value</var>'s [=DOMException/name=] to <var>serialized</var>.\[[Name]].</li>
     <li>Set <var>value</var>'s [=DOMException/message=] to <var>serialized</var>.\[[Message]].</li>
+    <li>Set <var>value</var>'s [=DOMException/cause=] to the [=sub-deserialization=] of the value of <var>serialized</var>.\[[Cause]].</li>
```
domenic, line 14708:
```suggestion
    <li>Set <var>value</var>'s [=DOMException/cause=] to the [=sub-deserialization=] of <var>serialized</var>.\[[Cause]].</li>
```

domenic (approved):
LGTM with nit

yuki3:
I'm happy to implement `cause` attribute in Chromium.

index.bs
legendecas: Thanks, fixed!

PR #1184, opened 2022-08-23
draft: Interaction with the Stage 2 Record & Tuple proposal for ECMAScript

rricard:
Hello, this draft PR is a prototype spec text that attempts to define a minimal interaction between the [Record & Tuple proposal for ECMAScript](https://github.com/tc39/proposal-record-tuple) and WebIDL. The proposal is currently at Stage 2 **so there is no intent to merge this PR at the moment** but the goal is to open a conversation for our approach here.

In this PR, we want to keep things minimal: we do not intend just yet to add Record & Tuple to WebIDL types, we are only defining a conversion algorithm when records are put in place of objects (option bags notably) and when tuples are similarly put in place of arrays...

If the need arises later to have an API defined with WebIDL that return a Record or a Tuple for instance, then more work would be needed...

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1184.html" title="Last updated on Aug 23, 2022, 1:17 PM UTC (4fb12ed)">Preview</a> | <a href="https://whatpr.org/webidl/1184/b3b44cc...4fb12ed.html" title="Last updated on Aug 23, 2022, 1:17 PM UTC (4fb12ed)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7747,8 +7747,8 @@ values are represented by ECMAScript Object values (including [=function objects
 <h4 id="es-dictionary">Dictionary types</h4>
 
 IDL [=dictionary type=] values are represented
-by ECMAScript Object values.  Properties on
-the object (or its prototype chain) correspond to [=dictionary members=].
+by ECMAScript Object or Record values. Properties on the
+object (or its prototype chain) or the record correspond to [=dictionary members=].
 
 <div id="es-to-dictionary" algorithm="convert an ECMAScript value to dictionary">
 
@@ -7756,7 +7756,7 @@ the object (or its prototype chain) correspond to [=dictionary members=].
     to an IDL [=dictionary type=] value by
     running the following algorithm (where |D| is the [=dictionary type=]):
 
-    1.  If <a abstract-op>Type</a>(|esDict|) is not Undefined, Null or Object, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+    1.  If <a abstract-op>Type</a>(|esDict|) is not Undefined, Null, Object or Record, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
     1.  Let |idlDict| be an empty [=ordered map=], representing a dictionary of type |D|.
     1.  Let |dictionaries| be a list consisting of |D| and all of |D|'s [=inherited dictionaries=],
         in order from least to most derived.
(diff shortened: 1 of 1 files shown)


rricard:
I should normally have been cleared for participation by Bloomberg that should have signed the agreement.

annevk:
Hey @rricard, Participation data looks correct. It'll be marked verified once this is ready to land. We usually do that jointly with the first contribution. I've also ensured Build runs.

rricard:
Excellent thank you for checking this! We'll clear that out then once it is ready to land which is not the case yet...

bathos:
I notice the [record\<K, V\> conversion](https://webidl.spec.whatwg.org/#es-record) isn’t changed here yet. That’s probably just because it’s a work in progress, but it would be funny if we couldn’t use em for that, so I figured it may be worth pointing out just in case it was overlooked.

```js
let headers = new Headers(#{ "content-type": "record/record; record=record" });
```

As with dictionary member types, the `V` type of `record<K,V>` might not be compatible with the primitive-members-only nature of ES Record/Tuple. In practice I’m pretty sure most usages are a natural match, though. From the first few pages of matches for `record<K,V>` usage in specs, most of them had `V` in the string/number/tuple-of-the-same zone:

```
(sequence<sequence<ByteString>> or record<ByteString, ByteString>)
record<DOMString, sequence<DOMString>>
record<DOMString, double>
record<USVString, double>
record<USVString, (USVString or undefined)>
record<USVString, (USVString or sequence<USVString>)>
record<DOMString, GPUSize64> // GPUSize64 = unsigned long long
```

bathos:
Another missing conversion is [`any`](https://webidl.spec.whatwg.org/#es-any). This one seems important because adding a new ES language type would cause it to have undefined behavior. The defined behavior could be to throw a TypeError if the goal is to be conservative pending a clearer picture of what’s needed.

Union conversion and the overload resolution algorithms are affected similarly I think.

rricard:
> That’s probably just because it’s a work in progress, but it would be funny if we couldn’t use em for that, so I figured it may be worth pointing out just in case it was overlooked.

> Another missing conversion is [`any`](https://webidl.spec.whatwg.org/#es-any). This one seems important because adding a new ES language type would cause it to have undefined behavior. 

Thank you for this, I definitely overlooked both, will try to add them soon!

nicolo-ribaudo:
The proposal has been abandoned.

PR #1200, opened 2022-09-27
Meta: fix the grammar post processor on Review Drafts

### LINKED ISSUE #1195: pp-webidl fails on Review Drafts
We're running into a subtle issue with Review Drafts. This first surfaced in #1194 but also affects the prior RD. Due to various settings in Bikeshed the resulting HTML ends up differing slightly between LS and RD which causes the JS to trip.

The simplest of these is probably
```js
document.querySelector("div[data-fill-with=\"grammar-index\"]").innerHTML = wrap(output);
```
failing due to that `div` element no longer being present. This is happening to several other similar `div` elements as well, but those end up failing silently. The result though is that various grammar productions do not end up in the final copy of the standard.

@tabatkins do you know what could cause this? I looked a bit at `slimBuildArtifact` in Bikeshed and various other settings, but nothing really jumped out to me.

For debugging, you can compare:

* https://webidl.spec.whatwg.org/review-drafts/2022-03/
* https://webidl.spec.whatwg.org/commit-snapshots/e5afaf75c697ef14df26a2fe4557fb789ac60b49/

You can find the script at `<script class="remove">` in the source of the first one as it essentially failed server-side when generating the draft.

domenic:
Closes #1195 by using a new attribute, grammar-fill-with="", which won't get removed by Bikeshed's "slim build artifact" setting.

This will not work until https://github.com/tobie/webidl-grammar-post-processor/pull/2 is merged and released as v1.0.0.

***
<a href="https://whatpr.org/webidl/1200.html" title="Last updated on Sep 28, 2022, 6:56 AM UTC (1ead4cb)">Preview</a> | <a href="https://whatpr.org/webidl/1200/fb27b8a...1ead4cb.html" title="Last updated on Sep 28, 2022, 6:56 AM UTC (1ead4cb)">Diff</a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ local: index.bs
 
 deploy: index.bs
 	curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh
-	POST_BUILD_STEP='node ./check-grammar.js "$$DIR/index.html" && npm run pp-webidl -- --input "$$DIR/index.html"' \
+	POST_BUILD_STEP='node ./check-grammar.js "$$DIR/index.html" && npm run webidl-grammar-post-processor -- --input "$$DIR/index.html"' \
 	bash ./deploy.sh
 
 review: index.bs
--- a/index.bs
+++ b/index.bs
@@ -1258,15 +1258,15 @@ No [=extended attributes=] defined in this specification are applicable to [=inc
 
 </div>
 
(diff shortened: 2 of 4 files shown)


annevk (approved):
Looks good. Curious to see if updating the old RD works. (I added a fix to make that more likely.)

PR #1205, opened 2022-09-29
Make PaintOptions example a better role model.

### LINKED ISSUE #1204: PaintOptions example should not have nullable optional fields
There's an example in §2.7 Dictionaries as follows:

```idl
dictionary PaintOptions {
  DOMString? fillPattern = "black";
  DOMString? strokePattern = null;
  Point position;
};
```

This example is giving bad design advice since both `fillPattern` and `strokePattern` are both nullable *and* optional. There's no need for them to be nullable.

* `fillPattern` has a default value, which means if omitted, it will be `"black"`, but it can also be explicitly set to `null`. It's not clear why you would design something that can be set to `null`, but with a non-null default value.
* `strokePattern` is a little more reasonable, having a default value of `null`, but this is still unusual. It's best practice to have it simply be optional with no default, so if omitted it will simply be missing from the dictionary, rather than having a `null` value.

Therefore I think it should be rewritten to avoid the nullability:

```idl
dictionary PaintOptions {
  DOMString fillPattern = "black";
  DOMString strokePattern;
  Point position;
};
```

(At a higher level, this feels like a strange design for a `drawRectangle` method, taking the "position" — presumably the top-left corner — in an optional `PaintOptions` dict, but the width and height as direct arguments. I might also suggest making both the `options` argument and the `position` member required, which helps demonstrate when `required` is a good choice. But it's not clear whether that's the intention behind the design.)

mgiuca:
Do not have members that are both optional and nullable.

Fixes #1204.

***
<a href="https://whatpr.org/webidl/1205.html" title="Last updated on Sep 29, 2022, 6:59 AM UTC (5dcc4d0)">Preview</a> | <a href="https://whatpr.org/webidl/1205/9e4ce33...5dcc4d0.html" title="Last updated on Sep 29, 2022, 6:59 AM UTC (5dcc4d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5036,8 +5036,8 @@ No [=extended attributes=] are applicable to dictionaries.
         };
 
         dictionary PaintOptions {
-          DOMString? fillPattern = "black";
-          DOMString? strokePattern = null;
+          DOMString fillPattern = "black";
+          DOMString strokePattern;
           Point position;
         };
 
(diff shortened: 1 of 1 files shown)


annevk (approved):
Thanks!

annevk:
Assuming you're still at Google, you'll need to make your membership of googlers public as per https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-your-membership-in-organizations/publicizing-or-hiding-organization-membership to help the IPR bot.

mgiuca:
Thanks @annevk . I'm not actually in the google org (though I still work for Google). I'm checking to see what the process is.

mgiuca:
Update: Joined the googlers org and the check passed now.

PR #1211, opened 2022-10-08
Allow DOMException subclasses to be used as exceptions

domenic:
See discussion in https://github.com/whatwg/webidl/issues/1168#issuecomment-1183851493.

This also includes some updates to all exception creation and throwing, such as reflecting how messages are actually passed along in practice, or using Web IDL infrastructure to create DOMException instances instead of Construct()ing them.

It seems some places were already doing this sort of thing, such as https://w3c.github.io/webtransport/#web-transport-error-interface and I think some media thing. For WebTransport, it seems like they aren't violating the letter of the current spec since they were never throwing such `WebTransportError`s or treating them like exceptions, instead just passing them through streams machinery and so on. But I think it's a good idea to codify this.

Note that WebTransportError's constructor violates the proposal here of requiring such errors to always take a `message` as their first constructor argument. Not sure exactly what to do about that...

   * Chrome has a spec being built that would directly use this.
   * I am hopeful other engines think this is a reasonable idea, e.g. based on the WebTransport precedent?
   * Nothing here is testable directly; we'd instead write tests for specific subclasses.
   * I don't think we need to preemptively do this; it would come along as part of the bugs for implementing such subclasses.

***
<a href="https://whatpr.org/webidl/1211.html" title="Last updated on Jun 5, 2023, 6:44 AM UTC (2867a8b)">Preview</a> | <a href="https://whatpr.org/webidl/1211/5dcedce...2867a8b.html" title="Last updated on Jun 5, 2023, 6:44 AM UTC (2867a8b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1048,7 +1048,7 @@ are [=included=] by their [=host interface=].
 
 Issue: [=interface mixin member|Member=] order isn't clearly specified,
 in particular when [=interface mixins=] are defined in separate documents.
-It is discussed in <a href="https://github.com/heycam/webidl/issues/432">issue #432</a>.
+It is discussed in <a href="https://github.com/whatwg/webidl/issues/432">issue #432</a>.
 
 No [=extended attributes=] defined in this specification are applicable to [=includes statements=].
 
@@ -1380,7 +1380,7 @@ Constants can appear on [=interfaces=] and [=callback interfaces=].
     named integer codes in the style of an enumeration.  The Web platform
     is moving away from this design pattern in favor of the use of strings.
     Editors who wish to use this feature are strongly advised to discuss this
-    by <a href="https://github.com/heycam/webidl/issues/new?title=Intent%20to%20use%20Constants">filing an issue</a>
+    by <a href="https://github.com/whatwg/webidl/issues/new?title=Intent%20to%20use%20Constants">filing an issue</a>
     before proceeding.
 </p>
 
@@ -2442,7 +2442,7 @@ See [[#interface-object]] for details on how a [=constructor operation=] is to b
     new object that [=implements=] the interface.
     It would take either zero or one argument.
 
-    Issue(heycam/webidl#698): It is unclear whether the <code class="idl">NodeList</code> interface
+    Issue(whatwg/webidl#698): It is unclear whether the <code class="idl">NodeList</code> interface
     object would implement a \[[Construct]] internal method. In any case, trying to use it as a
     constructor will cause a {{TypeError}} to be thrown.
 
@@ -3927,7 +3927,7 @@ Prose accompanying an [=interface=] with a [=pair iterator=] must define a [=/li
     or (b) with a forEach method that instead invokes its callback like
     Set.prototype.forEach (where the key is the same as the value).
     If you're creating an API that needs such a forEach method, please
-    <a href="https://github.com/heycam/webidl/issues/new?title=Enhancement%20request%20for%20Iterables">file an issue</a>.
+    <a href="https://github.com/whatwg/webidl/issues/new?title=Enhancement%20request%20for%20Iterables">file an issue</a>.
 
 </div>
 
@@ -4117,7 +4117,7 @@ async iterator's <code>return()</code> method is invoked. This most commonly occ
 
 <p class="note">We could add a similar hook for <code>throw()</code>. So far there has been no need,
 but if you are creating an API that needs such capabilities, please
-<a href="https://github.com/heycam/webidl/issues/new?title=Enhancement%20request%20for%20Async%20Iterables">file an issue</a>.
+<a href="https://github.com/whatwg/webidl/issues/new?title=Enhancement%20request%20for%20Async%20Iterables">file an issue</a>.
 
 The prose may also define <dfn export>asynchronous iterator initialization steps</dfn>. These
 receive the instance of the [=interface=] being iterated, the newly-created iterator object, and a
@@ -4898,21 +4898,14 @@ No [=extended attributes=] are applicable to dictionaries.
 
 <h3 id="idl-exceptions">Exceptions</h3>
 
-An <dfn id="dfn-exception" export>exception</dfn> is a type of object that
-represents an error and which can be thrown or treated as a first
-class value by implementations.  Web IDL does not allow exceptions
-to be defined, but instead has a number of pre-defined exceptions
-that specifications can reference and throw in their definition of
-operations, attributes, and so on.  Exceptions have an
-<dfn id="dfn-exception-error-name" for="exception" export>error name</dfn>,
-a {{DOMString}},
-which is the type of error the exception represents, and a
-<dfn id="dfn-exception-message" for="exception" export>message</dfn>, which is an optional,
-user agent-defined value that provides human readable details of the error.
+An <dfn id="dfn-exception" export>exception</dfn> is a type of object that represents an error and
+which can be thrown or treated as a first class value by implementations. Web IDL has a number of
+pre-defined exceptions that specifications can reference and throw in their definition of
+operations, attributes, and so on. Custom exception types can also be defined, as [=interfaces=]
+that [=interface/inherit=] from {{DOMException}}.
 
-There are two kinds of exceptions available to be thrown from specifications.
-The first is a <dfn id="dfn-simple-exception" export>simple exception</dfn>, which
-is identified by one of the following types:
+A <dfn id="dfn-simple-exception" export>simple exception</dfn> is identified by one of the
+following types:
 
 *   <dfn exception>EvalError</dfn>
 *   <dfn exception>RangeError</dfn>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5071,87 +5071,98 @@ No [=extended attributes=] are applicable to dictionaries.
...
+    would be:
+
+    <blockquote>
+        [=exception/Throw=] a "{{SyntaxError}}" {{DOMException}} indicating that the given value
+        had disallowed trailing spaces.
+    </blockquote>
+
+    <p>Such additional context is most helpful to implementers when it is not immediately obvious
+    why the exception is being thrown, e.g., because there are many different steps in the algorithm
+    which throw "{{SyntaxError}}" {{DOMException}}s. In contrast, if your specification throws a
```
annevk, line 5153:
```suggestion
    which throw a "{{SyntaxError}}" {{DOMException}}. In contrast, if your specification throws a
```

annevk (commented):
Let me introduce you to https://w3c.github.io/webrtc-pc/#dom-rtcerror which follows almost none of these guidelines and promises are rejected with instances of it already as far as I can tell.

index.bs
```diff
@@ -5071,87 +5071,98 @@ No [=extended attributes=] are applicable to dictionaries.
...
+    "{{NotAllowedError}}" {{DOMException}} immediately after checking if the user has provided
+    permission to use a given feature, it's fairly obvious what sort of [=exception/message=] the
+    implementation should construct, and so specifying it is not necessary.
+</div>
```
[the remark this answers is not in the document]
domenic: I think this is more about ensuring that the `message` value which implementers construct from the context given by the spec author, does not introduce tracking vectors. I'll add it over in that section; let me know what you think.

domenic:
What should we do about the non-compliant existing cases? Realistically I doubt anyone will update them. So I think we could either soften all the recommendations to "should", or we could retro-classify them as willful violations.

domenic:
The build says

> IDL ERROR LINE: 3 - Required argument "options" cannot follow optional argument "message"

but I cannot actually find a requirement in the Web IDL spec corresponding to this... I thought you could do `new WidgetError(undefined, opts)` to trigger the default value for the message. Anyone able to enlighten me?

index.bs
```diff
@@ -5071,87 +5071,104 @@ No [=extended attributes=] are applicable to dictionaries.
...
+The resulting behavior from creating and throwing an exception is language binding-specific.
+
+<p class=note>See [[#es-creating-throwing-exceptions]] for details on what creating and throwing an
+exception entails in the ECMAScript language binding.
 
-<h4 id="idl-DOMException-error-names">Error names</h4>
+
+<h4 id="idl-DOMException-error-names">{{DOMException}} error names</h4>
 
 The <dfn id="dfn-error-names-table" export>error names table</dfn> below lists all the allowed error names
```
dandrader, line 5173: Shouldn't this be rephrased, since other specs will now be able to introduce new ones?

annevk:
Thank you for filing those downstream issues! I suggest that if they cannot become compliant we call them out explicitly as exceptions (hah!) not to be repeated.

dandrader:
It's not clear to me whether a spec would be allowed to create a DOMException with a new error name (ie, not from that table) directly, without defining an interface that inherits from DOMException.

eg: Throw a "SomeSpecSpecificError" DOMException.
instead of: Throw a SpecError with some-error-code

annevk:
So for `WidgetErrorOptions` one thing is that the argument needs `= {}` at the end. Perhaps that is enough to silence widlparser? I would be somewhat surprised if the required dictionary argument were to blame.

domenic:
> It's not clear to me whether a spec would be allowed to create a DOMException with a new error name (ie, not from that table) directly, without defining an interface that inherits from DOMException.

The intention is you are not supposed to do that. And you are not supposed to define subclasses, just because you want a new name. Instead you are supposed to come coordinate on the Web IDL repo to determine if any existing names will suit your purpose, and if not, send a PR to add it to the table.

I agree making this explicit is a good idea, and I'll do so.

domenic:
> So for `WidgetErrorOptions` one thing is that the argument needs `= {}` at the end. Perhaps that is enough to silence widlparser? I would be somewhat surprised if the required dictionary argument were to blame.

The WidgetErrorOptions are intentionally not optional, since they have a required member. So we can't default them, I don't think.

dandrader:
> > It's not clear to me whether a spec would be allowed to create a DOMException with a new error name (ie, not from that table) directly, without defining an interface that inherits from DOMException.
> 
> The intention is you are not supposed to do that. And you are not supposed to define subclasses, just because you want a new name. Instead you are supposed to come coordinate on the Web IDL repo to determine if any existing names will suit your purpose, and if not, send a PR to add it to the table.

A spec will probably want error names that are rather specific to its problem space and thus not generally useful. I suppose the criteria for adding entries to this table is that the error name could be useful in several different contexts or problem domains? Would be good to clarify the criteria or rationale for adding new ones.

I also find it a bit confusing that a spec that defines a new interface inheriting from DOMException will also define a new, matching, error name, but that error name won't be present in that table. Or will it?

annevk:
I see, I think there used to be a requirement that optional cannot be followed by non-optional, but I cannot find that now.

> I also find it a bit confusing that a spec that defines a new interface inheriting from DOMException will also define a new, matching, error name, but that error name won't be present in that table. Or will it?

I think it would make sense to add these (as well as a reference to the class) to encourage reuse.

index.bs
```diff
@@ -14367,39 +14484,69 @@ A {{DOMException}} is represented by a
...
+
+    1.  Set |ex|'s [=DOMException/message=] to an [=implementation-defined=] message appropriate for
+        the exceptional situation. The calling specification may contain information to to help
+        implementations construct this message.
+
+        Implementations need to be cautious not to leak sensitive or secured information when
+        constructing this message, e.g., by including the URL of a cross-origin frame, or
+        information which could identify the user.
+
+    1.  Initialize |ex|'s additionally as described by the caller.
```
annevk, line 14540: I don't think this is English?

index.bs
```diff
@@ -5069,96 +5069,125 @@ No [=extended attributes=] are applicable to dictionaries.
...
+An <dfn id="dfn-exception" export>exception</dfn> is a type of object that represents an error and
+which can be thrown or treated as a first class value by implementations. Web IDL has a number of
+pre-defined exceptions that specifications can reference and throw in their definition of
+operations, attributes, and so on. Custom exception types can also be defined, as [=interfaces=]
+that [=interface/inherit=] from {{DOMException}}.
+
+In addition to their type, which is either one of the [=simple exceptions=], {{DOMException}}, or a
+derived interface of {{DOMException}}, all exceptions have a
+<dfn id="dfn-exception-message" for="exception" export>message</dfn>, which is an
+[=implementation-defined=] [=string=] that provides human readable details of the error.
```
annevk, line 5081:
```suggestion
[=implementation-defined=] [=string=] that provides human-readable details of the error.
```

index.bs
```diff
@@ -5069,96 +5069,125 @@ No [=extended attributes=] are applicable to dictionaries.
...
+    which throw a "{{SyntaxError}}" {{DOMException}}. In contrast, if your specification throws a
+    "{{NotAllowedError}}" {{DOMException}} immediately after checking if the user has provided
+    permission to use a given feature, it's fairly obvious what sort of [=exception/message=] the
+    implementation should construct, and so specifying it is not necessary.
+</div>
 
-<p class="warning">
-    The {{DOMException}} names marked as deprecated are kept for legacy purposes but their usage is discouraged.
-</p>
+The resulting behavior from creating and throwing an exception is language binding-specific.
```
annevk, line 5166:
```suggestion
The resulting behavior from creating and throwing an exception is language-binding specific.
```
(Or maybe hyphens everywhere?)

index.bs
```diff
@@ -5069,96 +5069,125 @@ No [=extended attributes=] are applicable to dictionaries.
...
+error names mean, and legacy numeric error code values.
+
+<p class="note">Interfaces [=interface/inheriting=] from {{DOMException}}, in the manner described
+in [[#idl-DOMException-derived-interfaces]], will have their own names, not listed in this table.
 
-Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
+When [=exception/creating=] or [=exception/throwing=] a {{DOMException}}, specifications must use
+one of these names. If a specification author believes none of these names are a good fit for their
+case, they must
+<a href="https://github.com/heycam/webidl/issues/new?title=DOMException%20name%20proposal">file an issue</a>
```
annevk, line 5184:
```suggestion
<a href="https://github.com/whatwg/webidl/issues/new?title=DOMException%20name%20proposal">file an issue</a>
```

index.bs
```diff
@@ -5069,96 +5069,125 @@ No [=extended attributes=] are applicable to dictionaries.
...
+pre-defined exceptions that specifications can reference and throw in their definition of
+operations, attributes, and so on. Custom exception types can also be defined, as [=interfaces=]
+that [=interface/inherit=] from {{DOMException}}.
+
+In addition to their type, which is either one of the [=simple exceptions=], {{DOMException}}, or a
+derived interface of {{DOMException}}, all exceptions have a
+<dfn id="dfn-exception-message" for="exception" export>message</dfn>, which is an
+[=implementation-defined=] [=string=] that provides human readable details of the error.
+{{DOMException}} instances also have an
+<dfn id="dfn-exception-error-name" for="exception" export>error name</dfn> [=string=], which
```
annevk, line 5083: It's weird that we have error name here, but then for `DOMException` we also define an internal name field. I don't think we need both? Message seems similarly duplicated.

index.bs
```diff
@@ -5342,6 +5371,94 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+    The definition for a {{DOMException}} derived interface which carries along an additional
+    "hardware error code", for interfacing with a hypothetical Widget hardware device, could look
+    something like this:
+
+    <pre highlight=webidl>
+        [Exposed=Window]
+        interface WidgetError : DOMException {
+          constructor(optional DOMString message = "", WidgetErrorOptions options);
+
+          readonly attribute unsigned long long hardwareErrorCode;
```
annevk, line 5408: Perhaps we ought to encourage an enum instead?

annevk (commented):
I found a couple nits, looks great overall.

index.bs
domenic: It looks like we have many instances of no hyphens at all, so I'll switch to that. I get the sense this is one of those questions that vary by style guide, but I can't seem to find the right combination of Google keywords to pin down any specific style guides for what's happening here (which is a noun phrase being used to modify an adjective).

index.bs
domenic: Also fixing all the others in the document

index.bs
domenic: The distinction is that these are defining the general concepts/types, and the DOMExceptions definitions are fields.

I think we could eliminate "error name", especially since in this PR I narrow it down to be DOMException-specific. I'll try that.

Eliminating "message" is a bit harder since it applies to all exceptions, not just DOMException. But I think it's doable.

index.bs
domenic: I'm kind of envisioning this being some opaque number spit out by the widget hardware. I can see how maybe that's not a good pattern to have as our only example, as we'd arguably want to have a web platform API abstract over that sort of thing (like the SmartCard proposal is doing).

Using an enum, to me, gets into #1219, which I don't consider quite settled yet.

We should probably change this to be some number spit out by a protocol, instead of by a widget, since that's a bit more respectable and has precedent in both RTCError and WebTransportError.

index.bs
```diff
@@ -5342,6 +5362,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+
+<div class=example id=example-domexception-derived-interface>
+    The definition for a {{DOMException}} derived interface which carries along an additional
+    "protocol error code", which is derived from what the server sent over some some hypothetical
+    network protocol "protocol X", could look something like this:
+
+    <pre highlight=webidl>
+        [Exposed=Window]
+        interface ProtocolXError : DOMException {
+          constructor(optional DOMString message = "", ProtocolXErrorOptions options);
```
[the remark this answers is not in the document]
domenic: That error is incorrect. https://github.com/plinss/widlparser/issues/79

index.bs
```diff
@@ -5342,6 +5362,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+    </pre>
+
+    Every <code>ProtocolXError</code> instance has an <dfn for="ProtocolXError">error code</dfn>,
+    a number.
+
+    <div algorithm="ProtocolXError constructor">
+        The <b><code>new ProtocolXError(|message|, |options|)</code></b> constructor steps are:
+
+        1.  Set [=this=]'s [=DOMException/name=] to "<code>ProtocolXError</code>".
+        1.  Set [=this=]'s [=DOMException/message=] to |message|.
```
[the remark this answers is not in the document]
domenic: There is no such concept in Web IDL.

index.bs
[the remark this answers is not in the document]
kainino0x: Ahh, I didn't realize this was the widlparser bug you were talking about. Sorry about that!

index.bs
```diff
@@ -5342,6 +5362,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+
+<div class=example id=example-domexception-derived-interface>
+    The definition for a {{DOMException}} derived interface which carries along an additional
+    "protocol error code", which is derived from what the server sent over some some hypothetical
+    network protocol "protocol X", could look something like this:
+
+    <pre highlight=webidl>
+        [Exposed=Window, Serializable]
+        interface ProtocolXError : DOMException {
+          constructor(optional DOMString message = "", ProtocolXErrorOptions options);
```
kainino0x, line 5397: IIUC `(DOMString or undefined) message`, with prose describing the defaulting to `""`, would be equivalent but work around the widlparser issue. Does this seem correct?
Context: https://github.com/gpuweb/gpuweb/pull/3710

index.bs
```diff
@@ -5342,6 +5362,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+        Their [=serialization steps=], given |value| and |serialized|, are:
+
+        1. Run the {{DOMException}} [=serialization steps=] given |value| and |serialized|.
+        1. Set |serialized|.\[[ErrorCode]] to |value|'s [=ProtocolXError/error code=].
+    </div>
+
+    <div algorithm="ProtocolXError deserialization steps">
+        Their [=deserialization steps=], given |value| and |serialized|, are:
+
+        1. Run the {{DOMException}} [=deserialization steps=] given |value| and |serialized|.
```
jan-ivar, line 5435: Arguments are reversed on deserialize.
```suggestion
        Their [=deserialization steps=], given |serialized| and |value|, are:

        1. Run the {{DOMException}} [=deserialization steps=] with |serialized| and |value|.
```

index.bs
```diff
@@ -5342,6 +5362,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+        The <b><code>errorCode</code></b> getter steps are to return [=this=]'s
+        [=ProtocolXError/error code=].
+    </div>
+
+    <code>ProtocolXError</code> objects are [=serializable objects=].
+
+    <div algorithm="ProtocolXError serialization steps">
+        Their [=serialization steps=], given |value| and |serialized|, are:
+
+        1. Run the {{DOMException}} [=serialization steps=] given |value| and |serialized|.
```
jan-ivar, line 5428: Nit: not sure what best practice is, but it seems odd to me to use "given" both on the calling and receiving end of an algorithm. I'd expect "with" on the calling end, like [two lines down from here](https://dom.spec.whatwg.org/#set-the-canceled-flag).

```suggestion
        Their [=serialization steps=], given |value| and |serialized|, are:

        1. Run the {{DOMException}} [=serialization steps=] with |value| and |serialized|.
```

index.bs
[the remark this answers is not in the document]
jan-ivar: Still, why can't we reference the[ DOMException constructor steps](https://webidl.spec.whatwg.org/#dom-domexception-domexception) (which already expects `this` to be set up) directly? E.g. [like this](https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/460.html#dom-webtransporterror-webtransporterror): _"Run the [new DOMException(message, name)](https://heycam.github.io/webidl/#dom-domexception-domexception) constructor steps on [this](https://webidl.spec.whatwg.org/#this) with message and "WebTransportError"."_

I concede the constructor naming suggests perhaps constructors aren't intended to be referenced directly like this — the "new" in the name seems to imply an implicit memory allocation and assignment to [=this=]. Is that how it's to be read? — If so, I'm having trouble doing that as well elsewhere in the PR where I DO want the allocation: _"Return a new [WebTransportError()](https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/460.html#dom-webtransporterror-webtransporterror) with message and options."_ (how can I put the args in the ())?

In lieu of that, I'm struggling to access [=DOMException/name=] which is different from {{DOMException/name}}. Will take any bikeshed help.

Sorry for dropping this request for PR help here.

index.bs
jan-ivar, line 5397:
```suggestion
          constructor(optional DOMString message = "", optional ProtocolXErrorOptions options);
```

domenic: This is incorrect; in the example they are not optional.

index.bs
domenic: "given" is fine here.

domenic:
Hey, the blocking build issue was fixed!! This is ready for review, and hopefully merging.

index.bs
```diff
@@ -4898,96 +4898,117 @@ No [=extended attributes=] are applicable to dictionaries.
...
 
-These correspond to all of the ECMAScript [=ECMAScript/error objects=]
-(apart from <l spec=ecmascript>{{SyntaxError}}</l> and {{Error}},
-which are deliberately omitted as they are reserved for use
-by the ECMAScript parser and by authors, respectively).
-The meaning of each [=simple exception=] matches
-its corresponding error object in the
-ECMAScript specification.
+These correspond to all of the ECMAScript [=ECMAScript/error objects=] (apart from
+<l spec=ecmascript>{{SyntaxError}}</l> and <l spec=ecmascript>{{Error}}</l>, which are deliberately
```
bathos, line 4917: pre-existing, but: AggregateError, too.

(Given there’s at least [one more in the pipeline](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=3000#sec-suppressederror-constructor), maybe keeping this in sync is too awkward and the gist could be conveyed without enumerating the excluded constructors...?)

index.bs
```diff
@@ -5176,6 +5197,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+        };
+    </pre>
+
+    Every <code>ProtocolXError</code> instance has an <dfn for="ProtocolXError">error code</dfn>,
+    a number.
+
+    <div algorithm="ProtocolXError constructor">
+        The <b><code>new ProtocolXError(|message|, |options|)</code></b> constructor steps are:
+
+        1.  Set [=this=]'s [=DOMException/name=] to "<code>ProtocolXError</code>".
```
kainino0x, line 5248: +1 to https://github.com/whatwg/webidl/pull/1211#discussion_r1085536857:
It seems `[=DOMException/name=]` should be `{{DOMException/name}}`, etc

domenic: No. We need to reference the internal slots (`[=DOMException/name=]`); you cannot set public getters (`{{DOMException/name}}`) to values.

kainino0x: OK, I had a suspicion that might be why. But I tried it, and those internal slots are not referenceable from the WebGPU spec. Is there a reason the reference is allowed in this spec but not ours?

domenic: They're not exported, for some reason! Let me fix that in this PR too.

kainino0x: Oho, for some reason I thought these were cross-references to the HTML spec, not internal references in the WebIDL spec. Thanks!

index.bs
annevk: Please either file a follow-up or address.

bathos: @annevk distinct issue here: https://github.com/whatwg/webidl/issues/1314

index.bs
```diff
@@ -5176,6 +5197,93 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
...
+certain rules, in order to have a predictable shape for developers. Specifically:
+
+*   The [=identifier=] of the [=interface=] must end with <code>Error</code>, and must not be any
+    of the names in the <a><code>DOMException</code> names table</a>.
+*   The [=interface=] must have a [=constructor operation=] which sets the instance's
+    [=DOMException/name=] to the interface's [=identifier=].
+*   Their [=constructor operation=] must take as its first parameter an [=optional argument|optional=]
+    {{DOMString}} named |message| defaulting to the empty string, and must set the instance's
+    [=DOMException/message=] to |message|.
+*   Their [=constructor operation=] should take as its second parameter a [=dictionary=] containing
```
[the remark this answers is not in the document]
domenic: Yeah. I'm not 100% sure, but I feel like being too prescriptive here is not necessary, since we don't know exactly what shape the additional information will come in.

PR #1215, opened 2022-10-11
Document what defaults are ok for dictionary values.

jyasskin:
***
<a href="https://whatpr.org/webidl/1215.html" title="Last updated on Oct 13, 2022, 5:14 AM UTC (9996240)">Preview</a> | <a href="https://whatpr.org/webidl/1215/e3d0c74...9996240.html" title="Last updated on Oct 13, 2022, 5:14 AM UTC (9996240)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4752,11 +4752,6 @@ supplied as an argument to an [=operation=]) to an IDL dictionary. Specification
 leave the members [=dictionary member/optional=] in all other cases, including when a dictionary
 type is used solely as the [=return type=] of [=operations=].
 
-[=dictionary member/Optional=] [=dictionary members=] can also be specified as having a
-<dfn id="dfn-dictionary-member-default-value" for="dictionary member" export>default value</dfn>
-(matching <emu-nt><a href="#prod-DefaultValue">DefaultValue</a></emu-nt>), which is the value used
-by default when author code or specification text does not provide a value for that member.
-
 A given dictionary value of type |D| can have [=map/entries=] for each of the dictionary members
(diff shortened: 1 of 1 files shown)


domenic:
This is kind of messy since the section you're modifying is defining the concept of default value, whereas the thing you're referencing is the grammar production for how default values manifest in the syntax. The syntax is discussed below already:

> If the identifier is followed by a U+003D EQUALS SIGN ("=") and a value (matching [DefaultValue](https://whatpr.org/webidl/1215.html#prod-DefaultValue)), then that gives the dictionary member its [default value](https://whatpr.org/webidl/1215.html#dfn-dictionary-member-default-value).

I assume you're doing this because the existing text was not clear to someone, so that probably means we should try to make default values clearer in some way, but I'm hesitant to do so by duplicating the "what is a default value's syntax" parts of the spec into the semantics parts.

jyasskin:
Yep, I couldn't find the place that said what defaults were possible without reading the whole dictionaries section. Thanks for pointing it out. This isn't a problem for [optional arguments](https://webidl.spec.whatwg.org/#dfn-optional-argument-default-value), which don't try to separate their syntax from their semantics. Dictionaries don't really have a separation between syntax and semantics either, so I'd be inclined to move the whole definition of default values down to the place that says what their values can be. Here's that patch in case you agree.

domenic:
I suspect I am happy with this, but PR preview is broken because Bikeshed. And that is also breaking the build. So, we'll revisit this at some point...

domenic (approved):
Nice, now that I can see the preview, this looks like a nice improvement.

PR #1216, opened 2022-10-11
"Quoted terminal symbols" are now monospaced

### LINKED ISSUE #976: Replace the word "quoted" from IDL grammar section
https://heycam.github.io/webidl/#idl-grammar

>Thus, the input text “long” is tokenized as the quoted terminal symbol long rather than an identifier called "long", and “.” is tokenized as the quoted terminal symbol . rather than an other.

This is from when terminal symbols were actually quoted with `"`, and now that's not true anymore as they are just monospaced.

saschanaz:
Fixes #976

Some other candidates are there, but I think the term `monospaced` is the most intuitive when looking at the grammar:

>*CallbackOrInterfaceOrMixin* `::`
>    `callback` *CallbackRestOrInterface*
>    `interface` *InterfaceOrMixin*

It's clear which are the monospaced terminal symbols.

***
<a href="https://whatpr.org/webidl/1216.html" title="Last updated on Oct 13, 2022, 6:22 AM UTC (2bf7d7c)">Preview</a> | <a href="https://whatpr.org/webidl/1216/bc7cb7a...2bf7d7c.html" title="Last updated on Oct 13, 2022, 6:22 AM UTC (2bf7d7c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14896,12 +14896,12 @@ text is “<span class="input">a1</span>”, it is tokenized as a single <emu-t 
 and not as a separate <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> and <emu-t class="regex"><a href="#prod-integer">integer</a></emu-t>.
 If the longest possible match could match one of the above named terminal symbols or
 one of the other terminal symbols from the grammar, it must be tokenized as the latter.
-Thus, the input text “<span class="input">long</span>” is tokenized as the quoted terminal symbol
+Thus, the input text “<span class="input">long</span>” is tokenized as the terminal symbol
 <emu-t>long</emu-t> rather than an <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> called "<code>long</code>",
-and “<span class="input">.</span>” is tokenized as the quoted terminal symbol
+and “<span class="input">.</span>” is tokenized as the terminal symbol
 <emu-t>.</emu-t> rather than an <emu-t class="regex"><a href="#prod-other">other</a></emu-t>.
 
-The IDL syntax is case sensitive, both for the quoted terminal symbols
+The IDL syntax is case sensitive, both for the monospaced terminal symbols
 used in the grammar and the values used for
 <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> terminals.  Thus, for
 example, the input text “<span class="input">Const</span>” is tokenized as


saschanaz:
Not sure what you mean, do you want just "terminal symbol"? I'm not sure that works...

annevk:
Yeah, why would that not work?

saschanaz:
Because "terminal symbols" does not specifically point to the symbol literals?

annevk:
But it's clear what kind of symbol literal is being discussed, no? With `long` vs an identifier whose value is "`long`". Not sure why a specific name is needed given that context.

And also, there is no actual specific name defined for them in the specification and using the style is not great from an accessibility perspective. I supposed based on that we should adjust some earlier bits in this section as well.

saschanaz:
For the upper paragraph I guess you're right, but it's less clear for "The IDL syntax is case sensitive, both for the terminal symbols used in the grammar and the values used for identifier terminals". Both are terminals, anyway...

I'd prefer removing the words from the upper paragraph and keep it for the lower one.

annevk:
They are both terminal symbols, but if you say that `long` is not an identifier "`long`" how is it unclear which terminal symbol you meant? But if we truly want to distinguish them we should give them different names. E.g., generic terminal symbols and specific terminal symbols.

saschanaz:
> but if you say that `long` is not an identifier "`long`"

That's the upper paragraph one that I agree to remove the extra words 😛

index.bs
```diff
@@ -14898,12 +14898,12 @@ text is “<span class="input">a1</span>”, it is tokenized as a single <emu-t
...
-Thus, the input text “<span class="input">long</span>” is tokenized as the quoted terminal symbol
+Thus, the input text “<span class="input">long</span>” is tokenized as the monospaced terminal symbol
 <emu-t>long</emu-t> rather than an <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> called "<code>long</code>",
-and “<span class="input">.</span>” is tokenized as the quoted terminal symbol
+and “<span class="input">.</span>” is tokenized as the monospaced terminal symbol
 <emu-t>.</emu-t> rather than an <emu-t class="regex"><a href="#prod-other">other</a></emu-t>.
 
-The IDL syntax is case sensitive, both for the quoted terminal symbols
+The IDL syntax is case sensitive, both for the monospaced terminal symbols
 used in the grammar and the values used for
```
annevk, line 14907: Ah I misunderstood. I thought you meant the paragraphs even earlier that also talk about fonts.

Perhaps here we can simply reuse the line from above "other terminal symbols from the grammar"?

index.bs
```diff
@@ -14898,12 +14898,12 @@ text is “<span class="input">a1</span>”, it is tokenized as a single <emu-t
...
 <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> terminals.  Thus, for
```
saschanaz, line 14908: Like this?

```suggestion
The IDL syntax is case sensitive, both for the the values used for
<emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> terminals and for other terminal symbols from the grammar. Thus, for
```

There are still other terminal symbols that are not identifiers nor literals, though. 🤔

annevk: Hmm okay. I guess we can go with monospaced for now here and file a follow-up on revamping this to use better terms and not rely on styling to distinguish these?

annevk (approved):

index.bs
annevk: Filed #1217.

index.bs
```diff
@@ -14896,12 +14896,12 @@ text is “<span class="input">a1</span>”, it is tokenized as a single <emu-t
 and not as a separate <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> and <emu-t class="regex"><a href="#prod-integer">integer</a></emu-t>.
 If the longest possible match could match one of the above named terminal symbols or
 one of the other terminal symbols from the grammar, it must be tokenized as the latter.
-Thus, the input text “<span class="input">long</span>” is tokenized as the quoted terminal symbol
+Thus, the input text “<span class="input">long</span>” is tokenized as the terminal symbol
 <emu-t>long</emu-t> rather than an <emu-t class="regex"><a href="#prod-identifier">identifier</a></emu-t> called "<code>long</code>",
-and “<span class="input">.</span>” is tokenized as the quoted terminal symbol
+and “<span class="input">.</span>” is tokenized as the terminal symbol
```
bathos, line 14901: The example in this line is not exactly a “real” terminal symbol of the grammar. Though it’s still present as an alternative of `Other`, it’s not used elsewhere and therefore has no meaning — if it were removed, everything would stay exactly the same because  `Other` would still include it via `other`, which is also only used in `Other` and only currently _doesn’t_ include `.` it  because `Other` _does_ include it. :)

Though the example was already present, the new language seems to make this situation even more confusing and probably justifies removing the vestigial `.` and `-` alternatives of `Other`.

annevk: Wanna create a follow-up PR @bathos? :-)

bathos: Afraid I’m [unable to](https://github.com/whatwg/sg/issues/93#issuecomment-1120799691).

Issue #1168, opened 2022-07-13
New error name: OptOutError


rsolomakhin:
Hello,

Would it be possible to add an "OptOutError" to the list of DOMException error names? 

[SPC](https://www.w3.org/TR/secure-payment-confirmation/) is experimenting with [opt out support](https://github.com/w3c/secure-payment-confirmation/issues/172) by returning an "AbortError: User has opted out of the process", but many different error conditions can return "AbortError" and checking for the error message is generally considered brittle. Perhaps an "OptOutError" can be used by other APIs that allow users to opt out, as well?

cc @jcemer-stripe @stephenmcgruer

bathos:
I’m curious whether use of multiple AbortSignals has been considered? That’s how I’ve ended up handling similar cases in userland generally and it permits assurances / control flow that an error code doesn’t.

stephenmcgruer:
@rsolomakhin - they're a somewhat* new mechanism on the web that allows asynchronous signalling that 'something' has aborted - see the [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal).

I'm not familiar with them being used as an output source from a Web API, however, only as an input to one. (That is, I'm used to them being a mechanism for the website to cause a browser operation to abort, rather than a browser operation to tell the website that it has or is aborting).

I do think there are other ways we could handle this result rather than a new DOMException type, however. I personally am partial to not considering opt-out a true error and just having the Secure Payment Confirmation API have a 'type' or 'status' flag on its success return object to the calling website.

That said, I am generally interested in whether we consider the set of DOMExceptions on the web 'fixed' or are still expanding it over time, and if we've put in any thought to whether DOMExceptions could carry more structured data to allow for better error handling by developers. The root of this issue was that we currently throw the same DOMException error type for two semantically different outcomes from Secure Payment Confirmation.

* Maybe not **that** new, I'm just old 🤣

bathos:
[AbortController & AbortSignal](https://dom.spec.whatwg.org/#aborting-ongoing-activities) are the web platform’s generic cancellation primitives. Some examples of their use in other specs:

- [Eyedropper](https://wicg.github.io/eyedropper-api/#ref-for-dom-colorselectionoptions-signal-1)
- [Fetch](https://fetch.spec.whatwg.org/#request-signal)
- [Geolocation Sensor](https://w3c.github.io/geolocation-sensor/#dictdef-readoptions)
[...]

PR #1231, opened 2022-11-11
Add OptOutError to the error names list

nickburris:
This PR adds a new OptOutError to the DOMException error names list. We plan to use this error name in the Secure Payment Confirmation API (see issue #1168). Since we don't need any additional attributes (i.e., we don't need a new class interface to extend DOMException), a new error name will suffice. The error name is generic so that it can be used by other APIs with similar use cases.

   * In the short term, this error name will be used by the Secure Payment Confirmation API implemented in Chrome, and can be used in a similar fashion by other APIs.
   * Usage of the error name will be tested by APIs (e.g. Secure Payment Confirmation) that use it.
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1383295
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed:
   * Will open a PR to add this to https://developer.mozilla.org/en-US/docs/Web/API/DOMException#error_names

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5339,6 +5339,11 @@ over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[D
             <td>The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.</td>
             <td>—</td>
         </tr>
+        <tr>
+            <td>"<dfn id="optouterror" exception export><code>OptOutError</code></dfn>"</td>
+            <td>User opted out of the process.</td>
+            <td>—</td>
+        </tr>
     </tbody>
 </table>
 


nickburris:
PTAL @rsolomakhin / @stephenmcgruer with your SPC API hat on (impl: https://chromium-review.googlesource.com/c/chromium/src/+/4021495)

PTAL @domenic with your webidl spec hat on, + @annevk thank you for your input on the issue, extending the error names list per your recommendation :)

rsolomakhin (commented):
LGTM!

domenic (approved):
This LGTM. Even if Secure Payment Confirmation is a single-implementer API, I think we should be lenient with merging values into this table, especially if they're designed to be generic in this way.

/cc @annevk @EdgarChen in case they disagree. I propose we give this a couple days in case others have thoughts then I'll merge.

annevk:
Yeah I think that's okay. @nickburris should make their membership of the googlers organization public.

PR #1270, opened 2023-02-21
Fix links, remove as many anchor definitions as possible

tabatkins:
This is purely an editorial change. It fixes the link errors the current spec has (around Memory and Memory/buffer), and additionally removes a large number of the explicit anchors defined in the spec. For the most part these removals were simple; in a few places I fixed the linking syntax down in the spec body as well.

The abstract-op anchors left behind just aren't being detected by WebRef; [I opened an issue about them](https://github.com/w3c/webref/issues/888). When that's fixed we should be able to remove that entire block.

A number of the remaining anchors are things that should probably be fixed in ECMAScript, but I don't have bandwidth to do that right now. I haven't done a full accounting of the remaining anchor values; some of them might also be removable right now.

***
<a href="https://whatpr.org/webidl/1270.html" title="Last updated on Feb 28, 2023, 7:56 PM UTC (4f171d4)">Preview</a> | <a href="https://whatpr.org/webidl/1270/8728881...4f171d4.html" title="Last updated on Feb 28, 2023, 7:56 PM UTC (4f171d4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4927,7 +4927,7 @@ is identified by one of the following types:
 *   <dfn exception>URIError</dfn>
 
 These correspond to all of the ECMAScript [=ECMAScript/error objects=]
-(apart from {{SyntaxError}} and {{Error}},
+(apart from <l spec=ecmascript>{{SyntaxError}}</l> and {{Error}},
 which are deliberately omitted as they are reserved for use
 by the ECMAScript parser and by authors, respectively).
 The meaning of each [=simple exception=] matches
@@ -4996,10 +4996,10 @@ for {{DOMException}}, a description, and legacy code values.
 Note: If an error name is not listed here, please file a bug as indicated at the top of this specification and it will be addressed shortly. Thanks!
 
 Note: Don't confuse the "{{SyntaxError!!exception}}" {{DOMException}} defined here
-with ECMAScript's {{SyntaxError}}.
+with ECMAScript's <l spec=ecmascript>{{SyntaxError}}</l>.
 "{{SyntaxError!!exception}}" {{DOMException}} is used to report parsing errors in web APIs,
 for example when parsing selectors,
-while the ECMAScript {{SyntaxError}} is reserved for the ECMAScript parser.
+while the ECMAScript <l spec=ecmascript>{{SyntaxError}}</l> is reserved for the ECMAScript parser.
 To help disambiguate this further,
 always favor the "{{SyntaxError!!exception}}" {{DOMException}} notation
 over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[DOM]]
--- a/index.bs
+++ b/index.bs
@@ -4,6 +4,7 @@ H1: Web IDL
 Shortname: webidl
 Text Macro: TWITTER webidl
 Text Macro: LATESTRD 2022-09
+Text Macro: COMMIT-SHA foo
 Abstract: This standard defines an interface definition language, Web IDL, that can be used to describe interfaces that
 Abstract: are intended to be implemented in web browsers.
 Translation: ja https://triple-underscore.github.io/WebIDL-ja.html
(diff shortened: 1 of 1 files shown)


tabatkins:
Oh hm, it's failing due to a badly-encoded URL, which I think is a Bikeshed bug I noticed before. I'll go investigate.

tabatkins:
Nah not a Bikeshed bug, it's a data source bug. Reported at https://github.com/w3c/webref/issues/889

index.bs
```diff
@@ -12,72 +12,16 @@ Complain About: accidental-2119 no
 
 <pre class="link-defaults">
 spec: infra; type: dfn; text: list
-spec: url; type: interface; text: URL
-spec: dom; type: interface; text: Document
 spec: ecma-262; type: dfn; for: /; text: internal method
 spec: ecma-262; type: dfn; for: /; text: internal slot
 spec: ecma-262; type: dfn; for: /; text:realm
+spec:wasm-js-api-2;
```
annevk, line 18: No variant without a version? cc @Ms2ger

annevk (commented):
Wow, this looks lovely.

index.bs
tabatkins: Yeah, it'll be fixed separately, but in the meantime this link-default fixes the error locally.

index.bs
```diff
@@ -6936,8 +6869,8 @@ in ECMAScript, as defined by the <cite>ECMAScript Language Specification</cite>
 [[!ECMA-262]].
 
 Unless otherwise specified, objects defined in this section are ordinary objects as described in
-[=ECMA-262 Ordinary object internal methods and internal slots=], and if the
-object is a [=function object=], [=ECMA-262 Built-in function objects=].
+[[ecmascript#sec-ordinary-object-internal-methods-and-internal-slots]], and if the
+object is a [=function object=], [[ecmascript#sec-built-in-function-objects]].
```
TimothyGu, line 6873: This is wonderful. Thanks!

index.bs
```diff
@@ -7655,7 +7587,7 @@ values are represented by ECMAScript Object values.
     An ECMAScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{object}} value by running the following algorithm:
 
-    1.  If <a abstract-op>Type</a>(|V|) is not Object, then [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+    1.  If <a abstract-op>Type</a>(|V|) is not Object, then [=ECMAScript/throw=] a {{TypeError}}.
```
TimothyGu, line 7590: This is another name clash between ES and Web IDL. Previously it linked to ECMAScript, but now it goes to Web IDL. It's mostly fine since Web IDL's `{{TypeError}}` is basically a redirect to ES. But since these algorithm steps are written in the style of ECMAScript (including the `throw` reference), I'd still prefer to use the ECMAScript link if possible.

index.bs
```diff
@@ -5154,10 +5087,10 @@ for {{DOMException}}, a description, and legacy code values.
...
 Note: Don't confuse the "{{SyntaxError!!exception}}" {{DOMException}} defined here
-with ECMAScript's {{ECMAScript/SyntaxError}}.
+with ECMAScript's {{SyntaxError}}.
 "{{SyntaxError!!exception}}" {{DOMException}} is used to report parsing errors in web APIs,
 for example when parsing selectors,
-while the ECMAScript {{ECMAScript/SyntaxError}} is reserved for the ECMAScript parser.
+while the ECMAScript {{SyntaxError}} is reserved for the ECMAScript parser.
 To help disambiguate this further,
 always favor the "{{SyntaxError!!exception}}" {{DOMException}} notation
 over just using {{SyntaxError!!exception}} to refer to the {{DOMException}}. [[DOM]]
```
TimothyGu, line 5096: The new `{{SyntaxError}}` references are incorrect: it goes to the unrelated `"SyntaxError"` `DOMException` in Web IDL rather than ES. In fact, this paragraph explains exactly how they are different.

index.bs
tabatkins: Ohhh, I missed that distinction, indeed. That can be fixed. (Using `<l spec=ecmascript>{{TypeError}}</l>`.)

domenic:
Looks like the webref data source bug is fixed, so this is ready for @tabatkins to fix up the exception references and then we should be good to go.

tabatkins:
All right, the TypeError and SyntaxError refs are fixed, and now that WebRef has fixed a lot of the missing dfns, I've done another pass thru the anchors and removed every single one that we no longer need. I specifically confirmed for all of these new removals that they didn't change any behavior; two or three now link more directly to a term in the first paragraph of a section, rather than the section itself, but I figured that was fine. The rest all link to the same anchors as before (just using the multipage version of the spec instead).

TimothyGu (approved):
Thanks so much! I looked through the changes and everything looks good to me.

There were a few TypeError links that changed from ES to Web IDL, but I think they should've been IDL in the first place, so thanks for fixing those also.

Issue #335, opened 2017-03-29
Revamp ByteString, DOMString, and USVString


annevk:
I think the goal should be to map these to

* byte sequence
* JavaScript string
* scalar value string

as defined in the Infra Standard. We'll most likely need to copy at the boundary for ByteString, but for DOMString and USVString that depends on the resolution of https://github.com/whatwg/infra/issues/91. Unless we get more opinions there I suspect we'll end up copying for all three, because that's most in line with how things are done (I realize this is not a great argument).

https://heycam.github.io/webidl/#dfn-obtain-unicode can be removed and any usage of it can use https://infra.spec.whatwg.org/#javascript-string-convert instead.

Existing issues here that we should be sure to take into account: #84, #141, and #236.

jyasskin:
I was talking to @domfarolino about this topic last week: I concluded that a USVString "is" a [scalar value string](https://infra.spec.whatwg.org/#scalar-value-string), so specs can use it as an Infra string without a conversion step.

annevk:
I decided to close this based on @jyasskin's work in #1287 as the copy-at-boundary discussion is much larger. For now making it clear these are aliases seems sufficient.

PR #1287, opened 2023-04-04
Editorial: Define the WebIDL string types in terms of Infra types.

jyasskin:
This might be sufficient for #335, but I think it's useful for spec authors to go roughly this far even if there's more to do to completely fix that issue. 

cc/ @domfarolino 

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1287.html" title="Last updated on Apr 21, 2023, 3:33 PM UTC (5690199)">Preview</a> | <a href="https://whatpr.org/webidl/1287/a35410d...5690199.html" title="Last updated on Apr 21, 2023, 3:33 PM UTC (5690199)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5846,7 +5846,7 @@ The {{USVString}} type
 corresponds to [=scalar value strings=].
 Depending on the context,
 these can be treated as sequences of either 16-bit unsigned integer [=code units=]
-or [=scalar value=] [=code points=].
+or [=scalar values=].
 
 There is no way to represent a constant {{USVString}}
 value in IDL, although {{USVString}} [=dictionary member=] [=dictionary member/default values=]
--- a/index.bs
+++ b/index.bs
@@ -2099,15 +2099,14 @@ the value is the IDL {{undefined}} value.
              :  {{DOMString}}
              :  an [=enumeration=] type
              :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
-                is the sequence of 16 bit unsigned integer [=code units=]
-                corresponding to the UTF-16 encoding of |S|.
+                is |S|.
              :  {{ByteString}}
-             :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
-                is the sequence of 8 bit unsigned integer code units
-                corresponding to the UTF-8 encoding of |S|.
+             :: Assert: |S| doesn't contain any [=code points=] higher than U+00FF.
+
+                The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
+                is the [=isomorphic encoding=] of |S|.
              :  {{USVString}}
-             :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
-                is the [=scalar value string=] whose [=code points=] are |S|.
+             :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token is |S|.
         </dl>
 </div>
 
@@ -5795,9 +5794,6 @@ The {{bigint}} type is an arbitrary integer type, unrestricted in range.
 <h4 oldids="dom-DOMString" id="idl-DOMString" interface>DOMString</h4>
 
 The {{DOMString}} type corresponds to [=strings=].
-The sequence of unsigned 16-bit integers in a [=string=]
-is commonly interpreted as a UTF-16 encoded string [[!RFC2781]]
-although this is not required.
 
 Note: Note also that <emu-val>null</emu-val>
 is not a value of type {{DOMString}}.
@@ -7475,10 +7471,8 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
 <div id="USVString-to-es" algorithm="convert an USVString to an ECMAScript value">
 
     The result of [=converted to an ECMAScript value|converting=]
-    an IDL {{USVString}} value to an ECMAScript
-    value is the String
-    value that represents the sequence of [=code units=]
-    in the IDL {{USVString}}.
+    an IDL {{USVString}} value |S| to an ECMAScript
+    value is |S|.
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -11,7 +11,9 @@ Complain About: accidental-2119 no
 </pre>
 
 <pre class="link-defaults">
-spec: infra; type: dfn; text: list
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7471,12 +7474,11 @@ value when its bit pattern is interpreted as an unsigned 64 bit integer.
...
-    to an ECMAScript value by running the following algorithm:
-
-    1.  Let |scalarValues| be the sequence of [=scalar values=] the {{USVString}} represents.
-    1.  Let |string| be the sequence of [=code units=] that results from encoding |scalarValues| in UTF-16.
-    1.  Return the String value that represents the same sequence of [=code units=] as |string|.
+    The result of [=converted to an ECMAScript value|converting=]
+    an IDL {{USVString}} value to an ECMAScript
+    value is the String
+    value that represents the sequence of [=code units=]
+    in the IDL {{USVString}}.
```
domfarolino, line 7481: Do we have to consider the String length limit in ECMAScript at all? https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type

domfarolino (approved):

index.bs
```diff
@@ -2097,15 +2099,15 @@ the value is the IDL {{undefined}} value.
              :  {{DOMString}}
              :  an [=enumeration=] type
              :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
-                is the sequence of 16 bit unsigned integer code units
-                (hereafter referred to just as <dfn id="dfn-code-unit">code units</dfn>)
+                is the sequence of 16 bit unsigned integer [=code units=]
                 corresponding to the UTF-16 encoding of |S|.
```
annevk, line 2103: I think we can just say that the value is |S|. We already know the value is a scalar value string from how we collect it. And those are compatible with strings.

index.bs
```diff
@@ -2097,15 +2099,15 @@ the value is the IDL {{undefined}} value.
...
              :  {{ByteString}}
              :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
                 is the sequence of 8 bit unsigned integer code units
                 corresponding to the UTF-8 encoding of |S|.
```
annevk, line 2107: If we want to continue to treat these as byte sequences (I could see an argument for isomorphic string, though that would be a bigger change with downstream impact on Fetch and such), we should use assert that |S| doesn't contain any code points higher than U+00FF and use isomorphic encode. UTF-8 is plain wrong.

index.bs
```diff
@@ -2097,15 +2099,15 @@ the value is the IDL {{undefined}} value.
...
              :  {{USVString}}
-             :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token is |S|.
+             :: The value of the <emu-t class="regex"><a href="#prod-string">string</a></emu-t> token
+                is the [=scalar value string=] whose [=code points=] are |S|.
```
annevk, line 2110: I think the original text was already accurate here.

index.bs
```diff
@@ -5791,9 +5794,9 @@ The {{bigint}} type is an arbitrary integer type, unrestricted in range.
 
 <h4 oldids="dom-DOMString" id="idl-DOMString" interface>DOMString</h4>
 
-The {{DOMString}} type corresponds to
-the set of all possible sequences of [=code units=].
-Such sequences are commonly interpreted as UTF-16 encoded strings [[!RFC2781]]
+The {{DOMString}} type corresponds to [=strings=].
+The sequence of unsigned 16-bit integers in a [=string=]
+is commonly interpreted as a UTF-16 encoded string [[!RFC2781]]
 although this is not required.
```
annevk, line 5800: I think we should drop the bit about UTF-16 while here.

index.bs
annevk: Do we have to define this at all given how Infra defines them to be compatible?

annevk (commented):
I mainly found problems with the existing text. Thanks for tackling this!

index.bs
jyasskin: Done.

index.bs
jyasskin: I don't see a definition for "isomorphic string"? I added the assertion and isomorphic encoding. The assertion is redundant with one inside https://infra.spec.whatwg.org/#isomorphic-encode, but I think it's worth having one here too, so people know what they can't write in default arguments.

index.bs
jyasskin: We have to define something, but I can simplify this further.

jyasskin (commented):
Thanks for the review!

index.bs
```diff
@@ -5818,7 +5816,7 @@ can be set to [=value of string literal tokens|the value=] of a
 <h4 oldids="dom-ByteString" id="idl-ByteString" interface>ByteString</h4>
 
 The {{ByteString}} type
-corresponds to the set of all possible sequences of bytes.
+corresponds to [=byte sequences=].
 Such sequences might be interpreted as UTF-8 encoded strings [[!RFC3629]]
 or strings in some other 8-bit-per-code-unit encoding, although this is not required.
```
annevk, line 5821: I wonder if we should remove this. In particular because when something returns a ByteString it's not expected that would not be treated as an isomorphic string. (Isomorphic string is indeed not defined at the moment, but perhaps at some point we should add it to Infra as the equivalent to the 8-bit strings implementations have when ASCII string isn't sufficient.)

Okay to leave this for a follow-up though.

index.bs
```diff
@@ -5845,10 +5843,10 @@ can be set to [=value of string literal tokens|the value=] of a
...
 
 The {{USVString}} type
-corresponds to the set of all possible sequences of
-[=scalar values=],
-which are all of the Unicode code points apart from the
-surrogate code points.
+corresponds to [=scalar value strings=].
+Depending on the context,
+these can be treated as sequences of either 16-bit unsigned integer [=code units=]
+or [=scalar value=] [=code points=].
```
annevk, line 5849:
```suggestion
or [=scalar values=].
```
(Otherwise we're doing the near equivalent of a round circle.)

annevk (approved):
Thanks again for tackling this, really appreciate it!

index.bs
annevk: Filed #1295 on this.

PR #1293, opened 2023-04-13
Export "an exception was thrown"

caraitto:
***
<a href="https://whatpr.org/webidl/1293.html" title="Last updated on Apr 13, 2023, 7:59 PM UTC (3bdbc1e)">Preview</a> | <a href="https://whatpr.org/webidl/1293/a35410d...3bdbc1e.html" title="Last updated on Apr 13, 2023, 7:59 PM UTC (3bdbc1e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14289,7 +14289,7 @@ must propagate to the caller, and if
 not caught there, to its caller, and so on.
 
 Per [[#conventions]], an algorithm specified in this document may intercept thrown exceptions, either by specifying
-the exact steps to take if <dfn>an exception was thrown</dfn>, or by explicitly handling [=abrupt completions=].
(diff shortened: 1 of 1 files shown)


jyasskin:
This is for use in https://github.com/WICG/turtledove/pull/538.

domenic:
This was intended to be used internal to Web IDL only, which deals with some of the complexities surrounding ECMAScript vs. Web IDL exceptions. However, new specs (and indeed most of Web IDL itself, 3 instances aside) tend to just use "catch any exceptions", as explained in https://infra.spec.whatwg.org/#algorithm-control-flow .

So, I'd rather remove this term from the ecosystem, than encourage its spread.

jyasskin:
FWIW, using Infra's "catch" works for me. FYI @domfarolino. Should we `<dfn>` that "catch"?

PR #1311, opened 2023-05-30
Give SharedArrayBuffer a dedicated type

### LINKED ISSUE #865: How to return a SharedArrayBuffer object

### LINKED ISSUE #961: It is unclear if [AllowShared] BufferSource is valid
The spec says 
"A type that is not a buffer source type must not be associated with the [AllowShared] extended attribute."
and
"The buffer source types are ArrayBuffer, DataView, and the typed array types."

BufferSource is a union type
typedef (ArrayBufferView or ArrayBuffer) BufferSource;

Yet https://heycam.github.io/webidl/#AllowShared has an example using [AllowShared] BufferSource.

annevk:
Generally IDL objects have been 1:1 with ECMAScript objects, except for SharedArrayBuffer. That was instead represented as [AllowShared] ArrayBuffer, although that includes ArrayBuffer as well. This created a number of challenges, e.g., how to return a SharedArrayBuffer object.

So we make these changes here that will also require corresponding downstream fixes:

* SharedArrayBuffer is now its own type.
* [AllowShared] only applies to buffer view types.
* AllowSharedBufferSource takes over from [AllowShared] BufferSource.

Fixes #865 and fixes #961.

   * Chromium per https://github.com/whatwg/webidl/issues/961#issuecomment-892616473
   * WebKit
   * N/A
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1455098
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1838639
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=258131
   * Deno: https://github.com/denoland/deno/issues/19506
   * Node.js: https://github.com/nodejs/node/issues/48459
   * webidl2.js: https://github.com/w3c/webidl2.js/issues/746
   * widlparser: https://github.com/plinss/widlparser/issues/82

***
<a href="https://whatpr.org/webidl/1311.html" title="Last updated on Jun 15, 2023, 2:21 PM UTC (9ad215c)">Preview</a> | <a href="https://whatpr.org/webidl/1311/5dcedce...9ad215c.html" title="Last updated on Jun 15, 2023, 2:21 PM UTC (9ad215c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5465,6 +5465,8 @@ The <dfn id="dfn-string-type" export>string types</dfn> are
 {{DOMString}}, all [=enumeration types=],
 {{ByteString}} and {{USVString}}.
 
+The <dfn export>buffer types</dfn> are {{ArrayBuffer}} and {{SharedArrayBuffer}}.
+
 The <dfn id="dfn-typed-array-type" export>typed array types</dfn> are
 {{Int8Array}},
 {{Int16Array}},
@@ -5478,10 +5480,9 @@ The <dfn id="dfn-typed-array-type" export>typed array types</dfn> are
 {{Float32Array}}, and
 {{Float64Array}}.
 
-The <dfn id="dfn-buffer-source-type" export>buffer source types</dfn>
-are {{ArrayBuffer}},
-{{DataView}},
-and the [=typed array types=].
+The <dfn export>buffer view types</dfn> are {{DataView}} and the [=typed array types=].
+
+The <dfn id="dfn-buffer-source-type" export>buffer source types</dfn> are the [=buffer types=] and the [=buffer view types=].
 
 The {{object}} type,
 all [=interface types=], and
@@ -6328,27 +6329,28 @@ For any type, the [=extended attributes associated with=] it must only contain
 
 There are a number of types that correspond to sets of all possible non-null
 references to objects that represent a buffer of data or a view on to a buffer of
-data.  The table below lists these types and the kind of buffer or view they represent.
+data. The table below lists these types and the kind of buffer or view they represent.
 
 <table class="vert data">
     <tr><th>Type</th><th>Kind of buffer</th></tr>
     <tr><td><dfn id="idl-ArrayBuffer" interface>ArrayBuffer</dfn></td><td>An object that holds a pointer (which can be null) to a buffer of a fixed number of bytes</td></tr>
-    <tr><td><dfn id="idl-DataView" interface>DataView</dfn></td><td>A view on to an {{ArrayBuffer}} that allows typed access to integers and floating point values stored at arbitrary offsets into the buffer</td></tr>
+    <tr><td><dfn id="idl-SharedArrayBuffer" interface>SharedArrayBuffer</dfn></td><td>An object that holds a pointer (which cannot be null) to a shared buffer of a fixed number of bytes</td></tr>
+    <tr><td><dfn id="idl-DataView" interface>DataView</dfn></td><td>A view on to a [=buffer type=] that allows typed access to integers and floating point values stored at arbitrary offsets into the buffer</td></tr>
     <tr><td><dfn id="idl-Int8Array" interface>Int8Array</dfn>,<br/>
             <dfn id="idl-Int16Array" interface>Int16Array</dfn>,<br/>
             <dfn id="idl-Int32Array" interface>Int32Array</dfn>,<br/>
             <dfn id="idl-BigInt64Array" interface>BigInt64Array</dfn></td>
-        <td>A view on to an {{ArrayBuffer}} that exposes it as an array of two's complement signed integers of the given size in bits</td></tr>
+        <td>A view on to a [=buffer type=] that exposes it as an array of two's complement signed integers of the given size in bits</td></tr>
     <tr><td><dfn id="idl-Uint8Array" interface>Uint8Array</dfn>,<br/>
             <dfn id="idl-Uint16Array" interface>Uint16Array</dfn>,<br/>
             <dfn id="idl-Uint32Array" interface>Uint32Array</dfn>,<br/>
             <dfn id="idl-BigUint64Array" interface>BigUint64Array</dfn></td>
-        <td>A view on to an {{ArrayBuffer}} that exposes it as an array of unsigned integers of the given size in bits</td></tr>
+        <td>A view on to a [=buffer type=] that exposes it as an array of unsigned integers of the given size in bits</td></tr>
     <tr><td><dfn id="idl-Uint8ClampedArray" interface>Uint8ClampedArray</dfn></td>
-        <td>A view on to an {{ArrayBuffer}} that exposes it as an array of unsigned 8-bit integers with clamped conversions</td></tr>
+        <td>A view on to a [=buffer type=] that exposes it as an array of unsigned 8-bit integers with clamped conversions</td></tr>
     <tr><td><dfn id="idl-Float32Array" interface>Float32Array</dfn>,<br/>
             <dfn id="idl-Float64Array" interface>Float64Array</dfn></td>
-        <td>A view on to an {{ArrayBuffer}} that exposes it as an array of IEEE 754 floating point numbers of the given size in bits</td></tr>
+        <td>A view on to a [=buffer type=] that exposes it as an array of IEEE 754 floating point numbers of the given size in bits</td></tr>
 </table>
 
 Note: These types all correspond to classes defined in ECMAScript.
@@ -6362,6 +6364,7 @@ in [[#es-buffer-source-types]].
 <pre class="grammar" id="prod-BufferRelatedType">
     BufferRelatedType :
         "ArrayBuffer"
+        "SharedArrayBuffer"
         "DataView"
         "Int8Array"
         "Int16Array"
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8701,9 +8755,9 @@ a reference to the same object that the IDL value represents.
 
 <div class="warning">
     Extreme care must be taken when writing specification text that
-    [=ArrayBuffer/writes=] into an {{ArrayBuffer}} or {{ArrayBufferView}}, as the underlying data
+    [=buffer type/writes=] into an [=buffer source type=], as the underlying data
```
bathos, line 8758:
```suggestion
    [=buffer type/writes=] into a [=buffer source type=], as the underlying data
```

index.bs
```diff
@@ -6362,6 +6364,7 @@ in [[#es-buffer-source-types]].
 <pre class="grammar" id="prod-BufferRelatedType">
     BufferRelatedType :
         "ArrayBuffer"
+        "SharedArrayBuffer"
```
bathos, line 6367: This becomes a non-identifier token by virtue of appearing anywhere like this, so it needs to be added as an alternative for the `Other` production (used for the “anything goes” extended attribute cover grammar) as well.

index.bs
```diff
@@ -8646,30 +8696,33 @@ a reference to the same object that the IDL value represents.
...
+    <dfn id=buffersource-underlying-buffer export for="buffer source type">underlying buffer</dfn>
+    of a [=buffer source type=] |bufferSource| is the value returned by the following steps:
 
-    1.  If |bufferSource| is an {{ArrayBuffer}}, return |bufferSource|.
+    1.  If |bufferSource| is a [=buffer type=], then return |bufferSource|.
     1.  Let |esBufferSource| be the result of [=converted to an ECMAScript value|converting=]
         |bufferSource| to an ECMAScript value.
     1.  Return the result of [=converted to an IDL value|converting=]
-        |esBufferSource|.\[[ViewedArrayBuffer]] to an IDL value of type {{ArrayBuffer}}.
+        |esBufferSource|.\[[ViewedArrayBuffer]] to an IDL value of type [=buffer type=].
```
domenic, line 8719: This doesn't seem right. You need to switch on whether [[ViewedArrayBuffer]] IsSharedArrayBuffer() or not, and then convert appropriately.

index.bs
```diff
@@ -8718,6 +8772,8 @@ a reference to the same object that the IDL value represents.
     1.  Let |esArrayBuffer| be the result of [=converted to an ECMAScript value|converting=]
         |arrayBuffer| to an ECMAScript value.
     1.  Assert: [$IsSharedArrayBuffer$](|esArrayBuffer|) is false.
+        <!-- We'll keep this in for now as we're generally bad at distinguishing IDL and JS
+             objects. -->
```
domenic, line 8776: What's the problem? We converted in step 1.

index.bs
```diff
@@ -8728,8 +8784,9 @@ a reference to the same object that the IDL value represents.
 </div>
 
 <div algorithm>
-    A {{BufferSource}} |bufferSource| is <dfn for="BufferSource" export>detached</dfn> if the
-    following steps return true:
+    A [=buffer source types=] |bufferSource| is
```
domenic, line 8787:
```suggestion
    A [=buffer source type=] |bufferSource| is
```

index.bs
```diff
@@ -5465,6 +5465,8 @@ The <dfn id="dfn-string-type" export>string types</dfn> are
 {{DOMString}}, all [=enumeration types=],
 {{ByteString}} and {{USVString}}.
 
+The <dfn export>buffer types</dfn> are {{ArrayBuffer}} and {{SharedArrayBuffer}}.
```
domenic, line 5468: So, Web IDL is quite picky about distinguishing types. In particular, a "type" like "string type" or "buffer type" refers to something in the category of types. It does not refer to things in the category of values.

Thus, you really shouldn't say things like "If _bufferSource_ is a buffer type". Because a value can never be a type. A value can be an instance of a type, I guess. (But I don't see much use of that language for the other types defined in this section?)

For example, look at all the reference sites for https://whatpr.org/webidl/1311.html#dfn-numeric-type , and notice how they're specifically talking about types (e.g. return types, overload resolution algorithm internals) and not about values.

index.bs
```diff
@@ -8646,30 +8696,33 @@ a reference to the same object that the IDL value represents.
 </div>
 
 <div algorithm>
-    The <dfn export for="BufferSource">byte length</dfn> of a {{BufferSource}} |bufferSource| is
-    the value returned by the following steps:
+    The <dfn id=buffersource-byte-length export for="buffer source type">byte length</dfn> of a
+    [=buffer source type=] |bufferSource| is the value returned by the following steps:
```
domenic, line 8700: So for example, I think using the typedef here is more correct, because if |bufferSource| were a "buffer source type", then it would either be the type "DataView" or the type "ArrayBuffer" or the type .... It would not be any specific IDL object.

index.bs
```diff
@@ -8745,6 +8802,8 @@ a reference to the same object that the IDL value represents.
     1.  Let |esArrayBuffer| be the result of [=converted to an ECMAScript value|converting=]
         |arrayBuffer| to an ECMAScript value.
     1.  Assert: [$IsSharedArrayBuffer$](|esArrayBuffer|) is false.
+        <!-- We'll keep this in for now as we're generally bad at distinguishing IDL and JS
+             objects. -->
```
domenic, line 8806: Again this seems fine?

index.bs
```diff
@@ -8882,34 +8946,34 @@ See the rules for converting ECMAScript values to IDL [=buffer source types=] in
     <pre highlight="webidl">
```
domenic, line 8946: Maybe update the intro sentence to explain that you're actually showing all possible combinations.

(Preexisting issue.)

index.bs
```diff
@@ -14367,6 +14433,19 @@ The {{BufferSource}} typedef is used to represent objects
 that are either themselves an {{ArrayBuffer}} or which
 provide a view on to an {{ArrayBuffer}}.
 
+Note: [{{AllowShared}}] cannot be used in conjunction as {{ArrayBuffer}} does not support it.
```
domenic, line 14436:
```suggestion
Note: [{{AllowShared}}] cannot be used with {{BufferSource}} as {{ArrayBuffer}} does not support it.
```

domenic (commented):
Biggest issue is the mass conversion from using the typedefs to using the "X type" definitions, which aren't really suitable since we're talking about instances of types, not types themselves.

index.bs
```diff
@@ -8646,30 +8696,33 @@ a reference to the same object that the IDL value represents.
...
 <div algorithm>
-    The <dfn export for="BufferSource">byte length</dfn> of a {{BufferSource}} |bufferSource| is
-    the value returned by the following steps:
+    The <dfn id=buffersource-byte-length export for="buffer source type">byte length</dfn> of a
+    [=buffer source type=] |bufferSource| is the value returned by the following steps:
 
     1.  Let |esBufferSource| be the result of [=converted to an ECMAScript value|converting=]
         |bufferSource| to an ECMAScript value.
     1.  If |esBufferSource| has a \[[ViewedArrayBuffer]] internal slot, then return
         |esBufferSource|.\[[ByteLength]].
```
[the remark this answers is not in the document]
domenic: You're right, thanks!!

index.bs
annevk: Is it? The way it defines these types is in terms of objects, no? E.g., "buffer source types" references the same ArrayBuffer the typedef does. I'm not sure we can expect people to understand they're fundamentally different if the fundamental concepts are shared.

https://dontcallmedom.github.io/webdex/b.html#buffer%20source%20types%40%40webidl%25%25dfn shows there is some existing "confusion" around this.

The way I see it the types are comparable to classes and from that perspective and how we deal with those elsewhere in the web platform I think it all works.

index.bs
```diff
@@ -8621,7 +8694,7 @@ a reference to the same object that the IDL value represents.
 
 <div algorithm>
     To <dfn id="dfn-get-buffer-source-copy" export lt="get a copy of the buffer source|get a copy of the bytes held by the buffer source">get a copy of the bytes held by the buffer source</dfn>
-    given a {{BufferSource}} |bufferSource|:
+    given a [=buffer source type=] |bufferSource|:
```
domenic, line 8697:
```suggestion
    given a [=buffer source type=] instance |bufferSource|:
```

index.bs
```diff
@@ -6328,27 +6329,47 @@ For any type, the [=extended attributes associated with=] it must only contain
...
+    <tr>
+        <td><dfn id="idl-Uint32Array" interface>Uint32Array</dfn>
+    <tr>
+        <td><dfn id="idl-BigUint64Array" interface>BigUint64Array</dfn>
+    <tr>
+        <td><dfn id="idl-Uint8ClampedArray" interface>Uint8ClampedArray</dfn>
+        <td>A view on to a [=buffer type=] instance that exposes it as an array of unsigned 8-bit integers with clamped conversions
+    <tr>
+        <td><dfn id="idl-Float32Array" interface>Float32Array</dfn>
+        <td rowspan=2>A view on to a [=buffer type=] that exposes it as an array of IEEE 754 floating point numbers of the given size in bits
```
domenic, line 6370:
```suggestion
        <td rowspan=2>A view on to a [=buffer type=] instance that exposes it as an array of IEEE 754 floating point numbers of the given size in bits
```

domenic (approved):

index.bs
```diff
@@ -8646,30 +8719,35 @@ a reference to the same object that the IDL value represents.
...
+    1.  If [$IsSharedArrayBuffer$](|esBuffer|) is false, then return the result of
+        [=converted to an IDL value|converting=] |esBuffer| to an IDL value of type {{ArrayBuffer}}.
+    1.  Return the result of [=converted to an IDL value|converting=] |esBuffer| to an IDL value of
+        type {{SharedArrayBuffer}}.
 </div>
 
-<div algorithm="ArrayBuffer/write">
-    To <dfn export for="ArrayBuffer">write</dfn> a [=byte sequence=] |bytes| into an {{ArrayBuffer}}
-    |arrayBuffer|, optionally given a <dfn export for="ArrayBuffer/write">|startingOffset|</dfn>
+<div algorithm="buffer type/write">
```
annevk, line 8747:
```suggestion
<div>
```

PR #1323, opened 2023-06-20
Editorial: Handle exceptions thrown while converting 'invoke's return value.

jyasskin:
Without this, the exception skips the "clean up" steps, which seems bad. I expect everyone's already implemented this correctly.

***
<a href="https://whatpr.org/webidl/1323.html" title="Last updated on Jun 21, 2023, 7:18 PM UTC (13f6a82)">Preview</a> | <a href="https://whatpr.org/webidl/1323/95f57dd...13f6a82.html" title="Last updated on Jun 21, 2023, 7:18 PM UTC (13f6a82)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14186,7 +14186,8 @@ the special value “missing”, which represents a missing optional argument.
         |callResult| and jump to the step labeled <a href="#call-user-object-operation-return"><i>return</i></a>.
     1.  Set |completion| to the result of [=converted to an IDL value|converting=]
         |callResult|.\[[Value]] to an IDL value of the same type as the operation's
-        return type.
+        return type. If this throws an exception, set |completion| to the completion value
+        representing the thrown exception.
     1.  <i id="call-user-object-operation-return">Return:</i> at this
         point |completion| will be set to an IDL value or an [=abrupt completion=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -14186,7 +14186,8 @@ the special value “missing”, which represents a missing optional argument.
         |callResult| and jump to the step labeled <a href="#call-user-object-operation-return"><i>return</i></a>.
     1.  Set |completion| to the result of [=converted to an IDL value|converting=]
         |callResult|.\[[Value]] to an IDL value of the same type as the operation's
```
domenic, line 14188: While here, maybe fix this to not say "operation's" for all three, and instead reference _callable_'s return type where appropriate?

domenic (approved):
(Note for when we merge: bugfixes aren't editorial.)

Happy to merge this as-is or with the additional fix I suggested. Thanks for finding it.

In practice, I suspect most cases where these algorithms are invoked, do not have return types where a conversion would fail. That is, most callback function types are void or something easy to convert to (e.g. boolean). I wonder which specs, if any, have more restrictive return types in such cases, and whether they're prepared to handle such conversion failures.

index.bs
jyasskin: Done.

PR #1324, opened 2023-06-20
Make call and invoke either return an IDL value or throw.

jyasskin:
They used to return JS Completions, contrary to the claim at their definitions.

***
<a href="https://whatpr.org/webidl/1324.html" title="Last updated on Jun 20, 2023, 9:08 PM UTC (5d50915)">Preview</a> | <a href="https://whatpr.org/webidl/1324/c45a0f3...5d50915.html" title="Last updated on Jun 20, 2023, 9:08 PM UTC (5d50915)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14188,12 +14188,12 @@ the special value “missing”, which represents a missing optional argument.
         |callResult|.\[[Value]] to an IDL value of the same type as the operation's
         return type.
     1.  <i id="call-user-object-operation-return">Return:</i> at this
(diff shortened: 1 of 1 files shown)


domenic (approved):
It's not fully clear this is necessary, because of the implicit conversion between completion values and ECMAScript return values/thrown exceptions, which most of the Web IDL spec somewhat assumes.

However, I think most of the spec tries to stay rigorous about the separation between IDL values and ECMAScript values. So for that improvement alone, let's merge this.

PR #1353, opened 2023-08-16
Add distinguishable state for cb and dictionary-like

### LINKED ISSUE #1191: Allow dictionaries and callbacks to be distinguishable
https://webidl.spec.whatwg.org/#dfn-distinguishable

It isn't clear to me why callbacks and dictionaries can't be distinguished.

I'm designing an API like:

```js
whatever(callback);
```

But there's a strong possibility that options will be added in future, and I want to avoid this pattern:

```js
whatever(() => {
  // Quite
  // a
  // bit
  // of
  // code
}, { behaveDifferently: true });
```

And instead go for:

```js
whatever({
  behaveDifferently: true,
  callback() {
    // Quite
    // a
    // bit
    // of
    // code
  }
});
```

But it seems like I can't overload from `whatever(callback)` to `whatever({ callback })`.

I would have assumed that an argument could be treated as a callback if it's callable.

vmpstr:
Fixes #1191

   * Chromium
   * …
   * ...
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1353.html" title="Last updated on Oct 5, 2023, 7:21 PM UTC (c3bf764)">Preview</a> | <a href="https://whatpr.org/webidl/1353/84c4ccc...c3bf764.html" title="Last updated on Oct 5, 2023, 7:21 PM UTC (c3bf764)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3465,7 +3465,7 @@ the following algorithm returns <i>true</i>.
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td>
-            <td>
+            <td>(c)
             <td>●
         <tr>
             <th>dictionary-like
@@ -3503,6 +3503,15 @@ the following algorithm returns <i>true</i>.
             <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
             overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a dictionary-like category based on the
+            [=ECMAScript/callable=] state. When selecting a type for the
+            purposes of [=overload resolution algorithm=] or [=union type=]
+            selection, and the possible types are [=callback function=] and
+            disctionary-like category type, then the type is a [=callback
+            function=] if it is [=ECMAScript/callable=]. Otherwise, it is a
+            dictionary-like category type.
     </ol>
 
     <div class="example" id="example-distinguishability-diff-types">
--- a/index.bs
+++ b/index.bs
@@ -3505,13 +3505,13 @@ the following algorithm returns <i>true</i>.
             advice about using unions of these types</a>.
         1.  A [=callback function=] that does not have
             [{{LegacyTreatNonObjectAsNull}}] extended attribute is
-            distinguishable from a dictionary-like category based on the
-            [=ECMAScript/callable=] state. When selecting a type for the
-            purposes of [=overload resolution algorithm=] or [=union type=]
-            selection, and the possible types are [=callback function=] and
-            disctionary-like category type, then the type is a [=callback
-            function=] if it is [=ECMAScript/callable=]. Otherwise, it is a
-            dictionary-like category type.
+            distinguishable from a dictionary-like category object based on
+            whether the object is [=ECMAScript/callable=]. When selecting a
+            type for the purposes of [=overload resolution algorithm=] or
+            [=union type=] selection, and the possible types are [=callback
+            function=] and disctionary-like category type, then the type is a
+            [=callback function=] if the object is [=ECMAScript/callable=].
+            Otherwise, it is a dictionary-like category type.
     </ol>
 
     <div class="example" id="example-distinguishability-diff-types">
--- a/index.bs
+++ b/index.bs
@@ -3505,7 +3505,7 @@ the following algorithm returns <i>true</i>.
             advice about using unions of these types</a>.
         1.  A [=callback function=] that does not have
             [{{LegacyTreatNonObjectAsNull}}] extended attribute is
-            distinguishable from a dictionary-like category type.
+            distinguishable from a type in the dictionary-like category.
             <div class="note">
               For example, when converting an ECMAScript value to [=union
               type=] which includes a [=callback function=] and a
--- a/index.bs
+++ b/index.bs
@@ -3505,12 +3505,12 @@ the following algorithm returns <i>true</i>.
             advice about using unions of these types</a>.
         1.  A [=callback function=] that does not have
             [{{LegacyTreatNonObjectAsNull}}] extended attribute is
-            distinguishable from a dictionary-like category object based on
-            whether the object is [=ECMAScript/callable=]. When selecting a
+            distinguishable from a dictionary-like category type based on
+            whether the type is [=ECMAScript/callable=]. When selecting a
             type for the purposes of [=overload resolution algorithm=] or
             [=union type=] selection, and the possible types are [=callback
             function=] and disctionary-like category type, then the type is a
-            [=callback function=] if the object is [=ECMAScript/callable=].
+            [=callback function=] if the type is [=ECMAScript/callable=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3503,6 +3503,15 @@ the following algorithm returns <i>true</i>.
             <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
             overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a dictionary-like category based on the
+            [=ECMAScript/callable=] state. When selecting a type for the
```
bathos, line 3509: Minor suggestion but: I think I'd avoid the phrasing “the callable state” because “state” implies “statefulness”? ES ensures/requires that “does/doesn’t implement [[Call]]” invariantly be an “eternal characteristic” of any ES value.

bathos:
_This comment only intends to make a design decision implied by this PR explicit to other readers._

The [LTNOAN] annotation causes only _nullable_ types with the annotated callback function as their inner type to have the problematically expanded value space — and more specifically, only when used as an attribute’s type. In every other context, an [LTNOAN]-annotated callback function behaves exactly the same as any other.

This PR will leave all [LTNOAN]-annotated callback functions indistinguishable from dictionary-likes even in contexts where [LTNOAN] is otherwise ignored/meaningless.

That seems like a reasonable call:

- It’s much simpler than trying to account for all those quirky conditions
- It’s usually okay (or even desirable) to leave _legacy_ features in the dust when introducing new features

Again, this comment’s just an “FYI” thing in case other folks _do_ see a problem with this approach — as far as I can tell, it works.

index.bs
vmpstr: Good call. Done.

index.bs
```diff
@@ -3503,6 +3503,15 @@ the following algorithm returns <i>true</i>.
...
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a dictionary-like category object based on
+            whether the object is [=ECMAScript/callable=]. When selecting a
+            type for the purposes of [=overload resolution algorithm=] or
+            [=union type=] selection, and the possible types are [=callback
+            function=] and disctionary-like category type, then the type is a
+            [=callback function=] if the object is [=ECMAScript/callable=].
+            Otherwise, it is a dictionary-like category type.
```
jeremyroman, line 3514: For clarity, even though this pull request doesn't change it, the overload resolution algorithm and algorithm "convert an ECMAScript value an to IDL union type" already handle this correctly in the Callable(V) case, as I understand it?

index.bs
```diff
@@ -3503,6 +3503,15 @@ the following algorithm returns <i>true</i>.
             <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
             overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a dictionary-like category object based on
+            whether the object is [=ECMAScript/callable=]. When selecting a
```
jeremyroman, line 3509:
```suggestion
        1.  A [=callback function=] that does not have
            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
            distinguishable from a dictionary-like type. When selecting a
```

I think the second sentence clarifies that the Callable algorithm is used, but what's normatively important in this algorithm is that callback functions without LTNOAN are distinguishable from dictionary-like types. (Since this algorithm operates on WebIDL types rather than ECMAScript values, it cannot actually normatively use Callable here.) Separating them with a full stop makes that clearer to me.

As a nit, the normative sentence here should probably discuss the types rather than the objects, for the same reason.

jeremyroman (approved):
Makes sense to me (though my approval isn't significant).

index.bs
vmpstr: I understand that the ECMAScript value is the ting that can be Callable, and it seems that at least the union resolution will convert a Callable ECMAScript value to a callable type. Otherwise, it will convert it to a dictionary-like type?

I'm not sure what you mean about a full stop separating things though.

vmpstr: Ok, I've updated this to explicitly separate normative text from non-normative notes. I don't think a punctuation mark is sufficient to imply a switch from normative to non-normative text. 

Let me know if this reads better

jeremyroman: lgtm thanks

index.bs
```diff
@@ -3503,6 +3503,16 @@ the following algorithm returns <i>true</i>.
             <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
             overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a dictionary-like category type.
```
domenic, line 3508:
```suggestion
            distinguishable from a type in the dictionary-like category.
```

domenic (approved):
Looks good with a wording nit.

Can we get multi-implementer interest here somehow? In particular, making sure this doesn't break something deep in their binding system, for example.

If the API you intend to use this for has multi-implementer interest, then recruiting those people to come over here and say "yes, no problem with changing Web IDL to accommodate that API" would be perfect.

yuki3:
> Looks good with a wording nit.
> 
> Can we get multi-implementer interest here somehow? In particular, making sure this doesn't break something deep in their binding system, for example.
> 
> If the API you intend to use this for has multi-implementer interest, then recruiting those people to come over here and say "yes, no problem with changing Web IDL to accommodate that API" would be perfect.

FWIW, I've implemented [the corresponding change](https://chromium-review.googlesource.com/c/chromium/src/+/4978263) in Chromium. It's already been landed.

lukewarlow:
Fwiw I have an implementation of this change within WebKit and it's rather trivial. @keithamus will probably be the one who ends up actually upstreaming it. Idk what signal is needed here for implementer interest, potentially can just wait and mark this as 2 implementers once the code is merged upstream?

annevk:
@vmpstr would you be willing to file the implementation bugs for this and update OP afterwards?

keithamus:
I have filed https://bugs.webkit.org/show_bug.cgi?id=267533 & https://bugzilla.mozilla.org/show_bug.cgi?id=1874665 as I intend to work on these.

annevk:
That's great. We'll need bugs for the tooling-related projects as well though. Less worried about browsers as they will get here through downstream features.

vmpstr:
Sorry for the late reply. I'm happy to file implementation bugs, but I don't know which tooling-related projects you mean. Is there a canonical list?

saschanaz:
It's in the PR template, the last two ones specifically.

vmpstr:
Ah thanks, I'll file issues and update here when I've done so

PR #1359, opened 2023-09-10
Editorial: Use CreateDataPropertyOrThrow abstract op in more places

shvaikalesh:
A step like `! CreateDataProperty(...)` asserts only that `[[DefineOwnProperty]]` is an ordinary method that can't throw, but doesn't assert that property creation is a guaranteed success because `[[DefineOwnProperty]]` only returns boolean status (it is [`Object.defineProperty`](https://tc39.es/ecma262/#sec-object.defineproperty) that throws a `TypeError`).

This change replaces these steps with `! CreateDataPropertyOrThrow(...)`, asserting that property is defined successfully since all its call sites operate on newly-created objects with no non-configurable properties.

ECMA-262 has no occurrences of `! CreateDataProperty(...)`.

Also, this change replaces "Call" with "Perform", which is a new convention of ECMA-262.

***
<a href="https://whatpr.org/webidl/1359.html" title="Last updated on Sep 10, 2023, 8:27 PM UTC (4f0ccdf)">Preview</a> | <a href="https://whatpr.org/webidl/1359/bc4a416...4f0ccdf.html" title="Last updated on Sep 10, 2023, 8:27 PM UTC (4f0ccdf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7740,7 +7740,7 @@ up on the ECMAScript object are not necessarily the same as the object's propert
             1.  If |V|[|key|] [=map/exists=], then:
                 1.  Let |idlValue| be |V|[|key|].
                 1.  Let |value| be the result of [=converted to an ECMAScript value|converting=] |idlValue| to an ECMAScript value.
-                1.  Perform [=!=] <a abstract-op>CreateDataProperty</a>(|O|, |key|, |value|).
+                1.  Perform [=!=] <a abstract-op>CreateDataPropertyOrThrow</a>(|O|, |key|, |value|).
 
                 <p class="note">Recall that if |member| has a [=dictionary member/default value=],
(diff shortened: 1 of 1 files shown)


domenic:
I still see the following; should we change them too?

- https://whatpr.org/webidl/1359.html#ref-for-sec-createdataproperty
- https://whatpr.org/webidl/1359.html#ref-for-sec-createdataproperty%E2%91%A0

shvaikalesh:
@domenic Oh no, https://whatpr.org/webidl/1359.html#ref-for-sec-createdataproperty asserts on the result, so it pretty much the same thing, while for https://whatpr.org/webidl/1359.html#ref-for-sec-createdataproperty%E2%91%A0 I have a different change in mind.

domenic (approved):

PR #1379, opened 2023-12-19
Editorial: Refer to ECMA-262 as JavaScript rather than ECMAScript

### LINKED ISSUE #863: Editorial: align with HTML on calling ECMAScript JavaScript
Currently there's some confusion already about this in the specification though the overwhelming majority still calls it ECMAScript. Aligning with HTML on JavaScript would make sense to me though.

jeremyroman:
This is consistent with how the HTML Living Standard, also a a WHATWG work product, refers to it and is how most developers would be familiar with it.

Included in this change:
* a note about this convention
* renaming various algorithms and sections (and adding aliases for pre-existing IDs)
* renaming variables and other things which refer to ECMAScript in an abbreviation

Fixes #863.

***
<a href="https://whatpr.org/webidl/1379.html" title="Last updated on Dec 27, 2023, 3:53 PM UTC (0e12027)">Preview</a> | <a href="https://whatpr.org/webidl/1379/10e559d...0e12027.html" title="Last updated on Dec 27, 2023, 3:53 PM UTC (0e12027)">Diff</a>


jeremyroman:
@annevk @domenic PTAL? I'm hopeful that preserving most of the IDs as old IDs (or replicating the effect of `oldids` by using the same trick of adding empty spans, for cases where that's not supported by Bikeshed) will minimize the disruption to other specs.

index.bs
```diff
@@ -1533,7 +1535,7 @@ on which they appear.  It is language binding specific whether
         };
     </pre>
 
-    the constant value can be accessed in ECMAScript  either as
+    the constant value can be accessed in JavaScript  either as
```
annevk, line 1538:
```suggestion
    the constant value can be accessed in JavaScript either as
```

index.bs
```diff
@@ -6964,11 +6967,11 @@ to its caller, and so on.
 </div>
 
 
-<h3 id="es-environment">ECMAScript environment</h3>
+<h3 id="js-environment" oldids="es-environment">JavaScript environment</h3>
```
annevk, line 6970: I think keeping the old IDs and not adding new IDs would be fine as well. I personally prefer that as it shows the document has some history.

annevk (approved):
Thanks for tackling this @jeremyroman!

annevk:
We should "Fixes #863." to the end of the final commit message. (Seems PR Preview won the race a couple times there.)

index.bs
jeremyroman: Done.

index.bs
jeremyroman: There's some history of using `oldids` (e.g., adding `es-undefined` and adding `es-void` as an old ID, when WebIDL renamed the concept of `void`). If the consensus is that we should continue to use `es-` for IDs then I can unwind that, though.

PR #1387, opened 2024-02-07
Update ongoing promise in async iterator return() method

MattiasBuelens:
@yuki3 found [an interesting edge case](https://chromium-review.googlesource.com/c/chromium/src/+/5263418/6#message-d786cceaee26217e9a67d9201f5e10d1c0c4a34e) related to async iterators.

In Firefox and Node.js (where async iteration on `ReadableStream` is already supported), when you run this snippet:
```js
const readable = new ReadableStream({
  start(c) {
    c.enqueue("a");
    c.enqueue("b");
  },
  cancel() {
    console.warn("cancelled");
  }
});

const it = readable.values();
const p1 = it.next().then(x => console.log("next 1 resolved with", x));
const p2 = it.return().then(x => console.log("return resolved with", x));
const p3 = it.next().then(x => console.log("next 2 resolved with", x));
```
you get:
```
next 1 resolved with { value: 'a', done: false }
cancelled
next 2 resolved with { done: true, value: undefined }
return resolved with { done: true, value: undefined }
```

This is quite surprising: the second `next()` call resolves *before* the `return()` call, even though it was called *after* `return()`. Intuitively, we would expect all async iterator calls to get queued. Indeed, if you do the same thing with an async generator:
```js
async function* test() {
  try {
    yield "a";
    yield "b";
  } finally {
    console.warn("cancelled");
  }
}

const it = test();
const p1 = it.next().then(x => console.log("next 1 resolved with", x));
const p2 = it.return().then(x => console.log("return resolved with", x));
const p3 = it.next().then(x => console.log("next 2 resolved with", x));
```
you get the results in the same order as they were called:
```
next 1 resolved with { value: 'a', done: false }
cancelled
return resolved with { value: undefined, done: true }
next 2 resolved with { value: undefined, done: true }
```

Yuki and I believe this to be a bug in the Web IDL specification. More precisely, the `return()` method should update the ongoing promise, such that future calls to `next()` and `return()` are properly chained. This aligns more closely with the behavior of async generators.

---

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12794,8 +12794,6 @@ The \[[Prototype]] [=/internal slot=] of an [=asynchronous iterator prototype ob
             |interface|, given |object|'s [=default asynchronous iterator object/target=], |object|,
             and |value|.
 
-    1.  Let |returnStepsPromise| be null.
-
     1.  Let |ongoingPromise| be |object|'s [=default asynchronous iterator object/ongoing promise=].
 
     1.  If |ongoingPromise| is not null, then:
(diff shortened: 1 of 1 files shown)


domenic (approved):
LGTM! I think this is a bugfix and I hope someone from WebKit and/or Mozilla can quickly step up to approve. I take it that it has Chromium support already.

annevk (approved):
Makes sense that they should be aligned. Please file implementation bugs.

PR #1392, opened 2024-03-11
Add `{{StringContext}}` extended attribute

lukewarlow:
Add a `{{StringContext}}` extended attribute that can be defined on DOMString and USVString.

This is to hook up the Trusted Types validation during the ES->IDL type
conversion to avoid issues with its default policy.

See https://github.com/w3c/trusted-types/issues/248,
https://github.com/w3c/trusted-types/issues/176

The actual logic of the string validation calls into HTML, and will be something similar to https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context.

   * Chromium
   * Gecko
   * … 
   * Chromium: Shipping
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1508286 - General trusted types issue
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=266630 - General Trusted Types issue
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: Trusted types itself is documented and the specifics of this IDL attribute probably aren't relevant.

***
<a href="https://whatpr.org/webidl/1392.html" title="Last updated on Mar 14, 2024, 12:59 PM UTC (36ddffb)">Preview</a> | <a href="https://whatpr.org/webidl/1392/5a2802b...36ddffb.html" title="Last updated on Mar 14, 2024, 12:59 PM UTC (36ddffb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12055,7 +12055,7 @@ in which case they are exposed on every object that [=implements=] the interface
 
                 <dt>Otherwise</dt>
                 <dd>
-                    1.  If the conversion is to an IDL type [=extended attribute associated with|associated with=] the
+                    1.  If |attribute|'s type is [=extended attribute associated with|associated with=] the
                         [{{StringContext}}] extended attribute, then set |V| to the result of [=validate the string in context=], passing
                         [=this=], |V|, the {{StringContext}} extended attribute [=identifier=], and the [=identifier=]
                         of the [{{StringContext}}] extended attribute [=related construct=].
--- a/index.bs
+++ b/index.bs
@@ -10268,7 +10268,7 @@ modifies how the value is converted to the IDL type, causing additional value va
 adhere to the context the string is used in.
 
 The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
-must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>" and "<code>TrustedScriptURL</code>".
+must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>", and "<code>TrustedScriptURL</code>".
 
 [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
 a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
--- a/index.bs
+++ b/index.bs
@@ -7588,7 +7588,7 @@ value when its bit pattern is interpreted as an unsigned 64-bit integer.
         [=this=], |V|, the {{StringContext}} extended attribute [=identifier=], and the [=identifier=]
         of the [{{StringContext}}] extended attribute [=related construct=].
 
-        Note: That algorithm may [=ECMAScript/throw=] a {{ECMAScript/TypeError}}.
+        Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
 
     1.  If |V| is <emu-val>null</emu-val> and the conversion is to an IDL type
         [=extended attribute associated with|associated with=] the [{{LegacyNullToEmptyString}}] extended
@@ -10286,10 +10286,10 @@ attribute appears in is its <dfn>related construct</dfn>.
 A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
 the [{{StringContext}}] extended attribute.
 
-See the rules for converting ECMAScript values to the IDL types in [[#es-DOMString]] and [[#es-USVString]]
+See the rules for converting ECMAScript values to the IDL types in [[#js-DOMString]] and [[#js-USVString]]
 for the specific requirements that the use of [{{StringContext}}] entails.
 
-<div class="example">
+<div class="example" id="example-7b5c2e9f">
 
     In the following [=IDL fragment=],
     a [=variadic=] [=operation=] is declared
@@ -11100,7 +11100,7 @@ allowed.  The security check takes the following three inputs:
 
 Note: The HTML Standard defines how a security check is performed. [[!HTML]]
 
-Certain algorithms in [[#es-type-mapping]] are defined to
+Certain algorithms in [[#js-type-mapping]] are defined to
 <dfn id="dfn-validate-the-string-in-context" export>validate the string in context</dfn> on a given
 value. This check is used to determine whether a given value
 is appropriate for its {{StringContext}}. This validation takes the following four inputs:
@@ -11111,7 +11111,7 @@ is appropriate for its {{StringContext}}. This validation takes the following fo
 1.  the {{StringContext}} [=identifier=], and
 1.  the [=identifier=] of the operation or attribute.
 
-The algorithm returns an ECMAScript String value, or [=ECMAScript/throws=] a {{ECMAScript/TypeError}}.
+The algorithm returns an ECMAScript String value, or [=JavaScript/throws=] a <l spec=ecmascript>{{TypeError}}</l>.
 
 Note: The HTML Standard defines how the validation is performed. [[!HTML]]
 
--- a/index.bs
+++ b/index.bs
@@ -7583,13 +7583,6 @@ value when its bit pattern is interpreted as an unsigned 64-bit integer.
     A JavaScript value |V| is [=converted to an IDL value|converted=]
     to an IDL {{DOMString}} value by running the following algorithm:
 
-    1.  If the conversion is to an IDL type [=extended attribute associated with|associated with=] the
-        [{{StringContext}}] extended attribute, then set |V| to the result of [=validate the string in context=], passing
-        [=this=], |V|, the {{StringContext}} extended attribute [=identifier=], and the [=identifier=]
-        of the [{{StringContext}}] extended attribute [=related construct=].
-
-        Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
-
     1.  If |V| is <emu-val>null</emu-val> and the conversion is to an IDL type
         [=extended attribute associated with|associated with=] the [{{LegacyNullToEmptyString}}] extended
         attribute, then return the {{DOMString}} value that represents the empty string.
--- a/index.bs
+++ b/index.bs
@@ -10270,7 +10270,7 @@ adhere to the context the string is used in.
 The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
 must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>", and "<code>TrustedScriptURL</code>".
 
-[{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
+The [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
 a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
 extended attribute must not appear in a [=read only=] attribute.
 
@@ -11146,7 +11146,7 @@ Note: The HTML Standard defines how the validation is performed. [[!HTML]]
                 extended attribute [=identifier=], and the [=identifier=]
                 of the [=operation=] or [=extended attribute=] of the first entry in |S|.
 
-                Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+                Note: That algorithm can [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
             1.  Append to |values| the result of [=converted to an IDL value|converting=]
                 |V| to IDL type |type|.
         1.  Set |i| to |i| + 1.
@@ -11354,7 +11354,7 @@ Note: The HTML Standard defines how the validation is performed. [[!HTML]]
                 extended attribute [=identifier=], and the [=identifier=]
                 of |callable|.
 
-                Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+                Note: That algorithm can [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
             1.  Append to |values| the result of [=converted to an IDL value|converting=]
                 |V| to IDL type |type|.
         1.  Set |i| to |i| + 1.
@@ -12050,7 +12050,7 @@ in which case they are exposed on every object that [=implements=] the interface
                         [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
                         extended attribute [=identifier=], and |id|.
 
-                        Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+                        Note: That algorithm can [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
                     1.  |idlValue| is the result of [=converted to an IDL value|converting=] |V| to an
                         IDL value of |attribute|'s type.
                 </dd>
--- a/index.bs
+++ b/index.bs
@@ -10277,6 +10277,9 @@ extended attribute must not appear in a [=read only=] attribute.
 A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
 the [{{StringContext}}] extended attribute.
 
+The [{{StringContext}}] extended attribute must not be used on only some of the [=overloaded|overloads=] of an
+[=overloaded=] [=operation=].
+
 <div class="example" id="example-7b5c2e9f">
 
     In the following [=IDL fragment=],
--- a/index.bs
+++ b/index.bs
@@ -6354,7 +6354,7 @@ The following extended attributes are <dfn for="extended attributes">applicable 
 [{{AllowShared}}],
 [{{Clamp}}],
 [{{EnforceRange}}],
-[{{LegacyNullToEmptyString}}] and
+[{{LegacyNullToEmptyString}}], and
 [{{StringContext}}].
 
 
@@ -10280,7 +10280,7 @@ must be one of "<code>html</code>", "<code>script-url</code>" and "<code>script<
 [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
 a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
 extended attribute must not appear in a [=read only=] attribute. The [=regular attribute=] or
-a [=regular operation=] argument that the type annotated with the [{{StringContext}}] extended
+[=regular operation=] argument that the type annotated with the [{{StringContext}}] extended
 attribute appears in is its <dfn>related construct</dfn>.
 
 A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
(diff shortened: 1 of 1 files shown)


lukewarlow:
Fork of #841 will address review comments here

index.bs
```diff
@@ -11092,8 +11141,16 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
+        1. Otherwise:
+            1.  If |type| is an IDL type [=extended attribute associated with|associated with=] the
+                [{{StringContext}}] extended attribute, then set |V| to the result of performing
+                [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
+                extended attribute [=identifier=], and the [=identifier=]
+                of the [{{StringContext}}] extended attribute [=related construct=].
+
+                Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+            1.  Append to |values| the result of [=converted to an IDL value|converting=]
+                |V| to IDL type |type|.
```
annevk, line 11153: I don't understand how this works. The overload resolution algorithm is an algorithm for finding the relevant operation, but somehow this assumes the relevant operation is already known because it's somehow magically tied to the StringContext attribute (despite nothing making it so).

I recommend attempting without relying on "related construct" being a thing. At least it seems to me that is what an implementation would have to do as well. Or perhaps they are taking shortcuts and it stops working the moment you have overloaded methods that each take StringContext?

lukewarlow: > I don't understand how this works. The overload resolution algorithm is an algorithm for finding the relevant operation, but somehow this assumes the relevant operation is already known because it's somehow magically tied to the StringContext attribute (despite nothing making it so).

I don't fully understand what you mean by this?

If the type has a StringContext we know it must be either a regular attribute or regular operation, so in this case a regular operation, which is the related construct here.

Are you saying not to rely on related construct "magic"?

As for implementation, WebKit doesn't fully correlate to these algorithms because of how it's code generator works, so I have the identifier context as a given.

Ms2ger: If the type has a StringContext we know it must be the type of an argument to a regular operation. We don't know *which* operation, though. Conceptually, a type doesn't have a pointer back to where it is used. I think you want to pass in additional information from `#dfn-create-operation-function`.

lukewarlow: I've removed related construct, I realised we had the identifier information in all 3 cases

index.bs
```diff
@@ -11092,8 +11139,16 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
                 then append to |values| that default value.
             1.  Otherwise, append to |values| the special value “missing”.
-        1.  Otherwise, append to |values| the result of [=converted to an IDL value|converting=]
-            |V| to IDL type |type|.
+        1. Otherwise:
+            1.  If |type| is an IDL type [=extended attribute associated with|associated with=] the
+                [{{StringContext}}] extended attribute, then set |V| to the result of performing
+                [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
+                extended attribute [=identifier=], and the [=identifier=]
+                of the [=operation=] or [=extended attribute=] of the first entry in |S|.
```
lukewarlow, line 11147: Because we only need the identifier which is the same for all overloads we don't actually need to worry about which entry in the overload set is used.

lukewarlow: Yeah based on my understanding you can't really have overloads. Like where you need the type to be different in a sub class we just switch back to using a union type and handling it manually rather than using this attribute. cc @koto in case I'm wrong.

Happy to add a note or something to make it clearer, not really sure what it should say though?

yuki3:
> I think this looks reasonable, although I think we probably should note the overload limitations more explicitly.
> 
> @EdgarChen @petervanderbeken @yuki3 any of you want to take a look as well?

Looks good to me, too.

index.bs
annevk: I guess I would add a paragraph to where StringContext is defined that outlines the limitations.

index.bs
```diff
@@ -10259,6 +10261,36 @@ that does specify [{{SecureContext}}].
...
+<h4 id="StringContext" extended-attribute lt="StringContext">[StringContext]</h4>
+
+If the [{{StringContext}}] [=extended attribute=] appears on {{DOMString}} or {{USVString}}, it
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>", and "<code>TrustedScriptURL</code>".
+
+[{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
```
Ms2ger, line 10273:
```suggestion
The [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
```

index.bs
```diff
@@ -11292,8 +11347,16 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
-        1.  Otherwise, append to |values| the result of
-            [=converted to an IDL value|converting=] |V| to IDL type |type|.
+        1. Otherwise:
+            1.  If |type| is an IDL type [=extended attribute associated with|associated with=] the
+                [{{StringContext}}] extended attribute, then set |V| to the result of performing
+                [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
+                extended attribute [=identifier=], and the [=identifier=]
+                of |callable|.
+
+                Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
```
Ms2ger, line 11357:
```suggestion
                Note: That algorithm can [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
```

index.bs
```diff
@@ -11092,8 +11139,16 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
-        1.  Otherwise, append to |values| the result of [=converted to an IDL value|converting=]
-            |V| to IDL type |type|.
+        1. Otherwise:
+            1.  If |type| is an IDL type [=extended attribute associated with|associated with=] the
+                [{{StringContext}}] extended attribute, then set |V| to the result of performing
+                [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
+                extended attribute [=identifier=], and the [=identifier=]
+                of the [=operation=] or [=extended attribute=] of the first entry in |S|.
+
+                Note: That algorithm may [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
```
Ms2ger, line 11149:
```suggestion
                Note: That algorithm can [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
```

index.bs
lukewarlow: I've added a paragraph hopefully it makes sense. Happy to adjust it if you'd like.

yuki3:
cc @natechapin as he's taking over the Blink-V8 bindings work (including interop work) from me.

index.bs
```diff
@@ -11092,8 +11142,16 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
                 then append to |values| that default value.
             1.  Otherwise, append to |values| the special value “missing”.
-        1.  Otherwise, append to |values| the result of [=converted to an IDL value|converting=]
-            |V| to IDL type |type|.
+        1. Otherwise:
+            1.  If |type| is an IDL type [=extended attribute associated with|associated with=] the
+                [{{StringContext}}] extended attribute, then set |V| to the result of performing
+                [=validate the string in context=], passing [=this=], |V|, the {{StringContext}}
+                extended attribute [=identifier=], and the [=identifier=]
+                of the [=operation=] or [=extended attribute=] of the first entry in |S|.
```
mbrodesser-Igalia, line 11150: "and the identifier of the operation or extended attribute of the first entry in S" is confusing. Is the operation's identifier or the extended attribute's identifier passed to "validate the string in context"?

mbrodesser-Igalia: It's either other (https://whatpr.org/webidl/1392/5a2802b...36ddffb.html#dfn-effective-overload-set).

mbrodesser-Igalia: Consider adding a note or feel free to resolve this 'conversation'.

index.bs
```diff
@@ -10259,6 +10261,39 @@ that does specify [{{SecureContext}}].
...
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>", and "<code>TrustedScriptURL</code>".
+
+The [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
+a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
+extended attribute must not appear in a [=read only=] attribute.
+
+A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
```
mbrodesser-Igalia, line 10277: In the future nullable types might require support to. At least chromium doesn't need them now.

lukewarlow: https://w3c.github.io/trusted-types/dist/spec/#setting-slot-values - does require use of `ScriptString?` so this might need updating to account for it.

index.bs
```diff
@@ -10259,6 +10261,39 @@ that does specify [{{SecureContext}}].
...
 </div>
 
+<h4 id="StringContext" extended-attribute lt="StringContext">[StringContext]</h4>
+
+If the [{{StringContext}}] [=extended attribute=] appears on {{DOMString}} or {{USVString}}, it
+modifies how the value is converted to the IDL type, causing additional value validation to
+adhere to the context the string is used in.
+
+The [{{StringContext}}] extended attribute must [=takes an identifier|take an identifier=]. The [=identifier=]
+must be one of "<code>TrustedHTML</code>", "<code>TrustedScript</code>", and "<code>TrustedScriptURL</code>".
```
annevk, line 10271: or?

index.bs
```diff
@@ -11056,6 +11091,21 @@ allowed.  The security check takes the following three inputs:
...
+
+1.  the [=platform object=] on
+    which the operation invocation or attribute access is being done,
+1.  the value to validate,
+1.  the {{StringContext}} [=identifier=], and
+1.  the [=identifier=] of the operation or attribute.
+
+The algorithm returns an ECMAScript value, or [=JavaScript/throws=] a <l spec=ecmascript>{{TypeError}}</l>.
+
+Note: The HTML Standard defines how the validation is performed. [[!HTML]]
```
annevk, line 11107: I looked into this a bit more and this description appears insufficient.

1. It's not clear it always ends up throwing a TypeError. Doesn't this involve userland for some bits?
2. It seems this algorithm might also end up reporting violations per step 6.1 of https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-compliant-string.

I don't think "validate" is a sufficiently clear term given the implications. "Validate" makes me expect a side-effect-free-boolean-returning operation, not something that can result in network traffic (and events?).

---

Apart from the above, we'll need some test coverage to ensure that this is invoked at the correct time relative to other exceptions that can be thrown. This will be observable if this can throw non-TypeError exceptions, but even if this always throws a TypeError (and we ignore the message field) it's observable through network traffic.

annevk (commented):
I think this is mostly good now, but on closer inspection I think the "validate" aspect needs a bit more flushing out.

index.bs
lukewarlow: Good catch!

index.bs
lukewarlow: > It's not clear it always ends up throwing a TypeError. Doesn't this involve userland for some bits?

Yeah you're correct this can throw any error type.

> It seems this algorithm might also end up reporting violations per step 6.1

Also correct it can report CSP violations.

> I don't think "validate" is a sufficiently clear term given the implications. "Validate" makes me expect a side-effect-free-boolean-returning operation, not something that can result in network traffic (and events?).

Maybe verify instead? I'm not sure what prior art there is for this sort of thing.

> Apart from the above, we'll need some test coverage to ensure that this is invoked at the correct time relative to other exceptions that can be thrown. This will be observable if this can throw non-TypeError exceptions, but even if this always throws a TypeError (and we ignore the message field) it's observable through network traffic.

Additional tests to ensure timing wrt to other errors makes sense here. cc @mbrodesser-Igalia 

As for the network traffic aspect, I'm not familiar with the details of how CSP violation reports work and whether their timing is an observable concern.

annevk: The string can also be modified, right? It says it returns an ECMAScript value, but doesn't it really return a string?

I think "verify" is okay. I guess in theory we want this to be able to be used for non-Trusted Types purposes.

There's various things that matter with respect to timing:

1. Does it all happen at the correct point in time relative to other exceptions that may be thrown. I.e., do implementations handle the checks in the correct order.
2. What further side effects may arise from this check happening. It seems that https://w3c.github.io/webappsec-csp/#report-violation always queues a task so we don't have to worry about that for CSP. And I think we don't have to worry in the general case because this happens before any specification algorithms are involved. Although if we add more of these utilities around the edges of IDL we'll run into conflicts at some point of course.

Ms2ger: > doesn't it really return a string?

Doesn't it return the argument directly in step 2?

annevk: It overwrites it first in 1.2 as far as I can tell. At least judging from https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context.

lukewarlow: > If platformObject’s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) has a [trusted type policy factory](https://w3c.github.io/trusted-types/dist/spec/#window-trusted-type-policy-factory):

In the case where this is false it would return the raw value not a string.

index.bs
```diff
@@ -10259,6 +10261,39 @@ that does specify [{{SecureContext}}].
...
+
+The [{{StringContext}}] extended attribute may only annotate a type of a [=regular attribute=] or
+a [=regular operation=] argument. A type annotated with the [{{StringContext}}]
+extended attribute must not appear in a [=read only=] attribute.
+
+A type that is not {{DOMString}} or {{USVString}} must not be [=extended attributes associated with|associated with=]
+the [{{StringContext}}] extended attribute.
+
+The [{{StringContext}}] extended attribute must not be used on only some of the [=overloaded|overloads=] of an
+[=overloaded=] [=operation=].
```
petervanderbeken, line 10281: I find this sentence a bit weird, because it's used on arguments, not on the operation itself. Does this mean that in the case of overloads, either all `DOMString`/`USVString` arguments need to be annotated or none? Or just that if the overloads have a `DOMString`/`USVString` argument in the same position it should be annotated the same way in all overloads?

index.bs
lukewarlow: I've updated the TT spec to just use a union type there rather than the extended attribute, because that solves dealing with nullable types.

index.bs
annevk: Perhaps it makes more sense to return null in that case. Having something return an Infra string, an ES value (that's always the input value), or throw is a tad confusing.

lukewarlow: That's probably fine, I wonder if this is a path that would realistically be hit, I think the factory would always exist in the codepaths where this attribute is used, as it's on workers and windows.

Only hesitation is returning null could be destructive. If it's not a path you're likely to ever hit then fine but if you are in a place where this could happen in the "happy path" that might not be ideal?

petervanderbeken: https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-compliant-string returns either a string (various steps stringify), `null` or `undefined`.

index.bs
lukewarlow: Off the top of my head I *think* it's that if they're in the same position they have to always or never be annotated. It might also be the case that you can't have an overload where it's an annotated string in one branch, and a different type entirely in another.

The attribute has a side effect and that means these limitations have to be in place.

I'll be honest I'm not super familar with the IDL overloading so I might be missing the point entirely.

@annevk can probably explain it better.

Happy to reword this paragraph to make more sense :)

petervanderbeken:
I do think that we need to make `[StringContext]`-annotated strings not distinguishable from `object`, `interface-like`, `callback function`, `dictionary-like` and `sequence-like`. So we probably need to add another letter in the table, and use it in the `string types` row to explain that these are only distinguishable if the string type doesn't have a `[StringContext]` extended attribute.

index.bs
annevk: If you don't expect the code path to be used then you need to use Assert.

But given @petervanderbeken's comments I'm starting to wonder if having this as an extended attribute is a good after all. I guess I'm back at https://github.com/whatwg/webidl/pull/841#issuecomment-590988522 where I wonder if we should put most of the complexity in the specification algorithms.

cc @domenic @koto

lukewarlow: Also cc @mbrodesser-Igalia as it will impact his implementation in Gecko

lukewarlow: After discussion in the matrix I'm on board with the idea of just using union types and updating algorithms, and closing this PR out

lukewarlow: @koto or @otherdaniel do you forsee chrome having any issues with this proposed change? It'd be good to get the specs and webkit implementation.

Tldr drop the IDL extended attribute and just use unions everywhere we need enforcement (updating the call site directly).

index.bs
mbrodesser-Igalia: > https://w3c.github.io/trusted-types/dist/spec/#setting-slot-values - does require use of `ScriptString?` so this might need updating to account for it.

Btw, Chromium doesn't reflect the spec here (https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/html_script_element.idl;l=24-52;drc=d1cf902be6fc9ae8654fe5a6c466dfb51f782197).

lukewarlow: Yeah that section specifically is going through a rewrite, Chrome doesn't match it currently and won't in its new form either so that is one change to the chromium implementation that will be needed. (Very minor web facing changes that shouldn't impact anyone)

mbrodesser-Igalia: > I've updated the TT spec to just use a union type there rather than the extended attribute, because that solves dealing with nullable types.

Where was the TT spec updated? This patch still adds `StringContext` extended attribute.

lukewarlow: https://github.com/w3c/trusted-types/pull/484 - in this PR the textContent one specifically was updated to use a union. I've done a follow up commit that changes them all to unions (which I can easily revert if we decide against that)

index.bs
lukewarlow: If it helps the discussion I've got a draft commit to update the WebKIt implementation to use union types instead of IDL magic. This doesn't account for DOM timers (webkit impl is different to spec). But otherwise replaces all existing usages of IDL magic.

https://github.com/webkit/webkit/compare/lukewarlow:trusted-types-use-union-over-attribute

lukewarlow:
Based on discussions here, in Matrix and a video call we had with Koto and Daniel I'm going to go ahead and close this PR. If Daniel discovers a reason why the observable timing differences between IDL and call site aren't web compatible we can revisit this.

PR #1397, opened 2024-03-25
Add `async iterable<T>` type to WebIDL

lucacasonato:
This commit lifts the async iterator processing logic from the
`ReadableStreamFromIterable` operation in the Streams spec, to a new WebIDL type
`async iterable<T>`.

This will clean up the streams spec, and enable a change in the Fetch spec to
directly allow taking async iterable as request/response bodies, without users
having to pass them to `ReadableStream.from` first.

   * [WebKit](https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$qSV6av8uKIJ18hBHKYT0Qw4R1g1U4YKtEkdM2O5aVEM?via=matrix.org&via=mozilla.org&via=igalia.com)
   * [Blink](https://github.com/whatwg/streams/pull/1310#issuecomment-2232217685)
   * [Gecko](https://github.com/whatwg/streams/pull/1310#issuecomment-2232899304)
   * I don't know? https://github.com/whatwg/webidl/pull/1397#issuecomment-2265330623
   * Chromium: https://issues.chromium.org/issues/356891478
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1911295
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=277548
   * Deno: https://github.com/denoland/deno/issues/24849
   * Node.js: https://github.com/nodejs/node/issues/54175
   * webidl2.js: https://github.com/w3c/webidl2.js/pull/775
   * widlparser: https://github.com/plinss/widlparser/pull/85

***
<a href="https://whatpr.org/webidl/1397.html" title="Last updated on Mar 12, 2025, 7:30 AM UTC (60c6648)">Preview</a> | <a href="https://whatpr.org/webidl/1397/acbbd5f...60c6648.html" title="Last updated on Mar 12, 2025, 7:30 AM UTC (60c6648)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3467,7 +3467,7 @@ the following algorithm returns <i>true</i>.
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td>
-            <td>
+            <td>(c)
             <td>●
         <tr>
             <th>dictionary-like
@@ -3505,6 +3505,16 @@ the following algorithm returns <i>true</i>.
             <a href="#limit-bigint-numeric-overloading">a separate restriction on their use in
             overloading</a> below. Please also note <a href="#limit-bigint-numeric-unions">the
             advice about using unions of these types</a>.
+        1.  A [=callback function=] that does not have
+            [{{LegacyTreatNonObjectAsNull}}] extended attribute is
+            distinguishable from a type in the dictionary-like category.
+            <div class="note">
+              For example, when converting an ECMAScript value to [=union
+              type=] which includes a [=callback function=] and a
+              dictionary-like type, if the value is [=ECMAScript/callable=],
+              then it is converted to a [=callback function=]. Otherwise, it is
+              converted to a dictionary-like type.
+            </div>
     </ol>
 
     <div class="example" id="example-distinguishability-diff-types">
--- a/index.bs
+++ b/index.bs
@@ -6212,7 +6212,13 @@ only [=list/items=] that are of type |T|.
 <h4 id="idl-async-iterable-type" lt="async iterable" dfn export>Async iterable types — async iterable&lt;|T|&gt;</h4>
 
 The <dfn lt="async iterable type" export>async iterable&lt;|T|&gt;</dfn> type is a parameterized
-type whose values are references to objects that can produce an [=async iterator=].
+type whose values are references to objects that can produce an asynchronously iterable, possibly infinite,
+sequence of values of type |T|.
+
+Unlike [=sequences=], which are fixed-length lists where all values are known in advance, the asynchronously
+iterable sequences created by async iterables are lazy. Their values may be produced asynchronously
+only during iteration, and thus the values or length might not be known at the time the async
+iterable is created.
 
 Async iterables are passed by reference in language bindings where they are represented by an object.
 This means that passing an async iterable to a [=platform object=] will result in a reference to the
@@ -6220,29 +6226,15 @@ async iterable being kept by that object. Similarly, any async iterable returned
 object will be a reference to the same object and modifications made to it will be visible to the
 platform object. This is in contrast to sequences, which are always passed by value.
 
-Note: Async iterables can not be constructed from IDL. If returned from an operation, or used as the
+Note: Async iterables cannot be constructed from IDL. If returned from an operation, or used as the  
 type of a dictionary member, the async iterable will have originated from the host environment and
 have been turned into an IDL type via a language binding. Instead of returning an async iterable
-from an IDL operation, the operation may want to return an [=interface=] that has an
+from an IDL operation, the operation might want to return an [=interface=] that has an  
 [=asynchronously iterable declaration=].
 
 Async iterables must not be used as the type of an [=attribute=] or [=constant=].
 
 There is no way to represent an async iterable value in IDL.
-
-<h5 id="idl-async-iterator-object" lt="async-iterator" dfn export>Async iterator objects</h5>
-
-An <dfn lt="async iterator" export>async iterator&lt;|T|&gt;</dfn> value is a reference to an object
-that can produce a sequence of values asynchronously. The async iterator value is parameterized by a
-type |T|, which defines the types of values produced by this async iterator.
-
-Async iterators, unlike sequences, do not have a fixed length and can be infinite. Values are
-asynchronously produced as the async iterator is iterated over.
-
-Async iterators are values, not types, and thus cannot be used as a type in IDL. They can only be
-created from [=async iterable types=]. Thus, async iterators can not be used as the type of an IDL
-operation parameter, an IDL interface attribute, or an IDL dictionary member. Instead an
-[=async iterable type=] should be used.
 
 <h4 id="idl-record" lt="record" dfn export>Record types — record&lt;|K|, |V|&gt;</h4>
 
@@ -8183,13 +8175,17 @@ JavaScript Array values.
 
 <h4 id="js-async-iterable">Async iterable — async iterable&lt;|T|&gt;</h4>
 
-IDL <a lt="async iterable type">async iterable&lt;|T|&gt;</a> values are represented by a JavaScript
-object, a JavaScript method, and a flag indicating whether the method is expected to produce a
-<code>sync</code> or <code>async</code> iterator.
+In the JavaScript binding, IDL [=async iterable type|async iterable=] values are represented by
+a [=struct=] with the following [=struct/items=]:
+
+* <dfn for="JS async iterable">object</dfn>, a JavaScript value
+* <dfn for="JS async iterable">method</dfn>, a JavaScript value
+* <dfn for="JS async iterable">type</dfn>, either "<code>sync</code>" or "<code>async</code>"
+
 
 <div id="js-to-async-iterable" algorithm="convert a JavaScript value to async iterable">
     A JavaScript value |V| is [=converted to an IDL value|converted=]
-    to an IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
+    to an IDL <a lt="async iterable">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
 
     1.  If <a abstract-op>Type</a>(|V|) is not Object,
         [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
@@ -8197,43 +8193,47 @@ object, a JavaScript method, and a flag indicating whether the method is expecte
     1.  If |method| is undefined:
         1.  Set |syncMethod| to [=?=] <a abstract-op>GetMethod</a>(obj, {{%Symbol.iterator%}}).
         1.  If |syncMethod| is undefined, [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
-        1.  Return the IDL async iterable value that represents a reference to the JavaScript object
-            |V|, the JavaScript method |syncMethod|, and the flag <code>sync</code>.
-    1.  Return the IDL async iterable value that represents a reference to the JavaScript object
-        |V|, the JavaScript method |method|, and the flag <code>async</code>.
+        1.  Return an IDL [=async iterable=] value with [=JS async iterable/object=] set to |V|,
+            [=JS async iterable/method=] set to |syncMethod|, and [=JS async iterable/type=] set to
+            "<code>sync</code>".
+    1.  Return an IDL [=async iterable=] value with [=JS async iterable/object=] set to |V|,
+        [=JS async iterable/method=] set to |method|, and [=JS async iterable/type=] set to
+        "<code>async</code>".
 </div>
 
 <div id="async-iterable-to-js" algorithm="convert an async iterable to a JavaScript value">
-    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value is
+    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value |V| is
     [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
 
-    1.  Return the JavaScript object that represents the same async iterable as the IDL value.
+    1.  Return |V|'s [=JS async iterable/object=].
 </div>
 
 <h5 id="js-async-iterator">Async iterators</h5>
 
-IDL [=async iterator=] values are represented by JavaScript [=Iterator=] records.
+<dfn lt="async iterator" export>Async iterator</dfn> values are represented by a [=struct=] with the following [=struct/items=]:
 
-Async iterators can only be created from an [=async iterable type=].
+* <dfn for="JS async iterator">underlying record</dfn>, an [=Iterator=] record
+* <dfn for="JS async iterator">type parameter</dfn>, a IDL type representing the type of values produced by the async iterator
 
-<h5 id="js-async-iterator-iteration">Iterating async iterator</h5>
+An [=async iterable=] is not directly iterated over. Instead, it is first opened to create
+a [=async iterator=]. The [=async iterator=] can be asynchronously iterated over to produce values.
+
+<h5 id="js-async-iterator-iteration">Iterating async iterators</h5>
 
 [=Async iterables=] are not directly iterated over. They are first opened, to create a new
 [=async iterator=], and then the [=async iterator=] is iterated over.
 
 <div algorithm>
 
-    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
-    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|,
-    perform the following steps:
+    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an  
+    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|:  
 
-    1.  Let |object| be the |iterable|'s JavaScript object.
-    1.  Let |method| be the |iterable|'s JavaScript method.
-    1.  Let |iterator| be [=?=] <a abstract-op>GetIteratorFromMethod</a>(|object|, |method|).
-    1.  If |iterable| has a flag that shows that it is expected to produce a
-        <code>sync</code> iterator, then:
-        1. Set |iterator| to <a abstract-op>CreateAsyncFromSyncIterator</a>(|iterator|).
-    1.  Return an IDL [=async iterator=] value with type parameter |T| that represents |iterator|.
+    1.  Let |iterator| be [=?=] <a abstract-op>GetIteratorFromMethod</a>(|iterable|'s
+        [=JS async iterable/object=], |iterable|'s [=JS async iterable/method=]).
+    1.  If |iterable|'s [=JS async iterable/type=] is "<code>sync</code>", set |iterator| to
+        <a abstract-op>CreateAsyncFromSyncIterator</a>(|iterator|).
+    1.  Return a [=async iterator=] value with [=JS async iterator/underlying record=] set to |iterator| and
+        [=JS async iterator/type parameter=] set to |T|.
 
 </div>
 
@@ -8242,11 +8242,10 @@ Async iterators can only be created from an [=async iterable type=].
 <div algorithm>
 
     To <dfn id="async-iterator-get-next-value" export lt="get an async iterator next value">get the next value</dfn> of an
-    <code><a lt="async iterator">async iterator&lt;<var>T</var>&gt;</a></code> |iterator|,
-    perform the following steps:
+    <code><a lt="async iterator">async iterator</a></code> |iterator|:
 
-    1.  Let |iteratorRecord| be the [=Iterator=] represented by |iterator|.
-    1.  Let |nextResult| be the result of calling <a abstract-op>IteratorNext</a>(|iteratorRecord|).
+    1.  Let |nextResult| be the result of calling
+        <a abstract-op>IteratorNext</a>(|iterator|'s [=JS async iterator/underlying record=]).
     1.  If |nextResult| is an abrupt completion, return [=a promise rejected with=]
         |nextResult|.\[[Value]].
     1.  Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
@@ -8260,7 +8259,7 @@ Async iterators can only be created from an [=async iterable type=].
         1.  Otherwise:
             1.  Let |V| be [=?=] <a abstract-op>IteratorValue</a>(|iterResult|).
             1.  Let |value| be the result of [=converted to an IDL value|converting=] |V| to an IDL
-                value of type |T|.
+                value of type |iterator|'s [=JS async iterator/type parameter=].
             1.  Return |value|.
 
 </div>
@@ -8271,13 +8270,13 @@ Async iterators can only be created from an [=async iterable type=].
     <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|,
     with a reason |reason|, perform the following steps:
 
-    1.  Let |iteratorRecord| be the [=Iterator=] represented by |iterator|.
-    1.  Let |iteratorObj| be |iteratorRecord|.\[[Iterator]].
-    1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iteratorObj|, "return").
+    1.  Let |iteratorRecord| be |iterator|'s [=JS async iterator/underlying record=].
+    1.  Let |iteratorObj| be |iteratorRecord|.\[[Iterator]]
+    1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iteratorObj|, "<code>return</code>"). 
     1.  If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
         |returnMethod|.\[[Value]].
     1.  If |returnMethod| is <emu-val>undefined</emu-val>, return [=a promise resolved with=]
-        <emu-val>undefined</emu-val>.
+        {{undefined}}.
     1.  Let |returnResult| be <a abstract-op>Call</a>(|returnMethod|.\[[Value]], |iteratorObj|, « |reason| »).
     1.  If |returnResult| is an abrupt completion, return [=a promise rejected with=]
         |returnResult|.\[[Value]].
@@ -8286,7 +8285,7 @@ Async iterators can only be created from an [=async iterable type=].
         given |returnPromiseResult|:
         1.  If <a abstract-op>Type</a>(|returnPromiseResult|) is not Object, [=JavaScript/throw=] a
             <l spec=ecmascript>{{TypeError}}</l>.
-        1.  Return <emu-val>undefined</emu-val>.
+        1.  Return {{undefined}}.
 
 </div>
 
@@ -8294,11 +8293,12 @@ Async iterators can only be created from an [=async iterable type=].
 
     <code>concatN</code> is an [=operation=] that returns a promise that will be fulfilled with the
     concatenation of all the strings yielded by the async iterable passed to it. It stops
-    concatenating and closes the iterator once the async iterable has yielded N strings.
+    concatenating and closes the iterator once the async iterable has yielded <var ignore>maxN</var>
+    strings.  
 
     <pre>
         interface I {
-          Promise&lt;DOMString> concat(async iterable&lt;DOMString> strings, unsigned long maxN);
+          Promise&lt;DOMString> concatN(async iterable&lt;DOMString> strings, unsigned long maxN);  
         };
     </pre>
 
@@ -8319,7 +8319,7 @@ Async iterators can only be created from an [=async iterable type=].
                 1.  Set |n| to |n| + 1.
                 1.  If |n| is |maxN|, then:
                     1.  Let |finish| be the result of <a lt="close an async iterator">closing</a>
-                        |iterator| with reason <emu-val>undefined</emu-val>.
+                        |iterator| with reason {{undefined}}.
                     1.  [=React=] to |finish|:
                         -   If |finish| was fulfilled, [=resolve=] |promise| with |result|.
                         -   If |finish| was rejected with reason |r|, [=reject=] |promise| with |r|.
@@ -8333,7 +8333,6 @@ Async iterators can only be created from an [=async iterable type=].
 </div>
 
 
-</div>
 
 
 <h4 id="js-record" oldids="es-record">Records — record&lt;|K|, |V|&gt;</h4>
--- a/index.bs
+++ b/index.bs
@@ -8162,9 +8162,9 @@ Note: Instead of returning an async iterable from an IDL operation, the operatio
 
 <div algorithm>
 
-    To <dfn id="async-iterable-finish-iterating" export lt="finish iterating an async iterable">finish iterating</dfn> an
+    To <dfn id="async-iterable-close" export lt="close an async iterable">close</dfn> an
     <code><a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a></code> |iteratorRecord|,
-    perform the following steps:
+    with a reason |reason|, perform the following steps:
 
     1.  Let |iterator| be |iteratorRecord|.\[[Iterator]].
     1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iterator|, "return").
@@ -8172,7 +8172,7 @@ Note: Instead of returning an async iterable from an IDL operation, the operatio
         |returnMethod|.\[[Value]].
     1.  If |returnMethod| is <emu-val>undefined</emu-val>, return [=a promise resolved with=]
         <emu-val>undefined</emu-val>.
-    1.  Let |returnResult| be <a abstract-op>Call</a>(|returnMethod|.\[[Value]], |iterator|).
+    1.  Let |returnResult| be <a abstract-op>Call</a>(|returnMethod|.\[[Value]], |iterator|, « |reason| »).
     1.  If |returnResult| is an abrupt completion, return [=a promise rejected with=]
         |returnResult|.\[[Value]].
     1.  Let |returnPromise| be [=a promise resolved with=] |returnResult|.\[[Value]].
@@ -8188,7 +8188,7 @@ Note: Instead of returning an async iterable from an IDL operation, the operatio
 
     <code>concatN</code> is an [=operation=] that returns a promise that will be fulfilled with the
     concatenation of all the strings yielded by the async iterable passed to it. It stops
-    concatenating and finishes the iterator once the async iterable has yielded N strings.
+    concatenating and closes the iterator once the async iterable has yielded N strings.
 
     <pre>
         interface I {
@@ -8211,7 +8211,8 @@ Note: Instead of returning an async iterable from an IDL operation, the operatio
                 1.  Set |result| to the result of concatenating |result| and |v|.
                 1.  Set |n| to |n| + 1.
                 1.  If |n| is |maxN|, then:
-                    1.  Let |finish| be the result of <a lt="finish iterating an async iterable">finishing iterating</a> |iterable|.
+                    1.  Let |finish| be the result of <a lt="close an async iterable">closing</a>
+                        |iterable| with reason <emu-val>undefined</emu-val>.
                     1.  [=React=] to |finish|:
                         -   If |finish| was fulfilled, [=resolve=] |promise| with |result|.
                         -   If |finish| was rejected with reason |r|, [=reject=] |promise| with |r|.
--- a/index.bs
+++ b/index.bs
@@ -68,6 +68,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ecmascript
             text: internal slot
         text: own property; url: sec-own-property
         text: PromiseCapability; url: sec-promisecapability-records
+        text: Iterator; url: sec-iterator-records
         text: element size; url: table-the-typedarray-constructors
 urlPrefix: https://tc39.es/proposal-resizablearraybuffer/; spec: RESIZABLE-BUFFERS-PROPOSAL
     type: abstract-op
@@ -3350,6 +3351,7 @@ the following algorithm returns <i>true</i>.
                 <th><div><span>interface-like</span></div>
                 <th><div><span>callback function</span></div>
                 <th><div><span>dictionary-like</span></div>
+                <th><div><span>async iterable</span></div>
                 <th><div><span>sequence-like</span></div>
         </thead>
         <tr>
@@ -3365,10 +3367,12 @@ the following algorithm returns <i>true</i>.
             <td>●
             <td>
             <td>●
+            <td>●
         <tr>
             <th>boolean
             <td class="belowdiagonal">
             <td>
+            <td>●
             <td>●
             <td>●
             <td>●
@@ -3391,12 +3395,14 @@ the following algorithm returns <i>true</i>.
             <td>●
             <td>●
             <td>●
+            <td>●
         <tr>
             <th>bigint
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td>
+            <td>●
             <td>●
             <td>●
             <td>●
@@ -3417,6 +3423,7 @@ the following algorithm returns <i>true</i>.
             <td>●
             <td>●
             <td>●
+            <td>●
         <tr>
             <th>object
             <td class="belowdiagonal">
@@ -3426,6 +3433,7 @@ the following algorithm returns <i>true</i>.
             <td class="belowdiagonal">
             <td>
             <td>●
+            <td>
             <td>
             <td>
             <td>
@@ -3443,6 +3451,7 @@ the following algorithm returns <i>true</i>.
             <td>●
             <td>●
             <td>●
+            <td>●
         <tr>
             <th>interface-like
             <td class="belowdiagonal">
@@ -3453,6 +3462,7 @@ the following algorithm returns <i>true</i>.
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td>(a)
+            <td>●
             <td>●
             <td>●
             <td>●
@@ -3469,6 +3479,7 @@ the following algorithm returns <i>true</i>.
             <td>
             <td>(c)
             <td>●
+            <td>●
         <tr>
             <th>dictionary-like
             <td class="belowdiagonal">
@@ -3482,8 +3493,24 @@ the following algorithm returns <i>true</i>.
             <td class="belowdiagonal">
             <td>
             <td>●
+            <td>●
+        <tr>
+            <th>async iterable
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td class="belowdiagonal">
+            <td>
+            <td>
         <tr>
             <th>sequence-like
+            <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td class="belowdiagonal">
             <td class="belowdiagonal">
@@ -4059,7 +4086,7 @@ The following extended attributes are applicable to [=iterable declarations=]:
 </pre>
 
 
-<h4 id="idl-async-iterable">Asynchronously iterable declarations</h4>
+<h4 oldids="idl-async-iterable" id="idl-async-iterable-declaration">Asynchronously iterable declarations</h4>
 
 An [=interface=] can be declared to be asynchronously iterable by using an
 <dfn id="dfn-async-iterable-declaration" export>asynchronously iterable declaration</dfn>
@@ -5665,6 +5692,7 @@ are known as <dfn id="dfn-object-type" export>object types</dfn>.
         StringType Null
         identifier Null
         "sequence" "&lt;" TypeWithExtendedAttributes "&gt;" Null
+        "async iterable" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         "object" Null
         "symbol" Null
         BufferRelatedType Null
@@ -6166,6 +6194,26 @@ sequence is used.
 Any [=list=] can be implicitly treated as a <code>sequence&lt;|T|&gt;</code>, as long as it contains
 only [=list/items=] that are of type |T|.
 
+<!-- Note: if we ever add synchronous iterable types, we should add a note here about why sequences and iterables are not the same. -->
+
+<h4 id="idl-async-iterable-type" lt="async-iterable" dfn export>Async iterable types — async iterable&lt;|T|&gt;</h4>
+
+The <dfn lt="async iterable type" export>async iterable&lt;|T|&gt;</dfn> type is a parameterized
+type whose values are references to objects that asynchronously yield zero or more values of type
+|T|.
+
+Async iterables, unlike sequences, do not have a fixed length and can be infinite. Values are
+asynchronously produced as the async iterable is iterated over.
+
+Async iterable are passed by reference in language bindings where they are represented by an object.
+This means that passing an async iterable to a [=platform object=] will result in a reference to the
+async iterable being kept by that object. Similarly, any async iterable returned from a platform
+object will be a reference to the same object and modifications made to it will be visible to the
+platform object. This is in contrast to sequences, which are always passed by value.
+
+Async iterables must not be used as the type of an [=attribute=] or [=constant=].
+
+There is no way to represent an async iterable value in IDL.
 
 <h4 id="idl-record" lt="record" dfn export>Record types — record&lt;|K|, |V|&gt;</h4>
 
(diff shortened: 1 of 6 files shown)


saschanaz:
Has there been discussion somewhere publicly? 👀

lucacasonato:
@saschanaz No, not of any specifics. The background here is this issue: https://github.com/whatwg/fetch/issues/1291#issuecomment-2009010502. @annevk mentioned we'd need to figure out how to make this work IDL wise. This is my attempt at that (it enables updating the `BodyInit` overload in Fetch from:

```webidl
typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) XMLHttpRequestBodyInit;
typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;
```

to

```webidl
typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) XMLHttpRequestBodyInit;
typedef (ReadableStream or async iterable<BufferSource> or XMLHttpRequestBodyInit) BodyInit;
```

annevk:
I wonder how this ends up working in practice. Say you have `Blob or async iterable<T> or DOMString`. If you pass in a `Blob` with `@@asyncIterable` support, how is it handled? I guess `@@asyncIterable` comes later, probably before `toString`?

cc @petervanderbeken @natechapin

petervanderbeken:
What is being proposed exactly, to what internal type would it be converted? A reference to an `[Iterator Record](https://tc39.es/ecma262/#sec-iterator-records)?

lucacasonato:
> I wonder how this ends up working in practice. Say you have Blob or async iterable<T> or DOMString. If you pass in a Blob with @@asyncIterable support, how is it handled? I guess @@asyncIterable comes later, probably before toString?

@annevk I have specified this behaviour in an update to the overload resolution algorithm in this PR already. The gist is: an `async iterable<T>` has the same behaviour for distinct types as `sequence<T>` (which one could think of as the eager synchronous version of `async iterator<T>`).

I have updated the graphic in the spec that explains this too: 
image

In practice:

1. `async iterator<T>` can be distinguished from all types but `object` and `sequence<T>`
2. As with `sequence<T>`, a matching interface is considered more important than a platform object that implements the iterable, or async iterable protocol. So specifically for the `BodyInit` with `Blob` case: nothing changes.
3. As with `sequence<T>`, primitives like strings, even if they have `@@iterator` on their prototype, are not considered for `async iterator<T>` overloads, and are also not valid `async iterator<T>` values. `async iterator<T>` requires values are `Type(V) == Object`, in addition to having either `@@iterator` or `@@asyncIterator`.

---

> What is being proposed exactly, to what WebIDL type would it be converted? A reference to an [Iterator Record](https://tc39.es/ecma262/#sec-iterator-records)?

@petervanderbeken I am proposing the addition of an `async iterable<T>` type, which mirrors `sequence<T>` in many ways, but unlike it is:

- **lazy**: it can be possibly infinite length, with the JS -> IDL conversion not requiring synchronous transformation of all values of the underlying iterable. Instead values are lazily iterated off the underlying iterator, converted from JS to IDL, and yielded to IDL callers, as they request them.
- **asynchronous**: values can be produced asynchronously

If you imagine a matrix of JS stream primitives, with asynchronicity on one axis, and eagerness on the other, this fills out the square opposite of `sequence<T>`.

|    | **sync** | **async** |
| -- | -- | -- |
| eager | JS: `Array<T>`; IDL: `sequence<T>` | JS: `Promise<Array<T \| Promise<T>>>`; _`async sequence<T>`_ (doesn't exist) |
| lazy | JS: `{ @@iterator: Iterator<T> }`; IDL: _`iterable<T>`_ (doesn't exist) | JS: `{ @@asyncIterator: AsyncIterator<T> }`; IDL: `async iterable<T>` (this PR) | 

The use case for this type are Web APIs that take "streams" of values (of unknown length), that they then process in some way. A real Web API that does this right now is `ReadableStream.from()`. It takes objects that are iterables as a Web IDL `any` type, and [performs manual processing of this iterable](https://streams.spec.whatwg.org/#readable-stream-from-iterable). For one API, this is fine, but there are other proposed APIs that would also take async iterables as arguments:

- `new Response(asyncIterable)`: this overload of `Response`'s `BodyInit` aims to simplify interop of fetch bodies with Node's own stream APIs, which implement `@@asyncIterable` (see https://github.com/whatwg/fetch/issues/1291).
- `crypto.subtle.digest(algorithm, asyncIterable)`: this overload of subtle digest would allow digesting large sets of data without having to load them into memory entirely (see https://github.com/w3c/webcrypto/issues/73 and https://github.com/w3c/webcrypto/issues/73#issuecomment-1003741806).

Introduction of this `async iterable<T>` simplifies all of these, because the entirety of the stream iteration logic is now pulled into Web IDL, where all callers can share it.

saschanaz:
> What is being proposed exactly, to what WebIDL type would it be converted? A reference to an [Iterator Record](https://tc39.es/ecma262/#sec-iterator-records)?

Per https://whatpr.org/webidl/1397/402a886...14f1bdb.html#js-async-iterable the result indeed seem to be iterator record. (Adding because I don't think the comment above actually answers this.)

annevk:
This design sounds good to me. It seems the Build is currently failing though. Could you write a corresponding PR for Streams to show it in action so we have some assurance it can be adopted?

cc @MattiasBuelens

index.bs
```diff
@@ -8064,6 +8112,122 @@ JavaScript Array values.
...
+    <code><a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a></code> |iteratorRecord|,
+    perform the following steps:
+
+    1.  Let |iterator| be |iteratorRecord|.\[[Iterator]].
+    1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iterator|, "return").
+    1.  If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
+        |returnMethod|.\[[Value]].
+    1.  If |returnMethod| is <emu-val>undefined</emu-val>, return [=a promise resolved with=]
+        <emu-val>undefined</emu-val>.
+    1.  Let |returnResult| be <a abstract-op>Call</a>(|returnMethod|.\[[Value]], |iterator|).
```
MattiasBuelens, line 8175: For [`ReadableStream.from(asyncIterable)`](https://streams.spec.whatwg.org/#readable-stream-from-iterable), we need a way to pass an optional `reason` to the `return()` call.

I propose we add an optional `value` argument to "finish iterating an async iterable".

MattiasBuelens:
> 3\. As with `sequence<T>`, primitives like strings, even if they have `@@iterator` on their prototype, are not considered for `async iterator<T>` overloads, and are also not valid `async iterator<T>` values. `async iterator<T>` requires values are `Type(V) == Object`, in addition to having either `@@iterator` or `@@asyncIterator`.

This would technically be a breaking change for `ReadableStream.from()`, since we currently do accept strings (see [this test](https://github.com/web-platform-tests/wpt/blob/3fd901dba4d461afda4cf9b692f8bd99fb05f4e1/streams/readable-streams/from.any.js#L21-L24)).

That said, I wrote that test more as a sanity check, so I don't expect much usage in the wild. We can consider making `ReadableStream.from()` throw on strings, or we could change its signature to accept `async iterable<any> or string` instead.

index.bs
lucacasonato: Done - will update `ReadableStream.from(asyncIterable)`.

lucacasonato:
@MattiasBuelens I was not aware of this behaviour in Streams. I think we should stick with the behaviour where `async iterator<T>` must be an object in this PR (mirroring `sequence<T>`). Overload resolution becomes tricky and rather confusing otherwise.

Let's discuss whether to accept or ignore `DOMString` arguments in the `ReadableStream.from()` PR.

lucacasonato:
@annevk I had already opened a draft PR to the streams spec, showing the integration yesterday. Here is a link: https://github.com/whatwg/streams/pull/1310.

lucacasonato:
Do we have Web IDL tests in WPT? I don't think so, right? I guess no tests specifically for this are relevant.

Considering this is mostly an editorial refactor, do we even need "implementer interest"? There is nothing to implement until an upstream spec uses this. I'll ask for support anyway though:

- @annevk You already expressed support on this design, can I take that as interest from WebKit?
- @saschanaz What are your thoughts? Does Mozilla have an opinion?
- For Chrome, maybe @domenic could chime in?
- Deno is interested in this.
- Node.js likely is too, considering that this unlocks `new Response(asyncIterable)`. @matteocollina can you comment?

Also @jasnell, any thoughts?

lucacasonato:
Got some feedback from @saschanaz elsewhere that the constraint that `async iterable<T>` is only valid in JS->IDL positions (rather than also in IDL->JS positions) should be more prominently highlighted. I'll address this tomorrow

domenic:
> Do we have Web IDL tests in WPT? I don't think so, right?

We do, but they're weird since they test things via APIs that use those features. https://github.com/web-platform-tests/wpt/tree/master/webidl/ecmascript-binding

> I guess no tests specifically for this are relevant.

I think linking to the tests for streams using this feature would be good. Especially if you can verify that everything you introduce in this PR is covered by some aspect of the streams tests. (Which is probably the case since the streams tests are quite thorough, IIRC.)

> Considering this is mostly an editorial refactor, do we even need "implementer interest"? There is nothing to implement until an upstream spec uses this. I'll ask for support anyway though:

Implementer support for Web IDL stuff like this is tricky and requires a judgement call. I think we should CC the bindings teams and give them a chance to weigh in, so @japhet and @yuki3 for Chromium. For example if they were concerned about how to implement this in the bindings layer, or if they were concerned that introducing this feature would be bad because then too many specs might use it and actually it should be rare.

However, since this doesn't require any immediate changes to any implementations, I think we shouldn't block on requiring such support, if we don't hear back.

domenic:
Regarding strings, I think we should follow the behavior of async iterator helpers's `AsyncIterator.from()`; /cc @bakkot. I cannot tell what that behavior is, however. `Iterator.from()` definitely [accepts strings](https://tc39.es/proposal-iterator-helpers/#sec-iterator.from). But `AynscIterator.from()` [calls](https://tc39.es/proposal-async-iterator-helpers/#sec-asynciterator.from) calls [GetIteratorFlattenable](https://tc39.es/proposal-iterator-helpers#sec-getiteratorflattenable) with a second parameter whose value is `async`, which is neither `accept-strings` nor `reject-strings`.

bakkot:
`AsyncIterator.from` will accept strings because `Iterator.from` does. But do note that it's specifically a coercion method. For something like `ReadableStream.from`, which is also a coercion method, you may wish to follow that precedent, but I'm not sure you'd want to do so for other async-iterable-taking things.

The way `AsyncIterator.from` works is to look up `Symbol.asyncIterator`, and if that fails then fall back to `Symbol.iterator` with the result wrapped the same way `for await` does (modulo some quibbles not relevant here), and if _that_ fails to use the `.next` method directly (which only _iterator_-taking things should do, which is to say, nothing except this specific method and `AsyncIterator.prototype.flatMap`).

And since it will not specifically reject strings, and strings have a `Symbol.iterator`, it will accept them, the same way `for await` does.

(Sorry the spec for `AsyncIterator.from` is out of date; I'm working through [other parts of the semantics](https://github.com/tc39/proposal-async-iterator-helpers/issues/4) and it doesn't really make sense to get the spec updated until those are sorted. But this question, at least, is settled.)

domenic:
Thanks for the quick response. I'm unsure whether we should think of `new Response(asyncIterable)` as a "coercion method", but I lean toward yes... and those are our two use cases so far, so, I'm unsure where that leaves us for the general Web IDL type.

bakkot:
Personally I would specify this by making `async iterable<T>` reject stirngs, and then adding `or USVString` to those places which should accept strings, just to ensure that future users of the type have actually decided to take strings on purpose instead of it happening by accident.

Plus if you take a string explicitly then you can use it directly instead of having to look up and invoke `String.prototype[Symbol.iterator]`, which is going to be... slower.

lucacasonato:
Thanks for the comments @domenic and @bakkot. I agree strongly with @bakkot that in the Web IDL type, we should only allow objects implementing the async iterable protocol for the `async iterable<T>` type. Anything else would be a significant hassle with a high chance for unexpected behaviour when plumbed into the overload / union resolution algorithm. Also there is already precedent in Web IDL that iterable taking types do not support iterating string (`sequence<T>` does not support iterating strings, but does iterate objects implementing the iterable protocol).

For the streams specific discussion, lets move it right here: https://github.com/whatwg/streams/pull/1310#issuecomment-2022484496

lucacasonato:
Folks, could you take another look? I have clarified the allowed positions for async iterables (namely not in attributes), and have added guidance that:
a) async iterable IDL values can not be constructed in IDL, and always must come from language bindings
b) that spec authors should generally return an interface with an async iterable declaration, not an async iterable IDL value

@annevk has confirmed implementer interest for WebKit on Matrix - is anyone else in favor, or opposed?

jasnell:
This LGTM

lucacasonato:
Gently pushing for reviews again :)

Still need positions from Chromium (@domenic) and Mozilla (@saschanaz).

bakkot:
TC39 [is considering](https://github.com/tc39/how-we-work/pull/152) rejecting strings in most (new) iterable-taking positions in general (and of course async iterable taking positions as well). It'll be discussed at the plenary at the end of the month, and I'm guessing it'll be adopted.

annevk:
@lucacasonato coupling this with the Streams PR I'm happy to consider Chromium and WebKit as interested and nobody opposed. This PR needs rebasing, the checklist in OP needs a bit more work, and maybe @MattiasBuelens can do a final pass after that. With that I think we're good to go.

saschanaz:
I'd like to ask @petervanderbeken for the position as he will be the implementer.

A parallel question for distinguishability: currently it says it's distinguishable with interfaces and dictionaries, but what happens if the interface has `async iterable<T>;`? sequence also says it's distinguishable, so I think I just lack some knowledge here 😛

lucacasonato:
@saschanaz The interface can be distinguished from other objects because it is branded. A union of `interface` and `async iterable<T>` will always attempt to match the `interface` first. This is also why `dictionary` and `interface` are distinguishable.

Union resolution goes like this in this case: `MyInterface | MyDictionary | async iterable<T> (assuming V has been checked to be an object):

1. Check if `V` is an instance of the interface. _it's an interface_
2. Check if `V` has `[Symbol.asyncIterable]`. _it's an async iterable<T>_
3. _it's a dictionary_

index.bs
```diff
@@ -8064,6 +8119,124 @@ JavaScript Array values.
...
+    to an IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
+
+    1.  If <a abstract-op>Type</a>(|V|) is not Object,
+        [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+    1.  Let |iteratorRecord| be [=?=] <a abstract-op>GetIterator</a>(|V|, async).
+    1.  Return |iteratorRecord|.
+</div>
+
+<div id="async-iterable-to-js" algorithm="convert an async iterable to a JavaScript value">
+    An IDL <a lt="async iterable type">async iterable&lt;|<var ignore>T</var>&gt;</a> value is
```
MattiasBuelens, line 8138: Typo:
```suggestion
    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value is
```

index.bs
```diff
@@ -11174,6 +11347,23 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
         1.  Otherwise: if <a abstract-op>Type</a>(|V|) is Object and
             there is an entry in |S| that has one of the
             following types at position |i| of its type list,
+            *   a [=async iterable type=]
```
MattiasBuelens, line 11350:
```suggestion
            *   an [=async iterable type=]
```

index.bs
```diff
@@ -11174,6 +11347,23 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
...
+
+            1.  Let |method| be [=?=] <a abstract-op>GetMethod</a>(|V|, {{@@asyncIterator}}).
+
+            |method| is not <emu-val>undefined</emu-val>, then remove from |S| all
+            other entries.
+
+        1.  Otherwise: if <a abstract-op>Type</a>(|V|) is Object and
+            there is an entry in |S| that has one of the
+            following types at position |i| of its type list,
+            *   a [=async iterable type=]
```
MattiasBuelens, line 11366:
```suggestion
            *   an [=async iterable type=]
```

index.bs
```diff
@@ -11342,11 +11532,11 @@ Note: The HTML Standard defines how a security check is performed. [[!HTML]]
     Generally, the inspection of the value at the distinguishing argument index does not have any
     side effects, and the only side effects in the overload resolution algorithm are the result of
     converting the JavaScript values to IDL values.
-    (An exception exists when one of the overloads has a [=sequence type=] or [=frozen array type=]
-    at the distinguishing argument index.
-    In this case, we attempt to get the {{@@iterator}} property to determine the appropriate
-    overload, and perform the conversion of the distinguishing argument separately before continuing
-    with the next step.)
+    (An exception exists when one of the overloads has a [=async iterable type=], [=sequence type=]
```
MattiasBuelens, line 11535:
```suggestion
    (An exception exists when one of the overloads has an [=async iterable type=], [=sequence type=]
```

index.bs
```diff
@@ -8064,6 +8119,124 @@ JavaScript Array values.
...
+object and a JavaScript [=Iterator=] record.
+
+<div id="js-to-async-iterable" algorithm="convert a JavaScript value to async iterable">
+    A JavaScript value |V| is [=converted to an IDL value|converted=]
+    to an IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
+
+    1.  If <a abstract-op>Type</a>(|V|) is not Object,
+        [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+    1.  Let |iteratorRecord| be [=?=] <a abstract-op>GetIterator</a>(|V|, async).
+    1.  Return |iteratorRecord|.
```
MattiasBuelens, line 8134: Consider making this more explicit:
```suggestion
    1.  Return the IDL async iterable value that represents a reference to the JavaScript object |V| and the JavaScript [=Iterator=] record |iteratorRecord|.
```

index.bs
```diff
@@ -8064,6 +8119,124 @@ JavaScript Array values.
...
+</div>
+
+<div id="async-iterable-to-js" algorithm="convert an async iterable to a JavaScript value">
+    An IDL <a lt="async iterable type">async iterable&lt;|<var ignore>T</var>&gt;</a> value is
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return the JavaScript object that represents the same async iterable as the IDL value.
+</div>
```
MattiasBuelens, line 8142: When would we need this? If I understand correctly, IDL interfaces should always add an async iterable declaration to themselves, rather than returning an async iterable from one of its operations.

Also, if the async iterable value can only have been created by converting it from a JavaScript value, is this value still usable when converting it back to a JavaScript value? The previous algorithm calls `GetIterator(V, async)`, which for a `ReadableStream` *locks the stream*. When we return the `ReadableStream` back to JavaScript, it'll still be locked by IDL and JavaScript won't be able to async-iterate the stream.

Would it be an option for us to just *disallow* converting an IDL async iterable type back to a JavaScript value?

MattiasBuelens (commented):
Looks pretty good! I still have a question about what it really means to "convert an IDL async iterable value to a JavaScript object" though, see below.

index.bs
MattiasBuelens: Hmm, maybe the real problem is that the IDL machinery should not be calling `GetIterator()` from within the conversion logic in the first place. 🤔 For example: if an IDL operation accepts *two* `async iterable<T>` values as parameters, and `GetIterator()` succeeds for the first value but fails for the second value (e.g. because it's a locked `ReadableStream`), then the operation errors immediately with a `TypeError` but the first value is now permanently "locked".

I think we should make "convert from a JavaScript object" *not* call `GetIterator()`:
* Make an IDL async iterable *only* represent a JavaScript object.
* Add a separate type for an "IDL async iterator", which is represented by a JavaScript `Iterator` record.
* Add an "open an async iterable" algorithm, which takes an IDL async iterable, calls `GetIterator()` and returns it as an IDL async iterator.
* Refactor the "get the next value of an async iterable" and "close an async iterable" algorithms to work on an IDL async iterator instead.

Thoughts?

lucacasonato: @MattiasBuelens I have attempted to implement this in [5533589](https://github.com/whatwg/webidl/pull/1397/commits/5533589f036f80ab6dbcff63ba1c5bad8dba03d2)

index.bs
```diff
@@ -8064,6 +8129,162 @@ JavaScript Array values.
...
+    A JavaScript value |V| is [=converted to an IDL value|converted=]
+    to an IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
+
+    1.  If <a abstract-op>Type</a>(|V|) is not Object,
+        [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+    1.  Let |method| be [=?=] <a abstract-op>GetMethod</a>(obj, %Symbol.asyncIterator%).
+    1.  If |method| is undefined:
+        1.  Set |syncMethod| to [=?=] <a abstract-op>GetMethod</a>(obj, %Symbol.iterator%).
+        1.  If |syncMethod| is undefined, [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+        1   Return the IDL async iterable value that represents a reference to the JavaScript object
```
MattiasBuelens, line 8148:
```suggestion
        1.  Return the IDL async iterable value that represents a reference to the JavaScript object
```

index.bs
```diff
@@ -8064,6 +8129,162 @@ JavaScript Array values.
...
+    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value is
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return the JavaScript object that represents the same async iterable as the IDL value.
+</div>
+
+<h5 id="idl-async-iterator" lt="async-iterator" dfn export>Async iterators</h5>
+
+IDL <a lt="async-iterator">async-iterator</a> values are represented by JavaScript [=Iterator=]
+records.
```
MattiasBuelens, line 8164:
```suggestion
<h5 id="idl-async-iterator" lt="async iterator" dfn export>Async iterators</h5>

IDL <a lt="async iterator">async iterator</a> values are represented by JavaScript [=Iterator=]
records.
```

index.bs
```diff
@@ -8064,6 +8129,162 @@ JavaScript Array values.
...
+Async iterators can only be created from an [=async iterable=] object.
+
+<h5 id="js-async-iterator-iteration">Iterating async iterator</h5>
+
+[=Async iterables=] are not directly iterated over. They are first opened, to create a new
+[=async iterator=], and then the [=async iterator=] is iterated over.
+
+<div algorithm>
+
+    To <dfn id="async-iterable-open" export lt="get an async iterable next value">get the next value</dfn> of an
```
MattiasBuelens, line 8175:
```suggestion
    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
```

MattiasBuelens (commented):
Much better! Just a few small typos left, otherwise LGTM.

index.bs
```diff
@@ -6177,6 +6206,42 @@ sequence is used.
...
+[=asynchronously iterable declaration=].
+
+Async iterables must not be used as the type of an [=attribute=] or [=constant=].
+
+There is no way to represent an async iterable value in IDL.
+
+<h5 id="idl-async-iterator-object" lt="async-iterator" dfn export>Async iterator objects</h5>
+
+An <dfn lt="async iterator" export>async iterator&lt;|T|&gt;</dfn> value is a reference to an object
+that can produce a sequence of values asynchronously. The async iterator value is is a parameterized
```
MattiasBuelens, line 6235:
```suggestion
that can produce a sequence of values asynchronously. The async iterator value is a parameterized
```

index.bs
```diff
@@ -34,6 +34,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ecmascript
     type: argument
         text: NewTarget; url: sec-built-in-function-objects
     type: abstract-op
+        text: CreateAsyncFromSyncIteator; url: sec-createasyncfromsynciterator
```
MattiasBuelens, line 37:
```suggestion
        text: CreateAsyncFromSyncIterator; url: sec-createasyncfromsynciterator
```

MattiasBuelens:
@lucacasonato I fixed the remaining Bikeshed issues [in this branch](https://github.com/lucacasonato/webidl/compare/async_iterator_type...MattiasBuelens:webidl:async_iterator_type). Could you pull those changes into your branch? 🙂

MattiasBuelens (approved):

index.bs
```diff
@@ -6177,6 +6207,42 @@ sequence is used.
 Any [=list=] can be implicitly treated as a <code>sequence&lt;|T|&gt;</code>, as long as it contains
 only [=list/items=] that are of type |T|.
 
+<!-- Note: if we ever add synchronous iterable types, we should add a note here about why sequences and iterables are not the same. -->
+
+<h4 id="idl-async-iterable-type" lt="async iterable" dfn export>Async iterable types — async iterable&lt;|T|&gt;</h4>
+
+The <dfn lt="async iterable type" export>async iterable&lt;|T|&gt;</dfn> type is a parameterized
+type whose values are references to objects that can produce an [=async iterator=].
```
domenic, line 6215: This doesn't seem to match how other parameterizable types are defined. I'd expect just "async iterable" as the dfn text.

index.bs
```diff
@@ -6177,6 +6207,42 @@ sequence is used.
...
+
+Async iterables are passed by reference in language bindings where they are represented by an object.
+This means that passing an async iterable to a [=platform object=] will result in a reference to the
+async iterable being kept by that object. Similarly, any async iterable returned from a platform
+object will be a reference to the same object and modifications made to it will be visible to the
+platform object. This is in contrast to sequences, which are always passed by value.
+
+Note: Async iterables can not be constructed from IDL. If returned from an operation, or used as the
```
domenic, line 6223:
```suggestion
Note: Async iterables cannot be constructed from IDL. If returned from an operation, or used as the
```

index.bs
```diff
@@ -6177,6 +6207,42 @@ sequence is used.
...
+type of a dictionary member, the async iterable will have originated from the host environment and
+have been turned into an IDL type via a language binding. Instead of returning an async iterable
+from an IDL operation, the operation may want to return an [=interface=] that has an
```
domenic, line 6226:
```suggestion
from an IDL operation, the operation might want to return an [=interface=] that has an
```

index.bs
```diff
@@ -6177,6 +6207,42 @@ sequence is used.
...
+<h5 id="idl-async-iterator-object" lt="async-iterator" dfn export>Async iterator objects</h5>
+
+An <dfn lt="async iterator" export>async iterator&lt;|T|&gt;</dfn> value is a reference to an object
+that can produce a sequence of values asynchronously. The async iterator value is parameterized by a
+type |T|, which defines the types of values produced by this async iterator.
+
+Async iterators, unlike sequences, do not have a fixed length and can be infinite. Values are
+asynchronously produced as the async iterator is iterated over.
+
+Async iterators are values, not types, and thus cannot be used as a type in IDL. They can only be
```
domenic, line 6242: This seems pretty confusing. What type to the created values have, then?

I think this whole section is JS language binding specific, since it relates to algorithms like "open" and "get the next value" which are only defined for the JS language binding. So maybe we can remove this confusing not-a-type from the Web IDL "Types" section (2.13), and make it something produced and consumed by those algorithms?

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
 </div>
 
 
+<h4 id="js-async-iterable">Async iterable — async iterable&lt;|T|&gt;</h4>
+
+IDL <a lt="async iterable type">async iterable&lt;|T|&gt;</a> values are represented by a JavaScript
+object, a JavaScript method, and a flag indicating whether the method is expected to produce a
+<code>sync</code> or <code>async</code> iterator.
```
domenic, line 8188: Can we give these fields proper `<dfn>`s? We're not consistent about this in general but I'd like to be here since this is one of the more complex fields we have. #idl-callback-interface has an example where it defines `callback context` (although it tries to be language-agnostic, which is probably not right).

Something like

```html
In the JavaScript binding, IDL [=async iterable type|async iterable=] values are represented by a [=struct=] with the following [=items=]:

* <dfn for="JS async iterable">object</dfn>, a JavaScript value
* <dfn for="JS async iterable">method</dfn>, a JavaScript value
* <dfn for="JS async iterable">type</dfn>, either "<code>sync</code>" or "<code>async</code>"
```

and then reference those struct items throughout.

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return the JavaScript object that represents the same async iterable as the IDL value.
+</div>
+
+<h5 id="js-async-iterator">Async iterators</h5>
+
+IDL [=async iterator=] values are represented by JavaScript [=Iterator=] records.
+
+Async iterators can only be created from an [=async iterable type=].
```
domenic, line 8217: Specifically by the "open" algorithm, right?

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+
+<h5 id="js-async-iterator-iteration">Iterating async iterator</h5>
```
domenic, line 8219:
```suggestion
<h5 id="js-async-iterator-iteration">Iterating async iterators</h5>
```

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+
+[=Async iterables=] are not directly iterated over. They are first opened, to create a new
+[=async iterator=], and then the [=async iterator=] is iterated over.
+
+<div algorithm>
+
+    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
+    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|,
+    perform the following steps:
```
domenic, line 8228:
```suggestion
    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|:
```

here and below

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+
+    1.  Let |object| be the |iterable|'s JavaScript object.
```
domenic, line 8230:
```suggestion
    1.  Let |object| be |iterable|'s JavaScript object.
```

here and below.

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+    1.  Let |method| be the |iterable|'s JavaScript method.
+    1.  Let |iterator| be [=?=] <a abstract-op>GetIteratorFromMethod</a>(|object|, |method|).
+    1.  If |iterable| has a flag that shows that it is expected to produce a
+        <code>sync</code> iterator, then:
+        1. Set |iterator| to <a abstract-op>CreateAsyncFromSyncIterator</a>(|iterator|).
```
domenic, line 8235: Nit: don't nest single-line if statements

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+<div id="async-iterable-to-js" algorithm="convert an async iterable to a JavaScript value">
+    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value is
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return the JavaScript object that represents the same async iterable as the IDL value.
+</div>
+
+<h5 id="js-async-iterator">Async iterators</h5>
+
+IDL [=async iterator=] values are represented by JavaScript [=Iterator=] records.
```
domenic, line 8215: Similarly I think we can formally define this as a struct that contains "type parameter" and "record" fields.

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+
+<div algorithm>
+
+    To <dfn id="async-iterator-close" export lt="close an async iterator">close</dfn> an
+    <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|,
+    with a reason |reason|, perform the following steps:
+
+    1.  Let |iteratorRecord| be the [=Iterator=] represented by |iterator|.
+    1.  Let |iteratorObj| be |iteratorRecord|.\[[Iterator]].
+    1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iteratorObj|, "return").
```
domenic, line 8276:
```suggestion
    1.  Let |returnMethod| be <a abstract-op>GetMethod</a>(|iteratorObj|, "<code>return</code>").
```

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+    1.  If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
+        |returnMethod|.\[[Value]].
+    1.  If |returnMethod| is <emu-val>undefined</emu-val>, return [=a promise resolved with=]
+        <emu-val>undefined</emu-val>.
```
domenic, line 8280:
```suggestion
        {{undefined}}.
```

(converting from ES-land to IDL-land)

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+
+<div class="example" id="example-js-async-iterable">
+
+    <code>concatN</code> is an [=operation=] that returns a promise that will be fulfilled with the
+    concatenation of all the strings yielded by the async iterable passed to it. It stops
+    concatenating and closes the iterator once the async iterable has yielded N strings.
+
+    <pre>
+        interface I {
+          Promise&lt;DOMString> concat(async iterable&lt;DOMString> strings, unsigned long maxN);
```
domenic, line 8301:
```suggestion
          Promise&lt;DOMString> concatN(async iterable&lt;DOMString> strings, unsigned long maxN);
```

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+            <l spec=ecmascript>{{TypeError}}</l>.
+        1.  Return <emu-val>undefined</emu-val>.
+
+</div>
+
+<div class="example" id="example-js-async-iterable">
+
+    <code>concatN</code> is an [=operation=] that returns a promise that will be fulfilled with the
+    concatenation of all the strings yielded by the async iterable passed to it. It stops
+    concatenating and closes the iterator once the async iterable has yielded N strings.
```
domenic, line 8297:
```suggestion
    concatenating and closes the iterator once the async iterable has yielded <var ignore>maxN</var> strings.
```

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+    1.  Let |step| be a sequence of steps that will be used to process the async iterable:
+        1.  Let |next| be the result of <a lt="get an async iterator next value">getting the next value</a> of |iterator|.
+        1.  [=React=] to |next|:
+            -   If |next| was fulfilled with value |v|:
+                1.  If |v| is [=end of iteration=], [=resolve=] |promise| with |result|.
+                1.  Set |result| to the result of concatenating |result| and |v|.
+                1.  Set |n| to |n| + 1.
+                1.  If |n| is |maxN|, then:
+                    1.  Let |finish| be the result of <a lt="close an async iterator">closing</a>
+                        |iterator| with reason <emu-val>undefined</emu-val>.
```
domenic, line 8322:
```suggestion
                        |iterator| with {{undefined}}.
```

(users of these algorithms should be staying in IDL land).

index.bs
```diff
@@ -8115,6 +8181,161 @@ JavaScript Array values.
...
+                    1.  Call |step|.
+            -   If |next| was rejected with reason |r|, [=reject=] |promise| with |r|.
+    1.  Call |step|.
+    1.  Return |promise|.
+
+</div>
+</div>
+
+
+</div>
```
domenic, line 8336: What is this closing?

domenic (commented):
Thanks for working on this! I think the theoretically-language-agnostic part is not quite separated nicely from the JS language part, and we can make the latter more rigorous. But the feedback is largely editorial.

index.bs
lucacasonato: Done. However I had to change `record` to `underlying record` because otherwise  multiple `[=record=]` (referring to the IDL type) that would emit link errors because multiple possible dfns of `record` existed (even if one of them was namespaced to `async iterator`). I couldn't figure out how to forciably link `[=record=]` to the IDL type.

index.bs
lucacasonato: Nothing

index.bs
lucacasonato: I moved it down to the JS language section, and added some other wording up here to express that IDL async iterables are themselves not being iterated over, but produce a value that is being iterated over (without referencing `[=async iterator=]`).

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return |V|'s [=JS async iterable/object=].
+</div>
+
+<h5 id="js-async-iterator">Async iterators</h5>
+
+<dfn lt="async iterator" export>Async iterator</dfn> values are represented by a [=struct=] with the following [=struct/items=]:
+
+* <dfn for="JS async iterator">underlying record</dfn>, a [=Iterator=] record
```
MattiasBuelens, line 8215:
```suggestion
* <dfn for="JS async iterator">underlying record</dfn>, an [=Iterator=] record
```

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+<div algorithm>
+
+    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
+    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|:
+
+    1.  Let |iterator| be [=?=] <a abstract-op>GetIteratorFromMethod</a>(|iterable|'s
+        [=JS async iterable/object=], |iterable|'s [=JS async iterable/method=]).
+    1.  If |iterable|'s [=JS async iterable/type=] is "<code>sync</code>", set |iterator| to
+        <a abstract-op>CreateAsyncFromSyncIterator</a>(|iterator|).
+    1.  Return a [=async iterator=] value with [=JS async iterator/underlying record=] set to |iterator| and
```
MattiasBuelens, line 8235:
```suggestion
    1.  Return an [=async iterator=] value with [=JS async iterator/underlying record=] set to |iterator| and
```

index.bs
```diff
@@ -6177,6 +6207,34 @@ sequence is used.
 Any [=list=] can be implicitly treated as a <code>sequence&lt;|T|&gt;</code>, as long as it contains
 only [=list/items=] that are of type |T|.
 
+<!-- Note: if we ever add synchronous iterable types, we should add a note here about why sequences and iterables are not the same. -->
+
+<h4 id="idl-async-iterable-type" lt="async iterable" dfn export>Async iterable types — async iterable&lt;|T|&gt;</h4>
+
+The <dfn lt="async iterable type" export>async iterable&lt;|T|&gt;</dfn> type is a parameterized
```
domenic, line 6214: I'm still not clear why we're using a different style here than the other parameterized type classes. We should say "An async iterable type", with no `<T>`.

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+<div id="async-iterable-to-js" algorithm="convert an async iterable to a JavaScript value">
+    An IDL <a lt="async iterable type">async iterable&lt;<var ignore>T</var>&gt;</a> value |V| is
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return |V|'s [=JS async iterable/object=].
+</div>
+
+<h5 id="js-async-iterator">Async iterators</h5>
+
+<dfn lt="async iterator" export>Async iterator</dfn> values are represented by a [=struct=] with the following [=struct/items=]:
```
domenic, line 8213:
```suggestion
<dfn lt="async iterator" export>Async iterators</dfn> are [=structs=] with the following [=struct/items=]:
```

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+
+* <dfn for="JS async iterator">underlying record</dfn>, an [=Iterator=] record
+* <dfn for="JS async iterator">type parameter</dfn>, an IDL type representing the type of values produced by the async iterator
+
+An [=async iterable=] is not directly iterated over. Instead, it is first opened to create
```
domenic, line 8218: This sentence is a dupe of the one right after it, so can be deleted.

Given that, this section is pretty small. How about just moving the definition of "async iterator" into the "Iterating async iterators" section, after that intro sentence?

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+
+<h5 id="js-async-iterator-iteration">Iterating async iterators</h5>
+
+[=Async iterables=] are not directly iterated over. They are first opened, to create a new
+[=async iterator=], and then the [=async iterator=] is iterated over.
+
+<div algorithm>
+
+    To <dfn id="async-iterable-open" export lt="open an async iterable">open</dfn> an
+    <code><a lt="async iterable type">async iterable&lt;<var>T</var>&gt;</a></code> |iterable|:
```
domenic, line 8229: I *think* we should be linking to "async iterable", not "async iterable type", since we're talking about instances? I'm not sure; you can check other cases where we distinguish to see if that's the practice.

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+        [=JS async iterator/type parameter=] set to |T|.
+
+</div>
+
+
+
+<div algorithm>
+
+    To <dfn id="async-iterator-get-next-value" export lt="get an async iterator next value">get the next value</dfn> of an
+    <code><a lt="async iterator">async iterator</a></code> |iterator|:
```
domenic, line 8245:
```suggestion
    [=async iterator=] |iterator|:
```

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+
+    1.  Let |nextResult| be the result of calling
```
domenic, line 8247:
```suggestion
    1.  Let |nextResult| be
```

index.bs
```diff
@@ -8115,6 +8173,168 @@ JavaScript Array values.
...
+                value of type |iterator|'s [=JS async iterator/type parameter=].
+            1.  Return |value|.
+
+</div>
+
+<div algorithm>
+
+    To <dfn id="async-iterator-close" export lt="close an async iterator">close</dfn> an
+    <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|,
+    with a reason |reason|, perform the following steps:
```
domenic, line 8271:
```suggestion
    To <dfn id="async-iterator-close" export lt="close an async iterator">close</dfn> an
    <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|
    with a reason |reason|:
```

domenic (commented):
This looks really nice now! Just a few more editorial tidbits. Also be sure to fill out the template in the OP.

index.bs
lucacasonato: Yes, promises seems to use the instance, not the type. I changed it.

lucacasonato:
@domenic What tests should I write for this? I don't think we have specific webidl tests in WPT, right? Only tests for features that use the webidl types.

index.bs
```diff
@@ -8115,6 +8173,162 @@ JavaScript Array values.
...
+</div>
+
+<div algorithm>
+
+    To <dfn id="async-iterator-close" export lt="close an async iterator">close</dfn> an
+    <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|,
+    with a reason |reason|:
+
+    1.  Let |iteratorRecord| be |iterator|'s [=JS async iterator/underlying record=].
+    1.  Let |iteratorObj| be |iteratorRecord|.\[[Iterator]]
```
domenic, line 8268:
```suggestion
    1.  Let |iteratorObj| be |iteratorRecord|.\[[Iterator]].
```

domenic (approved):
> What tests should I write for this? I don't think we have specific webidl tests in WPT, right? Only tests for features that use the webidl types.

This was discussed in https://github.com/whatwg/webidl/pull/1397#issuecomment-2021792761.

index.bs
```diff
@@ -8115,6 +8173,162 @@ JavaScript Array values.
...
+    An IDL <a lt="async iterable">async iterable&lt;<var ignore>T</var>&gt;</a> value |V| is
+    [=converted to a JavaScript value|converted=] to a JavaScript object as follows:
+
+    1.  Return |V|'s [=JS async iterable/object=].
+</div>
+
+<h5 id="js-async-iterator-iteration">Iterating async iterators</h5>
+
+An [=async iterable=] is not directly iterated over. Instead, it is first opened to create
+a [=async iterator=]. The [=async iterator=] can be asynchronously iterated over to produce values.
```
MattiasBuelens, line 8214:
```suggestion
an [=async iterator=]. The [=async iterator=] can be asynchronously iterated over to produce values.
```

MattiasBuelens (approved):
LGTM (aside from one last nit)

annevk:
Just to make sure I understand correctly, this is blocked on tests and this is best tested through the Streams change https://github.com/whatwg/streams/pull/1310 which is currently still in draft. So once the Streams change is ready and has test coverage (which should include coverage for the IDL bits) this is all good to go?

lucacasonato:
So during an experiment we ran for `new Response(async iterable<Uint8Array>)`, we discovered [another fun edge case](https://github.com/whatwg/fetch/issues/1291#issuecomment-2309773461) to think about: when having a union of `string | async iterable<Uint8Array>`, what should a boxed string resolve as? In the current spec text it resolves as a `async iterable<DOMString>`, which would cause an error during iteration. This happens because the boxed string is both an object, and has a `[Symbol.iterator]` method.

We have two options here:
1. handle this case specially in WebIDL by banning boxed string as a value for `async iterable`, like we do with primitives (kinda weird, but means this doesn't become a weird edge case only in fetch and is instead uniform across the platform)
2. handle this as a weird edge case in `fetch`, and in the future in any other `string | async iterator` WebIDL unions.

The big question is whether, assuming `ReadableStream.from("asd")` throws,  `ReadableStream.from(new String("asd"))` should throw, or create a stream of string chunks.

Any thoughts?

lucacasonato:
Copying over [a thought](https://github.com/whatwg/fetch/issues/1291#issuecomment-2309803657) from @annevk:

> Treating string objects and values in the same way whenever possible seems preferable. Which argues for 1 as otherwise each API would have to deal with this anew.

bakkot:
In TC39 we went the other direction, and said that String objects are just another kind of object. Having the check conceptually be "is a primitive" is the simplest mental model.

annevk:
That also seems reasonable, but the problem is that any API currently accepting a string would need to juggle a lot of complexity to safely adopt `async iterable<T>`.

Perhaps there's a third solution whereby `string | async iterable<T>` cases involving a String object are resolved by going down the `string` path? But if there's no string in the type union we'd treat a String object as an Object for the purposes of `async iterable<T>`.

lucacasonato:
I like the option where we just handle this in the union.

index.bs
```diff
@@ -8115,6 +8173,162 @@ JavaScript Array values.
...
+    1.  Return |V|'s [=JS async iterable/object=].
+</div>
+
+<h5 id="js-async-iterator-iteration">Iterating async iterators</h5>
+
+An [=async iterable=] is not directly iterated over. Instead, it is first opened to create
+an [=async iterator=]. The [=async iterator=] can be asynchronously iterated over to produce values.
+
+<dfn lt="async iterator" export>Async iterators</dfn> are [=structs=] with the following [=struct/items=]:
+* <dfn for="JS async iterator">underlying record</dfn>, an [=Iterator=] record
```
domfarolino, line 8217: ECMAScript routinely refers to these as the type "Iterator Record". The term is often linkable as "Iterator Record" too, so could we maybe follow suit and make this `[=Iterator Record=]` instead?

index.bs
```diff
@@ -8115,6 +8173,162 @@ JavaScript Array values.
...
+
+
+<div algorithm>
+
+    To <dfn id="async-iterator-get-next-value" export lt="get an async iterator next value">get the next value</dfn> of an
+    [=async iterator=] |iterator|:
+
+    1.  Let |nextResult| be
+        <a abstract-op>IteratorNext</a>(|iterator|'s [=JS async iterator/underlying record=]).
+    1.  If |nextResult| is an abrupt completion, return [=a promise rejected with=]
```
domfarolino, line 8243: All references of "abrupt completion" in the current standard link to the ES term. Could we do that too here?

index.bs
```diff
@@ -8115,6 +8173,162 @@ JavaScript Array values.
...
+                value of type |iterator|'s [=JS async iterator/type parameter=].
+            1.  Return |value|.
+
+</div>
+
+<div algorithm>
+
+    To <dfn id="async-iterator-close" export lt="close an async iterator">close</dfn> an
+    <code><a lt="async iterator">async iterator&lt;<var ignore>T</var>&gt;</a></code> |iterator|,
+    with a reason |reason|:
```
domfarolino, line 8265: Maybe clarify that `|reason|` is a JavaScript value, like we do in https://webidl.spec.whatwg.org/#a-promise-rejected-with?

lucacasonato:
We ran another experiment with changed behaviour where for unions of `string | async iterable<T>` the `String` object is resolved as `string` rather than as `async iterable<T>`. We have not gotten any regression reports so far. I am going to update this PR with that change.

twiss:
As @annevk pointed out in #1461, this would also be useful for https://github.com/w3c/webcrypto/pull/390 :)

lucacasonato:
@domenic Please take another look. As mentioned above, I had to add an exception for `string | async iterable<T>` overloads, so that string objects resolve as `string` rather than `async iterable<T>`.

This was discovered while we ran an experiement allowing `async iterable<T>` in `Request`/`Response` bodies.

I also realized I did not implement the union resolution previously (only the overload resolution). That is now fixed.

annevk:
I created #1464 to address the Bikeshed issues. Once that's merged any workarounds can be removed here.

domenic:
Maybe @domfarolino can take this, as I think he'll want to use it in the observable spec? (And maybe he's already written a version of it.

index.bs
```diff
@@ -8120,6 +8184,162 @@ JavaScript Array values.
...
+
+
+<div id="js-to-async-iterable" algorithm="convert a JavaScript value to async iterable">
+    A JavaScript value |V| is [=converted to an IDL value|converted=]
+    to an IDL <a lt="async iterable">async iterable&lt;<var ignore>T</var>&gt;</a> value as follows:
+
+    1.  If |V| [=is not an Object=], then
+        [=JavaScript/throw=] a <l spec=ecmascript>{{TypeError}}</l>.
+    1.  Let |method| be [=?=] <a abstract-op>GetMethod</a>(obj, {{%Symbol.asyncIterator%}}).
+    1.  If |method| is undefined:
```
domfarolino, line 8204: nit: For consistency with elsewhere (including within this PR), I guess we should use `<emu-val>undefined</emu-val>` here and below?

index.bs
```diff
@@ -8686,6 +8906,18 @@ that correspond to the union's [=member types=].
         1.  If |types| includes {{object}}, then return the IDL value
             that is a reference to the object |V|.
     1.  If |V| [=is an Object=], then:
+        1.  If |types| includes an [=async iterable type=], then
+            1. If |types| does not include a [=string type=] or |V| does not have a \[[StringData]] [=/internal slot=], then
```
domfarolino, line 8910: Can we just say `, then return the result of converting V to an async iterable`, linking to [your conversion steps](https://whatpr.org/webidl/1397.html#ref-for-dfn-convert-ecmascript-to-idl-value%E2%91%A2%E2%91%A1) that seem to do the same thing as the below substeps?

lucacasonato: Those steps throw when no iterable method is present. Here we want to not throw in that case

domfarolino: Oh duh 😅, thanks.

index.bs
lucacasonato:
```suggestion
    1.  If |method| is <emu-val>undefined</emu-val>:
```

domfarolino: Note you're still missing the instance two lines below.

domenic (approved):
This LGTM. Does anyone else want to review? Otherwise let's plan on merging this within a day or two.

index.bs
```diff
@@ -19,6 +19,7 @@ spec: ecmascript; type: dfn;
         text: constructor
         text: realm
 spec: dom; type: dfn; text: element
+spec: dom; type: dfn; for: /; text: document
```
annevk, line 22: I already fixed this differently.
```suggestion
```

annevk (commented):
Sounds good to me. Just one small nit.

PR #1413, opened 2024-06-26
Disallow non-attribute frozen arrays

### LINKED ISSUE #1399: Disallow FrozenArray<> in argument lists, and maybe dictionaries and return types

Argument lists should always accept sequences, not `FrozenArray<>`s. But some specs seem to use `FrozenArray<>`s as arguments. This never makes sense.

Similarly, it never makes sense to use `FrozenArray<>` as a type for an "input" dictionary. You should always use sequences.

---

Other "output" cases where we might want to disallow `FrozenArray<>`s are "output" dictionaries, and method return values.

In general, such APIs should be designed so that when you vend an object to the web developer, the developer can then mutate the object if they want. Taking an extra step to freeze them is unusual and kind of silly.

The exception is if some of these APIs are caching the `FrozenArray<>`s to avoid creating new JS Array values each time, for performance reasons. I'd be skeptical the performance issues here are really critical but it is a possible use case.

---

I'll file a variety of issues linking back to this one to see if we can clean up the ecosystem and then implement prohibitions in the spec.

- Using `FrozenArray<>` as an argument type:
    - [x] https://github.com/WICG/shared-storage/issues/143
    - [ ] https://github.com/WebAudio/web-audio-api/issues/2581
- Using `FrozenArray<>` in a dictionary:
    - [x] https://github.com/immersive-web/hit-test/issues/116
    - [x] https://github.com/WICG/shape-detection-api/issues/100
    - [x] https://github.com/immersive-web/webxr/issues/1368
- Using `FrozenArray<>` as a method return type
    - [x] https://github.com/immersive-web/hit-test/issues/117
    - [ ] https://github.com/w3c/ServiceWorker/issues/1722

domenic:
Also an an example of how to specify frozen array attributes.

Closes #1399.

   * Should be compatible with all code
   * Not needed
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: N/A
   * Node.js: N/A
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1413.html" title="Last updated on Jun 26, 2024, 2:22 AM UTC (9492715)">Preview</a> | <a href="https://whatpr.org/webidl/1413/f2b124e...9492715.html" title="Last updated on Jun 26, 2024, 2:22 AM UTC (9492715)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6539,6 +6539,44 @@ A <dfn id="dfn-frozen-array-type" export>frozen array type</dfn> is a parameteri
 type whose values are references to objects that hold a fixed length array
 of unmodifiable values.  The values in the array are of type |T|.
 
+Frozen array types must only be used as the type of [=regular attributes=] or [=static attributes=]
+defined on an [=interface=].
+
+<div class="example" id="example-frozen-array">
+    The following [=IDL fragment=] defines an [=interface=] with two frozen array attributes, one
+    [=read only=] and one not.
+
+    <pre highlight="webidl">
+        [Exposed=Window]
+        interface PersonalPreferences {
+            readonly attribute FrozenArray&lt;DOMString> favoriteColors;
+            attribute FrozenArray&lt;DOMString> favoriteFoods;
+
+            undefined randomizeFavoriteColors();
+        };
+    </pre>
+
+    The behavior of these attributes could be defined like so:
+
+    <blockquote>
+        Each <code>PersonalPreferences</code> has associated <b>favorite colors</b>, a {{FrozenArray}}&lt;{{DOMString}}>, initially equal to the result of [=creating a frozen array=] from « "<code>purple</code>", "<code>aquamarine</code>" ».
+
+        Each <code>PersonalPreferences</code> has an associated <b>favorite foods</b>, a {{FrozenArray}}&lt;{{DOMString}}>, initially equal to the result of [=creating a frozen array=] from the empty list.
+
+        The <code>favoriteColors</code> [=getter steps=] are to return [=this=]'s favorite colors.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6539,6 +6539,44 @@ A <dfn id="dfn-frozen-array-type" export>frozen array type</dfn> is a parameteri
...
+    <blockquote>
+        Each <code>PersonalPreferences</code> has associated <b>favorite colors</b>, a {{FrozenArray}}&lt;{{DOMString}}>, initially equal to the result of [=creating a frozen array=] from « "<code>purple</code>", "<code>aquamarine</code>" ».
+
+        Each <code>PersonalPreferences</code> has an associated <b>favorite foods</b>, a {{FrozenArray}}&lt;{{DOMString}}>, initially equal to the result of [=creating a frozen array=] from the empty list.
+
+        The <code>favoriteColors</code> [=getter steps=] are to return [=this=]'s favorite colors.
+
+        The <code>favoriteFoods</code> [=getter steps=] are to return [=this=]'s favorite foods.
+
+        The <code>favoriteFoods</code> [=setter steps=] are to set [=this=]'s favorite foods to [=the given value=].
```
annevk, line 6570: This just works? So if it's not frozen it ends up throwing or some such?

annevk (approved):

index.bs
domenic: The conversion steps for JS values to Web IDL `FrozenArray<T>` types will take any iterable, iterate over it, grab the values into an Infra list, then create a JS array from them, then freeze it, then package it up into a `FrozenArray<T>` pointer. So yeah, it just works.

PR #1419, opened 2024-07-16
Introduce a "transferable" algorithm for BufferSource

inexorabletash:
The new "transferable" algorithm allows an invoker to determine if a BufferSource can be detached without errors before invoking transfer. This is particularly helpful if several buffers are processed in a batch, as the validity of the operation can be determined up front, rather than during processing which could leave the buffers in a mix of states.

See discussion in [1]. A local copy of the algorithm was landed as [2] (as "detachable") but this seems like a useful addition to WebIDL.

1: https://github.com/webmachinelearning/webnn/issues/351
2: https://webmachinelearning.github.io/webnn/#buffersource-detachable

*Eliding the checklist since this is a helper algorithm not new functionality, but please ask if more details are desired*

***
<a href="https://whatpr.org/webidl/1419.html" title="Last updated on Jul 18, 2024, 5:31 PM UTC (08fc50a)">Preview</a> | <a href="https://whatpr.org/webidl/1419/3fb6ab4...08fc50a.html" title="Last updated on Jul 18, 2024, 5:31 PM UTC (08fc50a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9055,6 +9055,21 @@ a reference to the same object that the IDL value represents.
 </div>
 
 <div algorithm>
+    A [=buffer source type=] instance |bufferSource| is
+    <dfn for="BufferSource" export>detachable</dfn> if the following steps return true:
+
+    1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
+        |bufferSource| to a JavaScript value.
+    1.  If |jsArrayBuffer| has a \[[ViewedArrayBuffer]] internal slot, then set |jsArrayBuffer| to
+        |jsArrayBuffer|.\[[ViewedArrayBuffer]].
+    1.  If [$IsSharedArrayBuffer$](|jsArrayBuffer|) is true, then return false.
+    1.  If [$IsDetachedBuffer$](|jsArrayBuffer|) is true, then return false.
+    1.  If |jsArrayBuffer|.\[[ArrayBufferDetachKey]] is not <emu-val>undefined</emu-val>, then
+        return false.
+    1.  Return true.
+</div>
+
+<div algorithm>
     To <dfn for="ArrayBuffer" export>transfer</dfn> an {{ArrayBuffer}} |arrayBuffer|, optionally
     given a [=realm=] |targetRealm|:
 
--- a/index.bs
+++ b/index.bs
@@ -9056,7 +9056,7 @@ a reference to the same object that the IDL value represents.
 
 <div algorithm>
     A [=buffer source type=] instance |bufferSource| is
-    <dfn for="BufferSource" export>detachable</dfn> if the following steps return true:
+    <dfn for="BufferSource" export>transferable</dfn> if the following steps return true:
 
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |bufferSource| to a JavaScript value.


inexorabletash:
@domenic - could you please take a look? Low priority. 

Also, feedback on whether the checklist (re: implementation commitments, etc) is needed in this case would helpful.

index.bs
```diff
@@ -9054,6 +9054,21 @@ a reference to the same object that the IDL value represents.
...
+<div algorithm>
+    A [=buffer source type=] instance |bufferSource| is
+    <dfn for="BufferSource" export>detachable</dfn> if the following steps return true:
+
+    1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
+        |bufferSource| to a JavaScript value.
+    1.  If |jsArrayBuffer| has a \[[ViewedArrayBuffer]] internal slot, then set |jsArrayBuffer| to
+        |jsArrayBuffer|.\[[ViewedArrayBuffer]].
+    1.  If [$IsSharedArrayBuffer$](|jsArrayBuffer|) is true, then return false.
+    1.  If [$IsDetachedBuffer$](|jsArrayBuffer|) is true, then return false.
```
domenic, line 9066: Hmm, this step is a bit surprising. Is it helpful for your use case?

domenic (commented):
Awesome, happy to have this here. Just one question.

I don't think the PR template is necessary for this kind of definition.

index.bs
inexorabletash: Hopefully I'm not misunderstanding your question. The use case is over at https://webmachinelearning.github.io/webnn/#mlnamedarraybufferviews-transfer

Shorter version for explanatory purposes:

1. For each _buffer_ of _buffers_:
    1. If _buffer_ is not **detachable**, then throw a TypeError.
1. For each _buffer_ of _buffers_:
    1. Let _xfer_ be the result of [transferring](https://webidl.spec.whatwg.org/#arraybuffer-transfer) _buffer_'s underlying buffer.

The "IsDetachedBuffer" check is to ensure that [transferring](https://webidl.spec.whatwg.org/#arraybuffer-transfer) wouldn't assert ("_Assert: IsDetachedBuffer(jsArrayBuffer) is false"_).

This could alternately be done by the caller, e.g.:

1. For each _buffer_ of _buffers_:
    1. If _buffer_ has a [[ViewedArrayBuffer]] internal slot, then set _buffer_ to _buffer_.[[ViewedArrayBuffer]].
    1. If [IsDetachedBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-isdetachedbuffer)(_buffer_) is true, then throw a TypeError.
    1. If _buffer_ is not **detachable**, then throw a TypeError.
1. For each _buffer_ of _buffers_:
    1. Let _xfer_ be the result of [transferring](https://webidl.spec.whatwg.org/#arraybuffer-transfer) _buffer_'s underlying buffer.

... but it seems we'd want to keep this logic together in WebIDL. Would it be clearer if we named the new algorithm **transferable** instead?

(But maybe I'm answering the wrong question?)

inexorabletash: Okay - updated algorithm name to "transferable".

inexorabletash: ... and updated PR title and description

domenic (approved):

PR #1420, opened 2024-07-18
Throw, instead of assert, for transferring detached ArrayBuffers

domenic:
This makes the algorithm more user-friendly, as this is generally the desired behavior, and the algorithm can already throw for other reasons so callers are expected to handle that.

/cc @inexorabletash as this was inspired by your work in https://github.com/whatwg/webidl/pull/1419.

I audited [the callers](https://dontcallmedom.github.io/webdex/t.html#transfer%40%40ArrayBuffer%40dfn) and they are all fine with this. (Most are just referencing the concept of transferring, as these definitions are not yet widely used.)

***
<a href="https://whatpr.org/webidl/1420.html" title="Last updated on Jul 18, 2024, 1:23 AM UTC (bf45161)">Preview</a> | <a href="https://whatpr.org/webidl/1420/3fb6ab4...bf45161.html" title="Last updated on Jul 18, 2024, 1:23 AM UTC (bf45161)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9060,7 +9060,8 @@ a reference to the same object that the IDL value represents.
 
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
-    1.  Assert: [$IsDetachedBuffer$](|jsArrayBuffer|) is false.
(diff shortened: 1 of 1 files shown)


domenic:
I think you're right that this algorithm assumes same agent-cluster. Let me fix that in a separate PR.

PR #732, opened 2019-05-30
Mark DOMException as serializable

yutakahirano:
Mark DOMException as serialiable

Also define serialization and deserialization steps.
tests: https://github.com/web-platform-tests/wpt/pull/17095

Fixes #729

***
<a href="https://pr-preview.s3.amazonaws.com/yutakahirano/webidl/pull/732.html" title="Last updated on Jun 7, 2019, 7:51 AM UTC (0017a1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/heycam/webidl/732/4ef3509...yutakahirano:0017a1d.html" title="Last updated on Jun 7, 2019, 7:51 AM UTC (0017a1d)">Diff</a>

PR #1421, opened 2024-07-18
add [[ErrorData]] slot to DOMExceptions

ljharb:
This is an attempt to fix https://github.com/tc39/proposal-is-error/issues/9.

The intention is to make DOM Exceptions true subclasses of `Error`.

It's possible this PR needs additional changes to simplify/undo #732 - I'm happy to make whatever changes are appropriate.

   * Google
   * Firefox
   * https://github.com/web-platform-tests/wpt/pull/49133 
   * Chromium: https://issues.chromium.org/issues/382104870
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1923733
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=284506
   * Deno: gets it for free from v8
   * Node.js: gets it for free from v8
   * webidl2.js: n/a, i think?
   * widlparser: n/a, i think?
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/606

This is part of a stage 2 TC39 proposal for `Error.isError`. When it advances to stage 2.7, I'll file the appropriate implementation bugs.

***
<a href="https://whatpr.org/webidl/1421.html" title="Last updated on Dec 12, 2024, 5:00 AM UTC (fa40209)">Preview</a> | <a href="https://whatpr.org/webidl/1421/e7b2fe1...fa40209.html" title="Last updated on Dec 12, 2024, 5:00 AM UTC (fa40209)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13532,8 +13532,9 @@ the realm given as an argument.
             1.  Let |targetRealm| be [=?=] [$GetFunctionRealm$](|newTarget|).
             1.  Set |prototype| to the [=interface prototype object=] for |interface| in
                 |targetRealm|.
-    1.  Let |instance| be [$MakeBasicObject$](
-        « \[[Prototype]], \[[Extensible]], \[[Realm]], \[[PrimaryInterface]] »).
+    1.  Let |slots| be « \[[Prototype]], \[[Extensible]], \[[Realm]], \[[PrimaryInterface]] ».
+    1.  If |interface| is {{DOMException}}, append \[[ErrorData]] to |slots|.
+    1.  Let |instance| be [$MakeBasicObject$](|slots|).
     1.  Set |instance|.\[[Realm]] to |realm|.
     1.  Set |instance|.\[[PrimaryInterface]] to |interface|.
     1.  Set |instance|.\[[Prototype]] to |prototype|.
@@ -14547,6 +14548,7 @@ The [=class string=] of a [=namespace object=] is the [=namespace=]'s [=identifi
 In the JavaScript binding, the [=interface prototype object=] for {{DOMException}}
 has its \[[Prototype]] [=/internal slot=] set to the intrinsic object {{%Error.prototype%}},
 as defined in the [=create an interface prototype object=] abstract operation.
+It also has an \[[ErrorData]] slot, like all built-in exceptions.
 
 Additionally, if an implementation gives native {{Error}} objects special powers or
 nonstandard properties (such as a <code>stack</code> property),


piggyASS (commented):
Very bad spam

ljharb:
ping @domenic @annevk, this proposal is on the agenda for advancement next plenary, and i'd love to know that this direction is acceptable :-)

domenic:
Seems fine directionally. In general we want `DOMException` to behave as if its constructor had a `super()` call and I think this aligns with that.

Editorially I wonder if putting it as a step in https://webidl.spec.whatwg.org/#internally-create-a-new-object-implementing-the-interface would be better. (E.g. modifying step 4.)

I also don't understand the relationship to #732 (serializability).

ljharb:
How would you suggest I modify step 4?

The relationship to #732, as I understand it (but may indeed understand incorrectly) is that that PR was only necessary because DOMExceptions aren't proper Errors - but by making them proper Errors, then serialization of DOMExceptions should be automatic now?

domenic:
> How would you suggest I modify step 4?

By adding [[ErrorData]] to the argument list of MakeBasicObject, if the inclusive inherited interfaces of _interface_ includes DOMException.

> but by making them proper Errors, then serialization of DOMExceptions should be automatic now?

No, it's not automatic, and it should not be.

The relevant step is step 17 of https://html.spec.whatwg.org/#structuredserializeinternal. Thankfully, it's guarded by "and value is not a [platform object](https://webidl.spec.whatwg.org/#dfn-platform-object)", which means DOMException will properly go down to step 19 instead.

If we took the generic step 17 path for DOMException, we would lose the name information, and would look at the public message property instead of the internal [message](https://webidl.spec.whatwg.org/#domexception-message).

ljharb:
ah, gotcha. ok cool, i'll make those changes to step 4, thanks!

ljharb:
hmm - looking at the MakeBasicObject call, that seems like it's for all platform objects, not just DOMExceptions. Would something in https://webidl.spec.whatwg.org/#js-creating-throwing-exceptions make more sense?

If the MakeBasicObject call is correct, then how would I modify that to only add the slot for DOMExceptions and not for other things?

annevk:
I'd put the list passed to MakeBasicObject in a variable and then conditionally append [[ErrorData]] to it.

ljharb:
Thanks, updated! I left the original note I added as informative, but can remove it if preferred.

ljharb:
This proposal is now stage 3. What are the next steps to getting this PR merged?

domenic:
Implementer interest is the biggest checkbox in the OP that isn't checked. Maybe you can gather that in TC39? Ideally an explicit statement that the people implementing `Error.isError()` will also make sure the WPTs you've linked pass for `DOMException`.

syg:
For Chrome I'll make sure `Error.isError` returns true for `DOMException`s.

dminor:
For Firefox, we have https://bugzilla.mozilla.org/show_bug.cgi?id=1936385 filed, and the implementation looks straightforward, so we're also in support.

domenic (approved):
Awesome! Will merge once the rest of the checkboxes in the OP are filled out.

ljharb:
done, if i got the `n/a`'s right :-) thanks!

petamoriken:
For Deno, I have implemented it so that it returns `true` when V8 implements `Error.isError`.

PR #1422, opened 2024-07-19
Stop assuming "transfer" is same-agent cluster

domenic:
The previous text assumed that the AllocateArrayBuffer step would never throw, since it was basically spec fiction for retargeting a pointer. However, if we transfer across agent clusters, memory allocation will be needed. And even then, we shouldn't be normatively asserting that implementations use the pointer-retargeting strategy. Instead, we can explain it in a note.

***
<a href="https://whatpr.org/webidl/1422.html" title="Last updated on Jul 19, 2024, 1:30 AM UTC (6a37390)">Preview</a> | <a href="https://whatpr.org/webidl/1422/155ae10...6a37390.html" title="Last updated on Jul 19, 2024, 1:30 AM UTC (6a37390)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9036,9 +9036,10 @@ a reference to the same object that the IDL value represents.
         |arrayBuffer| to a JavaScript value.
     1.  Perform [=?=] [$DetachArrayBuffer$](|jsArrayBuffer|).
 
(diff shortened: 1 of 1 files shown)


annevk (approved):
Thanks!

PR #1424, opened 2024-07-31
Allow users of "invoke a callback function" to report the exception

### LINKED ISSUE #1423: Handle exceptions when calling callbacks by default

Splitting from https://github.com/whatwg/html/issues/958#issuecomment-226309871.

After https://github.com/whatwg/html/pull/10404 lands, HTML will have a solid foundation for reporting exceptions.

It turns out that most cases, when spec writers use Web IDL to [construct](https://webidl.spec.whatwg.org/#construct-a-callback-function) or [invoke](https://webidl.spec.whatwg.org/#invoke-a-callback-function) callback functions, or to [call a user object's operation](https://webidl.spec.whatwg.org/#call-a-user-objects-operation), they want errors to be reported to the global error handler; they don't want them to propagate out. Such reporting is a safer default.

We should add a new [optional named parameter](https://infra.spec.whatwg.org/#algorithm-params) _rethrowExceptions_ (default false) to these definitions.  When it is false, we introduce the following new behavior: any exceptions thrown by web developer code get caught and routed to HTML's new "report the exception". ("Web developer code" is roughly everything between the "prepare" and "cleanup" steps, including both argument conversion and the actual code invocation.)

Which global do we propagate to? @jeremyroman's research in https://github.com/whatwg/html/pull/10404 settled on the JavaScript function object's realm's global object.

---

Before merging this change, however, we need to have PRs ready for all call sites of construct/invoke/call a user object's operation, to prepare them for the new default. In most cases this will be deleting some spec code that try/catches the exception and reports it. So, this will be a multi-spec project.

We can use these links to find all callers:

- [Invoke](https://dontcallmedom.github.io/webdex/i.html#invoke%40%40webidl%25%25dfn)
- [Call a user object's operation](https://dontcallmedom.github.io/webdex/c.html#call%20a%20user%20object%E2%80%99s%20operation%40%40webidl%25%25dfn)
- [Construct](https://dontcallmedom.github.io/webdex/c.html#construct%40%40webidl%25%25dfn)

jeremyroman:
This makes it more concise for users to, as is frequently desired for undefined-returning callbacks especially, immediately catch and report the exception rather than needing to handle it themselves.

This isn't appropriate for other types of callback functions, where a result may be expected or the exception needs to be rethrown. Callers need to explicitly decide which behavior they want, unless the callback returns a promise type, in which case exceptions are turned into rejected promises implicitly.

Fixes #1423.

- [x] At least two implementers are interested (and none opposed): n/a; no normative change
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at: n/a; no normative change
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed: n/a; no normative change
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: n/a; no normative change

***
<a href="https://whatpr.org/webidl/1424.html" title="Last updated on Aug 2, 2024, 5:35 AM UTC (000a709)">Preview</a> | <a href="https://whatpr.org/webidl/1424/155ae10...000a709.html" title="Last updated on Aug 2, 2024, 5:35 AM UTC (000a709)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14400,19 +14400,19 @@ described in the previous section).
 
     To <dfn id="invoke-a-callback-function" export>invoke</dfn> a
     [=callback function type=] value |callable| with a [=Web IDL arguments list=] |args|,
-    exception behavior |exceptionBehavior| (either "`report`" or "`rethrow`"),
+    exception behavior |exceptionBehavior| (either "<code>report</code>" or "<code>rethrow</code>"),
     and an optional [=callback this value|callback this value=] |thisArg|,
     perform the following steps.
     These steps will either return an IDL value or throw an exception.
 
     The |exceptionBehavior| argument must be supplied if, and only if, |callable|'s
     [=return type=] is not a [=promise type=]. If |callable|'s return type is neither
-    {{undefined}} nor {{any}}, it may not be "`report`".
+    {{undefined}} nor {{any}}, it may not be "<code>report</code>".
 
     <div class="XXX">
         Until call sites are updated to respect this, specifications which fail to
-        provide a value here when it would be mandatory should be read to supply
-        "`rethrow`".
+        provide a value here when it would be mandatory should be understood as
+        supplying "<code>rethrow</code>".
     </div>
 
     1.  Let |completion| be an uninitialized variable.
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic ptal?

index.bs
```diff
@@ -14399,11 +14399,22 @@ described in the previous section).
...
-    [=callback function type=] value |callable| with a [=Web IDL arguments list=] |args|
+    [=callback function type=] value |callable| with a [=Web IDL arguments list=] |args|,
+    exception behavior |exceptionBehavior| (either "<code>report</code>" or "<code>rethrow</code>"),
     and an optional [=callback this value|callback this value=] |thisArg|,
     perform the following steps.
     These steps will either return an IDL value or throw an exception.
 
+    The |exceptionBehavior| argument must be supplied if, and only if, |callable|'s
+    [=return type=] is not a [=promise type=]. If |callable|'s return type is neither
+    {{undefined}} nor {{any}}, it may not be "<code>report</code>".
```
domenic, line 14410:
```suggestion
    {{undefined}} nor {{any}}, it must be "<code>rethrow</code>".
```

domenic (approved):
LGTM with nit.

PR #1427, opened 2024-08-22
export "is a platform object"

tabatkins:
This algo is being used in <https://github.com/w3c/css-houdini-drafts/pull/1127>, but since it's unexported (and thus not linking by default), the author of that PR is dropping down to a manual anchor.


domenic:
Per https://github.com/w3c/css-houdini-drafts/pull/1127/files#r1728342155 I don't think we should export this. This is an intentionally-internal operation and external specs should use higher-level constructs like overloading, callback types, and [=implements=].

PR #1483, opened 2025-04-08
Define ModuleNamespace to ECMA-262 Module Namespace Exotic Object

guybedford:
This came up in a PR addition to the WebAssembly ESM Integration in https://github.com/WebAssembly/esm-integration/pull/104 to have a WebIDL function definition returning a module namespace object (`namespaceInstance` below):

```webidl
interface Instance {
  constructor(Module module, optional object importObject);
  readonly attribute object exports;
  static Instance namespaceInstance(ModuleNamespace moduleNamespace);
};
```

To support this it would be beneficial to have an upstream definition of `ModuleNamespace` to ensure validations are handled at the WebIDL layer for functions taking module namespaces as arguments.

The PR has not landed yet, posting this up for initial feedback before the ESM Integration PR lands.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5730,6 +5730,7 @@ are known as <dfn id="dfn-object-type" export>object types</dfn>.
         "async iterable" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         "object" Null
         "symbol" Null
+        "ModuleNamespace" Null
         BufferRelatedType Null
         "FrozenArray" "&lt;" TypeWithExtendedAttributes "&gt;" Null
         "ObservableArray" "&lt;" TypeWithExtendedAttributes "&gt;" Null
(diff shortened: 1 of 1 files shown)


annevk:
This looks okay to me, but the template will have to be filled in. One caveat here I guess is that Wasm JS API bindings are not actually implemented through IDL in (most?) implementations.

guybedford:
Thanks for taking a look and I believe you're right on the implementation aspects. The alternative would be to just use `object` and then to define the validations in line, but this seemed better if possible, unless you can see any reasons against having this defined here.

I'll aim to get the Node.js implementation together and post back the template with an update when that's ready.

PR #1529, opened 2025-10-08
Fix `BufferSource` algorithms for shared and resizable buffers

### LINKED ISSUE #1312: Buffer source types infrastructure
The algorithms provided in https://webidl.spec.whatwg.org/#es-buffer-source-types are a little shaky as the [AllowResizable] and [AllowShared] extended attributes end up changing the IDL type. Thus in principle these algorithms cannot be used if you have an IDL object for a view that points to a shared buffer, for instance.

Properly changing the types to be maximally inclusive would have some downstream repercussions and we'd also want to be careful that whatever maximally-inclusive IDL typedef we create here can be used for any future expansions of these buffers as well, should they happen. (Unless we're happy to do this again, but meh.)

### LINKED ISSUE #1385: "byte length of a buffer source type" needs updating for resizable buffers

As of [this proposal](https://github.com/tc39/proposal-resizablearraybuffer) / [PR](https://github.com/tc39/ecma262/pull/3116), the `[[ByteLength]]` slot on TypedArrays and DataViews can be `~auto~` instead of an integer. Consumers of [the "byte length of a buffer source" operation](https://webidl.spec.whatwg.org/#buffersource-byte-length) (like [`encodeInto`](https://encoding.spec.whatwg.org/#dom-textencoder-encodeinto)) probably want [TypedArrayByteLength](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-typedarraybytelength) or [GetViewByteLength](https://tc39.es/ecma262/multipage/structured-data.html#sec-getviewbytelength) instead of the raw value of the `[[ByteLength]]` slot.

Also, when getting the length of an ArrayBuffer, this should probably use the [ArrayBufferByteLength](https://tc39.es/ecma262/multipage/structured-data.html#sec-arraybufferbytelength) abstract operation, not raw access of the slot.

There's some other raw uses of the `[[ByteLength]]` slot, like in ["write a byte sequence bytes into an ArrayBufferView"](https://webidl.spec.whatwg.org/#arraybufferview-write), which should probably also be updated.

cc @syg

MattiasBuelens:
Previously, many `BufferSource` algorithms were using internal slots such as `[[ByteOffset]]` and `[[ByteLength]]` directly. However, with the addition of `SharedArrayBuffer` (#353, #1311) and `[AllowResizable]` (#982), this comes with extra caveats:
* To get the byte length of a `SharedArrayBuffer`, we need to use [`ArrayBufferByteLength`](https://tc39.es/ecma262/multipage/structured-data.html#sec-arraybufferbytelength) to match [`SharedArrayBuffer.prototype.byteLength`](https://tc39.es/ecma262/multipage/structured-data.html#sec-get-sharedarraybuffer.prototype.bytelength).
* To get the byte length of a ["length-tracking" typed array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#behavior_when_viewing_a_resizable_buffer), we need to use [`TypedArrayByteLength`](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-typedarraybytelength). Similarly, for a length-tracking `DataView`, we need to use [`GetViewByteLength`](https://tc39.es/ecma262/multipage/structured-data.html#sec-getviewbytelength).
* To get the byte offset of a typed array or `DataView`, we first need to check whether the view is still in bounds using [`IsArrayBufferViewOutOfBounds`](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-isarraybufferviewoutofbounds).

"Byte length" and "byte offset" both mirror JavaScript's asymmetry between typed arrays and `DataView`s: an out-of-bounds typed array reports 0, whereas an out-of-bounds `DataView` throws a `TypeError`. That way, implementations can reuse their existing getters.

When reading the length of a buffer, we use `SEQ-CST` ordering to match ECMAScript. When reading the *contents* of a buffer, we keep using `UNORDERED`, as "get a copy of the bytes held by the buffer source" already did. See https://github.com/whatwg/webidl/pull/1529#issuecomment-3757018345.

Transferring an `ArrayBuffer` must also take into account whether it should remain resizable or not. Fortunately, we can use the new [`ArrayBufferCopyAndDetach`](https://tc39.es/ecma262/multipage/structured-data.html#sec-arraybuffercopyanddetach) operation for that.
* I added an optional `preserveResizability` parameter to "transfer an `ArrayBuffer`", so specifications can choose whether to preserve resizability or not. By default, this is not preserved. I intend to use it in Streams, see https://github.com/whatwg/streams/pull/1360.
* I removed the `realm` parameter of that algorithm, since `ArrayBufferCopyAndDetach` doesn't accept such a parameter. [According to WebDex](https://dontcallmedom.github.io/webdex/t.html#transfer%40%40ArrayBuffer%40dfn), no specs were actually using that parameter anyway.

Normative changes, which need a review of existing callers:
* "Byte length" now throws a `TypeError` for a `DataView` that is detached or out of bounds, where it previously returned a stale `[[ByteLength]]`. This is reachable without `[AllowResizable]`, simply by detaching the underlying buffer. "Get a copy of the bytes held by the buffer source" inherits this. Both algorithms have a note documenting it, but every existing caller is currently written as an infallible read.
* "Transfer" no longer takes a `realm` parameter. Per WebDex this is not a breaking change in practice, but it does change the exported signature.
* "Transfer" on a resizable `ArrayBuffer` now yields a fixed-length `ArrayBuffer` by default, instead of reusing the original buffer's resizability.

Relates to #1312.
Fixes #1385.

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/62301
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1529.html" title="Last updated on Aug 29, 2026, 2:41 PM UTC (aabd4f3)">Preview</a> | <a href="https://whatpr.org/webidl/1529/3c82f93...aabd4f3.html" title="Last updated on Aug 29, 2026, 2:41 PM UTC (aabd4f3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9316,6 +9316,7 @@ a reference to the same object that the IDL value represents.
     1.  Let |arrayBuffer| be |bufferSource|.
     1.  Let |offset| be 0.
     1.  Let |length| be |bufferSource|'s [=BufferSource/byte length=].
+    1.  If |length| is 0, then return the empty [=byte sequence=].
     1.  If |bufferSource| is a [=buffer view type=] instance:
         1.  Set |arrayBuffer| to |bufferSource|'s [=underlying buffer=].
         1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
--- a/index.bs
+++ b/index.bs
@@ -9321,8 +9321,8 @@ a reference to the same object that the IDL value represents.
         1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
     1.  If |length| is 0, then return the empty [=byte sequence=].
 
-        <p class="note">This step is not strictly necessary, since [=the range=] below is empty
-        when |length| is 0. It is spelled out as a fast path for implementations.
+        <p class="note">This step is not strictly necessary, since [=the exclusive range|the range=]
+        below is empty when |length| is 0. It is spelled out as a fast path for implementations.
     1.  Assert: |arrayBuffer| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
--- a/index.bs
+++ b/index.bs
@@ -9320,6 +9320,9 @@ a reference to the same object that the IDL value represents.
         1.  Set |arrayBuffer| to |bufferSource|'s [=BufferSource/underlying buffer=].
         1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
     1.  If |length| is 0, then return the empty [=byte sequence=].
+
+        <p class="note">This step is not strictly necessary, since [=the range=] below is empty
+        when |length| is 0. It is spelled out as a fast path for implementations.
     1.  Assert: |arrayBuffer| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
--- a/index.bs
+++ b/index.bs
@@ -9317,12 +9317,13 @@ a reference to the same object that the IDL value represents.
     1.  Let |arrayBuffer| be |bufferSource|.
     1.  Let |offset| be 0.
     1.  If |bufferSource| is a [=buffer view type=] instance:
-        1.  Set |arrayBuffer| to |bufferSource|'s [=underlying buffer=].
+        1.  Set |arrayBuffer| to |bufferSource|'s [=BufferSource/underlying buffer=].
         1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
-            <p class="note">This will throw an exception if |bufferSource| is
-            [=ArrayBufferView/out of bounds=].
     1.  If |length| is 0, then return the empty [=byte sequence=].
-    1.  [=/Assert=]: |arrayBuffer| is an {{ArrayBuffer}} or {{SharedArrayBuffer}} object.
+
+        <p class="note">This step is not strictly necessary, since [=the range=] below is empty
+        when |length| is 0. It is spelled out as a fast path for implementations.
+    1.  [=/Assert=]: |arrayBuffer| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
     1.  [=/Assert=]: [$IsDetachedBuffer$](|jsArrayBuffer|) is false.
@@ -9331,6 +9332,10 @@ a reference to the same object that the IDL value represents.
         set |bytes|[|i| &minus; |offset|] to [$GetValueFromBuffer$](|jsArrayBuffer|, |i|, Uint8,
         true, Unordered).
     1.  Return |bytes|.
+
+    <p class="note" id="note-get-buffer-source-copy-exceptions">This will throw an exception if
+    |bufferSource| is a {{DataView}} that is [=ArrayBufferView/out of bounds=], since reading its
+    [=BufferSource/byte length=] <a href="#note-BufferSource-byte-length-exceptions">throws</a>.
 </div>
 
 <div algorithm>
@@ -9362,7 +9367,8 @@ a reference to the same object that the IDL value represents.
     [=typed array type=] instance) or {{DataView/byteOffset|DataView.prototype.byteOffset}}
     (for a {{DataView}}). In particular, this deliberately mirrors JavaScript's asymmetry: a
     [=typed array type=] instance that is [=ArrayBufferView/out of bounds=] has a byte offset of
-    0, whereas a {{DataView}} that is [=ArrayBufferView/out of bounds=] instead throws.
+    0, whereas a {{DataView}} that is [=ArrayBufferView/out of bounds=] instead throws an
+    exception.
 </div>
 
 <div algorithm>
@@ -9392,6 +9398,12 @@ a reference to the same object that the IDL value represents.
     {{ArrayBuffer/byteLength|ArrayBuffer.prototype.byteLength}} and
     {{SharedArrayBuffer/byteLength|SharedArrayBuffer.prototype.byteLength}}
     (for a [=buffer type=] instance).
+
+    <p class="note" id="note-BufferSource-byte-length-exceptions">This will throw an exception if
+    |bufferSource| is a {{DataView}} that is [=ArrayBufferView/out of bounds=], which includes the
+    case where it is [=BufferSource/detached=]. Every other [=buffer source type=] instance
+    returns 0 in those cases. This mirrors JavaScript, where only
+    {{DataView/byteLength|DataView.prototype.byteLength}} throws.
 </div>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -9426,7 +9426,7 @@ a reference to the same object that the IDL value represents.
     <dfn id=arraybuffer-write-startingoffset export for="ArrayBuffer/write,SharedArrayBuffer/write">|startingOffset|</dfn>
     (default 0):
 
-    1.  Assert: |bytes|'s [=byte sequence/length=] ≤ |arrayBuffer|'s [=byte length=]
+    1.  Assert: |bytes|'s [=byte sequence/length=] ≤ |arrayBuffer|'s [=BufferSource/byte length=]
         &minus; |startingOffset|.
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
--- a/index.bs
+++ b/index.bs
@@ -9359,12 +9359,9 @@ a reference to the same object that the IDL value represents.
                 Seq-Cst).
             1.  If [$IsViewOutOfBounds$](|viewRecord|) is true, then return 0.
             1.  Return [$GetViewByteLength$](|viewRecord|).
-    1.  Otherwise, if [$IsSharedArrayBuffer$](|jsBufferSource|) is true:
-        1.  [=/Assert=]: |jsBufferSource| is a {{SharedArrayBuffer}}.
-        1.  Return [$ArrayBufferByteLength$](|jsBufferSource|, Seq-Cst).
     1.  Otherwise:
-        1.  [=/Assert=]: |jsBufferSource| is an {{ArrayBuffer}}.
-        1.  Return |jsBufferSource|.\[[ArrayBufferByteLength]].
+        1.  [=/Assert=]: |jsBufferSource| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
+        1.  Return [$ArrayBufferByteLength$](|jsBufferSource|, Seq-Cst).
 </div>
 <!-- TODO: should this be stricter around not being detached? -->
 
--- a/index.bs
+++ b/index.bs
@@ -9357,7 +9357,8 @@ a reference to the same object that the IDL value represents.
             1.  [=/Assert=]: |jsBufferSource| is a {{DataView}}.
             1.  Let |viewRecord| be [$MakeDataViewWithBufferWitnessRecord$](|jsBufferSource|,
                 Seq-Cst).
-            1.  If [$IsViewOutOfBounds$](|viewRecord|) is true, then return 0.
+            1.  If [$IsViewOutOfBounds$](|viewRecord|) is true, then [=JavaScript/throw=] a
+                <l spec=ecmascript>{{TypeError}}</l>.
             1.  Return [$GetViewByteLength$](|viewRecord|).
     1.  Otherwise:
         1.  [=/Assert=]: |jsBufferSource| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
--- a/index.bs
+++ b/index.bs
@@ -9320,9 +9320,6 @@ a reference to the same object that the IDL value represents.
         1.  Set |arrayBuffer| to |bufferSource|'s [=BufferSource/underlying buffer=].
         1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
     1.  If |length| is 0, then return the empty [=byte sequence=].
-
-        <p class="note">This step is not strictly necessary, since [=the range=] below is empty
-        when |length| is 0. It is spelled out as a fast path for implementations.
     1.  Assert: |arrayBuffer| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
     1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
         |arrayBuffer| to a JavaScript value.
--- a/index.bs
+++ b/index.bs
@@ -9351,6 +9351,7 @@ a reference to the same object that the IDL value represents.
             1.  [=/Assert=]: |jsBufferSource| is a [=typed array type=] instance.
             1.  Let |taRecord| be [$MakeTypedArrayWithBufferWitnessRecord$](|jsBufferSource|,
                 Seq-Cst).
+            1.  If [$IsTypedArrayOutOfBounds$](|taRecord|) is true, return 0.
             1.  Return [$TypedArrayByteLength$](|taRecord|).
         1.  Otherwise:
             1.  [=/Assert=]: |jsBufferSource| is a {{DataView}}.
--- a/index.bs
+++ b/index.bs
@@ -9351,7 +9351,7 @@ a reference to the same object that the IDL value represents.
             1.  [=/Assert=]: |jsBufferSource| is a [=typed array type=] instance.
             1.  Let |taRecord| be [$MakeTypedArrayWithBufferWitnessRecord$](|jsBufferSource|,
                 Seq-Cst).
-            1.  If [$IsTypedArrayOutOfBounds$](|taRecord|) is true, return 0.
+            1.  If [$IsTypedArrayOutOfBounds$](|taRecord|) is true, then return 0.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9294,19 +9294,15 @@ a reference to the same object that the IDL value represents.
...
-        1.  Set |length| to |jsBufferSource|.\[[ByteLength]].
-    1.  Otherwise:
-        1.  Assert: |jsBufferSource| is an {{ArrayBuffer}} or
-            {{SharedArrayBuffer}} object.
-        1.  Set |length| to |jsBufferSource|.\[[ArrayBufferByteLength]].
+    1.  Let |length| be |bufferSource|'s [=BufferSource/byte length=].
+    1.  If |bufferSource| is a [=buffer view type=] instance, then:
+        1.  Set |arrayBuffer| to |bufferSource|'s [=underlying buffer=].
+        1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
+    1.  Assert: |arrayBuffer| is an {{ArrayBuffer}} or {{SharedArrayBuffer}} object.
```
```suggestion
    1.  [=/Assert=]: |arrayBuffer| is an {{ArrayBuffer}} or {{SharedArrayBuffer}} object.
```

index.bs
```diff
@@ -9316,15 +9312,40 @@ a reference to the same object that the IDL value represents.
...
     |bufferSource| is the value returned by the following steps:
 
     1.  Let |jsBufferSource| be the result of [=converted to a JavaScript value|converting=]
         |bufferSource| to a JavaScript value.
-    1.  If |jsBufferSource| has a \[[ViewedArrayBuffer]] internal slot, then return
-        |jsBufferSource|.\[[ByteLength]].
-    1.  Return |jsBufferSource|.\[[ArrayBufferByteLength]].
+    1.  If |jsBufferSource| has a \[[ViewedArrayBuffer]] [=/internal slot=], then:
+        1.  If |jsBufferSource| has a \[[TypedArrayName]] [=/internal slot=]:
+            1.  Assert: |jsBufferSource| is a [=typed array type=] instance.
```
```suggestion
            1.  [=/Assert=]: |jsBufferSource| is a [=typed array type=] instance.
```

index.bs
```diff
@@ -9316,15 +9312,40 @@ a reference to the same object that the IDL value represents.
...
+            1.  Assert: |jsBufferSource| is a {{DataView}}.
+            1.  Let |viewRecord| be [$MakeDataViewWithBufferWitnessRecord$](|jsBufferSource|,
+                Seq-Cst).
+            1.  If [$IsViewOutOfBounds$](|viewRecord|) is true, return 0.
+            1.  Return [$GetViewByteLength$](|viewRecord|).
+    1.  Otherwise, if [$IsSharedArrayBuffer$](|jsBufferSource|) is true, then:
+        1.  Assert: |jsBufferSource| is a {{SharedArrayBuffer}}.
+        1.  Return [$ArrayBufferByteLength$](|jsBufferSource|, Seq-Cst).
+    1.  Otherwise:
+        1.  Assert: |jsBufferSource| is an {{ArrayBuffer}}.
```
```suggestion
        1.  [=/Assert=]: |jsBufferSource| is an {{ArrayBuffer}}.
```

index.bs
```diff
@@ -9429,22 +9446,20 @@ a reference to the same object that the IDL value represents.
...
         <l spec=ecmascript>{{TypeError}}</l>.
-    1.  Let |arrayBufferData| be |jsArrayBuffer|.\[[ArrayBufferData]].
-    1.  Let |arrayBufferByteLength| be |jsArrayBuffer|.\[[ArrayBufferByteLength]].
-    1.  Perform [=?=] [$DetachArrayBuffer$](|jsArrayBuffer|).
-    1.  If |targetRealm| is not given, let |targetRealm| be the [=current realm=].
-    1.  Let |jsTransferred| be [=?=]
-        [$AllocateArrayBuffer$](|targetRealm|.\[[Intrinsics]].\[[{{%ArrayBuffer%}}]], 0).
-    1.  Set |jsTransferred|.\[[ArrayBufferData]] to |arrayBufferData|.
-    1.  Set |jsTransferred|.\[[ArrayBufferByteLength]] to |arrayBufferByteLength|.
+    1.  If |preserveResizability| is true, then:
```
```suggestion
    1.  If |preserveResizability| is true:
```

annevk (commented):
I did an editorial pass.

MattiasBuelens (commented):
Quick self-review.

index.bs
```diff
@@ -9294,19 +9294,15 @@ a reference to the same object that the IDL value represents.
...
-            {{SharedArrayBuffer}} object.
-        1.  Set |length| to |jsBufferSource|.\[[ArrayBufferByteLength]].
+    1.  Let |length| be |bufferSource|'s [=BufferSource/byte length=].
+    1.  If |bufferSource| is a [=buffer view type=] instance:
+        1.  Set |arrayBuffer| to |bufferSource|'s [=underlying buffer=].
+        1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
+    1.  [=/Assert=]: |arrayBuffer| is an {{ArrayBuffer}} or {{SharedArrayBuffer}} object.
+    1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
+        |arrayBuffer| to a JavaScript value.
     1.  If [$IsDetachedBuffer$](|jsArrayBuffer|) is true, then return the empty
```
bakkot: If you pass a TA backed by a detached buffer, then the `Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].` step above will throw. It's weird that passing a TA backed by a detached buffer throws while passing the detached buffer itself returns the empty byte sequence (in this step). It doesn't fundamentally break anything, but it's inconsistent.

index.bs
```diff
@@ -9316,15 +9312,38 @@ a reference to the same object that the IDL value represents.
...
+            1.  [=/Assert=]: |jsBufferSource| is a [=typed array type=] instance.
+            1.  Let |taRecord| be [$MakeTypedArrayWithBufferWitnessRecord$](|jsBufferSource|,
+                Seq-Cst).
+            1.  If [$IsTypedArrayOutOfBounds$](|taRecord|) is true, then return 0.
+            1.  Return [$TypedArrayByteLength$](|taRecord|).
+        1.  Otherwise:
+            1.  [=/Assert=]: |jsBufferSource| is a {{DataView}}.
+            1.  Let |viewRecord| be [$MakeDataViewWithBufferWitnessRecord$](|jsBufferSource|,
+                Seq-Cst).
+            1.  If [$IsViewOutOfBounds$](|viewRecord|) is true, then return 0.
```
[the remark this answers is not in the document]
MattiasBuelens: That is quite horrifying! 😱

My intention is to match `(DataView|TypedArray|ArrayBuffer).prototype.byteLength` exactly, so implementors can reuse their existing implementations for those getters. I'll change it to match ECMAScript, and add a note that we're just copying their idiocies. 😅

index.bs
```diff
@@ -9316,15 +9312,38 @@ a reference to the same object that the IDL value represents.
...
+            1.  Return [$GetViewByteLength$](|viewRecord|).
+    1.  Otherwise:
+        1.  [=/Assert=]: |jsBufferSource| is an {{ArrayBuffer}} or a {{SharedArrayBuffer}}.
+        1.  Return [$ArrayBufferByteLength$](|jsBufferSource|, Seq-Cst).
```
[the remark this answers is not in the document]
MattiasBuelens: Right, [ArrayBuffer.prototype.byteLength](https://tc39.es/ecma262/multipage/structured-data.html#sec-get-arraybuffer.prototype.bytelength) does a detached check before reading the byte length slot. I'll add that check here too and return 0. 👍

index.bs
MattiasBuelens: I've moved the detached check all the way to the top. I also check if *length* is zero, so reading byte offset later on won't throw.

Funnily enough, reading *bufferSource's byte length* itself can still throw for an out-of-bounds `DataView` [(as you already reported)](https://github.com/whatwg/webidl/pull/1529#discussion_r2663169745)... Do we want "get a copy of the bytes held" to propagate that error? Or should we return an empty byte sequence in that case? 🤔

MattiasBuelens: I changed it so we read both *byte length* and *byte offset* before we check if length is zero. That way, any exceptions from reading those properties will be thrown first.

MattiasBuelens:
@annevk @bakkot Could you re-review? Thanks. 🙏

annevk:
AI found a number of substantive issues that appear to be correct: https://gist.github.com/annevk/123a3994ba82ec664b787559803cdabf (it's also incorrect about some things though; our Bikeshed setup would definitely fail on link errors).

index.bs
```diff
@@ -9473,11 +9530,12 @@ a reference to the same object that the IDL value represents.
             <a href="#note-ArrayBuffer-detach-exceptions">explained in that algorithm's
             definition</a>;
         *   |arrayBuffer| is already [=BufferSource/detached=];
-        *   Sufficient memory cannot be allocated in |realm|. Generally this will only be the case
-            if |realm| is in a different [=agent cluster=] than the one in which |arrayBuffer| was
-            allocated. If they are in the same [=agent cluster=], then implementations will just
-            change the backing pointers to get the same observable results with better performance
-            and no allocations.
+        *   Sufficient memory cannot be allocated in the [=current realm=], which is the [=realm=]
```
MattiasBuelens, line 9533: I'm not sure if we have to fully spell out the meaning of "current realm" here? Happy to remove that.

index.bs
```diff
@@ -9380,17 +9440,19 @@ a reference to the same object that the IDL value represents.
...
-        <i>[=ArrayBuffer/write/startingOffset=]</i> set to |jsView|.\[[ByteOffset]] +
-        |startingOffset|.
+    1.  [=ArrayBuffer/Write=] |bytes| into |view|'s [=BufferSource/underlying buffer=] with
+        <i>[=ArrayBuffer/write/startingOffset=]</i> set to
+        |view|'s [=ArrayBufferView/byte offset=] + |startingOffset|.
+
+    <p class="note">Callers must ensure that |view| is not [=ArrayBufferView/out of bounds=],
+    which also means that it is not [=BufferSource/detached=]. Otherwise, |view|'s
+    [=BufferSource/byte length=] and [=ArrayBufferView/byte offset=] could throw an exception,
+    and assertions must not be observable.
```
MattiasBuelens: Okay, this is just the AI copying the review remarks into the specification... 🙄 I'll remove this sentence.

MattiasBuelens (commented):
Quick self-review.

index.bs
```diff
@@ -9312,26 +9312,63 @@ a reference to the same object that the IDL value represents.
...
-        1.  Set |length| to |jsBufferSource|.\[[ArrayBufferByteLength]].
-    1.  If [$IsDetachedBuffer$](|jsArrayBuffer|) is true, then return the empty
-        [=byte sequence=].
+    1.  If |bufferSource| is a [=buffer view type=] instance:
+        1.  Set |arrayBuffer| to |bufferSource|'s [=BufferSource/underlying buffer=].
+        1.  Set |offset| to |bufferSource|'s [=ArrayBufferView/byte offset=].
+    1.  If |length| is 0, then return the empty [=byte sequence=].
+
+        <p class="note">This step is not strictly necessary, since [=the range=] below is empty
+        when |length| is 0. It is spelled out as a fast path for implementations.
```
annevk: This is necessary because of the second assert below.

Also, when _length_ is 0 you would get _offset_ + 0 &minus; 1 below, right? That also seems like a problem. This might be an existing problem in this document though.

index.bs
```diff
@@ -9380,17 +9440,15 @@ a reference to the same object that the IDL value represents.
     {{ArrayBufferView}} |view|, optionally given a
     <dfn export for="ArrayBufferView/write">|startingOffset|</dfn> (default 0):
 
-    1.  Let |jsView| be the result of [=converted to a JavaScript value|converting=] |view| to
-        a JavaScript value.
-    1.  Assert: |bytes|'s [=byte sequence/length=] ≤ |jsView|.\[[ByteLength]] &minus;
-        |startingOffset|.
+    1.  Assert: |view| is not [=ArrayBufferView/out of bounds=].
+        <p class="note">This implies that |view| is not [=BufferSource/detached=].
```
annevk, line 9448: We should add a newline before this note.

annevk (approved):
This looks okay to me now, but how do we want to handle tests and such?

MattiasBuelens:
We'll need some APIs that already have `[AllowResizable]` and/or `[AllowShared]` attributes. Looking around a bit, I found:
* `WebAssembly.validate()`, `.compile()` and `.instantiate()` accept `[AllowResizable] AllowSharedBufferSource bytes`, so that's ideal for testing the combination of both attributes. [See WASM-JS-API.](https://webassembly.github.io/spec/js-api/index.html#webassembly-namespace)
* Encoding is being updated to have `[AllowResizable]` on `TextDecoder.decode()` and `TextEncoder.encodeInto()`, see https://github.com/whatwg/encoding/pull/362. However, that hasn't landed yet.
* I'm planning to update Streams to support `[AllowResizable]` on `ReadableStreamBYOBReader.read(view)` in https://github.com/whatwg/streams/pull/1360. (In the future, I want to also add `[AllowShared]`.)

So I think the best way forward right now is to use the WebAssembly methods for testing. It'll be a bit awkward since the test inputs will have to be valid WASM modules, but I think I can make that work. I'll take a look. 🙂

MattiasBuelens:
I've opened https://github.com/web-platform-tests/wpt/pull/62301, which tests "get a copy of the bytes held by the buffer source" across all combinations of resizable/shared buffers and fixed-length/length-tracking/out-of-bounds views.

index.bs
MattiasBuelens: With the changes from #1632, this should no longer be a *necessary* step. I'll bring the note back.

MattiasBuelens:
I ran the new tests from https://github.com/web-platform-tests/wpt/pull/62301 against stable Chromium and Firefox, and there were already a few issues. 😅 

Chromium currently rejects `SharedArrayBuffer`s and `DataView`s on `WebAssembly` methods. see https://github.com/web-platform-tests/wpt/pull/62301#issuecomment-5476246270. Definitely a bug, since the Web IDL standard *as published today* already requires this to work. I filed [#555029264](https://issues.chromium.org/issues/555029264) on their end.

However, when Chromium decides to fix this bug, we'll need to decide what behavior they should be implementing for `DataView`s, specifically for **out-of-bounds DataViews**.
* The published Web IDL standard would read a stale `[[ByteLength]]` field - definitely wrong.
* Gecko (in Firefox 154) treats this as if the DataView were *empty*, matching the behavior for an out-of-bounds typed array. This appears to be the more "natural" behavior, as browsers can treat typed arrays and `DataView`s more or less the same.
* With this PR, browsers would need to *throw* instead. This aligns with [ECMAScript's `DataView.prototype.byteLength`](https://tc39.es/ecma262/multipage/structured-data.html#sec-get-dataview.prototype.bytelength), but requires browsers to have separate paths for typed arrays and `DataView`s instead of a single `ArrayBufferView` path.

Note that I haven't tested WebKit yet, since I don't have a macOS device. It might be interesting to know how they handle out-of-bounds `DataView`s, so we can better assess the impact of this change.

@bakkot Are we still happy with this change [as discussed earlier](https://github.com/whatwg/webidl/pull/1529#discussion_r2663169745)? If yes, I'll make sure to get implementer's interest from the Gecko folks, since this *will* change observable behavior in their implementation.

bakkot:
I'll review this in a day or two though I'm somewhat rusty on the webidl side of things.

FWIW, another possible fix for the DataView issue is to make those operations _not_ throw on DataViews in ECMAScript (and then have the web match). Making errors into non-errors is almost always web-compat. (Fun fact: ArrayBuffer.prototype.byteLength used to be spec'd to throw also; this was changed in https://github.com/tc39/ecma262/pull/2164.)

MattiasBuelens:
> FWIW, another possible fix for the DataView issue is to make those operations _not_ throw on DataViews in ECMAScript (and then have the web match). Making errors into non-errors is almost always web-compat. (Fun fact: ArrayBuffer.prototype.byteLength used to be spec'd to throw also; this was changed in [tc39/ecma262#2164](https://github.com/tc39/ecma262/pull/2164).)

Interesting! Looks like this was raised before: https://github.com/tc39/ecma262/issues/2635. I'll comment there to bring this back to TC39's attention.

bakkot:
I [added an agenda item](https://github.com/tc39/agendas/pull/2143) for TC39 to discuss this at the end of the month. I weakly expect it to get through, though the PR probably won't actually _land_ until a couple implementations have shipped, which will be another 3-6 months probably. Assuming we get consensus there I think it'd be fine to have this reflect the new intended behavior, ahead of the PR landing.

index.bs
```diff
@@ -9312,26 +9312,65 @@ a reference to the same object that the IDL value represents.
     To <dfn id="dfn-get-buffer-source-copy" export lt="get a copy of the buffer source|get a copy of the bytes held by the buffer source">get a copy of the bytes held by the buffer source</dfn>
     given a [=buffer source type=] instance |bufferSource|:
 
-    1.  Let |jsBufferSource| be the result of [=converted to a JavaScript value|converting=]
-        |bufferSource| to a JavaScript value.
-    1.  Let |jsArrayBuffer| be |jsBufferSource|.
+    1.  If |bufferSource| is [=BufferSource/detached=], then return the empty [=byte sequence=].
```
bakkot, line 9315: It's weird for this to treat detached and out of bounds differently - nothing in ecma262 does that. My preference would be to rectify this by replacing this "is detached" check to "is out of bounds", and have "is out of bounds" cover both detached and out of bounds ([like it does in ecma262](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-istypedarrayoutofbounds)).

In fact probably the right thing is to replace _every_ detachedness check with an OOB check covering both possibilities.

PR #1568, opened 2026-02-11
Add an extended attribute to allow propagating TC39's AsyncContext

andreubotella:
This PR updates WebIDL to add the `[PropagatesAsyncContext]` extended attribute. This is a part of the TC39 AsyncContext proposal, which allows storing state associated with an async flow of execution in JavaScript, and preserving it across different kinds of async continuations in both JS and the web platform.

This `[PropagatesAsyncContext]` extended attribute can only be applied to callback function types in operation arguments. When set, the callback context will also store an Async Context Mapping, which the callback will run in.

This allows a callback-taking operation to act like an async continuation, and propagate the state associated with the async flow of execution at the time that this operation is called to the callback.

This patch relies on HTML's PR whatwg/html#12152 to define the "run with Async Context Mapping" operation. This operation must be defined in a web specs rather than in the TC39 proposal because it deals with throwing in spec algorithms, which is not a concept in the TC39 specs.

This PR is part of whatwg/html#10432.

   * Chromium
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1568.html" title="Last updated on Aug 5, 2026, 12:53 PM UTC (3188eab)">Preview</a> | <a href="https://whatpr.org/webidl/1568/fad9b4c...3188eab.html" title="Last updated on Aug 5, 2026, 12:53 PM UTC (3188eab)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,6 +91,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ecmascript
         url: sec-object-type
             text: is an Object
             text: is not an Object
+urlPrefix: https://tc39.es/proposal-async-context/; spec: PROPOSAL-ASYNCCONTEXT
+    type: dfn
+        text: Async Context Mapping; url: async-context-mapping
+    type: abstract-op
+        text: AsyncContextSnapshot; url: sec-asynccontextsnapshot
 </pre>
 
 <pre class=biblio>
@@ -99,6 +104,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ecmascript
         "publisher": "Ecma",
         "href": "https://tc39.es/proposal-float16array/",
         "title": "Proposal to add float16 TypedArrays to JavaScript"
+    },
+    "PROPOSAL-ASYNCCONTEXT": {
+        "publisher": "Ecma",
+        "href": "https://tc39.es/proposal-async-context/",
+        "title": "AsyncContext"
     }
 }
 </pre>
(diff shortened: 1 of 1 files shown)


jasnell:
Spec text looks reasonable to me. Not a browser but from the perspective of both Node.js and Cloudflare Workers, consider this a +1.

nicolo-ribaudo:
I rebased and fixed up some things in this pull request. Note that it does not fully work when it comes to nullable types, and it's related to https://github.com/whatwg/webidl/issues/670. In the corresponding HTML PR we currently use `[PropagatesAsyncContext] FunctionStringCallback? _callback` in [DataTransferItem's getAsString](https://html.spec.whatwg.org/#datatransferitem), but `FunctionStringCallback?` is _not_ a callback type so we cannot technically use the `[PropagatesAsyncContext]` attribute on it.

I am considering doing a ad-hoc fix, even though eventually we should fix https://github.com/whatwg/webidl/issues/670 in general.

**EDIT:** I opened https://github.com/whatwg/webidl/pull/1624 to fix this in general.

shaseley:
Do [callback interfaces](https://webidl.spec.whatwg.org/#idl-callback-interfaces) need to be handled as well? Looking at these in [Chromium idl](https://source.chromium.org/search?q=%22callback%20interface%22%20f:idl$%20AND%20f:third_party%2Fblink%2Frenderer&ss=chromium%2Fchromium%2Fsrc), some we definitely wouldn't want to include (event listener), but some we might (file APIs).

nicolo-ribaudo:
Probably -- my only interaction with callback interfaces has only ever been through events and I completely forgot that they are used in other places too.

PR #1578, opened 2026-03-10
Update DOMException language to reflect standardized stack accessor

ljharb:
Since DOMException objects have [[ErrorData]], the Error.prototype.stack accessor from the Error Stack Accessor proposal naturally applies to them. Update the language to reflect that stack is now a specified accessor property rather than a nonstandard property that implementations "should" expose.

See https://github.com/tc39/proposal-error-stack-accessor/issues/9 and https://github.com/tc39/proposal-error-stack-accessor/issues/8

   * Reached TC39 stage 3 and implementers appear on board.
   * https://github.com/web-platform-tests/wpt/pull/58397
   * Chromium: Captcha :-( 
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2042301
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=315550
   * Deno: https://github.com/denoland/deno/issues/34388
   * Node.js: https://github.com/nodejs/node/issues/63579
   * webidl2.js: https://github.com/w3c/webidl2.js/issues/843
   * widlparser: https://github.com/plinss/widlparser/issues/96

***
<a href="https://whatpr.org/webidl/1578.html" title="Last updated on May 26, 2026, 7:16 AM UTC (627f1c6)">Preview</a> | <a href="https://whatpr.org/webidl/1578/e1300e9...627f1c6.html" title="Last updated on May 26, 2026, 7:16 AM UTC (627f1c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14925,10 +14925,10 @@ The [=class string=] of a [=namespace object=] is the [=namespace=]'s [=identifi
 In the JavaScript binding, the [=interface prototype object=] for {{DOMException}}
 has its \[[Prototype]] [=/internal slot=] set to the intrinsic object {{%Error.prototype%}},
 as defined in the [=create an interface prototype object=] abstract operation.
-It also has an \[[ErrorData]] slot, like all built-in exceptions.
+It also has \[[ErrorData]] and \[[Stack]] slots, like all built-in exceptions.
 
-Additionally, if an implementation gives native {{Error}} objects special powers or
-nonstandard properties (such as a <code>stack</code> property),
+Additionally, if an implementation gives native {{Error}} objects other special powers or
+nonstandard properties (such as <code>line</code> and <code>column</code> properties),
 it should also expose those on {{DOMException}} objects.
 
 <h4 id="js-exception-objects" oldids="es-exception-objects" dfn>Exception objects</h4>
@@ -15236,8 +15236,9 @@ Their [=serialization steps=], given <var>value</var> and <var>serialized</var>,
 <ol>
     <li>Set <var>serialized</var>.\[[Name]] to <var>value</var>'s [=DOMException/name=].</li>
     <li>Set <var>serialized</var>.\[[Message]] to <var>value</var>'s [=DOMException/message=].</li>
+    <li>Set <var>serialized</var>.\[[Stack]] to an implementation-defined string derived from <var>value</var>.\[[Stack]].</li>
     <li>User agents should attach a serialized representation of any interesting accompanying data
-    which are not yet specified, notably the <code>stack</code> property, to
+    which are not yet specified, notably the <code>line</code> and <code>column</code> properties, to
     <var>serialized</var>.</li>
 </ol>
 
@@ -15246,6 +15247,7 @@ Their [=deserialization steps=], given <var>value</var> and <var>serialized</var
 <ol>
     <li>Set <var>value</var>'s [=DOMException/name=] to <var>serialized</var>.\[[Name]].</li>
     <li>Set <var>value</var>'s [=DOMException/message=] to <var>serialized</var>.\[[Message]].</li>
+    <li>Set <var>value</var>.\[[Stack]] to <var>serialized</var>.\[[Stack]].</li>
     <li>If any other data is attached to <var>serialized</var>, then deserialize and attach it to
     <var>value</var>.</li>
 </ol>


annevk:
I'm not sure why this particular paragraph needs a change (doesn't it follow from the prior paragraph?), but I would expect changes to the serialization and deserialization steps here: https://webidl.spec.whatwg.org/#idl-DOMException

ljharb:
This one needs to change because stack will no longer be non-standard.

I'll add some changes to the linked paragraph, thanks :-)

index.bs
```diff
@@ -14927,9 +14927,11 @@ has its \[[Prototype]] [=/internal slot=] set to the intrinsic object {{%Error.p
 as defined in the [=create an interface prototype object=] abstract operation.
 It also has an \[[ErrorData]] slot, like all built-in exceptions.
 
-Additionally, if an implementation gives native {{Error}} objects special powers or
-nonstandard properties (such as a <code>stack</code> property),
-it should also expose those on {{DOMException}} objects.
```
annevk, line 14932: Now that I understand why you want to change this I think we should just change the example from "such as a `stack` property" to "such as a hypothetical `howToAddressThisError` property".

index.bs
```diff
@@ -14927,9 +14927,11 @@ has its \[[Prototype]] [=/internal slot=] set to the intrinsic object {{%Error.p
 as defined in the [=create an interface prototype object=] abstract operation.
 It also has an \[[ErrorData]] slot, like all built-in exceptions.
```
annevk: Maybe here we need to call out that it has a \[[Stack]] slot as well?

index.bs
ljharb: hmm - i can certainly do that, but is it really valuable to make up a hypothetical property name?

annevk: Reportedly (from feedback on the HTML PR) there are other non-standard properties that are real we could also use. We could also remove the parenthetical, but I think it provides some clarity.

ljharb: should we just mention line and column then?

annevk: If those are the ones that are missing, sure. (Sorry, I don't know exactly.)

ljharb: Done!

annevk:
I don't understand the latest force push. It seems to have regressed a bunch of work.

ljharb:
woof, yeah, that's clearly a mistake. i'll fix it.

index.bs
```diff
@@ -14925,11 +14925,14 @@ The [=class string=] of a [=namespace object=] is the [=namespace=]'s [=identifi
...
 
-Additionally, if an implementation gives native {{Error}} objects special powers or
-nonstandard properties (such as a <code>stack</code> property),
-it should also expose those on {{DOMException}} objects.
+Since {{DOMException}} objects have an \[[ErrorData]] slot, the
+<code>Error.prototype.stack</code> accessor and other {{Error}}-related behaviors
+gated on \[[ErrorData]] apply to them. Additionally, if an implementation gives
+native {{Error}} objects other special powers or nonstandard properties
+(such as <code>line</code> and <code>column</code> properties), it should
+also expose those on {{DOMException}} objects.
```
```suggestion
Additionally, if an implementation gives native {{Error}} objects other special powers or
nonstandard properties (such as <code>line</code> and <code>column</code> properties),
it should also expose those on {{DOMException}} objects.
```
I don't think we have any business here mentioning `stack` as it's handled at a different layer now.

index.bs
```diff
@@ -15246,6 +15250,7 @@ Their [=deserialization steps=], given <var>value</var> and <var>serialized</var
 <ol>
     <li>Set <var>value</var>'s [=DOMException/name=] to <var>serialized</var>.\[[Name]].</li>
     <li>Set <var>value</var>'s [=DOMException/message=] to <var>serialized</var>.\[[Message]].</li>
+    <li>Set <var>value</var>.\[[Stack]] to an implementation-defined value derived from <var>serialized</var>.\[[Stack]].</li>
```
```suggestion
    <li>Set <var>value</var>.\[[Stack]] to <var>serialized</var>.\[[Stack]].</li>
```
We only have to do the implementation-defined bit once. Just like in HTML.

annevk (approved):

annevk:
Once the relevant proposals are stage 3 this seems good.

ljharb:
The proposal is now stage 3, and test262 tests for it have been merged.

PR #1610, opened 2026-06-17
Editorial: Use CreateBuiltinFunction for (async) iterator return and next methods

nicolo-ribaudo:
> This makes it explicit what the `name` and `length` of these methods is.
> Many other functions created by WebIDL already use CreateBuiltinFunction.

Right now it was a bit implicit, because we were defining _steps_ taking 0/1 parameters, but not actual JS functions.

This is already tested by https://github.com/web-platform-tests/wpt/blob/8e113d970b5ef2b6c0b9a81f38dd6dbadaa31795/streams/readable-streams/async-iterator.any.js#L38-L39.

In ECMA-262 these `.return()`/`.next()` methods have length of 0 or 1 depending on whether they use their argument or not, so it's probably ok for WebIDL to have `.next()` have length of 0 and `.return(value)` have length of 1.

Note that both Chrome and Firefox currently fail this test (https://wpt.fyi/results/streams/readable-streams/async-iterator.any.html?label=master&label=experimental&aligned&q=stream), as they set the length of `.return()` to 0. Looking at precedent in ECMA-262, 1 is the right length here since the algorithm actually use the parameter.

cc @javifernandez

***
<a href="https://whatpr.org/webidl/1610.html" title="Last updated on Jun 17, 2026, 2:18 PM UTC (c9664b5)">Preview</a> | <a href="https://whatpr.org/webidl/1610/cca1103...c9664b5.html" title="Last updated on Jun 17, 2026, 2:18 PM UTC (c9664b5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12946,7 +12946,8 @@ must be {{%Iterator.prototype%}}.
 
     An [=iterator prototype object=] must have a <code class="idl">next</code> data property with attributes
     { \[[Writable]]: <emu-val>true</emu-val>, \[[Enumerable]]: <emu-val>true</emu-val>, \[[Configurable]]: <emu-val>true</emu-val> }
-    and whose value is a [=built-in function object=] that behaves as follows:
+    and whose value is the [=built-in function object=] [$CreateBuiltinFunction$](|nextSteps|, 0, "<code>next</code>", « »),
+    where |nextSteps| is the following algorithm:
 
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
Looks good

annevk (approved):
This is somewhat inconsistent with the existing style for these things, but we should probably align on this explicit style. I'll create a follow-up PR.

PR #1624, opened 2026-08-04
Propagate a type's extended attributes to a nullable inner type

### LINKED ISSUE #670: Is `long?` an integer type?

nicolo-ribaudo:
This patch makes types such as `[Clamp] long?` legal, propagating the extended attribute to the inner type of the nullable type.

It relaxes the definitions of `[AllowResizable]`, `[AllowShared]`, `[Clamp]` and `[EnforceRange]` so that they can be used as such; while it leaves `[LegacyNullToEmptyString]` as is since it does not really make sense to apply it to a nullable string (since then the attribute would be a no-op, given that the null is already handled).

Closes #670.

---

My motivation for doing this is that I will need it in https://github.com/whatwg/webidl/pull/1568.

---

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
   * webidl2.js: …
   * widlparser: …

***
<a href="https://whatpr.org/webidl/1624.html" title="Last updated on Sep 14, 2026, 1:41 PM UTC (20cc820)">Preview</a> | <a href="https://whatpr.org/webidl/1624/fad9b4c...20cc820.html" title="Last updated on Sep 14, 2026, 1:41 PM UTC (20cc820)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6492,7 +6492,7 @@ The following extended attributes are <dfn for="extended attributes">applicable 
     an IDL type |type| are determined as follows:
 
     1.  Let |extended attributes| be a new empty [=ordered set|set=].
-    1.  If |type| appears as part of a
+    1.  If |type| is the <emu-nt><a href="#prod-Type">Type</a></emu-nt> directly within a
         <emu-nt><a href="#prod-TypeWithExtendedAttributes">TypeWithExtendedAttributes</a></emu-nt>
         production, [=set/append=] each of the [=extended attributes=] present in the production's
         <emu-nt><a href="#prod-ExtendedAttributeList">ExtendedAttributeList</a></emu-nt> to
@@ -6536,10 +6536,9 @@ The following extended attributes are <dfn for="extended attributes">applicable 
                 };
             </pre>
         </div>
-    1.  If |type| appears as part of a <emu-nt><a href="#prod-Type">Type</a></emu-nt> production
-        directly within an <emu-nt><a href="#prod-Argument">Argument</a></emu-nt> production,
-        [=set/append=] to |extended attributes| all of the [=extended attributes=] present in the
-        production's
+    1.  If |type| is the <emu-nt><a href="#prod-Type">Type</a></emu-nt> directly within an
+        <emu-nt><a href="#prod-Argument">Argument</a></emu-nt> production, [=set/append=] to
+        |extended attributes| all of the [=extended attributes=] present in the production's
         <emu-nt><a href="#prod-ExtendedAttributeList">ExtendedAttributeList</a></emu-nt> that are
         [=applicable to types=].
 
@@ -6555,10 +6554,10 @@ The following extended attributes are <dfn for="extended attributes">applicable 
             [=applicable to types=]; otherwise [<code>XAttr</code>] applies to the argument, and not
             the argument's type.
         </div>
-    1.  If |type| appears as part of a <emu-nt><a href="#prod-Type">Type</a></emu-nt> production
-        directly within a <emu-nt><a href="#prod-DictionaryMember">DictionaryMember</a></emu-nt>
-        production, [=set/append=] to |extended attributes| all of the [=extended attributes=]
-        present in the production's
+    1.  If |type| is the <emu-nt><a href="#prod-Type">Type</a></emu-nt> directly within a
+        <emu-nt><a href="#prod-DictionaryMember">DictionaryMember</a></emu-nt> production,
+        [=set/append=] to |extended attributes| all of the [=extended attributes=] present in the
+        production's
         <emu-nt><a href="#prod-ExtendedAttributeList">ExtendedAttributeList</a></emu-nt> that are
         [=applicable to types=].
 
--- a/index.bs
+++ b/index.bs
@@ -6522,6 +6522,17 @@ The following extended attributes are <dfn for="extended attributes">applicable 
                 };
             </pre>
         </div>
+    1.  If |type| is the [=nullable types/inner type=] of a [=nullable type=] |N|, [=set/append=]
+        each of the [=extended attributes associated with=] |N| to |extended attributes|.
+
+        <div class="example" id="example-5feb96ff">
+            <pre class="idl">
+                [Exposed=Window]
+                interface I {
+                    undefined f([Clamp] long? arg);
+                };
+            </pre>
+        </div>
     1.  If |type| appears as part of a <emu-nt><a href="#prod-Type">Type</a></emu-nt> production
         directly within an <emu-nt><a href="#prod-Argument">Argument</a></emu-nt> production,
         [=set/append=] to |extended attributes| all of the [=extended attributes=] present in the
@@ -9578,7 +9589,8 @@ that is growable.
 
 The [{{AllowResizable}}] extended attribute must [=takes no arguments|take no arguments=].
 
-A type that is not a [=buffer source type=] must not be
+A type that is not a [=buffer source type=], and that is not a [=nullable type=] whose
+[=nullable types/inner type=] is a [=buffer source type=], must not be
 [=extended attributes associated with|associated with=] the [{{AllowResizable}}] extended attribute.
 
 See the rules for converting JavaScript values to IDL [=buffer source types=] in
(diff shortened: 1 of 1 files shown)


saschanaz:
I'm not sure I like this kind of syntactic ambiguity, can we force `([Clamp] long)?`

nicolo-ribaudo:
Probably yes (I actually first started doing that, before changing my mind :sweat_smile:). The main reasons why I went with this approach are that:
- it's a much larger change, affecting a lot of the WebIDL grammar
- it's kind of weird to have something that normally is space-separate having tighter precedence than something that is not

I can finish my attempt and open a PR to compare.

Note that the approach taken by this PR is similar to what we do for union types, e.g. `[Attr] (SomeType or undefined) foo`

nicolo-ribaudo:
Actually, I'm not sure it would be possible*. There are currently productions that are not nullable (e.g. `Promise<T>`), which are however children of `TypeWithExtendedAttributes`. If I made `TypeWithExtendedAttributes` a child of "something ?", i.e.
```
NullableType:
  TypeWithExtendedAttributes `?`
  TypeWithExtendedAttributes

TypeWithExtendedAttributes :
  ExtendedAttributes Type
  Type

Type :
  ...
``` 

\* well, it is, but it further complicates the grammar because I'd need to have a `TypeWithExtendedAttributesForInsideNullable` and `TypeWithExtendedAttributesForOutsideNullable` with different allowed types inside of them.

nicolo-ribaudo:
@saschanaz Another problem with that approach is with types in arguments. Per https://github.com/whatwg/webidl/issues/691, we can have both extended attributes that apply to arguments as well as extended attributes that apply to types. This means that in an argument `[AttrForArg, AttrForType] Type? x`, `[AttrForArg, AttrForType]` _cannot_ bind tighter than `?`, because `AttrForArg` needs to apply to the whole argument.

saschanaz:
By forcing I was thinking about "this extended attribute is valid only on ..." type of thing rather than forcing syntactically. That should be doable.

> This means that in an argument `[AttrForArg, AttrForType] Type? x`, `[AttrForArg, AttrForType]` _cannot_ bind tighter than `?`, because `AttrForArg` needs to apply to the whole argument.

It can be `[AttrForArg] ([AttrForType] Type)? x` right? Yes it's kinda awkward to read, but maybe less ambiguous?

nicolo-ribaudo:
> By forcing I was thinking about "this extended attribute is valid only on ..." type of thing rather than forcing syntactically. That should be doable.

Isn't this what this PR is doing? It marks the extended attributes with, e.g.

> A type that is not an [=integer type=], and that is not a [=nullable type=] whose [=nullable types/inner type=] is an [=integer type=], must not be [=extended attributes associated with|associated with=] the [{{Clamp}}] extended attribute.

saschanaz:
Yes except the propagation through nullable type.

nicolo-ribaudo:
Ok I'm very confused because that would be effectively "not do this PR", I think maybe I misunderstood your original comment:

> I'm not sure I like this kind of syntactic ambiguity, can we force `([Clamp] long)?`

Where you suggesting "can we force `[Clamp] long?` to be parsed with `[Clamp] ...` having higher priority than `... ?`, effectively as if it was `([Clamp] long)?`, or were you suggesting to actually allow writing parentheses there?

I interpreted it as the first one but I'm now guessing you meant the second one? 😅

annevk (commented):
This seems like the right direction, WebGL already relies on it (it has `[AllowShared] ArrayBufferView?`), and engines already behave this way.

Unless @saschanaz is prepared to drive the alternative approach and get the world to adopt that, I think we should go ahead with the direction this PR proposes. It's also not clear the alternative is better as it would allow types such as `(long)` potentially.

AI found a couple of issues:

- For each of the four attributes there are two kinds of sentence: a "must not be associated with" restriction, and an "appears on X, it creates a new IDL type…" definition. The PR edits only the former. The clean fix is to change "appears on" → "is associated with" in those six sentences, which also repairs the same staleness for the existing union propagation.
- There is no NullableType production. Null : "?" | ε (index.bs:5845) hangs off DistinguishableType : PrimitiveType Null (index.bs:5770). So in long?, the only type node is the whole DistinguishableType; the inner long is a PrimitiveType
  — not a Type, SingleType, or DistinguishableType. It exists only as the §6187 concept.
- UnionMemberType : ExtendedAttributeList DistinguishableType (index.bs:5758) is not a TypeWithExtendedAttributes production, and no step consumes it. Step 3 only pushes down from the union. So [AllowShared] ArrayBufferView written as a
  union member has nothing associating the attribute with it — which is what Web IDL's own AllowSharedBufferSource typedef does at index.bs:15087. That's a hole in the standard today, in the algorithm this PR edits.

nicolo-ribaudo:
> - For each of the four attributes there are two kinds of sentence: a "must not be associated with" restriction, and an "appears on X, it creates a new IDL type…" definition. The PR edits only the former. The clean fix is to change "appears on" → "is associated with" in those six sentences, which also repairs the same staleness for the existing union propagation.

I think this is fine, actually. Those sentences (e.g. second paragraph of https://webidl.spec.whatwg.org/#AllowResizable) are descriptive; the actual normative behavior is given by _associated with_ which already covers all cases.

We can make it more precise and use _is associated with_ rather than _appears on_, but that's a pre-existing issue as it covers every case of extended attributes moving from a type to a nested one (e.g. with type unions). Would you want it here or in a separate editorial PR?

> - There is no NullableType production. Null : "?" | ε (index.bs:5845) hangs off DistinguishableType : PrimitiveType Null (index.bs:5770). So in long?, the only type node is the whole DistinguishableType; the inner long is a PrimitiveType
— not a Type, SingleType, or DistinguishableType. It exists only as the §6187 concept.

I agree with this observation, but do you think it implies some concrete change for this PR?

> - UnionMemberType : ExtendedAttributeList DistinguishableType (index.bs:5758) is not a TypeWithExtendedAttributes production, and no step consumes it. Step 3 only pushes down from the union. So [AllowShared] ArrayBufferView written as a
union member has nothing associating the attribute with it — which is what Web IDL's own AllowSharedBufferSource typedef does at index.bs:15087. That's a hole in the standard today, in the algorithm this PR edits.

This is a pre-existing bug on `main`, right? Extended attributes are meant to be put on the whole union (and then they are associated with its members) rather than on the members themselves.

index.bs
```diff
@@ -6522,6 +6522,17 @@ The following extended attributes are <dfn for="extended attributes">applicable
                 };
             </pre>
         </div>
+    1.  If |type| is the [=nullable types/inner type=] of a [=nullable type=] |N|, [=set/append=]
+        each of the [=extended attributes associated with=] |N| to |extended attributes|.
+
+        <div class="example" id="example-5feb96ff">
```
```suggestion
        <div class="example" id="example-extended-attribute-association">
```

saschanaz:
> I interpreted it as the first one but I'm now guessing you meant the second one? 😅

I did mean the second one, but I realize that needs parser change as currently the rules do not allow a single item inside parentheses - it must always be a union right now. Maybe changing that is not worth just to fix this 🤔

nicolo-ribaudo:
- "Introduce concept of a "potentially nullable" type" solves the first of those three points by also making things a bit nicer.
- "Clairfy that the inner types are "directly within" ..." I think solves the second point, but I'm not 100% sure it does not have unintended side effects by making things stricter.
- For the third point, there is already https://github.com/whatwg/webidl/issues/827. How it interacts with potentially nullable types really depends on what solution we'll have for that issue.

index.bs
```diff
@@ -6536,10 +6536,10 @@ The following extended attributes are <dfn for="extended attributes">applicable
                 };
             </pre>
         </div>
-    1.  If |type| is the <emu-nt><a href="#prod-Type">Type</a></emu-nt> directly within an
-        <emu-nt><a href="#prod-Argument">Argument</a></emu-nt> production, [=set/append=] to
-        |extended attributes| all of the [=extended attributes=] present in the production's
-        <emu-nt><a href="#prod-ExtendedAttributeList">ExtendedAttributeList</a></emu-nt> that are
+    1.  If |type| is the type of an argument |A|, [=set/append=] to |extended attributes| all of
```
nicolo-ribaudo, line 6539: This wording ("type of an argument", and "type of a dictionary member" below) is already used in a few other places.

